PR #588, opened 2025-03-16
Update Ballerina extension readme images

anuruddhal:
- Update Ballerina extension readme images


NipunaRanasinghe (commented):
@anuruddhal IMO its better to have a focused view of the diagrams instead of having the entire vscode window (like in the old screenshots)? wdyt?

NipunaRanasinghe (approved):

PR #635, opened 2026-06-01
Improve repo security

gigara:
## Summary

- **SHA-pin all GitHub Actions**: Replace mutable tags (`v4`, `master`, `0.35.0`, etc.) with immutable 40-character commit SHAs across all 12 workflow and composite-action files (70 `uses:` entries). Prevents silent tag mutations from affecting CI.
- **Husky pre-commit hook**: Replicate the strict-versioning gate from `wso2-vscode-extensions` — validates that staged `package.json` files contain no `^` or `~` version prefixes. Hook is auto-installed via `rush.json` `postRushInstall` event so every developer gets it after `rush install`.
- **Fix pre-existing violations**: Remove `^`/`~` prefixes in `packages/trace-visualizer` and `packages/ballerina-grammar` so the hook passes from day one.
- **Fix `publish-vsix.yml` workflow error**: `secrets` context is not available in `if:` expressions — move the secret into `env:` and guard with a shell-level empty-check instead.
- **Misc**: Upgrade erroneous `actions/upload-artifact@v7` ref to v4; remove invalid `archive: false` parameter (not a valid v4 input); pin `aquasecurity/trivy-action@master` to the same `0.35.0` SHA used elsewhere.

## Test plan

- [ ] PR build passes (`pull-request.yml`)
- [ ] Verify husky hook runs on next `rush install` by checking `.husky/pre-commit` is present
- [ ] Confirm no `uses:` entries with bare tags remain: `grep -r "uses:" .github/ | grep -v "@[0-9a-f]\{40\}"`

### DIFF
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -57,7 +57,7 @@ runs:
   using: "composite"
   steps:
     - name: Setup Rush
-      uses: gigara/setup-rush@v1.2.0
+      uses: gigara/setup-rush@7ba3446572d15f5eacc85de383ad9679dfb7223c # v1.2.0
       with:
         pnpm: 10.11.0
         node: 22.x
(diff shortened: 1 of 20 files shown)


NipunaRanasinghe (approved):

NipunaRanasinghe:
@coderabbitai review

PR #646, opened 2026-06-17
Normalize tool-call inputs to prevent `tool_use.input` string bug

lochana-chathura:
$subject.

Added `normalizeToolCallInputs()` in `ai-utils.ts`, a defensive guard that walks every assistant message before it is sent to the API, finds any `tool-call` block where `input` is a string (or null/undefined), tries `JSON.parse()` to recover the object, and falls back to `{}` if that also fails. Called in two places:

- `prepareStep` in `AgentExecutor.ts`: covers SDK-accumulated messages during in-step streaming (the main failure path).
-  `populateHistoryForAgent` in `ai-utils.ts`: covers cross-stage AI Chat history loaded from `chatStateStorage`.

Fixes https://github.com/wso2/product-integrator/issues/1741


lochana-chathura:
Closing as the issue was resolved with a proxy side fix.

PR #664, opened 2026-07-02
Improve LS test stability and fix dependency version drift

NipunaRanasinghe:
## Summary
- Scope the LS test Ballerina dependency cache key to the PR's base branch (`github.base_ref`) instead of the PR ref (`github.ref_name`, which resolves to `<pr-number>/merge`), matching upstream's caching strategy. Every PR previously got its own always-cold cache namespace instead of sharing one warm cache per target branch.
- Drop the branch-agnostic third-level cache restore-key fallback, which could inject an unrelated branch's stale cache state — upstream has no equivalent.
- Fix `PublishArtifactsSubscriberTest.assertArtifacts()` to gate on `isJsonPermutation()` (normalized JSON comparison) instead of raw `Map.equals()`, matching the already-correct pattern in `ArtifactsTest` in the same file. The raw-equals gate was version-sensitive to fields (like package icon URLs) that the normalized comparison correctly treats as environment-dependent noise.
- Bump `ballerina/http` 2.16.3 → 2.16.4 across `flow-model-generator` and `service-model-generator` test fixtures. The cache-key fix above removed an accidental stale-cache safety net that had been masking this drift across most of the test suite.
- Bump `stdlibTimeVersion` 2.8.0 → 2.8.1 and add `stdlibXmlDataVersion=1.6.3` (new `io.ballerina.lib:data.xmldata-ballerina` pin), matching versions actually resolved during test runs. Ports [ballerina-platform/ballerina-language-server#955](https://github.com/ballerina-platform/ballerina-language-server/pull/955).

### DIFF
--- a/.github/actions/ls-test/action.yml
+++ b/.github/actions/ls-test/action.yml
@@ -40,7 +40,6 @@ runs:
         key: ${{ runner.os }}-ballerina-${{ inputs.cache-branch }}-${{ hashFiles('**/gradle.properties') }}
         restore-keys: |
           ${{ runner.os }}-ballerina-${{ inputs.cache-branch }}-
-          ${{ runner.os }}-ballerina-
 
     - name: Test with Gradle
       shell: bash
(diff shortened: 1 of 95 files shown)


samithkavishke (approved):
LGTM

PR #666, opened 2026-07-03
Improve test dependency management in LS

pasindufernando1:
## Purpose

Decouples the language-server test suite from floating library versions.

Today each test compiles real Ballerina projects and resolves their connector/standard-library
dependencies from Ballerina Central at whatever version happens to be latest when the tests run. That
makes the suite behave like a set of integration tests bound to a moving target: whenever any of those
libraries publishes a new release, the recorded expectations stop matching and a batch of otherwise
unrelated tests goes red, intermittently blocking CI. This PR makes the tests resolve from a fixed,
locked, offline dependency set so the runs are deterministic and independent of upstream releases.

Tracking issue: https://github.com/wso2-enterprise/integration-engineering/issues/1830

## Approach

- **Single locked dependency project** (`build-config/ballerina_dependencies`) declares every package
  the tests need. One `bal build` resolves the whole graph and locks every version — transitives
  included — in `Dependencies.toml`.
- **Build-owned offline cache**: the build provisions those locked packages into a dedicated Ballerina
  home (`BALLERINA_HOME_DIR`); the test JVMs resolve only from it and never contact Central.
- **Offline enforcement** (`-Dls.test.offline`): test resolution is gated offline. Production
  resolution is unchanged.
- **SOFT locking** when loading balas so their baked transitive versions re-resolve to the
  locked/cached versions instead of demanding the exact baked version.

## Test migration

- Regenerated the test snapshots onto the locked versions; dropped the committed `Dependencies.toml`
  lockfiles inside individual test source projects and removed pinned version selectors so they resolve
  from the shared locked set.
- Aligned the declared versions in `Ballerina.toml` with the resolved lock.
- Investigated and disabled the deprecated `ballerinax/ai` test cases (to be migrated by the AI team),
  and aligned the AI model / embedding / vector-store listings with `dependent_modules_fallback.json`.

## Known upstream issues surfaced

- Connectors loaded from their compiled form (BIR) fall back to type (zero) default values instead of
  the symbolic defaults; affected snapshots were conformed to the fallback output —
  https://github.com/wso2/product-integrator/issues/1893
- `ballerinax/salesforce` and `ballerinax/wso2.controlplane` fail to load from their BIR under the
  provisioned distribution; the affected cases are skipped with the issue referenced in the code —
  https://github.com/wso2/product-integrator/issues/1920


packages/ballerina-language-server/build.gradle
```diff
@@ -85,8 +85,8 @@ subprojects {
         ballerinaStdLibs "io.ballerina.stdlib:os-ballerina:${stdlibOsVersion}"
         ballerinaStdLibs "io.ballerina.stdlib:uuid-ballerina:${stdlibUuidVersion}"
         ballerinaStdLibs "io.ballerina.stdlib:auth-ballerina:${stdlibAuthVersion}"
-        ballerinaStdLibs "io.ballerina.stdlib:oauth2-ballerina:${stdlibOAuth2Version}"
-        ballerinaStdLibs "io.ballerina.lib:data.jsondata-ballerina:${stdlibJsonDataVersion}"
+        ballerinaStdLibs "io.ballerina.stdlib:oauth2-ballerina:${stdlibOauth2Version}"
+        ballerinaStdLibs "io.ballerina.lib:data.jsondata-ballerina:${stdlibJsondataVersion}"
```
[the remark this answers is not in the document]
pasindufernando1: Addressed

packages/ballerina-language-server/build.gradle
```diff
@@ -106,51 +106,58 @@ if (moduleVersion.indexOf('-') != -1) {
...
+        def currentHash = java.security.MessageDigest
+            .getInstance("SHA-256")
+            .digest(tomlFile.bytes)
+            .collect { String.format("%02x", it) }
+            .join()
+
+        if (hashFile.exists() && hashFile.text.trim() == currentHash) {
+            logger.lifecycle("Ballerina.toml unchanged — skipping dependency resolution")
+            return
+        }
```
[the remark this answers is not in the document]
pasindufernando1: Addressed

packages/ballerina-language-server/build.gradle
```diff
@@ -340,25 +347,8 @@ task pack(type: com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar) {
...
-    dependsOn pullBallerinaModule('trigger.shopify', ballerinaxShopifyTriggerVersion)
-    dependsOn pullBallerinaModule('trigger.hubspot', ballerinaxHubspotTriggerVersion)
-    dependsOn pullBallerinaModule('np', ballerinaxNp)
-    dependsOn pullBallerinaModule('workflow', ballerinaWorkflowVersion, "ballerina")
-}
-
 test {
-    dependsOn pullBallerinaModules
+    dependsOn resolveBallerinaDependencies
 }
```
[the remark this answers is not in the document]
pasindufernando1: Addressed

packages/ballerina-language-server/build.gradle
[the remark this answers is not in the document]
pasindufernando1: Addressed

packages/ballerina-language-server/build-config/ballerina_dependencies/Ballerina.toml
```diff
@@ -0,0 +1,179 @@
...
+
+[[dependency]]
+org = "ballerinax"
+name = "ai.azure"
+version = "1.1.2"
+
+[[dependency]]
+org = "ballerinax"
+name = "ai.anthropic"
+version = "1.1.2"
```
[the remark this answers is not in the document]
pasindufernando1: Addressed

pasindufernando1:
@coderabbitai review

.DS_Store
NipunaRanasinghe, line 1: `.DS_Store` is already listed in the root `.gitignore` (line 61) and no `.DS_Store` is tracked anywhere on `main`, so this must have been force-added. Could you drop it from the branch?

.github/workflows/daily-build.yml
```diff
@@ -192,7 +192,7 @@ jobs:
       - name: Set Up Ballerina
         uses: ballerina-platform/setup-ballerina@97fd67111e3c3ecdfafff96fc24d95ad909ab299 # v1.1.4
         with:
-          version: 2201.13.3
+          version: 2201.13.4
 
       - name: Cache Ballerina dependencies
         uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
```
NipunaRanasinghe, line 193: It seems the cache-path migration landed in `.github/actions/ls-test/action.yml` but not here for some reason

`ls-windows-build` runs `./gradlew.bat build` (line 210). The root project applies `id "java"`, so `build` → `check` → `test`, and the new `allprojects { tasks.withType(Test) }` block in `build.gradle` points those test JVMs at `BALLERINA_HOME_DIR = build/ballerina_dependencies/home`. That directory isn't cached in this job, so every Windows daily run re-executes `bal build` over the whole locked graph against Central — while caching `C:\Users\runneradmin\.ballerina\repositories\`, which nothing reads anymore.

Suggest mirroring the composite action:

```yaml
path: packages/ballerina-language-server/build/ballerina_dependencies
key: ${{ runner.os }}-ballerina-${{ matrix.branch }}-${{ hashFiles('packages/ballerina-language-server/build-config/ballerina_dependencies/**') }}
```

and dropping `restore-keys` here too, for the same reason the action.yml comment gives — a fuzzy match would restore an older dependency set.

.github/workflows/daily-build.yml
```diff
@@ -99,7 +99,7 @@ jobs:
...
```
NipunaRanasinghe: Same stale path as the Windows job below, though it matters less here: `ls-ubuntu-pack` runs `pack -x check`, so no tests. Caching `/home/runner/.ballerina/repositories/` is now just dead weight — worth either repointing it or removing the step.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
```diff
@@ -326,9 +391,10 @@ public static Optional<SemanticModel> getSemanticModelFromWorkspace(Project proj
 
     public static ModuleInfo fetchVersionIfNotExists(ModuleInfo moduleInfo) {
         if (moduleInfo.version() == null) {
-            CentralAPI centralApi = RemoteCentral.getInstance();
-            return new ModuleInfo(moduleInfo.org(), moduleInfo.packageName(), moduleInfo.moduleName(),
-                    centralApi.latestPackageVersion(moduleInfo.org(), moduleInfo.packageName()));
+            String version = FORCE_OFFLINE
+                    ? cachedVersion(moduleInfo.org(), moduleInfo.packageName())
+                    : RemoteCentral.getInstance().latestPackageVersion(moduleInfo.org(), moduleInfo.packageName());
```
NipunaRanasinghe, line 396: This path degrades silently, which I think contradicts the intent stated on `FORCE_OFFLINE` above ("versions that are not cached fail visibly", line 78).

`cachedVersion()` returns `null` when the package isn't provisioned (line 136). That `null` flows straight into the new `ModuleInfo`, and `ModuleInfo.isComplete()` requires `version != null` — so the caller in `FunctionDataBuilder.build()` just skips the block. Net effect: forgetting to add a package to `build-config/ballerina_dependencies` surfaces as a mysteriously empty model rather than an error pointing at the missing dependency, which is the class of confusion this PR is otherwise removing.

Throwing here when `FORCE_OFFLINE && version == null`, with the org/name in the message, would match the documented behaviour and make a missing lock entry self-diagnosing.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/main/java/io/ballerina/architecturemodelgenerator/extension/persist/PersistERModelGeneratorService.java
```diff
@@ -68,8 +69,11 @@ public CompletableFuture<PersistERModelResponse> getPersistERModels(PersistERMod
             AtomicBoolean hasDiagnosticErrors = new AtomicBoolean(false);
             Map<String, Entity> entities = new HashMap<>();
             try {
-                // Persist model file should be loaded as a single file project
-                Project project = SingleFileProject.load(path);
+                // Persist model file should be loaded as a single file project. Tests (ls.test.offline) force offline
+                // resolution so compilation never pulls from Central; production keeps the default behaviour.
+                Project project = Boolean.getBoolean("ls.test.offline")
```
NipunaRanasinghe: This is the only site that reads the property inline instead of going through one of the three constants (`PackageUtil.FORCE_OFFLINE`, `CommonUtil.TEST_OFFLINE`, `RemoteCentral.OFFLINE`). `DataMappingModelConverterUtils` does the equivalent branch via `PackageUtil.isOffline()` — using that here too would keep the string in one place and avoid this site being missed in a future rename.

(No objection to the three constants themselves — they look module-boundary-driven, and the cross-references in their comments make that clear.)

NipunaRanasinghe (changes requested):

.DS_Store
pasindufernando1: Addressed

.github/workflows/daily-build.yml
pasindufernando1: Addressed

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
pasindufernando1: Addressed

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/main/java/io/ballerina/architecturemodelgenerator/extension/persist/PersistERModelGeneratorService.java
pasindufernando1: Addressed

.github/workflows/daily-build.yml
pasindufernando1: Addressed

packages/ballerina-language-server/misc/diagram-util/build.gradle
```diff
@@ -156,6 +156,7 @@ task buildAndCopyLangserverStdlib() {
 test {
     dependsOn copyRuntime
     dependsOn unpackJballerinaTools
+    dependsOn copyStdlibs
```
NipunaRanasinghe, line 159: Why do we need this task here now?

packages/ballerina-language-server/gradle.properties
```diff
@@ -82,43 +82,25 @@ persistCoreVersion=1.9.1
...
+stdlibMessagingVersion=1.0.0
+stdlibSoapVersion=2.3.1
+stdlibYamlVersion=0.8.0
 
 # Ballerina tool versions
 migrateTibcoVersion=1.2.13
-migrateMuleVersion=1.2.11
-
-ballerinaWorkflowVersion=0.7.0
+migrateMuleVersion=1.2.11
```
NipunaRanasinghe: Missing EoF newline

NipunaRanasinghe:
@nipunayf Appreciate your review on LS core changes

packages/ballerina-language-server/gradle.properties
pasindufernando1: Addressed

packages/ballerina-language-server/misc/diagram-util/build.gradle
pasindufernando1: They fill two different repositories the test JVM reads from, for two disjoint package sets:

- bal build (resolveBallerinaDependencies) → the offline central cache (BALLERINA_HOME_DIR), with the ballerinax/* connectors and any non-bundled ballerina/*.
- copyStdlibs (versions from gradle.properties) → the distribution (ballerina.home), with the standard libraries (http, crypto, ai, mime, log, …).

Neither can do the other's job: bal build only writes the central cache and won't pull a package it thinks the distribution bundles — so stdlibs the slim jballerina-tools is missing should be copied via copyStdlibs. Conversely copyStdlibs copies only io.ballerina.stdlib:*-ballerina artifacts, so the ballerinax connectors can only come from bal build.

packages/ballerina-language-server/gradle.properties
```diff
@@ -82,43 +82,25 @@ persistCoreVersion=1.9.1
...
-ballerinaxAiOllamaVersion = 1.1.2
-ballerinaxAiAzureVersion = 1.1.2
-ballerinaxAiAnthropicVersion = 1.1.2
-ballerinaxAiDeepseekVersion = 1.0.4
-ballerinaxAiMistralVersion = 1.1.2
+# Packages bundled in the full Ballerina distribution but NOT in the minimal
+# jballerina-tools distribution the tests run on. bal build resolves these from
+# the installed distribution without caching them, so they are injected into
+# every module's test distribution via the ballerinaStdLibs configuration.
+# Versions must match build-config/ballerina_dependencies/Dependencies.toml.
```
nipunayf, line 89: Can we keep all the version information in one place and auto-generate the rest from that file?

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/test/resources/get_service_init_model/source/sample2/main.bal
```diff
@@ -1,10 +1,10 @@
 import ballerina/mqtt;
 import ballerinax/rabbitmq;
-import ballerinax/salesforce;
+//import ballerinax/salesforce;
```
nipunayf, line 3: Is there a reason this is commented out? If so, can we mention it here or link to the relevant issue?

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
```diff
@@ -68,6 +70,41 @@
...
+     *   <li><b>Tests</b> ({@code FORCE_OFFLINE}): resolve offline with SOFT locking, so a bala's baked transitive
+     *       versions re-resolve to the locked/available versions in the cache instead of demanding the exact baked
+     *       version (which is often not provisioned).</li>
+     *   <li><b>Production</b>: replicate the original two-argument {@code BalaProject.loadProject} behaviour
+     *       ({@code sticky = true}), so production is unchanged.</li>
+     * </ul>
+     */
+    private static BuildOptions balaBuildOptions() {
+        if (FORCE_OFFLINE) {
+            return BuildOptions.builder().setOffline(true).setLockingMode(PackageLockingMode.SOFT).build();
```
nipunayf, line 103: This could break the LS for Ballerina dists <U13, as it was introduced in U13. Can we move this to `BaseCompilerApi` and provide the implementation there?

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
```diff
@@ -68,6 +70,41 @@
 public class PackageUtil {
 
     private static final String BALLERINA_HOME_PROPERTY = "ballerina.home";
+
+    // Test-only switch (set by the Gradle test tasks via -Dls.test.offline=true). When
+    // enabled, package resolution never contacts Ballerina Central: packages resolve
+    // only from the local repositories the build pre-populates, "latest version"
+    // lookups use the cached version instead of the remote one, and versions that are
+    // not cached fail visibly. Absent in production, so production stays online.
+    private static final boolean FORCE_OFFLINE = Boolean.getBoolean("ls.test.offline");
```
nipunayf, line 79: Can’t we use the same flag from `CommonUtils`?

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
```diff
@@ -101,6 +101,11 @@ public final class CommonUtil {
 
     public static final boolean COMPILE_OFFLINE;
 
+    // Set by the Gradle test tasks (-Dls.test.offline=true). When enabled, every site that would contact Ballerina
+    // Central is forced offline so tests resolve only from the build-provisioned Ballerina home. Defaults to false, so
+    // production behaviour is unchanged. See also FORCE_OFFLINE in PackageUtil and OFFLINE in RemoteCentral.
+    public static final boolean TEST_OFFLINE = Boolean.getBoolean("ls.test.offline");
```
nipunayf, line 107: Did we explore setting the flag directly from the `BallerinaLangServer` instance? similar to https://github.com/ballerina-platform/ballerina-language-server/pull/829/changes#diff-a70223dcd22e56e98a40e062bd1d3eb4ca8658b39afe5d7e994d73ae2f16ba2c:~:text=private%20boolean%20packageUtilOffline%3B

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/resources/vector_store_manager/source/vector_stores_with_existing_ai/Ballerina.toml
```diff
@@ -11,4 +11,4 @@ sticky = true
 [[dependency]]
 org="ballerina"
 name="ai"
-version="1.0.0"
+version="1.11.2"
```
nipunayf, line 14: Ideally, we should define the versions in one place. Why do we need to specify this here as well?

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/test/resources/get_design_model/source/project_11/Dependencies.toml
nipunayf, line 1: Can we create a separate issue to remove these Dependencies.toml files?

nipunayf (commented):
Didn’t review each test file, assuming they only contained version changes.

NipunaRanasinghe (approved):
Approving as the pending suggestions are non-blocking. 

@pasindufernando1 please create an GH issue to track the remaining improvements and lets make sure to address those in a separate PR.

packages/ballerina-language-server/gradle.properties
pasindufernando1: Will be fixed with https://github.com/wso2-enterprise/integration-engineering/issues/2141

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/test/resources/get_service_init_model/source/sample2/main.bal
pasindufernando1: Will be fixed with https://github.com/wso2-enterprise/integration-engineering/issues/2141

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
pasindufernando1: Will be fixed with https://github.com/wso2-enterprise/integration-engineering/issues/2141

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
pasindufernando1: Will be fixed with https://github.com/wso2-enterprise/integration-engineering/issues/2141

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
pasindufernando1: Will be fixed with https://github.com/wso2-enterprise/integration-engineering/issues/2141

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/resources/vector_store_manager/source/vector_stores_with_existing_ai/Ballerina.toml
pasindufernando1: Will be fixed with https://github.com/wso2-enterprise/integration-engineering/issues/2141

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/test/resources/get_design_model/source/project_11/Dependencies.toml
pasindufernando1: Will be fixed with https://github.com/wso2-enterprise/integration-engineering/issues/2141

pasindufernando1:
> Approving as the pending suggestions are non-blocking.
> 
> @pasindufernando1 please create an GH issue to track the remaining improvements and lets make sure to address those in a separate PR.

Created. Will be tracked via https://github.com/wso2-enterprise/integration-engineering/issues/2141

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/resources/vector_store_manager/source/vector_stores_with_existing_ai/Ballerina.toml
pasindufernando1: These pins predate this PR — the fixtures already declared `ai`, and this PR only refreshed the version to match the one the test distribution provides. So it is a refresh rather than a new source of truth, and there is no harm in keeping them.

That said, this one is not needed at all (the `import ballerina/ai as _;` in connections.bal is what the fixture exercises), so removed it. The other fixtures that pin `ai` need checking case by case — some expected models encode the pinned version.

PR #669, opened 2026-07-06
Revamp migration prompts with source first strategy

### LINKED ISSUE #1802: Revamp migration too AI enhancement prompts (source-first strategy)
### Current Limitation

The current Stage 1 prompt is output-driven: it scans Ballerina files for // TODO / // FIXME comments and reads source files reactively to resolve them. This is fragile because the rule-based tool silently drops constructs without leaving a TODO (e.g. RAML specs, DataWeave files with complex logic, TIBCO process variables, property files). The agent ends up verifying the migration tool's honesty rather than verifying the migration itself.

The new strategy is source-driven: the source project is the ground truth. The Ballerina output is a skeleton/scaffold that guides structure but does not define correctness. The agent reads every source construct and verifies it is correctly represented in Ballerina, fixing gaps whether or not a TODO exists. The goal is complete migration (everything in source is in Ballerina), not idiomatic Ballerina (that is a separate refactoring stage).

### Suggested Improvement

Revamp AI enhancement prompt using source-first strategy to get better results

### Version

_No response_

lochana-chathura:
$subject.

Fixes https://github.com/wso2/product-integrator/issues/1802

## Summary

- Revamped migration prompts to use source-first verification.
- Added MuleSoft and TIBCO source context across migration requests, settings, workflows, and prompts.
- Added six migration stages for source inventory, implementation, diagnostics, tests, validation, and refactoring.
- Improved coverage checks for source constructs, configuration, and migration gaps without generated TODO or FIXME comments.
- Updated multi-package progress tracking to use the generated stage count.

### DIFF
--- a/packages/ballerina-extension/src/features/ai/migration/orchestrator.ts
+++ b/packages/ballerina-extension/src/features/ai/migration/orchestrator.ts
@@ -40,6 +40,7 @@ import {
     AI_MIGRATION_DIR,
     ActiveMigrationSessionLocal,
     EnhanceTomlData,
+    MigrationContext,
     MIGRATION_PROJECT_ROOT_KEY,
     PackageEnhancementResult,
     PENDING_ENHANCEMENT_TTL_MS,
@@ -82,6 +83,7 @@ export function readEnhanceToml(projectRoot: string): EnhanceTomlData | null {
         const currentStageMatch = content.match(/currentStage\s*=\s*(\d+)/);
         const multiProjectMatch = content.match(/multiProject\s*=\s*(true|false)/);
         const keepStructureMatch = content.match(/keepStructure\s*=\s*(true|false)/);
+        const sourcePlatformMatch = content.match(/sourcePlatform\s*=\s*"(mule|tibco)"/);
 
         // Parse completedPackages array
         const completedPackagesMatch = content.match(/completedPackages\s*=\s*\[([^\]]*)\]/);
@@ -102,6 +104,7 @@ export function readEnhanceToml(projectRoot: string): EnhanceTomlData | null {
             currentStage: currentStageMatch ? parseInt(currentStageMatch[1], 10) : undefined,
             multiProject: multiProjectMatch ? multiProjectMatch[1] === "true" : undefined,
             keepStructure: keepStructureMatch ? keepStructureMatch[1] === "true" : undefined,
+            sourcePlatform: sourcePlatformMatch?.[1] as 'mule' | 'tibco' | undefined,
         };
     } catch {
         return null;
@@ -121,6 +124,7 @@ export function writeEnhanceToml(
     currentStage?: number,
     multiProject?: boolean,
     keepStructure?: boolean,
+    sourcePlatform?: 'mule' | 'tibco',
 ): void {
     const dir = path.join(projectRoot, AI_MIGRATION_DIR);
     if (!fs.existsSync(dir)) {
@@ -141,8 +145,9 @@ export function writeEnhanceToml(
     if (currentStage !== undefined) {
         content += `currentStage = ${currentStage}\n`;
     }
-    // When multiProject / keepStructure are not explicitly provided, preserve existing values from disk.
-    const existing = multiProject === undefined || keepStructure === undefined ? readEnhanceToml(projectRoot) : undefined;
+    // When optional fields are not explicitly provided, preserve existing values from disk.
+    const needExisting = multiProject === undefined || keepStructure === undefined || sourcePlatform === undefined;
+    const existing = needExisting ? readEnhanceToml(projectRoot) : undefined;
     const effectiveMultiProject = multiProject !== undefined ? multiProject : existing?.multiProject;
     if (effectiveMultiProject !== undefined) {
         content += `multiProject = ${effectiveMultiProject}\n`;
@@ -150,6 +155,10 @@ export function writeEnhanceToml(
     const effectiveKeepStructure = keepStructure !== undefined ? keepStructure : existing?.keepStructure;
     if (effectiveKeepStructure) {
         content += `keepStructure = true\n`;
+    }
+    const effectiveSourcePlatform = sourcePlatform !== undefined ? sourcePlatform : existing?.sourcePlatform;
+    if (effectiveSourcePlatform) {
+        content += `sourcePlatform = "${effectiveSourcePlatform}"\n`;
     }
     fs.writeFileSync(filePath, content);
 }
@@ -223,6 +232,19 @@ export function getMigrationSourcePathForProject(projectRoot: string): string | 
         return data.sourcePath;
     }
     return undefined;
+}
+
+/**
+ * Builds a `MigrationContext` from the persisted toml for the given project root.
+ * This is the single place that translates toml fields into prompt parameters —
+ * all stage call sites use this instead of passing loose flags.
+ */
+export function buildMigrationContext(projectRoot: string): MigrationContext {
+    const data = readEnhanceToml(projectRoot);
+    return {
+        sourcePlatform: data?.sourcePlatform ?? 'unknown',
+        keepStructure: data?.keepStructure ?? false,
+    };
 }
 
 /**
@@ -772,7 +794,7 @@ export async function runMigrationAgent(): Promise<void> {
                 const fullPkgPath = path.join(projectRoot, pkgRelPath);
                 const pkgName = readPackageName(fullPkgPath) ?? pkgRelPath;
                 const manifest = buildCrossPackageManifest(projectRoot, packagePaths, pkgRelPath);
-                const stages = getPerProjectEnhancementStages(pkgName, pkgRelPath, pkgIdx, packagePaths.length, manifest);
+                const stages = getPerProjectEnhancementStages(pkgName, pkgRelPath, pkgIdx, packagePaths.length, manifest, buildMigrationContext(projectRoot));
                 if (!resumeInjected) {
                     injectResumePreamble(projectRoot, stages);
                     resumeInjected = true;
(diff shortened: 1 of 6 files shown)


lochana-chathura (commented):
LGTM

KavinduZoysa (dismissed):

NipunaRanasinghe:
@coderabbitai review

packages/ballerina-extension/src/features/ai/migration/prompts.ts
```diff
@@ -241,19 +375,17 @@ ${keepStructure ? `### Original Source File Structure Preserved
 
 This project was migrated with **\`--keep-structure\`** enabled. Each \`.bal\` file corresponds
 to one original source file. The source filename and its directory path are encoded into the \`.bal\`
-filename — for example, MuleSoft's \`foo/bar.xml\` becomes \`foo_bar.bal\`; TIBCO may use a similar
-but different encoding. **Do not assume the \`.bal\` filename exactly matches the source filename.**
+filename. For ${platformName === 'MuleSoft Mule 3/4' ? "MuleSoft: `foo/bar.xml` becomes `foo_bar.bal`" : "TIBCO: `foo/Bar.bwp` becomes a similarly named `.bal` file"}.
```
NipunaRanasinghe: This branches on the display string (`platformName === 'MuleSoft Mule 3/4'`) rather than `context.sourcePlatform`. `getPlatformName('unknown')` returns `'MuleSoft Mule 3/4 / TIBCO BusinessWorks'`, which fails this `===` check, so the ternary falls into the TIBCO (`.bwp`) branch.

Per `MigrationContext` in `types.ts`, `'unknown'` is exactly the value every migration project created before this PR has (a TOML that predates the `sourcePlatform` field). So every pre-existing migration — MuleSoft or TIBCO alike — is currently told to expect `.bwp`-style filenames here.

Suggested fix (branch on `context.sourcePlatform` directly, same as `getPlatformName`/`getPlatformExpertise` already do):
```ts
const filenameEncodingHint = context.sourcePlatform === 'mule'
    ? "MuleSoft: `foo/bar.xml` becomes `foo_bar.bal`"
    : context.sourcePlatform === 'tibco'
    ? "TIBCO: `foo/Bar.bwp` becomes a similarly named `.bal` file"
    : "MuleSoft: `foo/bar.xml` becomes `foo_bar.bal`; TIBCO: `foo/Bar.bwp` becomes a similarly named `.bal` file";
```
then reference `filenameEncodingHint` in the template literal instead of the inline ternary.

packages/ballerina-extension/src/features/ai/migration/prompts.ts
```diff
@@ -421,75 +664,123 @@ Your sole focus is achieving **zero error-level compilation diagnostics** across
...
 - Output the final diagnostics count (should be 0 errors).
 - List any best-effort approximations you made to fix diagnostics.
 
-Then stop. Stage 3 will handle test files.`;
+Then stop. Stage 4 will handle test files.`;
 }
 
 // ---------------------------------------------------------------------------
-// Stage 3 — Test Refinement
+// Stage 4 — Test Refinement
```
NipunaRanasinghe: Section comment still says "Test Refinement" — the actual stage name everywhere else (including two lines below, in this same function's returned heading) is "Test Migration & Gap Coverage". Cosmetic only, but worth a one-line fix for the next reader:
```ts
// Stage 4 — Test Migration & Gap Coverage
```

packages/ballerina-extension/src/features/ai/migration/prompts.ts
```diff
@@ -100,40 +110,51 @@ their exports are correct as declared.`;
...
+            name: `[${packageName}] Stage 3 — Zero Compilation Diagnostics`,
+            prompt: shared + "\n\n" + getStage3Prompt(context) + "\n\n" + CROSS_PACKAGE_ISOLATION_NOTE,
             agentLimits: { maxSteps: 100, maxOutputTokens: 16384 },
         },
         {
-            name: `[${packageName}] Stage 3 — Test Refinement`,
-            prompt: shared + "\n\n" + getStage3Prompt(),
-            agentLimits: { maxSteps: 100, maxOutputTokens: 16384 },
+            name: `[${packageName}] Stage 4 — Test Migration & Gap Coverage`,
+            prompt: shared + "\n\n" + getStage4Prompt(context),
```
NipunaRanasinghe: Stages 3, 5, and 6 append `CROSS_PACKAGE_ISOLATION_NOTE` (lines 140, 150, 155); Stage 2 and this Stage 4 entry don't, even though all six stages run under the same isolated per-package compilation. Stage 4 specifically instructs the agent to run diagnostics and require zero errors across all test files — if a test references a sibling package, the agent hits an expected `BCE2003`-style error here with no instruction to ignore it. Worth appending the note to both:
```ts
prompt: shared + "\n\n" + getStage2Prompt(context) + "\n\n" + CROSS_PACKAGE_ISOLATION_NOTE,
...
prompt: shared + "\n\n" + getStage4Prompt(context) + "\n\n" + CROSS_PACKAGE_ISOLATION_NOTE,
```

packages/ballerina-extension/src/features/ai/migration/prompts.ts
```diff
@@ -298,106 +436,216 @@ Process one package at a time. Read other packages for context but only modify t
...
+**Phase A: Load your work plan from Stage 1**
+
+The Stage 1 inventory is in your context (from the previous stage output). Use it as your work list.
+Process source files in this priority order:
+1. ❌ Missing — constructs the tool silently dropped (highest priority)
+2. ⚠️ Partial — constructs with \`// TODO\` or \`// FIXME\` markers
+3. ✅ Covered — verify correctness of already-translated constructs (spot-check, don't skip)
+
+If Stage 1 output is not in your context (e.g. this is a resumed session), call
+\`migration_source_list(".")\` to re-establish the source structure, then proceed.
```
NipunaRanasinghe, line 573: This premise doesn't hold on a normal run — nothing carries Stage 1's output into Stage 2's context, so this is always the fallback path.

`runStagesForPackage` (`orchestrator.ts:595`) constructs a new executor per stage:

```ts
params: { usecase: stage.prompt, fileAttachmentContents: [], isPlanMode: false },
...
await new AgentExecutor(config).run();
```

with the docstring at `:589` stating it outright — *"Each stage gets a fresh `AgentExecutor` (fresh context window)."* The only cross-stage channels are `injectResumePreamble`, which touches `stages[0]` only (`:525`), and a same-stage partial transcript on resume (`:622`). Neither forwards stage *i*'s final text to stage *i+1*. `chatStorage` is `undefined` whenever `fromAIChat` is false, which is every migration-panel call site (`:813`, `:842`, `:878`), and `:251` documents history seeding as now a no-op.

The consequence is quiet rather than broken: Stage 1 spends up to 50 steps building the inventory, it's discarded, and Stage 2 takes the line 572 branch and re-derives structure from `migration_source_list(".")`. What's lost is everything Stage 1 uniquely produces — the per-file ✅/⚠️/❌ coverage verdicts and the ranked `GAPS REQUIRING IMPLEMENTATION IN STAGE 2` list. Since the source-first strategy is the point of the PR, it's worth making this deterministic rather than leaving it to a fallback framed as a resume-only edge case (Stage 4's Preparation step 1 has the same dependency).

Two ways out:

1. **Persist it.** Have Stage 1 write the inventory to a file under `AI_MIGRATION_DIR` and have Stages 2 and 4 read it. This currently conflicts with line 498 (*"this is working notes for Stage 2, NOT a file"*) and the read-only framing, so it'd need that carve-out — but it survives resume for free.
2. **Inject it.** Have the orchestrator capture Stage 1's transcript and prepend it to the Stage 2 and Stage 4 prompts. The machinery already exists: `transcriptWriter.readStageTranscript` plus the `injectResumePreamble` prompt-prefix pattern.

Either way, the fallback is worth keeping as a genuine backstop — just not as the default path.

packages/ballerina-extension/src/features/ai/migration/orchestrator.ts
```diff
@@ -1404,8 +1428,8 @@ export async function runWizardMigrationEnhancement(): Promise<void> {
                             packageIndex: packagePaths.length,
                             totalPackages: packagePaths.length,
                             packageName: "",
-                            stageOffset: packagePaths.length * 4,
-                            totalStagesOverall: packagePaths.length * 4 + 1,
+                            stageOffset: packagePaths.length * stagesPerPackage,
+                            totalStagesOverall: packagePaths.length * stagesPerPackage + 1,
```
NipunaRanasinghe, line 1432: Deriving the count instead of hardcoding `4` is the right move, but `stagesPerPackage` can still be `0` here, and the value it replaced never could be.

It's declared `0` at line 1336 and only assigned at line 1367 — inside the loop body, after the completed-package guard at line 1355:

```ts
if (completedPackages.has(pkgRelPath)) {
    eventHandler({ ... });
    results.push({ packagePath: pkgRelPath, success: true });
    continue;
}
```

When every package is in `completedPackages` but workspace validation hasn't run yet — the resume case this block exists to serve — the guard fires on every iteration, the assignment never executes, and validation reports `stageOffset: 0`, `totalStagesOverall: 1`. The wizard progress bar drops from "N of 13" to "0 of 1" at the last step of a resumed run.

Worth noting the earlier suggestion to use `stages.length` directly wouldn't work: it's `const`-scoped to the loop body and unreachable from here. `stagesPerPackage` is the right shape, it just needs a value that doesn't depend on the loop having run — computing the count once before the loop from the same generator, or a small `getEnhancementStageCount(context)` helper alongside the stage factories, both do it.

Secondary point on the same assignment: it captures the *last* processed package's count and multiplies by `packagePaths.length`. That's correct today since every package gets the same six stages, but `getPerProjectEnhancementStages` takes `packageIndex` and `totalPackages`, so a future per-package variation would make this silently wrong. Computing the count once up front sidesteps that too.

NipunaRanasinghe (approved):
Approved with some minor comments. @lochana-chathura @KavinduZoysa its okay to address them in a separate follow up PR

PR #670, opened 2026-07-06
Support creating workflow activities from connections; generate workflows and activities in dedicated files

hasithaa:
## Purpose

Add a **Create Activity from a Connection** flow to the workflow Call Activity panel, and harden how workflow and activity source is generated so that adding one never corrupts the other.

A user picks an existing connection, selects one of its actions (remote/resource call), and fills a small form. The backend derives the activity signature from the action and generates a `@workflow:Activity` wrapper. The activity is then callable via `ctx->callActivity(...)` like any other activity.

## What's included

**Create activity from a connection (wizard)**
- Pick a connection → pick an action → a signature form driven by the LS analysis of the action.
- Each derived parameter is a checkbox: **checked** it becomes an input to the activity; **unchecked** it takes a fixed value baked into the action call. Required-parameter fixed values use a **type-aware** input (a `string` opens the text box with a type pill; other types keep the expression editor), grouped under their checkbox.
- Optional **"Connection as parameter"**: expose the connection as the activity's first parameter (built-in-activity style) instead of closing over the module-level connection.
- Default activity name is the **module-prefixed action** (e.g. `httpPost`, `mysqlQuery`); the generated `@display` label matches the name.
- Single **Create Activity** action that creates the activity and returns to the activity list (the two-step Next/stepper flow was dropped).

Example generated activity (`functions.bal`):

```ballerina
# Create a new resource or submit data to a resource for processing.
# + path - Resource path
# + return - Type of the variable
@workflow:Activity
@display {label: "httpPost", iconPath: "https://.../ballerina_http.png"}
isolated function httpPost(string path) returns json|error {
    json result = check httpClient->post(path, {});
    return result;
}
```

**Dedicated files for workflows and activities (safety by design)**
- New `@workflow:Workflow` functions are generated in **`workflow.bal`** (created if absent); activities are generated in **`functions.bal`**. Because a workflow and its activities live in separate files, the imports one needs (e.g. an `http:Client` parameter → `import ballerina/http`) never shift the other's line positions — the earlier "Cannot find a line with the character offset" failures are prevented structurally.

**Workflow function signature**
- New workflow functions declare `workflow:Context` as the mandatory first parameter, followed by the input type when set — e.g. `function order(workflow:Context ctx, OrderInput input) returns error?`.

**Workflow creation form**
- Consistent copy with the Add-Artifact card; the form asks only for the **name** and **input type** (the description field was removed as it is not needed to define a workflow).

## Approach

- **LS**: `ActionSignatureAnalyzer` + `ConnectionActionProvider` analyze an action (`workflowManager/analyzeActivityAction`); `ActivityGenerator` emits the wrapper (`workflowManager/genActivity`) into `functions.bal`. `SourceBuilder` routes `WORKFLOW` → `workflow.bal` and activities → `functions.bal`; `WorkflowBuilder` emits the `workflow:Context` parameter. A new `ACTIVITY_CALL` node/search surfaces the created activities.
- **RPC**: the two new APIs are exposed through the extended language client and the bi-diagram RPC layer; the extension applies the returned text edits.
- **UX**: `NewActivityFromConnection` drives the wizard; the shared side-panel `Form` gains a validation-reset on removed fields and an indent option; `FlowDiagram` re-resolves the insertion point from a fresh flow model after creation.

## Tests

- LS: `AnalyzeActivityActionTest`, `GenActivityTest` (remote/resource actions, no-description case, connection-as-parameter → asserts the `http` import lands in `functions.bal`, and routing regardless of the requested file). `SourceGeneratorTest` `workflow_to_source*` goldens cover new-workflow generation in `workflow.bal` with the `workflow:Context` parameter.


hasithaa:
## Update: visualize the connection on activity calls

Added a distinct node kind so activity calls that use a connection are shown with a connection arrow in the diagram (like remote action calls), instead of a plain activity node.

**New node kind `CONNECTION_ACTIVITY_CALL`** — applies to:
- Built-in activities (`callRestAPI` / `callSoapAPI` / `sendEmail`), which always wrap a connection.
- User activities whose **first parameter is a connection client** (exactly what the "Create Activity from a Connection" flow generates).

Regular activities (no connection parameter) stay `ACTIVITY_CALL`.

### LS
- `CodeAnalyzer` now detects the connection first-parameter of a called activity (`WorkflowUtil.resolveConnectionClass`) and stores it under `CONNECTION_KEY` with connector metadata + icon, rather than as a plain `$connection` data argument. Built-in activity calls are marked `CONNECTION_ACTIVITY_CALL` too.
- `ActivityCallBuilder` builds a connection selector for connection-backed user activities and preserves the node kind through `build()` (which re-runs `setConstData()` — this is why the kind previously reset to `ACTIVITY_CALL`). `Codedata.Builder` gained a `node()` getter for that.
- This also fixes a latent bug: a connection parameter was previously keyed `$connection` and would have been emitted as an invalid `$connection:` argument.

### Diagram
- `CONNECTION_ACTIVITY_CALL` routes to the API-call node (`NodeFactoryVisitor` + `SizingVisitor`), so it renders with the same connection arrow + endpoint bubble as remote action calls, using the connection variable as the label and the connector icon in the bubble.

### Tests
- Built-in activity node-template baselines updated to `CONNECTION_ACTIVITY_CALL`.
- New flow-model analysis fixture (using a **local** client class, so it's version-drift-free) asserting a user activity with a connection first-parameter is reconstructed as `CONNECTION_ACTIVITY_CALL` carrying the connection.
- Workflow node-template, source-generation, and the activity fixtures all pass; frontend (`ballerina-core`, `bi-diagram`) typechecks.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/ActivityWizardSteps.tsx
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * The two-step progress bar of the create-activity-from-connection flow, shown after an action is
+ * selected: create the activity, then wire the activity call. (The connection/action selection uses
+ * the shared connection list, so no step bar is shown there.)
+ */
+export function ActivityWizardSteps({ activeStep }: { activeStep: 1 | 2 }): JSX.Element {
+    return (
+        <StepsContainer>
+            {STEP_LABELS.map((label, index) => (
+                <Step key={label} active={activeStep === index + 1}>
```
madushajg: Instead of implementing this ourselves, let's use the existing Stepper component from the UI Toolkit to keep the experience consistent and avoid duplicating functionality.
https://github.com/wso2/vscode-extensions/blob/main/workspaces/common-libs/ui-toolkit/src/components/Stepper/Stepper.tsx

packages/ballerina-visualizer/src/views/BI/FlowDiagram/NewActivityFromConnection.tsx
```diff
@@ -0,0 +1,647 @@
...
+                <>
+                    <ActivityWizardSteps activeStep={1} />
+                    <ArtifactForm
+                        preserveFieldOrder={true}
+                        fileName={fileName}
+                        targetLineRange={{ startLine: { line: 0, offset: 0 }, endLine: { line: 0, offset: 0 } }}
+                        fields={fields}
+                        onSubmit={handleActivitySubmit}
+                        onBack={() => setPanelView(PanelView.CONNECTION_LIST)}
+                        submitText={"Create Activity"}
```
madushajg, line 704: Here the user will be navigated to the next step where it calls the activity. However there can be users who  only create the activity this time and call it later. For such users it is better to have an option 'Create Activity' along with 'Call Activity'.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/ActivityWizardSteps.tsx
hasithaa: Due to the complexity, I have removed the multi-step path.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/NewActivityFromConnection.tsx
hasithaa: Due to the complexity, I have removed the multi-step path. Now it is always Create Activity.

hasithaa: Due to the complexity, I have removed the multi-step path. Now it is always Create Activity.

madushajg (dismissed):
LGTM

madushajg:
@coderabbitai review

madushajg (dismissed):
LGTM

hasithaa:
This is the final DX.

https://github.com/user-attachments/assets/6c7e4d01-a406-485c-b8a2-3496e8e2ea4f

madushajg (approved):

PR #671, opened 2026-07-06
Fix MCP tool Configure error with @mcp:Tool description annotation

dan-niles:
Fixes https://github.com/wso2/product-integrator/issues/1832

Duplicate of ballerina-platform/ballerina-language-server#938 for this repository (the language-server code now lives under `packages/ballerina-language-server`).

Clicking **Configure** on an MCP tool generated with `@mcp:Tool { description: "..." }` crashed the visualizer (`Field types are not defined`). The parent annotation handler created an `annotTool` property with null `types`, and `parseDocumentation` only reads doc comments, so `toolDescription` was never populated.

`McpFunctionBuilder` now hoists a string-literal `description` from the annotation into `toolDescription` on read, while rewriting `annotTool.value` to preserve any other fields (`schema` kept byte-for-byte). It also seeds `annotTool.types`, sets the function `kind` to `REMOTE` so `updateFunction` dispatches to the MCP builder, and inserts a fresh doc-comment block on save when the source had none.

Fixtures added for annotation-only, annotation + schema, and doc-comment forms (read + update).

Updated MCP tool handling so annotation-based tools round-trip correctly with doc-comment-based tools.

- Extracts `@mcp:Tool` descriptions into `toolDescription` while preserving fields such as `schema`
- Marks MCP tools as remote functions and initializes MCP tool types
- Generates or updates doc comments when source documentation is missing
- Adds fixture coverage for annotation-only, annotation-with-schema, and doc-comment tools


dan-niles:
Closing this since this issue is no longer reproducible in 5.1.0 alpha 3 and this implementation uses `McpFunctionBuilder` which is no longer available in LS.

https://github.com/user-attachments/assets/3b5229c1-36ab-4c18-b7d1-bb6f6aaab102

PR #685, opened 2026-07-14
Recover in-progress Copilot chat when the panel is closed and reopened

### LINKED ISSUE #1862: Chat doesn't recover properly if you close and reopen the panel while it's still generating
### Current Limitation

If you close the Copilot panel (or switch away and it gets torn down) while the agent is still working, the generation keeps running in the background, but the UI doesn't reconnect to it cleanly when you come back.

A few things go wrong. When you reopen, whatever the agent produced while the panel was closed is just gone from the chat. You either see a blank reply or a new half-finished bubble that picks up from wherever the stream happens to be. In some cases the reply is lost for good, even after the run finishes, because the visible chat history only gets saved when the panel itself is open to catch it. On top of that, the "generating..." state doesn't reflect reality. A run can still be going while the panel looks idle, and if the "finished" signal is ever missed, the panel can sit there spinning forever with no way out except reloading.

The MI Copilot handles this situation well today. BI doesn't really have the machinery for it, so recovery is mostly accidental rather than something we designed.

### Suggested Improvement

Make an in-progress run something the UI can reliably pick back up. A few pieces that get us there:

- Save the visible chat history on the backend as the run progresses, so a reply is never lost just because the panel wasn't open to record it.
- Give the UI a way to ask "is a run happening, and what did I miss?" when it reopens, and let it replay those missed steps so the conversation comes back complete instead of half-there.
- Reflect the real state on reopen (show "generating" if it actually is), and add a lightweight safety net so a dropped "finished" message can't leave the panel stuck spinning.

### Version

_No response_

SasinduDilshara:
### Issue
If you close the Copilot panel while the agent is still generating, the run keeps
running in the extension host but the UI can't reconnect. On reopen the reply
produced while the panel was closed is missing, the "generating…" state doesn't
reflect reality, a pending clarifying question is silently skipped, and a missed
"finished" signal can leave the panel spinning forever.

### Fix
- Record each generation's UI event stream in an in-memory journal in the extension
  host, and add a `getActiveGenerationState` RPC so a reopened panel can ask "is a run
  in progress, and what did I miss?" and replay the missed events through the existing
  renderer — bringing the conversation back complete and correctly ordered.
- Reflect the real run state on reopen (show the spinner only when a run is genuinely
  active) and add a lightweight watchdog so a missed terminal signal can't strand the
  spinner.
- Keep in-chat prompts (clarify / approvals) alive when the panel closes so the run
  pauses instead of being skipped, and re-surface the still-pending prompt on reopen.
- Queue live events while the journal is replaying and drain them in order afterward,
  so live streaming can't interleave with replay; and stop posting notifications to a
  disposed panel (the journal already captured them), which removes the "Webview is
  disposed" log flood.

Fixes https://github.com/wso2/product-integrator/issues/1862

### DIFF
--- a/packages/ballerina-core/src/rpc-types/ai-panel/index.ts
+++ b/packages/ballerina-core/src/rpc-types/ai-panel/index.ts
@@ -43,6 +43,8 @@ import {
     CheckpointInfo,
     AbortAIGenerationRequest,
     UsageResponse,
+    ActiveGenerationState,
+    GetActiveGenerationStateRequest,
     OpenFileDiffRequest,
     WebToolApprovalRequest,
     CompactConversationRequest,
@@ -136,6 +138,7 @@ export interface AIPanelAPI {
     updateChatMessage: (params: UpdateChatMessageRequest) => Promise<void>;
     getActiveTempDir: () => Promise<string>;
     getUsage: () => Promise<UsageResponse | undefined>;
+    getActiveGenerationState: (params: GetActiveGenerationStateRequest) => Promise<ActiveGenerationState>;
     openFileDiff: (params: OpenFileDiffRequest) => void;
     approveWebTool: (params: WebToolApprovalRequest) => Promise<void>;
     declineWebTool: (params: WebToolApprovalRequest) => Promise<void>;
--- a/packages/ballerina-core/src/rpc-types/ai-panel/interfaces.ts
+++ b/packages/ballerina-core/src/rpc-types/ai-panel/interfaces.ts
@@ -543,6 +543,12 @@ export interface ActiveGenerationState {
      * these through its normal renderer to rebuild everything it missed while it was closed.
      */
     events?: ChatNotify[];
+    /**
+     * Request ids of interactive prompts (clarify/approval/etc.) that are still awaiting a user
+     * response. On reopen the panel re-surfaces the journaled prompt for these (so the question can
+     * be answered) and skips prompts already resolved earlier in the run.
+     */
+    pendingRequestIds?: string[];
 }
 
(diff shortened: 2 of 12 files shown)


RNViththagan:
One design thought, non-blocking: this PR treats **closing** the panel as "keep the run going in the background and reconnect later." IMO closing (the `X`) usually signals the opposite intent — the user wants to *terminate* the process, not detach from it. Overloading close to mean "continue in background" can be surprising (the run keeps consuming quota/compute after the user thinks they've stopped it).

If we want to support a genuine "keep it running in the background" flow, I'd rather make that an explicit, separate action — e.g. **minimize / collapse the side panel** — and keep close as a real terminate. That keeps the two intents distinct:
- **Minimize (collapse panel):** run continues, reconnect + replay on expand — exactly what this PR enables.
- **Close (`X`):** abort the run and tear down.

Not a blocker for this PR; more of a follow-up UX improvement we could track separately.

IsuruMaduranga:
Closing in favor of #696, which now consolidates both fixes for wso2/product-integrator#1862. The strongest parts of this PR's design were ported there with credit — the interactive-prompt lifecycle (`cancelOnPanelClose` + `pendingRequestIds` with replay-skip of resolved prompts), the replay/live race gate + queue, the turn-aware `ensureAssistantMessage` fix, the retryable buffer-until-persisted recovery for runs that finish while the panel is closed, and the disposed-panel send guard — combined with #696's seq-based incremental polling (recovers missed events, not just the spinner), bounded buffers, and per-event `generationId` stamping. Thanks @SasinduDilshara — the prompt-lifecycle handling here was the most complete treatment of that problem.

PR #686, opened 2026-07-14
Persist Copilot chat history incrementally (append-only JSONL)

SasinduDilshara:
### Issue

BI Copilot rewrote the entire chat thread to disk on every mutation: it re-serialized the whole conversation into a single `thread.json` and rewrote the full file, synchronously, on each new message / tool step / turn update. Because the file grows with the conversation, per-turn cost scaled with the total history (O(n²) over a session) and blocked the extension host — causing noticeable lag in long, tool-heavy chats.

### Fix

Wire `ChatStateStorage` to the new append-only JSONL store API (see wso2/vscode-extensions#2418), so each mutation persists a single record instead of rewriting the whole thread:

- **Hot paths append one record** — persisting after every agent step (`updateGeneration`), plus add / remove / review-state / restore-to-checkpoint.
- **Thread creation and the per-turn checkpoint flush stay full compacted writes** (they also act as natural compaction points).
- **Load path is unchanged** — the store replays the log transparently and returns the same `PersistedThread`, so nothing downstream changes.

Bumps the `copilot-utilities` submodule to the engine change.

Depends on wso2/vscode-extensions#2418 (merge that first).

Fixes https://github.com/wso2/product-integrator/issues/1861

### DIFF
--- a/packages/ballerina-extension/src/views/ai-panel/chatStateStorage.ts
+++ b/packages/ballerina-extension/src/views/ai-panel/chatStateStorage.ts
@@ -387,6 +387,7 @@ export class ChatStateStorage {
                 }
 
                 // Ensure at least the default thread exists
+                let createdFallbackDefault = false;
                 if (threads.size === 0) {
                     const defaultThread: ChatThread = {
                         id: 'default',
@@ -396,6 +397,7 @@ export class ChatStateStorage {
                         updatedAt: Date.now(),
                     };
                     threads.set('default', defaultThread);
+                    createdFallbackDefault = true;
                 }
 
                 workspaceState = {
@@ -405,6 +407,13 @@ export class ChatStateStorage {
                 };
 
                 this.storage.set(projectRootPath, workspaceState);
+                // Establish the thread log (head + meta) for a freshly synthesized
+                // default thread so its name/identity survive reload. Without this,
+                // the first mutation would only append a generation record and the
+                // thread name would be lost on replay.
+                if (createdFallbackDefault) {
+                    this.flushThread(projectRootPath, 'default');
+                }
                 console.log(`[ChatStateStorage] Restored workspace from disk: ${projectRootPath} with ${threads.size} thread(s)`);
             } else {
                 // Create fresh default thread
--- a/packages/ballerina-extension/src/views/ai-panel/chatStateStorage.ts
+++ b/packages/ballerina-extension/src/views/ai-panel/chatStateStorage.ts
@@ -285,6 +285,56 @@ export class ChatStateStorage {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


IsuruMaduranga:
This PR (together with the engine change in wso2/vscode-extensions#2418) ends up writing the same messages to thread.jsonl several times within a single turn.

updateGeneration runs after every agent step, and each time it re-appends the whole generation (the entire modelMessages array so far) under the same gen.id. So a 3 step turn writes something like:

```
{"t":"head",...}
{"t":"meta",...}
{"t":"gen","gen":{"id":"msg-42","modelMessages":[user, a1, t1]}}                 // after step 1
{"t":"gen","gen":{"id":"msg-42","modelMessages":[user, a1, t1, a2, t2]}}          // after step 2
{"t":"gen","gen":{"id":"msg-42","modelMessages":[user, a1, t1, a2, t2, a3, t3]}}  // after step 3
```

A message from step k of an S step turn gets written S-k+1 times, so we're writing roughly O(S^2) bytes per turn, and it's all synchronous on the extension host. compactThread cleans it up afterwards, but that runs synchronously too (read, replay, rewrite) every ~64 appends or on load, and it only exists to undo the duplication we just wrote.

Instead of duplicating and then compacting, can we just append the new messages each step? Keep the small mutable fields (uiResponse, reviewState, etc.) in a lightweight header record and write each message once:

```
{"t":"gen","gen":{"id":"msg-42","userPrompt":..., "reviewState":..., "metadata":...}}
{"t":"msg","genId":"msg-42","message":{user}}
{"t":"msg","genId":"msg-42","message":{a1}}
{"t":"msg","genId":"msg-42","message":{t1}}
```

On load you group the msg records by genId in order. Each step only writes what's new, the heavy records are never rewritten, and there's nothing left to compact.

SasinduDilshara:
Good catch — you're right, and I've reworked it to the message-level design you described.

The store now writes each message once:
- `gen` — lightweight generation header (everything except `modelMessages`), appended only when a header field actually changes.
- `msg` — one model message, appended as it arrives.
- `msgs` — replace the whole message list, but only when the messages are rewritten rather than extended (e.g. server-side context compaction); detected cheaply via a length / last-message change so the normal per-step path never hits it.

`appendGeneration` still takes the full generation (so `ChatStateStorage` is unchanged), but internally diffs against per-generation tracking and appends only the delta. An S-step turn now costs O(S) to persist instead of O(S²), and there's no per-step duplication left for compaction to undo (compaction now only reclaims superseded headers and tombstones).

Engine change is in wso2/vscode-extensions#2418 (updated); this PR just bumps the submodule to it. Added tests asserting each message is written exactly once, the header is written only when it changes, and the reset path.

PR #689, opened 2026-07-14
Show workflow event connections on the project overview page

### LINKED ISSUE #1887: [Bug]: Navigation from built-in workflow activity nodes on the project overview does not work
### Description

The project overview page renders the activities called by a durable workflow as clickable nodes, and clicking a node navigates to the activity function's source. For **built-in activities** (`Call REST API`, `Call SOAP API`, `Send Email (SMTP)` from the `ballerina/workflow.activity` module), this navigation does not work: the activity's source location points into the pulled module (bala), not into the project, so the click either does nothing or fails to open a view.

Navigation should be disabled for nodes whose source lives outside the current project instead of offering a click affordance that goes nowhere.

### Steps to Reproduce

1. Create a durable workflow that calls a built-in activity, e.g. `ctx->callActivity(activity:callRestAPI, {...}, retryPolicy = workflow:ManualRetry)`.
2. Open the project overview page — the built-in activity appears as a node next to the workflow.
3. Click the activity node → navigation fails (code lives in the `workflow.activity` module, not the project).

### Fix

Addressed in wso2/ballerina-vscode#689 as part of the overview UX rework: the activities section is removed from the overview — only the workflow (events, human tasks) and the connections derived from its activities are rendered — which removes the broken navigation surface.

### Version

Ballerina Integrator (workflow overview preview, wso2/ballerina-vscode#689)

hasithaa:
## Purpose

Part of wso2/product-integrator#1666
Resolves wso2/product-integrator#1887

Workflows can be started by any trigger and receive data from any piece of code, but the project overview page did not show any of these relationships — workflow nodes rendered as isolated boxes with no edges. This PR derives the full workflow topology through static code analysis in the design model generator and renders it on the overview diagram:

```
                                              ┌───────────────────────┐
[resource calls workflow:run(foo)] ────────▶ ▶│ <icon> foo            │
                                              ├───────────────────────┤
[resource workflow:sendData(foo, "bar")] ───▶ │ bar      Data Event   │──▶ [Connection]
[resource workflow:sendData(foo, "baz")] ───▶ │ baz      Data Event   │──▶ [Another Connection]
                                              │ approve  Human Task   │
                                              └───────────────────────┘
```

https://github.com/user-attachments/assets/eebaf9ff-8bda-4830-9b02-d479307d9fe2

## Approach

The change spans the two existing layers:

1. **Design model (language server, `architecture-model-generator`)** — the `CodeAnalyzer` / `DesignModelGenerator` pass is extended to recognize the workflow API surface and emit richer `Workflow` nodes plus a new top-level `Activity` node.
2. **Overview diagram (`component-diagram` package)** — the workflow entry node renders the derived events/human tasks, a new activity node type is introduced, and the layout/edge rules are extended.

### Code analysis rules (language server)

The analyzer resolves every call site against the semantic model (org `ballerina`, module `workflow`) rather than matching on names alone:

| Construct | Rule |
| --- | --- |
| **Workflow node** | Module-level function annotated `@workflow:Workflow` (existing behavior). |
| **Data events** | The workflow function's optional third parameter — a `record {\| future<T> ...; \|}` — declares the events the workflow waits on. Each field becomes a *Data Event* row carrying its name and payload type `T`. |
| **Run triggers** | `workflow:run(fn, ...)` call sites. The first argument is resolved to the workflow function symbol. Attribution is propagated through the existing transitive call graph (helper functions, service-class methods), so a resource that runs a workflow via a helper is still linked. Attribution is recorded per resource/remote function as well as per service/automation. |
| **Data senders** | `workflow:sendData(fn, id, dataName, data)` call sites. `dataName` is resolved from a string literal or a string constant (positional or named argument) and matched against the target workflow's declared events. Senders are recorded per function (`workflowSendData`: workflow → event names) and propagated through the call graph. If the event name cannot be resolved statically, the sender is attached to the workflow itself so the edge is never silently dropped. |
| **Human tasks** | `ctx->awaitHumanTask(taskName, ...)` calls inside the workflow body (callee verified to be `workflow:Context`). The task name is resolved from a literal/constant, with a generic fallback label. |
| **Activities** | `ctx->callActivity(fn, args)` calls inside the workflow body. User-defined activities are pre-populated from `@workflow:Activity` functions and referenced by UUID (deduplicated across workflows); functions from the builtin `workflow.activity` module are surfaced with friendly labels (Call REST API, Call SOAP API, Send Email (SMTP)). |
| **Activity → connection** | Union of (a) the activity function's own transitive connection usage, reusing the existing connection graph, and (b) connections passed to the activity as `callActivity` arguments (e.g. an `http:Client` parameter), resolved via the existing `ConnectionFinder`. |

Model additions: `Workflow.events` (with per-event sender back-references), `Workflow.humanTasks`, `Workflow.activities` (UUID refs), new top-level `DesignModel.activities`, and `Function`/`ResourceFunction.workflowSendData`.

### Diagram rules (overview page)

- **Columns**: `listeners | services & automation | workflows | connections`, left to right. The workflow column collapses when empty, so diagrams without workflows are unchanged. Workflows are positioned vertically near the entry points that reference them. Activities are analyzed but **not rendered** — the overview shows only the workflow (events, human tasks) and the connections its activities use.
- **Workflow box**: header (icon + name) with an explicit **play-button target** on the left border — all `workflow:run` edges point at it to make the trigger semantics obvious; one row per **Data Event** (payload type shown on hover) with its own inbound port; one row per **Human Task**.
- **Edges**:
  - *run*: from the exact resource/remote function row → play button. Falls back to a service-level edge only when no rendered row carries the link (e.g. a private helper). Resources hidden behind "Show More Resources" route through that button's port, mirroring connection-edge behavior.
  - *sendData*: from the exact resource/remote function row → the specific Data Event row (same fallback rules).
  - *connections*: workflow box → each connection used by the workflow's activities (derived from the activity analysis; deduplicated per workflow). Activity boxes themselves are not drawn — built-in activities live in the `ballerina/workflow.activity` module, so an activity node offered a navigation target outside the project (wso2/product-integrator#1887).

### Artifact naming

The workflow artifact is now presented as **Durable Workflow** across the UI (add-artifact card — with a description and a tooltip explaining that execution state is persisted at every step so the workflow can wait on events, timers, and human tasks and resume after crashes — the overview node label, the form title, and the flow-diagram workflow list).

## Tests

- New fixture `get_design_model/source/project_13` + `config/project_with_workflow_events.json` covering: `run` from a service and transitively from `main`, `sendData` matched to two events, `awaitHumanTask`, a user activity using a global connection internally, and an activity receiving the connection as a `callActivity` argument.
- `DesignModelGeneratorTest` assertions extended for events/human tasks/activities (`assertActivities`, null-safe field checks).
- Component-diagram jest snapshots regenerated for the updated rendering.
- Verified manually end-to-end with a packaged VSIX on sample projects (single/multiple resources, collapsed resource lists, builtin + user-defined activities).


packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/test/resources/get_design_model/config/project_with_workflow_events.json
```diff
@@ -0,0 +1,309 @@
...
+              "connections": [],
+              "workflows": [],
+              "workflowSendData": {
+                "d1e5ae82-aece-9da5-91cb-790d3644d8ab": [
+                  "paymentReceived"
+                ]
+              }
+            }
+          ],
+          "absolutePath": "/orders ",
```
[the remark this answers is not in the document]
hasithaa: That value accurately reflects the generator's output — toSourceCode() carries the last path token's trailing trivia, a pre-existing behavior present in every service fixture ("/api " etc.).

packages/ballerina-extension/test/runLibsIntegrationTest.ts
samithkavishke, line 1: Is this related to the issue?

samithkavishke:
- Currently, the architecture diagram view only consists of three layers. With this feature, that structure has changed, so it would be better to introduce a consistent visualization that aligns with the integration views.
- The current activity nodes are also missing the source view and other related information.

hasithaa:
Note on CI: the `Run Ballerina LS tests (ubuntu-latest)` job is failing on pre-existing flakes unrelated to this PR (which changes no language-server code — the diff is confined to `component-diagram`, `ballerina-visualizer`, and one label in `ballerina-core`). Five attempts hit four different unrelated tests: a Java heap-space OOM in `AvailableNodesTest` (`caller.json`); the roaming `Failed to initialize the compiler plugin ballerina:http … ctxData is null` error in `DesignModelGeneratorTest` (`connections_for_service_classes.json`, twice) and in `SearchTest` (`types/submodule_special_chars.json`); and an `Index 1 out of bounds` in `NodeTemplateTest` (`function_call-userEscaped.json`). The windows LS job, the diagram snapshot tests (the check this PR actually touches — previously failing, now green), the repo build, and all four e2e groups pass. The `main` Daily Build has also been failing daily for the past week, consistent with suite-level flakiness rather than a regression in this PR.

packages/ballerina-extension/test/runLibsIntegrationTest.ts
hasithaa: Good catch — the changes to `runAiTest.ts` and `runLibsIntegrationTest.ts` were general dedup (migrating those entrypoints to the shared `runTests` helper) that this PR does not need: the new workflow snapshot tests run through the main `runTest.ts` entrypoint only. Reverted both files to their upstream state in 3a73f8b1eb. The runner changes the new tests do need are kept: the `env-webpack-helper` path fix in `runTest.ts`, the `.md` require hook and `TEST_FILE_GREP` file filter in `test/index.ts` (the test host cannot load webpack raw-loader imports otherwise, and the full mocha run breaks without filtering), and the shared helper installing extension dependencies into the launched extensions dir so the extension activates in the test instance. All extension workflow tests, the component-diagram jest suite, and the full rush build pass after the revert.

hasithaa:
@samithkavishke @hevayo @kanushka @rdulmina  This is the latest updated diagram. 

image

hasithaa:
CI status on d3341e82ac: the earlier types-manager regression from the first perf attempt is confirmed fixed — **windows LS is green** and the ubuntu run has zero `recordConfig` failures.

The remaining ubuntu LS failure is new upstream breakage unrelated to this PR: **`ballerina/ai` 1.12.0 was published to central today at 06:36 UTC**, and the 18 failing tests (`model_provider_generate*`, `agent.json`, vector-store manager, …) all fail with `codedata.version: actual "1.12.0", expected "1.11.2"` — the goldens pin 1.11.2 while the non-sticky central resolution now returns 1.12.0 on runners with a cold module cache (windows passed because its cache still held 1.11.2). Every PR and main build will hit this until the goldens are refreshed for ai 1.12.0 (or the test projects pin the ai version); reruns won't help since central keeps serving the new latest.

kanushka:
> @samithkavishke @hevayo @kanushka @rdulmina This is the latest updated diagram.
> 
> image

This diagram is better because it only shows the top-level components to the user.  However, we still offer the option to view more details by clicking on each component.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/resources/agents_manager/config/agent_call_flow_node_8.json
```diff
@@ -909,6 +937,34 @@
             },
             "defaultValue": "NO_FILTER"
           },
+          "executeToolCallsInParallel": {
```
samithkavishke, line 940: Can you verify whether this is caused only by the version update or by something else?

samithkavishke (commented):
Language server changes look good to me. Just one minor comment: the AI changes and the package resolution through the new cache mechanism don't seem to be related to the original issue this pull request addresses. If possible, it would be better to drop those changes.
Other than that, everything looks good. The resolution logic also seems fine to me. My only concern is that the TTL is set to 2 minutes, which might negatively affect the user experience.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/resources/agents_manager/config/agent_call_flow_node_8.json
hasithaa: This is a newly added feature from the AI module.

hasithaa:
Split the version/build churn out of this PR into #697 ("Stabilize flow-model test goldens for ai 1.12.0 / workflow 0.7.0 and raise test heap"), which targets `main` on its own. That PR carries only test goldens + Gradle build config (2g test heap, ai 1.12.0 golden refresh + pin, workflow 0.7.0 golden refresh) — the changes that were failing CI here purely due to upstream module drift (ai 1.12.0 and workflow 0.7.0 released after this branch's goldens were baked; `main`'s own nightly Build has been red on the same drift).

This branch is now rebased on top of #697 and contains **only the workflow-overview feature** (event/activity-connection overview, Durable Workflow rename) plus the designer-load perf fix. Once #697 merges to `main`, this PR's diff auto-reduces to feature-only. Also brought up to date with the latest `main`.

madushajg (approved):

PR #690, opened 2026-07-14
Fix workflow node UX issues: Run Workflow input typing, flow-node diagnostics, and duplicate variable validation

### LINKED ISSUE #1879: [Workflow] Run Workflow node loses the workflow function's input type when reopened from source
### Description

When a Run Workflow node is freshly added from the palette, its **Input** field is typed from the selected workflow function's declared input parameter (e.g. a record type) — `WorkflowRunBuilder.setConcreteTemplateData` derives it from the function symbol.

However, once the `workflow:run(...)` call is written to source and the diagram is regenerated, `CodeAnalyzer.visit(FunctionCallExpressionNode)` rebuilds the node's form properties from the **library signature of `workflow:run`** (`map<anydata>? input`), losing the workflow function's specific input type. The form then rejects valid inputs with `incompatible types: expected 'map<anydata>?', found 'string'` (surfaced when the workflow declares a `string`/record input), and an internal `processFunction` expression property leaks into the form.

### Steps to Reproduce

1. Create a `@workflow:Workflow` function with a record input, e.g. `orderWorkflow(workflow:Context ctx, OrderInput input)`.
2. Add a **Run Workflow** node from the palette → the Input field correctly shows `OrderInput`.
3. Save, then reopen the node form (diagram regenerated from source).
4. The Input field now shows `map<anydata>?` (and a raw `processFunction` field), and entering a valid `OrderInput` value/expression is flagged as incompatible.

### Expected Behavior

The Input field should always be typed from the selected workflow function's declared input parameter, on both the template path and the source-read path. The `processFunction` reference is internal (carried in `codedata.symbol`) and should not appear as a form field.

### Fix

Implemented on branch `general-ux-bugs`: `CodeAnalyzer.populateWorkflowRunProperties(...)` post-processes WORKFLOW_RUN nodes on the source-read path — re-derives the `input` property type from the workflow function symbol (first parameter that is a subtype of `anydata`), keeps the current argument value, drops the `processFunction` property, and removes the `input` property when the workflow declares no input.

Related module-side fix: `ballerina/workflow` now accepts any `anydata` input in `workflow:run()` and validates it at compile time (workflow module 0.7.0).

### Version

ballerina-vscode (BI) main; ballerina/workflow 0.6.0

hasithaa:
## Purpose

Fix three workflow (BI designer) UX defects around the Run Workflow node form, flow-node diagnostics, and variable-name validation.

Resolves wso2/product-integrator#1879, wso2/product-integrator#1880, wso2/product-integrator#1881

### 1. Run Workflow node: derive the `input` type from the workflow function on the source-read path (wso2/product-integrator#1879)
A freshly added Run Workflow node types its **Input** field from the selected workflow function's declared input parameter, but once the `workflow:run(...)` call is written to source and re-read, `CodeAnalyzer` rebuilt the form properties from the `workflow:run` library signature (`map<anydata>? input`), losing the specific type and leaking an internal `processFunction` field into the form. The new `CodeAnalyzer.populateWorkflowRunProperties(...)` re-derives the `input` property type from the workflow function symbol (matching the template path in `WorkflowRunBuilder`), keeps the current argument value, drops the `processFunction` property, and removes `input` when the workflow declares no input.

### 2. Sort diagnostics by position so compiler-plugin diagnostics attach to flow nodes (wso2/product-integrator#1880)
`DiagnosticHandler` advances a single cursor while nodes are visited in source order, so it requires position-sorted diagnostics. Package compilation appends compiler-plugin diagnostics (e.g. the workflow plugin's `WORKFLOW_1xx` codes) after the semantic-phase ones, breaking the within-file order — plugin diagnostics were silently skipped and never attached to Run Workflow / Call Activity / Send Data / Wait Data / Human Task nodes. The handler now sorts diagnostics by (file, start line, start offset) before iterating. `HUMAN_TASK` is also added to `WORKFLOW_NODE_KINDS` in `bi-diagram`.

### 3. Report duplicate variable names for new flow nodes (wso2/product-integrator#1881)
The identifier duplicate check excluded any visible symbol matching the property's original value. That exclusion is only correct for the edit/rename flow (where the match is the node's own variable); for **new** nodes it masked genuine duplicates whenever the default or typed name collided with an existing variable, so forms such as Call Activity accepted duplicate result variable names without any error. The exclusion now applies only when the property is bound to an existing declaration (property codedata carries a `lineRange`). `WorkflowRunBuilder` also generates a unique default (`workflowId`, `workflowId1`, ...) instead of a hardcoded name.

## Automation tests

- New `DiagnosticHandlerTest` covering the out-of-order diagnostics case (plugin diagnostics appended after semantic ones).
- `ModelGeneratorTest` expected configs updated for the behavior changes: `workflow_run.json` (input typed from the workflow function, `processFunction` property removed) and `workflow_activity_and_wait.json` (nodes now correctly carry `hasDiagnostics: true`).

## Related PRs

- ballerina-platform/module-ballerina-workflow#63 — module-side counterpart: `workflow:run()` accepts any `anydata` input, and new compile-time validations (`WORKFLOW_130`–`WORKFLOW_136`) produce the diagnostics that change (2) surfaces on the flow nodes.

### DIFF
--- a/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
+++ b/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
@@ -131,6 +131,7 @@ import io.ballerina.compiler.syntax.tree.WhileStatementNode;
 import io.ballerina.flowmodelgenerator.core.model.Branch;
 import io.ballerina.flowmodelgenerator.core.model.Codedata;
 import io.ballerina.flowmodelgenerator.core.model.CommentProperty;
+import io.ballerina.flowmodelgenerator.core.model.Diagnostics;
 import io.ballerina.flowmodelgenerator.core.model.FlowNode;
 import io.ballerina.flowmodelgenerator.core.model.FormBuilder;
 import io.ballerina.flowmodelgenerator.core.model.ItemOption;
@@ -3363,13 +3364,7 @@ public class CodeAnalyzer extends NodeVisitor {
         }
 
         // The workflow's input parameter is the first parameter that is a subtype of anydata.
-        TypeSymbol inputType = null;
-        for (ParameterSymbol param : workflowFuncSymbol.typeDescriptor().params().orElse(List.of())) {
-            if (param.typeDescriptor().subtypeOf(semanticModel.types().ANYDATA)) {
-                inputType = param.typeDescriptor();
-                break;
-            }
-        }
+        TypeSymbol inputType = WorkflowRunBuilder.findWorkflowInputType(workflowFuncSymbol, semanticModel);
         if (inputType == null) {
             // The workflow function declares no input; drop the library-derived input property.
             currentProps.remove(WorkflowRunBuilder.INPUT_KEY);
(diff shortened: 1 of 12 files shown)


madushajg (approved):
LGTM

PR #694, opened 2026-07-17
Guide Copilot to add @sql:Column when mapping records to DB tables

ravindu25:
fixes: https://github.com/wso2/product-integrator/issues/1809
Screenshot 2026-07-17 at 3 13 12 PM

## Problem
When Copilot generates a Ballerina record for a SQL table, it uses camelCase
field names (the Ballerina convention) while SQL columns are snake_case.
Without `@sql:Column {name: "..."}`, the sql client cannot bind columns to
fields at runtime. Copilot was not adding the annotation.

## Changes
- **New** `flow-model-generator-ls-extension/.../copilot/instructions/ballerina/sql/library.md`
  — instructs the model to annotate each field of a DB-table record with `@sql:Column`.
- **Fix** `function-registry.ts` `getExternalRecords`: preserve `instructions`
  and `readme` (previously only `services`/`annotations` were copied).

## Root cause — why the instruction alone wasn't enough
Per-library `instructions` reach the model via two paths:
- `toMaximizedLibrariesFromLibJson` (library has selected functions) — already copied `instructions`.
- `getExternalRecords` (library pulled in only as a type dependency) — did **not**.

In a DB flow the client is `ballerinax/mysql`; `ballerina/sql` contributes
only types, so it takes the external path and its `library.md` was dropped —
the model never saw the rule. Other libraries with a `library.md` (ai,
workflow, test) are always primary/selected, so they went through the main
path and were unaffected. This is why the gap surfaced now.

### DIFF
--- a/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/main/resources/copilot/instructions/ballerina/sql/library.md
+++ b/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/main/resources/copilot/instructions/ballerina/sql/library.md
@@ -4,19 +4,15 @@ The `ballerina/sql` library provides the core types, errors, and annotations for
 relational database access. Database-specific clients (`ballerinax/mysql`,
 `ballerinax/postgresql`, `ballerinax/mssql`, etc.) build on top of it.
 
-## CRITICAL: Annotate every field of a database-table record with `@sql:Column`
+## Mapping record fields to database columns with `@sql:Column`
 
-Whenever you define a Ballerina record type that represents a row of a database
-table, you MUST add a `@sql:Column {name: "<exact_column_name>"}` annotation to
-**every** field of that record, mapping the (camelCase) Ballerina field to its
-exact database column name (conventionally snake_case).
+When a SQL query result is mapped into a Ballerina record, the client matches
+each database column to a record field **by name**. Ballerina record fields use
+camelCase by convention, while database columns are commonly snake_case, so the
+names frequently do not match.
 
-Apply this rule to **all** fields, always:
-- Do it even if a field's name already matches its column name.
-- Do it even if no query has been written yet.
-
-The SQL client binds columns to record fields **by exact name**, so annotating
-every field with its column name guarantees the mapping is always correct.
+Use the `@sql:Column {name: "<column_name>"}` annotation to map a record field to
+its database column **whenever the field name differs from the column name**:
 
 ```ballerina
 import ballerina/sql;
@@ -28,3 +24,6 @@ type Customer record {|
     string customerName;
 |};
 ```
+
+A field whose name already matches its column exactly (e.g. a column `id` mapped
+to a field `id`) does not need the annotation.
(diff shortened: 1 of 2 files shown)


packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/main/resources/copilot/instructions/ballerina/sql/library.md
```diff
@@ -0,0 +1,30 @@
...
+The `ballerina/sql` library provides the core types, errors, and annotations for
+relational database access. Database-specific clients (`ballerinax/mysql`,
+`ballerinax/postgresql`, `ballerinax/mssql`, etc.) build on top of it.
+
+## CRITICAL: Annotate every field of a database-table record with `@sql:Column`
+
+Whenever you define a Ballerina record type that represents a row of a database
+table, you MUST add a `@sql:Column {name: "<exact_column_name>"}` annotation to
+**every** field of that record, mapping the (camelCase) Ballerina field to its
+exact database column name (conventionally snake_case).
```
RNViththagan: is @sql:Column actually needed on every field, or only when the field name differs from the column name?

RNViththagan: @SasinduDilshara

RNViththagan: Following up on this — since these instructions double as Copilot's source of truth (and future "ask" features will lean on them), I think we should be precise rather than just simple here.

Technically, `@sql:Column` is only *required* when the field name doesn't match the column name (e.g. `customerId` vs `customer_id`). For a field that already matches (like `id`), it's optional.

Suggestion — split this into two tiers instead of one blanket "CRITICAL"/mandatory rule:
- **Required**: annotate any field whose name differs from its column name (covers almost every camelCase↔snake_case field in practice).
- **Recommended**: annotate the rest too, for consistency and to guard against future renames — but frame it as a best practice, not a requirement.

That keeps Copilot's own understanding accurate instead of it treating a convention as a hard rule.

ravindu25: I have updated the library.md file

ravindu25:
After removing the 'MUST' rule from the library.md, I ran 10 tests with the same prompt to check if the copilot adds annotations. It successfully did so when required. I have shared screenshots here of 5 of them:

Screenshot 2026-07-20 at 10 56 33 AM
Screenshot 2026-07-20 at 10 57 28 AM
Screenshot 2026-07-20 at 11 00 41 AM
Screenshot 2026-07-20 at 11 01 49 AM
Screenshot 2026-07-20 at 11 02 38 AM

RNViththagan (approved):
LGTM

PR #707, opened 2026-07-23
Add explore samples e2e test cases

samithkavishke:
## Purpose
Add explore samples E2E test cases

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary

- Added end-to-end coverage for browsing, filtering, searching, downloading, and opening Ballerina samples.
- Added authoring scenarios for sample discovery and project validation.
- Added the Playwright sample test suite to Group 2.
- Improved test stability with polling and explicit validation for asynchronous views and new windows.

### DIFF
--- a/packages/ballerina-extension/e2e-test/e2e-authoring/scenarios/samples/scenario.md
+++ b/packages/ballerina-extension/e2e-test/e2e-authoring/scenarios/samples/scenario.md
@@ -1,0 +1,38 @@
+Use Samples in the WSO2 Integrator
+
+From the "Get Started" welcome page, open the samples browser via the "Explore"
+card, filter and search the list, then use one of the built-in samples and
+verify the extracted project's integration overview loads in a new window.
+
+## Steps
+
+| # | Action | Verification |
+|---|--------|-------------|
+| 1 | Click "Get Started", then click "Explore" on the "Explore Pre-built Integrations and Samples" card | "Browse Samples" heading is visible with an initial results count |
+| 2 | Click the "Sample" type filter button | Results count is smaller than the initial "All" count |
+| 3 | Click "All" again and type "Hello World" into the search box | Exactly the "Hello World Service" sample card is shown |
+| 4 | Click "Use this" on the sample card, confirm the download directory, then choose "New Window" | Downloaded sample opens in a new VS Code window and its integration overview ("Add Artifact") becomes visible |
+
+## Gaps
+
+- The scenario as originally written used older wording ("built in samples
+  button", "All button", "use this button") from an earlier `SamplesView`
+  implementation in `packages/ballerina-visualizer` (Download button, category
+  `<select>`, native `showOpenDialog`). The UI that actually ships today comes
+  from the `WSO2.wso2-integrator` prerequisite extension's "Browse Samples"
+  page — a "Sample" / "Pre-built Integrations" type filter, a free-text search
+  box, a category dropdown, and per-card "Use this" buttons. The steps above
+  and the promoted spec target this current implementation.
+- "Use this" calls `window.showOpenDialog` for the download directory. The
+  harness sets `files.simpleDialog.enable: true`, so this renders as an
+  in-workbench dialog reachable via the host workbench page (a
+  `getByRole('button', { name: 'Select Folder' })`), not a real native OS
+  picker — no extra harness support was needed.
+- After extraction, VS Code shows a modal "Where would you like to open the
+  project?" with "Current Window" / "New Window". "New Window" was used here:
+  it spawns a genuinely new Electron window, picked up via
+  `vscode.windows()` (comparing the window count before/after, since
+  `firstWindow()` only resolves once). "Current Window" was tried first during
+  authoring and left the workspace in a confusing nested-root state (the test
+  workspace's own folder is the sample's parent directory), so it was dropped
+  in favor of "New Window".
--- a/packages/ballerina-extension/e2e-test/e2e-authoring/scenarios/samples/steps/02_filter_type.step.js
+++ b/packages/ballerina-extension/e2e-test/e2e-authoring/scenarios/samples/steps/02_filter_type.step.js
@@ -5,12 +5,21 @@
   // "Pre-built Integrations" — the resulting count must be strictly smaller
   // than the unfiltered "All" count captured in step 1.
   await guestClick(frame.getByRole('button', { name: 'Sample', exact: true }));
-  await window.waitForTimeout(1000);
 
-  const snap = await snapshot();
-  const match = snap.match(/(\d+) results?/);
-  if (!match) throw new Error(`could not find a results count after filtering by "Sample":\n${snap}`);
-  const filteredCount = Number(match[1]);
+  // Poll for the count to settle below the initial "All" count instead of a
+  // fixed sleep — the filtered list updates asynchronously.
+  const readCount = async () => {
+    const snap = await snapshot();
+    const match = snap.match(/(\d+) results?/);
+    if (!match) throw new Error(`could not find a results count after filtering by "Sample":\n${snap}`);
+    return Number(match[1]);
+  };
+  const deadline = Date.now() + 15000;
+  let filteredCount = await readCount();
+  while (filteredCount >= globalThis.initialSampleCount && Date.now() < deadline) {
+    await window.waitForTimeout(300);
+    filteredCount = await readCount();
+  }
   console.log(`sample-only count: ${filteredCount} (was ${globalThis.initialSampleCount})`);
   if (!(filteredCount < globalThis.initialSampleCount)) {
     throw new Error(`expected filtered count (${filteredCount}) to be less than initial count (${globalThis.initialSampleCount})`);
--- a/packages/ballerina-extension/e2e-test/e2e-authoring/scenarios/samples/steps/02_filter_type.step.js
+++ b/packages/ballerina-extension/e2e-test/e2e-authoring/scenarios/samples/steps/02_filter_type.step.js
@@ -1,0 +1,20 @@
+{
+  const frame = await getBIWebview();
+
+  // "Sample" restricts the view to the built-in samples, excluding
+  // "Pre-built Integrations" — the resulting count must be strictly smaller
+  // than the unfiltered "All" count captured in step 1.
+  await guestClick(frame.getByRole('button', { name: 'Sample', exact: true }));
+  await window.waitForTimeout(1000);
+
+  const snap = await snapshot();
+  const match = snap.match(/(\d+) results?/);
+  if (!match) throw new Error(`could not find a results count after filtering by "Sample":\n${snap}`);
+  const filteredCount = Number(match[1]);
+  console.log(`sample-only count: ${filteredCount} (was ${globalThis.initialSampleCount})`);
+  if (!(filteredCount < globalThis.initialSampleCount)) {
+    throw new Error(`expected filtered count (${filteredCount}) to be less than initial count (${globalThis.initialSampleCount})`);
+  }
+
+  console.log('done: 02_filter_type');
+}
--- a/packages/ballerina-extension/e2e-test/e2e-authoring/scenarios/samples/steps/04_use_sample.step.js
+++ b/packages/ballerina-extension/e2e-test/e2e-authoring/scenarios/samples/steps/04_use_sample.step.js
@@ -1,0 +1,36 @@
+{
+  const frame = await getBIWebview();
+  const card = frame.getByRole('article').filter({ hasText: 'Hello World Service' });
+  await guestClick(card.getByRole('button', { name: 'Use this' }));
+
+  // "Use this" triggers window.showOpenDialog for the download directory.
+  // The harness renders this as VS Code's in-workbench simple file dialog
+  // (files.simpleDialog.enable), not a native OS picker, so it is reachable
+  // through the host page rather than the sample-view webview. It defaults
+  // to the current workspace folder, which is fine for this scenario.
+  const selectFolderButton = window.getByRole('button', { name: 'Select Folder' });
+  await selectFolderButton.waitFor({ state: 'visible', timeout: 20000 });
+  await selectFolderButton.click();
+  console.log('confirmed sample download directory');
+
+  // Wait for the download+extract to finish, then choose "New Window" on the
+  // "Where would you like to open the project?" prompt.
+  const newWindowButton = window.getByRole('button', { name: 'New Window' });
+  await newWindowButton.waitFor({ state: 'visible', timeout: 60000 });
+  await newWindowButton.click();
+  console.log('sample downloaded; opening in a new window');
+
+  // A brand-new Electron window opens for the extracted sample project.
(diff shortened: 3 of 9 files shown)


packages/ballerina-extension/e2e-test/e2e-authoring/scenarios/samples/steps/04_use_sample.step.js
```diff
@@ -0,0 +1,38 @@
...
+  console.log('sample downloaded; opening in a new window');
+
+  // A brand-new Electron window opens for the extracted sample project.
+  // Playwright's firstWindow() only fires for the very first window, so
+  // ensureWorkbench()/waitForGuest() pick up the newest open window instead.
+  window = null;
+  const projectFrame = await waitForGuest(BI_INTEGRATOR_LABEL, 120000);
+  await waitForText('Add Artifact', 60000);
+  const snap = await snapshot();
+  if (!snap.includes('hello-world-service') && !snap.includes('Add Artifact')) {
```
NipunaRanasinghe, line 32: This guard can never fire. `waitForText('Add Artifact', 60000)` on line 30 already succeeded, so `snap.includes('Add Artifact')` is true and the second conjunct is false — the `&&` makes the whole condition permanently false.

Presumably the intent was `||`, but since `'Add Artifact'` is already covered by the `waitForText` above, the only assertion that adds anything here is the project identifier:

```js
if (!snap.includes('hello-world-service')) {
    throw new Error(`sample project overview did not appear after opening the sample:\n${snap}`);
}
```

packages/ballerina-extension/e2e-test/e2e-playwright-tests/samples/samples.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+            logStep('Waiting for the sample\'s integration overview to load');
+            const deadline = Date.now() + 120000;
+            let projectWebView;
+            while (Date.now() < deadline && !projectWebView) {
+                const openWindows = vscode!.windows().filter((w) => !w.isClosed());
+                for (const candidate of [...openWindows].reverse()) {
+                    try {
+                        await candidate.waitForLoadState('domcontentloaded', { timeout: 3000 });
+                        const frame = await switchToIFrame(BI_INTEGRATOR_LABEL, candidate, 5000);
+                        if (frame && await frame.getByText('Add Artifact').isVisible({ timeout: 3000 }).catch(() => false)) {
```
NipunaRanasinghe, line 140: Three separate problems in this loop:

**1. Nothing ties the matched window to the downloaded sample.** The only check is the generic text `Add Artifact`, which is the standard overview affordance — any open window showing a `WSO2 Integrator` webview satisfies it. The test can report success without the sample having been opened. Assert on something sample-specific (`hello-world-service`) as well.

**2. Combined with the window never being closed, retries can pass falsely.** `e2e-test/playwright.config.js:9` sets `retries: 2`. If attempt 1 fails anywhere after the "New Window" click on line 122, that Electron window stays open. On attempt 2 this loop scans *all* open windows and can match the stale window from attempt 1 — which already shows the overview — and pass without the flow actually working. Snapshot the window list immediately before line 122 and only consider windows that aren't in that set:

```ts
const before = new Set(vscode!.windows());
await newWindowButton.click();
// ...
const openWindows = vscode!.windows().filter((w) => !w.isClosed() && !before.has(w));
```

Also worth closing the spawned window in an `afterAll` — otherwise it plus its language server stay resident for the rest of Group 2 (this spec is registered last, so the blast radius is small today, but it breaks as soon as anything is added after it).

**3. `isVisible({ timeout: 3000 })` on line 140 does not wait.** Playwright marks that option `@deprecated This option is ignored` — the call returns immediately. It happens to work because of the 2s poll on line 149, but the intent reads wrong; drop the option or use `expect(...).toBeVisible({ timeout })`.

Minor, same line: `getByText('Add Artifact')` is unqualified. If it ever matches more than one node the strict-mode error is swallowed by the bare `catch {}` on line 144 and the loop silently spins the full 120s before failing with a misleading message. Add `.first()`.

packages/ballerina-extension/e2e-test/e2e-playwright-tests/samples/samples.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+
+            const initialCount = await readResultCount();
+            expect(initialCount).toBeGreaterThan(0);
+            logStep(`Initial sample count: ${initialCount}`);
+
+            logStep('Clicking the "Sample" type filter (built-in samples only)');
+            await samplesWebView.getByRole('button', { name: 'Sample', exact: true }).click({ force: true });
+            const filteredCount = await waitForResultCountBelow(initialCount);
+            logStep(`Sample-only count: ${filteredCount}`);
+            expect(filteredCount).toBeLessThan(initialCount);
```
NipunaRanasinghe, line 88: `waitForResultCountBelow` returns as soon as the count is below `below`, and that includes `0` while the list is re-rendering. So `expect(filteredCount).toBeLessThan(initialCount)` also passes on a filter that renders "0 results" — i.e. a genuinely broken "Sample" filter is indistinguishable from a working one here.

Add a lower bound:

```ts
expect(filteredCount).toBeGreaterThan(0);
expect(filteredCount).toBeLessThan(initialCount);
```

Same hole in `02_filter_type.step.js:24`.

packages/ballerina-extension/e2e-test/e2e-authoring/scenarios/samples/steps/02_filter_type.step.js
```diff
@@ -0,0 +1,29 @@
...
+    return Number(match[1]);
+  };
+  const deadline = Date.now() + 15000;
+  let filteredCount = await readCount();
+  while (filteredCount >= globalThis.initialSampleCount && Date.now() < deadline) {
+    await window.waitForTimeout(300);
+    filteredCount = await readCount();
+  }
+  console.log(`sample-only count: ${filteredCount} (was ${globalThis.initialSampleCount})`);
+  if (!(filteredCount < globalThis.initialSampleCount)) {
```
NipunaRanasinghe, line 24: Same gap as in the promoted spec: the `while` on line 19 exits as soon as the count drops below `initialSampleCount`, which includes `0` mid-render, so a filter that renders "0 results" passes this check. Assert a lower bound too:

```js
if (!(filteredCount > 0 && filteredCount < globalThis.initialSampleCount)) {
```

packages/ballerina-extension/e2e-test/e2e-playwright-tests/samples/samples.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+ * under the License.
+ */
+import { expect, test } from '@playwright/test';
+import { BI_INTEGRATOR_LABEL, BI_WEBVIEW_NOT_FOUND_ERROR, getWebview, initTest, logStep, page, vscode } from '../utils/helpers';
+import { switchToIFrame } from '@wso2/playwright-vscode-tester';
+
+export default function createTests() {
+    test.describe.serial('Use Samples in the WSO2 Integrator', {
+    }, async () => {
+        initTest(false);
```
NipunaRanasinghe, line 25: This is the only active `initTest(false)` in the suite — every other spec calls `initTest()` or `initTest(true, ...)`, and the one other occurrence (`project-overview/project-creation.spec.ts:28`) is commented out. Worth a comment explaining why, because the resulting state is subtle:

`wipeAndRecreateDataFolder()` runs unconditionally (`utils/helpers/setup.ts:548`) and deletes `data/` including `data/test_project`, but with `newProject=false` the template is *not* restored. So after the reload the window's workspace root no longer exists on disk. That empty state is presumably what makes the "Get Started" welcome page appear, but the test is relying on it implicitly.

It also makes the comment on line 112-113 inaccurate — the "already-open test workspace" was just deleted, so where the sample actually lands is unspecified. See the separate note on line 115.

packages/ballerina-extension/e2e-test/e2e-playwright-tests/samples/samples.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+            // "Use this" triggers window.showOpenDialog for the download
+            // directory. The harness renders VS Code's in-workbench simple file
+            // dialog (files.simpleDialog.enable) rather than a native OS picker,
+            // so it's reachable through the host workbench page, not the
+            // samples webview. Confirming it keeps the default directory
+            // (the already-open test workspace), which is fine for this flow.
+            logStep('Confirming the sample download directory');
+            const selectFolderButton = workbenchPage.getByRole('button', { name: 'Select Folder' });
+            await selectFolderButton.waitFor({ state: 'visible', timeout: 20000 });
+            await selectFolderButton.click();
```
NipunaRanasinghe, line 117: "Confirming it keeps the default directory (the already-open test workspace)" doesn't hold: with `initTest(false)` on line 25, `wipeAndRecreateDataFolder()` has just deleted `data/test_project`, so the workspace root doesn't exist on disk at this point and the dialog's default is whatever VS Code falls back to. The test then extracts a sample project into an unspecified location and never cleans it up.

Type an explicit destination into the simple dialog (a dedicated temp dir under `data/`) instead of accepting the default, and remove it in an `afterAll`. That also removes the nested-Ballerina-package hazard that `scenario.md` flags as the reason "Current Window" was dropped.

The rest of this comment (the `files.simpleDialog.enable` mechanism) is accurate — confirmed at `playwright-vscode-tester/src/browser.ts:91`.

packages/ballerina-extension/e2e-test/e2e-playwright-tests/samples/samples.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+            const searchBox = samplesWebView.getByRole('textbox', { name: 'Text field' });
+            await searchBox.click({ force: true });
+            await searchBox.fill('Hello World');
+
+            const sampleCard = samplesWebView.getByRole('article').filter({ hasText: 'Hello World Service' });
+            await sampleCard.waitFor({ state: 'visible', timeout: 15000 });
+            // The search must narrow the list down to exactly this one sample —
+            // otherwise "Use this" below could click the wrong card if more
+            // than one result matched "Hello World Service".
+            await expect(samplesWebView.getByRole('article')).toHaveCount(1);
```
NipunaRanasinghe, line 102: `toHaveCount(1)` couples the test to the exact contents of the remote sample catalog — the day someone adds a second sample whose title contains "Hello World" (a gRPC or GraphQL variant, say), this fails even though the flow under test is fine.

The stated reason for the assertion ("`Use this` below could click the wrong card") is already covered: `sampleCard` on line 97 filters by text, so a second match makes the `.click()` on line 106 fail with a strict-mode violation that names both candidates. I'd drop line 102 and rely on that.

packages/ballerina-extension/e2e-test/e2e-authoring/scenarios/samples/scenario.md
```diff
@@ -0,0 +1,38 @@
...
+| 3 | Click "All" again and type "Hello World" into the search box | Exactly the "Hello World Service" sample card is shown |
+| 4 | Click "Use this" on the sample card, confirm the download directory, then choose "New Window" | Downloaded sample opens in a new VS Code window and its integration overview ("Add Artifact") becomes visible |
+
+## Gaps
+
+- The scenario as originally written used older wording ("built in samples
+  button", "All button", "use this button") from an earlier `SamplesView`
+  implementation in `packages/ballerina-visualizer` (Download button, category
+  `<select>`, native `showOpenDialog`). The UI that actually ships today comes
+  from the `WSO2.wso2-integrator` prerequisite extension's "Browse Samples"
```
NipunaRanasinghe, line 22: This gap correctly identifies that the UI under test belongs to `WSO2.wso2-integrator` rather than this repo, but it stops short of the consequence, which is the main thing a future maintainer needs to know.

`prepareExtensionsForLaunch` (`e2e-playwright-tests/utils/helpers/setup.ts:188`) downloads that extension from the Marketplace at whatever version is current — nothing is pinned. So every selector this scenario depends on ("Explore", "Browse Samples", the "Sample"/"All" filter, "Use this", the `N results` string, `role=article`) can change in a release of a different repo and break Group 2 with no commit here. The count assertions go further and depend on the live remote catalog: `filteredCount < initialCount` requires the catalog to always contain at least one non-sample entry, and `toHaveCount(1)` requires exactly one card matching "Hello World Service".

Please say that explicitly here — two external, unpinned dependencies (extension version, catalog contents), and what to check first when this test starts failing.

Also worth stating: this repo's own `packages/ballerina-visualizer/src/views/BI/SamplesView` remains uncovered by any e2e test. The gap notes it's a different implementation but reads as if it's dead code; if it still ships on some path, that's a coverage hole rather than a non-issue.

packages/ballerina-extension/e2e-test/e2e-playwright-tests/samples/samples.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+                throw new Error(BI_WEBVIEW_NOT_FOUND_ERROR);
+            }
+            await welcomeWebView.waitForLoadState();
+            await welcomeWebView.getByRole('heading', { name: 'WSO2 Integrator' }).waitFor({ timeout: 60000 });
+
+            logStep('Clicking "Explore" to open the samples browser');
+            await welcomeWebView.getByRole('button', { name: 'Explore', exact: true }).click();
+
+            const samplesWebView = await getWebview('Welcome', page);
+            await samplesWebView.getByRole('heading', { name: 'Browse Samples' }).waitFor({ timeout: 60000 });
```
NipunaRanasinghe, line 55: Nit: `getWebview('Welcome', page)` on line 54 re-resolves the same `iframe[title="Welcome"]` that `welcomeWebView` already points at — the samples page renders into that same webview. `switchToIFrame` has a hardcoded 2s settle plus `getWebview`'s own 1s wait, so this costs ~3s for no new frame. The `waitFor` on the "Browse Samples" heading is what actually gates on the view swap, and it works just as well against `welcomeWebView`.

packages/ballerina-extension/e2e-test/e2e-playwright-tests/samples/samples.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import { expect, test } from '@playwright/test';
+import { BI_INTEGRATOR_LABEL, BI_WEBVIEW_NOT_FOUND_ERROR, getWebview, initTest, logStep, page, vscode } from '../utils/helpers';
+import { switchToIFrame } from '@wso2/playwright-vscode-tester';
+
+export default function createTests() {
+    test.describe.serial('Use Samples in the WSO2 Integrator', {
```
NipunaRanasinghe, line 23: Nit: `describe.serial` has no effect with a single test in the block. Harmless and consistent with the rest of the suite, so fine to leave if you expect to add more sample cases here.

dulajdilshan:
@samithkavishke Please attach the correct issue/ticket with the appropriate milestone

PR #713, opened 2026-07-24
Use append-only JSONL persistence for Copilot chat history

### LINKED ISSUE #1861: Move chat-history persistence from full-file JSON rewrites to append-only JSONL
### Current Limitation

Right now, every time anything happens in a conversation (a new message, a tool call, an update to the current turn), BI Copilot saves the whole chat thread back to disk from scratch. It rebuilds the entire history into one big JSON file and rewrites the full thing, every single time.

That's fine early on, but the file grows with the conversation, and the save happens synchronously. So the longer a chat gets, the more work each little update costs, because we're re-serializing and rewriting the whole history just to tack on one new line. In a long session this can add up to noticeable lag, since the extension is repeatedly writing an ever-larger file for what should be a tiny change.

### Suggested Improvement

Switch to an append-only format (JSONL), where each new message or step is just added as a single line instead of rewriting the whole file. Adding a turn then costs about the same whether the conversation is two messages long or two hundred, because we only write the new part rather than everything that came before. On load we replay the lines to rebuild the conversation, and we can occasionally compact the file to clean up anything that's been superseded.

This is also how the MI Copilot already stores its history, so it's a proven approach rather than something new. The large per-turn snapshots already live in their own separate files, so they aren't affected by this change.

### Version

_No response_

SasinduDilshara:
### Solution

Rewires the Ballerina Copilot chat-state store (`packages/ballerina-extension/src/views/ai-panel/chatStateStorage.ts`) to persist chat history through the incremental append-only API instead of rewriting the whole thread file on every mutation.

- Per-generation add/update appends only the changed header and new messages (`appendGeneration`).
- Generation removal appends a tombstone (`removeGenerationRecord`); checkpoint restore appends a truncation (`truncateFromGeneration`).
- Full rewrites remain only for thread creation.

This eliminates the O(n²) per-turn write cost described in the issue.

> **Depends on** wso2/vscode-extensions#2431 (append-only API in `copilot-utilities`). The submodule pointer must be bumped to the merged engine commit before this builds/merges.

Fixes https://github.com/wso2/product-integrator/issues/1861

### DIFF
--- a/packages/ballerina-extension/src/views/ai-panel/chatStateStorage.ts
+++ b/packages/ballerina-extension/src/views/ai-panel/chatStateStorage.ts
@@ -285,6 +285,48 @@ export class ChatStateStorage {
     }
 
     /**
+     * Persist a single generation incrementally (append-only).
+     * Appends only the changed header and/or the model messages not yet
+     * written — instead of rewriting the whole thread. Use for every
+     * per-generation mutation. Relies on the store's per-generation tracking,
+     * which is primed by loadThread()/saveThread() (thread creation + load).
+     */
+    private persistGeneration(projectRootPath: string, threadId: string, generation: Generation): void {
+        const thread = this.storage.get(projectRootPath)?.threads.get(threadId);
+        if (!thread) {
+            return;
+        }
+        try {
+            this.persistenceStore.appendGeneration(
+                projectRootPath,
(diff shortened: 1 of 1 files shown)


SasinduDilshara:
Close this for now and will prioritize later

PR #714, opened 2026-07-24
Provide library annotations to Copilot from the Semantic Model

SasinduDilshara:
### Purpose

Adds annotation details to the Copilot's library context, sourced dynamically from the compiler Semantic Model. Previously the Copilot only received `SERVICE`/`OBJECT_METHOD` annotation definitions from the static service-index and no per-symbol annotation usage at all.

### What this does

- **Definition catalog — all attachment points.** The annotation catalog now covers every attach point (`function`, `type`, `record field`, `parameter`, `return`, `listener`, etc.), extracted from the Semantic Model. It **supplements** the existing service-index catalog (which keeps owning `SERVICE`/`OBJECT_METHOD`) and de-duplicates, so existing output is unchanged.
- **Per-symbol attachments.** Annotations actually attached to API elements (e.g. `@display`, `@constraint`) are now surfaced on functions, parameters, record fields, type definitions, and clients.
- **Frontend rendering.** These are rendered into the Ballerina-syntax library context fed to the LLM (`to-syntax-string.ts`), with attach-point labels broadened and per-symbol attachments emitted inline.

### Testing

- Added `CopilotAnnotationTest` verifying the catalog now includes non-service attach points (e.g. `ballerina/http` `Payload`/`Header`/`Query` on `PARAMETER`/`RECORD_FIELD`) and that per-symbol attachments (e.g. `@display`) are extracted.
- Verified end-to-end in the Copilot: prompts that require recently-added annotations (e.g. `@log:Sensitive {strategy: {replacement}}`) are now produced correctly, whereas they were not before this change.

Fixes https://github.com/wso2/product-integrator/issues/1929

### DIFF
--- a/packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
+++ b/packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
@@ -488,6 +488,7 @@ function selectClients(originalClients: Client[], funcResponse: GetFunctionRespo
             name: originalClient.name,
             description: originalClient.description,
             functions: [],
+            annotations: originalClient.annotations,
         };
 
         const output: (RemoteFunction | ResourceFunction)[] = [];
--- a/packages/ballerina-extension/src/features/ai/utils/libs/library-types.ts
+++ b/packages/ballerina-extension/src/features/ai/utils/libs/library-types.ts
@@ -29,11 +29,18 @@ export interface Link {
 
 export type Category = "internal" | "external";
 
+export interface AnnotationAttachment {
+    name: string;
+    module?: string;
+    value?: string;
+}
+
 export interface Parameter {
     name: string;
     description: string;
     type: Type;
     default?: string;
+    annotations?: AnnotationAttachment[];
 }
 
 export interface ParameterDef {
@@ -59,6 +66,7 @@ export interface Field {
     type: Type;
     default?: string;
     isDeprecated?: boolean;
+    annotations?: AnnotationAttachment[];
 }
 
 export interface UnionValue {
@@ -76,6 +84,7 @@ export interface TypeDefinitionBase {
     description: string;
     type: string;
     isDeprecated?: boolean;
+    annotations?: AnnotationAttachment[];
 }
 
 export interface ConstantTypeDefinition extends TypeDefinitionBase {
@@ -113,6 +122,7 @@ export interface AbstractFunction {
     parameters: Parameter[];
     return: Return;
     isDeprecated?: boolean;
+    annotations?: AnnotationAttachment[];
 }
 
 export interface ResourceFunction extends AbstractFunction {
@@ -139,6 +149,7 @@ export interface Client {
     description: string;
     functions: (RemoteFunction | ResourceFunction)[];
     isDeprecated?: boolean;
+    annotations?: AnnotationAttachment[];
 }
 
 export interface Listener {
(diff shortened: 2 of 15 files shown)


SasinduDilshara:
### Note for reviewers — why this touches 15 files

The change is two related but separable parts. Splitting them out makes the diff much easier to read:

**Part A — annotation definition catalog (the vocabulary).** 4 files:
- `CopilotLibraryManager.java` — merge Semantic-Model annotations into the existing service-index catalog, de-duplicated by `(name, attachmentPoint)`. The service-index keeps owning `SERVICE`/`OBJECT_METHOD`, so existing output is byte-identical.
- `LibraryModelConverter.java` — converts the annotation's constraint `TypeSymbol` into the existing `Type` model (with links), so the config record resolves like any other type.
- `to-syntax-string.ts` — broadens `ATTACHMENT_POINT_LABELS`. **This one is load-bearing, not cosmetic:** `renderAnnotation` returns `null` for any unrecognised attachment point, so without it every new catalog entry would be extracted, serialized, shipped over LSP, and then silently discarded before reaching the LLM.
- `CopilotAnnotationTest.java` — tests.

**Part B — per-symbol annotation attachments (the ground truth).** The remaining files. This surfaces the annotations actually attached to each API element (functions, parameters, record fields, type definitions, clients), read via `Annotatable.annotAttachments()`.

Most of Part B's file count is unavoidable boilerplate: each attachment has to be carried on the object that owns it, so `LibraryFunction`, `Parameter`, `Field`, `TypeDef`, and `Client` each gain a field + getter + setter (**9-11 lines each, nothing else**). `AnnotationAttachment.java` is the DTO, `AnnotationAttachmentExtractor.java` holds the extraction/value-rendering logic, `library-types.ts` mirrors the DTO in TS, and `function-registry.ts` is a **1-line** fix so client-level annotations survive the existing filtering step.

The `AnnotationSymbol` branch was added inside `SymbolProcessor`'s existing `moduleSymbols()` walk rather than as a new loop in the manager, so the catalog costs **zero additional iteration** over module symbols.

### Why Part B is worth including rather than deferring

These annotations are runtime-enforced via `constraint:validate`, so omitting them causes generated code that compiles but fails at runtime. Real examples from published packages:

```ballerina
// ballerinax/twitter, ballerinax/paypal.orders
@constraint:String {maxLength: 2048, minLength: 1}
string query;

// ballerinax/azure.openai.chat, ballerinax/azure.openai.responses
@constraint:String {maxLength: 64}
string safety_identifier?;

// @constraint:Array {minLength: 1} — empty array is invalid
// @constraint:Int {minValue: 100, maxValue: 599} — type-level bound
```

Without Part B the Copilot sees `string query;` and none of its bounds — so it can emit `query: ""` against a `minLength: 1` field. **40 packages** in a single local Central cache carry `@constraint` on fields or types, including `ballerina/ai`, `ballerinax/github`, `ballerinax/jira`, `ballerinax/discord`, `ballerinax/openai.*`, `ballerinax/azure.openai.*`, and `ballerinax/mailchimp.*`.

Happy to split this into two PRs if reviewers would prefer to land Part A first.

### Verification

- `CopilotAnnotationTest` confirms the catalog now includes non-service attachment points (`ballerina/http`: `Payload`/`Header`/`Query` on `PARAMETER`/`RECORD_FIELD`, `Cache` on `RETURN`) and that per-symbol attachments are extracted (`ballerinax/salesforce`: `@display {label: "Salesforce", iconPath: "icon.png"}`).
- Verified live in the Copilot with an A/B run on the same prompt: a prompt requiring a recently-added annotation (`@log:Sensitive {strategy: {replacement: "****"}}`) is generated correctly on this branch and **not** on the base branch, which holds LLM knowledge and README content constant.

PR #719, opened 2026-07-27
Add nightly build support

samithkavishke:
Three related problems in the release automation:

1. **The daily build verified everything but published nothing consumable.** `daily-build.yml` ran the full LS matrix (pack, test, Windows build, Trivy) plus the extension build and e2e suite, but the resulting VSIX only ever existed as a workflow run artifact. Anyone wanting to test the latest `main` had to open a workflow run, find the artifact, and unzip it — and had no stable link to point others at.
2. **The repository was transferred from `wso2/ballerina-vscode` to `ballerina-platform/ballerina-vscode`, and the workflows still hardcoded the old name.** Because `github.repository` now evaluates to the new name, every `if: github.repository == 'wso2/ballerina-vscode'` guard was silently false. That disabled the nightly publish job, all three failure notifications, and the main-sync PR — producing green runs that quietly did nothing. The `repo:`/`REPO:` values were only still working because GitHub redirects renamed repositories.
3. **The `release` composite action was fragile.** It parsed the release ID out of JSON with `sed`, derived the version by cutting the VSIX filename at its last `-`, and never checked whether the VSIX asset upload actually succeeded.

- Publish every successful daily build to a **rolling `nightly` GitHub pre-release**, so there is always one predictable place to get the latest build of `main`.
- Give nightly VSIXes a **`-SNAPSHOT` version** so a nightly build is never mistaken for a release build of the same version.
- Point every workflow at the correct repository, and make the dynamic references rename-proof.
- Harden the `release` action so a failed release or upload fails the job instead of reporting success.

### 1. Rolling `nightly` pre-release

`.github/actions/release/action.yml` gains an optional `tag` input that switches it between two modes:

| `tag` input | Release tag | Release name | Behaviour |
|---|---|---|---|
| empty (default) | `v<version>` | `Release v<version>` | unchanged — a new release per version |
| e.g. `nightly` | `nightly` | `Nightly Build (v<version>)` | deletes the existing release **and** git tag ref first, then recreates |

`daily-build.yml` passes `tag: nightly`. Because the tag is reused, the download URL stays stable across nights. Releases are created with `prerelease: true`, so nightlies carry the GitHub **Pre-release** badge and never appear as "Latest". The Google Chat notification links were updated to use the resolved `releaseTag` rather than a hardcoded `v<version>`, so the card's "Download VSIX" button points at the right place in both modes.

The nightly publish job is gated on **all** verification jobs — `ls-ubuntu-pack`, `ls-ubuntu-test`, `ls-windows-build`, `ls-trivy`, and `Build` — so a Trivy or test failure blocks the nightly.

### 2. `-SNAPSHOT` versioning for nightlies

A new `snapshot` input threads through `build.yml` → `actions/build` → `actions/updateVersion`, where it adds a `Set snapshot version` step producing `<version>-SNAPSHOT` (e.g. `5.13.0-SNAPSHOT`).

- It is **mutually exclusive** with the existing timestamp suffix: the `Set prerelase version` step now also requires `snapshot != 'true'`, so the two can never both apply.
- `-SNAPSHOT` is valid semver, and `vsce package` validates the version with nothing more than `semver.valid()` (`@vscode/vsce/out/validation.js`), so packaging accepts it. Nightly VSIXes are never published to a marketplace, so the suffix never reaches one.
- The step strips any pre-existing prerelease id (`${currentVersion%%-*}`) so re-runs are idempotent rather than yielding `…-SNAPSHOT-SNAPSHOT`.

Existing version behaviour is untouched:

| Trigger | Resulting version |
|---|---|
| Daily build (new) | `5.13.0-SNAPSHOT` |
| `release-vsix.yml`, `isPreRelease=true` | `5.13.26072909` (timestamp as patch) |
| `release-vsix.yml`, `version=patch` | `5.12.4` |
| `release-vsix.yml`, `version=minor` | `5.14.0` |
| `release-vsix.yml`, `version=major` | `6.0.0` |

### 3. Filename parsing that survives a prerelease id

`actions/release` and `actions/dailyBuildNotification` both derived the version with `${file##*-}`, which cuts at the **last** dash. For `ballerina-5.13.0-SNAPSHOT.vsix` that yields `SNAPSHOT`, producing a release titled `Nightly Build (vSNAPSHOT)`. Both now strip the `<name>-` prefix and the `.vsix` suffix instead.

### 4. Hardened `Remove timestamp` guard

`updateVersion`'s strip step guarded on the length of the **full** version and then applied `${…:0:-8}` to the base. For `5.13.0-SNAPSHOT` the base is only 6 characters, and a negative length longer than the string is a fatal bash expansion error. The guard now measures the base version and requires `> 8`.

This also fixes a pre-existing bug it exposed: a version exactly 8 characters long (e.g. `5.13.100`) passed the old `>= 8` check, stripped to an empty string, was padded to `0`, and rewrote `package.json` to version **`0`**.

### 5. Correct repository references

- Fork guards (which need a literal to compare against) → `'ballerina-platform/ballerina-vscode'`: `daily-build.yml`, `release-vsix.yml`, `publish-vsix.yml`, `sync-main-with-releases.yml`.
- API targets → `${{ github.repository }}`, so a future rename cannot break them: `daily-build.yml`, `release-vsix.yml`, `publish-vsix.yml`.
- `CONTRIBUTING.md` clone URL → new org.

Deliberately left alone: `publisher: "wso2"`, `itemName=WSO2.ballerina`, and `open-vsx.org/extension/wso2/ballerina` are **marketplace publisher identities**, not repository references — changing them would orphan the published extension. The `wso2/vscode-extensions` links in the PR template correctly refer to the submodule's source repository.

### 6. Release-creation hardening (CodeRabbit review)

The release step was rewritten:

- `set -euo pipefail`; release ID extracted with `jq` instead of `sed`; `target_commitish` set explicitly.
- Token and other values passed via `env:` rather than interpolated into the script body.
- **Upload verification** — `curl` exits 0 on a non-2xx HTTP response, so `set -e` alone let a rejected upload (auth, size limit, rate limit) pass as success. This mattered most on the rolling path, where the previous nightly has already been deleted: an unchecked failure would leave an assetless `nightly` release and a notification linking to a 404. The upload response is now checked for an asset `id`, failing the step with the response body otherwise.
- **Bounded timeouts** — `--connect-timeout 15` everywhere, with `--max-time 60` for metadata calls and `--max-time 900` for the asset upload (the VSIX is ~75–97 MB, so a single blanket 60s would have broken every release).

**N/A** — this PR changes only GitHub Actions workflows, composite actions, and `CONTRIBUTING.md`. No extension or webview code is touched, so there is no UI surface.

- As a **developer or QA engineer**, I want to install the latest build of `main` from a single predictable GitHub release instead of hunting through workflow run artifacts.
- As a **developer**, I want an installed nightly to be identifiable as a nightly from its version string, so a `-SNAPSHOT` build is never confused with a released version.
- As a **maintainer**, I want nightly and release automation to actually run after the repository transfer, and to be notified when it fails, instead of seeing green runs that skipped their publish step.
- As a **release manager**, I want a failed VSIX upload to fail the workflow rather than silently produce a release with no artifact attached.

Successful daily builds are now published to a rolling `nightly` GitHub pre-release, versioned `<version>-SNAPSHOT`. The release tag is reused, so the latest nightly is always available at the same location. Also fixes release-automation jobs that stopped running after the repository transfer to `ballerina-platform`.

**N/A for product documentation** — the change is confined to CI/release automation and has no effect on extension behaviour or user-facing features.

Repository documentation updated in this PR: `CONTRIBUTING.md` clone URL now points at `ballerina-platform/ballerina-vscode`.

> Note: `.github/workflows/README.md` is stale independently of this PR (it documents `ls-build-master.yml` and `ls-trivy.yml`, neither of which exists). Left out to keep this PR scoped; worth a follow-up.

**N/A** — internal build automation with no training content impact.

**N/A** — CI/release plumbing only. Nothing here is part of any product surface that certification exams cover.

No new unit or integration tests: the change set is GitHub Actions YAML and shell, which the repo's jest and Playwright suites do not cover. Verification performed instead:

- **YAML validity** — all 8 workflows and all 7 composite actions parse.
- **Version stamping** — replayed the `updateVersion` steps end to end across every input combination (`isPreRelease` × `patch`/`minor`/`major`/`N/A` × `snapshot`), from clean, timestamped, and `-SNAPSHOT` starting versions. Confirmed the four documented outcomes, confirmed `-SNAPSHOT` is idempotent on re-run, and confirmed the new guard changes behaviour *only* for the two previously-broken shapes (`5.13.0-SNAPSHOT`, `5.13.100`) while leaving all real timestamped cases identical.
- **Filename parsing** — verified the new prefix/suffix strip against `-SNAPSHOT`, plain, and timestamped VSIX names, and that all three still match the existing `ls` glob.
- **`vsce` compatibility** — confirmed `vsce` validates the version via `semver.valid()` only, and that `5.13.0-SNAPSHOT` is valid semver.
- **Upload verification** — exercised the new check with a stubbed `curl` across four responses: success returns exit 0; auth error, size limit, and empty body each exit 1 and print the response body.

**Not yet exercised end to end:** the nightly path has not run as a real scheduled build, so the first scheduled run (or a manual `workflow_dispatch` of Daily Build) is the real confirmation. Recommend a manual dispatch before merge.

## Security checks
- Ran FindSecurityBugs plugin and verified report? **N/A** — no Java/source changes in this PR; the language server is untouched.
- Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? **yes** — all credentials remain `${{ secrets.* }}` / `${{ github.token }}` references.

Incidental hardening: the release step now passes the token and other values through `env:` instead of interpolating them directly into the shell script body, and the Trivy scan continues to gate the nightly publish.

**N/A** — no samples relate to build automation.

## Related PRs

None. CodeRabbit raised one actionable review comment on this PR (unverified VSIX asset upload in `.github/actions/release/action.yml`); it is addressed in commit "Address coderabbit suggestions" and described under Approach §6.

No data or code migration. Two operational notes:

1. The first nightly run **creates** the `nightly` git tag and release. Every subsequent run **deletes and recreates** both, so the tag intentionally does not preserve history — it always points at the newest build. The job needs `contents: write`, which it already has.
2. The repository-reference fixes re-enable jobs that had been silently skipped since the transfer (nightly publish, failure notifications, main-sync PR). Expect notification traffic to resume on the next failure.

- **CI (unchanged by this PR):** GitHub-hosted `ubuntu-latest` and `windows-latest`; JDK 21.0.3 (Temurin); Node 22.x; pnpm 10.11.0; Ballerina 2201.13.2 (extension) / 2201.13.3 (LS).
- **Local verification:** macOS (darwin 25.5.0), Node 22.22.1 — used for the YAML, version-stamping, filename-parsing, and upload-validation checks described under Automation tests.

- **`vsce` version validation is looser than the Marketplace.** `@vscode/vsce` only calls `semver.valid()`, so a prerelease identifier such as `-SNAPSHOT` packages fine. The Marketplace's stricter `x.y.z` expectation is why the existing prerelease scheme concatenates a timestamp onto the patch instead of using a `-beta.N` suffix — but since nightlies are never published to a marketplace, that constraint does not apply to them.
- **`npm version` parses loosely and normalises leading zeros.** `5.13.0` + `26072909` produces `5.13.026072909`, which strict semver rejects, but npm stores it as `5.13.26072909`. This is why existing prerelease tags read the way they do, and it is worth knowing before touching that logic.
- **Bash negative substring lengths are a trap.** `${var:0:-8}` is a fatal expansion error — not an empty string — when the negative length exceeds the string length. Under `shell: bash` (which runs with `-e`) that fails the whole step, so any such expression needs a length guard ahead of it.
- **`curl` exits 0 on HTTP errors.** `set -e` gives no protection against a 4xx/5xx response; the response body must be inspected explicitly (or `--fail` used). Worth auditing wherever a workflow talks to an API with `curl`.

## Summary

- Added nightly build support by stamping VSIX versions with a `-SNAPSHOT` prerelease suffix and publishing successful builds as a rolling `nightly` GitHub pre-release.
- Extended CI/build automation with a `snapshot` input that ensures consistent, idempotent version stamping for nightly builds.
- Hardened release automation to support reusable rolling releases via optional `tag` selection, improved VSIX version/filename parsing, and strengthened upload/asset handling and release metadata updates.
- Updated notification and parsing logic to correctly derive versions from nightly/prerelease VSIX filenames.
- Updated workflow guards, release target repositories, and URLs/conditions to reflect the post-transfer `ballerina-platform/ballerina-vscode` repository.
- Updated contributor documentation to use the new `ballerina-platform/ballerina-vscode` clone URL.

### DIFF
--- a/.github/actions/release/action.yml
+++ b/.github/actions/release/action.yml
@@ -67,23 +67,29 @@ runs:
         set -euo pipefail
         API="https://api.github.com/repos/${REPO}"
 
+        # Bounded timeouts so a hung GitHub API call can't stall the job until
+        # the job-level timeout. The metadata calls are small; the asset upload
+        # ships a ~100MB VSIX, so it gets a much larger budget.
+        CURL_API=(curl -sS --connect-timeout 15 --max-time 60)
+        CURL_UPLOAD=(curl -sS --connect-timeout 15 --max-time 900)
+
         # Rolling tag (e.g. nightly): delete the previous release + git tag so
         # the tag always points at the newest build.
         if [ "${ROLLING}" = "true" ]; then
-          existingId=$(curl -sS -H "Authorization:token ${GH_TOKEN}" \
+          existingId=$("${CURL_API[@]}" -H "Authorization:token ${GH_TOKEN}" \
             -H "Accept: application/vnd.github.v3+json" \
             "${API}/releases/tags/${TAG}" | jq -r '.id // empty')
           if [ -n "${existingId}" ]; then
             echo "Deleting existing release ${existingId} for tag ${TAG}"
-            curl -sS -X DELETE -H "Authorization:token ${GH_TOKEN}" \
+            "${CURL_API[@]}" -X DELETE -H "Authorization:token ${GH_TOKEN}" \
               "${API}/releases/${existingId}"
           fi
           # Remove the git tag ref too (ignore if it doesn't exist).
-          curl -sS -X DELETE -H "Authorization:token ${GH_TOKEN}" \
+          "${CURL_API[@]}" -X DELETE -H "Authorization:token ${GH_TOKEN}" \
             "${API}/git/refs/tags/${TAG}" || true
         fi
 
-        createResponse=$(curl -sS -X POST \
+        createResponse=$("${CURL_API[@]}" -X POST \
           -H "Accept: application/vnd.github.v3+json" \
           -H "Authorization:token ${GH_TOKEN}" \
           "${API}/releases" \
@@ -99,11 +105,23 @@ runs:
           exit 1
         fi
 
-        curl -sS -X POST \
+        # curl exits 0 on a non-2xx HTTP response, so 'set -e' alone would let a
+        # rejected upload (auth, size limit, rate limit) pass as success. That is
+        # especially bad on the rolling path: the previous nightly has already
+        # been deleted above, so an unchecked failure would leave an assetless
+        # 'nightly' release and a notification linking to a 404.
+        uploadResponse=$("${CURL_UPLOAD[@]}" -X POST \
           -H "Authorization:token ${GH_TOKEN}" \
           -H "Content-Type:application/octet-stream" \
           --data-binary "@${VSIX}" \
-          "https://uploads.github.com/repos/${REPO}/releases/${id}/assets?name=${VSIX}"
+          "https://uploads.github.com/repos/${REPO}/releases/${id}/assets?name=${VSIX}")
+        assetId=$(echo "${uploadResponse}" | jq -r '.id // empty')
+        if [ -z "${assetId}" ]; then
+          echo "Failed to upload ${VSIX} to release ${id}:" >&2
+          echo "${uploadResponse}" >&2
+          exit 1
+        fi
+        echo "Uploaded ${VSIX} as asset ${assetId}"
 
     - name: "Release Notification"
       shell: bash
--- a/.github/actions/release/action.yml
+++ b/.github/actions/release/action.yml
@@ -73,55 +73,109 @@ runs:
         CURL_API=(curl -sS --connect-timeout 15 --max-time 60)
         CURL_UPLOAD=(curl -sS --connect-timeout 15 --max-time 900)
 
-        # Rolling tag (e.g. nightly): delete the previous release + git tag so
-        # the tag always points at the newest build.
+        # On the rolling path (e.g. nightly) the release is REUSED rather than
+        # deleted and recreated. Nothing is destroyed until the replacement
+        # artifact is uploaded and verified, so a failed run always leaves the
+        # previous nightly downloadable. That matters because the whole point of
+        # a fixed tag is a URL that keeps working; deleting first would take
+        # 'releases/download/<tag>/...' offline until the next successful build.
+        existingId=""
+        existingAssets="[]"
         if [ "${ROLLING}" = "true" ]; then
-          existingId=$("${CURL_API[@]}" -H "Authorization:token ${GH_TOKEN}" \
+          existing=$("${CURL_API[@]}" -H "Authorization:token ${GH_TOKEN}" \
             -H "Accept: application/vnd.github.v3+json" \
-            "${API}/releases/tags/${TAG}" | jq -r '.id // empty')
-          if [ -n "${existingId}" ]; then
-            echo "Deleting existing release ${existingId} for tag ${TAG}"
-            "${CURL_API[@]}" -X DELETE -H "Authorization:token ${GH_TOKEN}" \
-              "${API}/releases/${existingId}"
-          fi
-          # Remove the git tag ref too (ignore if it doesn't exist).
-          "${CURL_API[@]}" -X DELETE -H "Authorization:token ${GH_TOKEN}" \
-            "${API}/git/refs/tags/${TAG}" || true
+            "${API}/releases/tags/${TAG}")
+          existingId=$(echo "${existing}" | jq -r '.id // empty')
+          existingAssets=$(echo "${existing}" | jq -c '.assets // []')
         fi
 
-        createResponse=$("${CURL_API[@]}" -X POST \
-          -H "Accept: application/vnd.github.v3+json" \
-          -H "Authorization:token ${GH_TOKEN}" \
-          "${API}/releases" \
-          -d "$(jq -n \
-            --arg tag "${TAG}" \
-            --arg name "${RELEASE_NAME}" \
-            --arg commit "${GITHUB_SHA}" \
-            '{tag_name: $tag, target_commitish: $commit, draft: false, name: $name, prerelease: true}')")
-        id=$(echo "${createResponse}" | jq -r '.id // empty')
-        if [ -z "${id}" ]; then
-          echo "Failed to create release:" >&2
-          echo "${createResponse}" >&2
-          exit 1
+        if [ -n "${existingId}" ]; then
+          echo "Reusing existing release ${existingId} for tag ${TAG}"
+          id="${existingId}"
+        else
+          createResponse=$("${CURL_API[@]}" -X POST \
+            -H "Accept: application/vnd.github.v3+json" \
+            -H "Authorization:token ${GH_TOKEN}" \
+            "${API}/releases" \
+            -d "$(jq -n \
+              --arg tag "${TAG}" \
+              --arg name "${RELEASE_NAME}" \
+              --arg commit "${GITHUB_SHA}" \
+              '{tag_name: $tag, target_commitish: $commit, draft: false, name: $name, prerelease: true}')")
+          id=$(echo "${createResponse}" | jq -r '.id // empty')
+          if [ -z "${id}" ]; then
+            echo "Failed to create release:" >&2
+            echo "${createResponse}" >&2
+            exit 1
+          fi
+        fi
+
+        # Asset names must be unique within a release, and a nightly VSIX has the
+        # same filename every run — so the replacement is uploaded under a temp
+        # name and renamed only after the old one is gone. Clear any temp asset
+        # left behind by a previously failed run, or this upload would 422.
+        staleId=$(echo "${existingAssets}" | jq -r --arg n "${VSIX}.new" \
+          '.[] | select(.name == $n) | .id' | head -n 1)
+        if [ -n "${staleId}" ]; then
+          echo "Removing stale temp asset ${staleId} (${VSIX}.new)"
+          "${CURL_API[@]}" -X DELETE -H "Authorization:token ${GH_TOKEN}" \
+            "${API}/releases/assets/${staleId}"
         fi
 
         # curl exits 0 on a non-2xx HTTP response, so 'set -e' alone would let a
-        # rejected upload (auth, size limit, rate limit) pass as success. That is
-        # especially bad on the rolling path: the previous nightly has already
-        # been deleted above, so an unchecked failure would leave an assetless
-        # 'nightly' release and a notification linking to a 404.
+        # rejected upload (auth, size limit, rate limit) pass as success, leaving
+        # a release whose artifact silently never arrived.
         uploadResponse=$("${CURL_UPLOAD[@]}" -X POST \
           -H "Authorization:token ${GH_TOKEN}" \
           -H "Content-Type:application/octet-stream" \
           --data-binary "@${VSIX}" \
-          "https://uploads.github.com/repos/${REPO}/releases/${id}/assets?name=${VSIX}")
+          "https://uploads.github.com/repos/${REPO}/releases/${id}/assets?name=${VSIX}.new")
         assetId=$(echo "${uploadResponse}" | jq -r '.id // empty')
         if [ -z "${assetId}" ]; then
           echo "Failed to upload ${VSIX} to release ${id}:" >&2
           echo "${uploadResponse}" >&2
           exit 1
         fi
+
+        # The new artifact is safely uploaded — only now retire the old one.
+        oldId=$(echo "${existingAssets}" | jq -r --arg n "${VSIX}" \
+          '.[] | select(.name == $n) | .id' | head -n 1)
+        if [ -n "${oldId}" ]; then
+          echo "Deleting superseded asset ${oldId} (${VSIX})"
+          "${CURL_API[@]}" -X DELETE -H "Authorization:token ${GH_TOKEN}" \
+            "${API}/releases/assets/${oldId}"
+        fi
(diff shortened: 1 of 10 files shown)


.github/actions/release/action.yml
```diff
@@ -23,25 +31,79 @@ runs:
...
+          echo "Failed to create release:" >&2
+          echo "${createResponse}" >&2
+          exit 1
+        fi
+
+        curl -sS -X POST \
+          -H "Authorization:token ${GH_TOKEN}" \
+          -H "Content-Type:application/octet-stream" \
+          --data-binary "@${VSIX}" \
+          "https://uploads.github.com/repos/${REPO}/releases/${id}/assets?name=${VSIX}"
```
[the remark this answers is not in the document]
samithkavishke: Addressed

NipunaRanasinghe:
@samithkavishke Please update the PR description

samithkavishke:
> @samithkavishke Please update the PR description

Added

NipunaRanasinghe:
@coderabbitai review

.github/workflows/daily-build.yml
```diff
@@ -266,6 +266,15 @@ jobs:
...
       runFastTests: true
       runBalE2ETests: true
+      # Daily builds ship only to a rolling GitHub `nightly` release, never the
+      # Marketplace, so instead of the isPreRelease timestamp suffix they get a
+      # '-SNAPSHOT' version (5.13.0-SNAPSHOT). It is valid semver so vsce
+      # accepts it, and it makes a nightly VSIX unmistakable next to a release
+      # build of the same version. The bundled LS is built from source
+      # (lsSource defaults to 'local'), so it always matches the extension
+      # version regardless of this.
+      isPreRelease: false
```
NipunaRanasinghe, line 287: `isPreRelease` was previously unset here, so it took `build.yml`'s `default: true`. It doesn't only select a version suffix — `actions/build/action.yml:138` exports it into the rush build environment, and the submodule's `workspaces/common-libs/scripts/package-vsix.js` reads it:

```js
const isPreRelease = process.env.isPreRelease === 'true';
const args = ['vsce', 'package', '--no-dependencies'];
if (isPreRelease) {
  args.push('--pre-release');
}
```

So nightlies move from `vsce package --pre-release` to plain `vsce package`, and a nightly VSIX is no longer marked as a pre-release extension build — the opposite of the previous daily-build behaviour, and it affects anything gated on pre-release status at install time.

This may well be required rather than accidental, if `--pre-release` and a `-SNAPSHOT` semver tag don't coexist in `vsce`. But the description says "Existing version behaviour is untouched", and the comment above justifies `isPreRelease: false` purely in terms of the timestamp suffix and the bundled LS, so as written a reader wouldn't know the packaging flag changed.

Could you either confirm it's intentional and note the `--pre-release` consequence in the comment, or keep the flag if nightlies are still meant to install as pre-release builds?

.github/actions/release/action.yml
```diff
@@ -23,25 +31,97 @@ runs:
...
+
+        # Bounded timeouts so a hung GitHub API call can't stall the job until
+        # the job-level timeout. The metadata calls are small; the asset upload
+        # ships a ~100MB VSIX, so it gets a much larger budget.
+        CURL_API=(curl -sS --connect-timeout 15 --max-time 60)
+        CURL_UPLOAD=(curl -sS --connect-timeout 15 --max-time 900)
+
+        # Rolling tag (e.g. nightly): delete the previous release + git tag so
+        # the tag always points at the newest build.
+        if [ "${ROLLING}" = "true" ]; then
```
NipunaRanasinghe, line 84: The upload verification further down is a real improvement, and its comment identifies exactly the right hazard — but deleting first means it only fixes half of it.

Order here is: delete release → delete tag → create release → upload asset. Failing the step on a bad upload stops the notification from linking to a 404, but it can't bring the deleted asset back. Between a failed upload and the next successful nightly, `releases/download/nightly/...` is dead and the `nightly` release sits there with no asset. Given the feature's selling point is "one predictable place to get the latest build", that's the window users would actually hit.

An alternative that closes it: look up the existing release, `PATCH /releases/{id}` to retarget `target_commitish`/`name`, upload the new asset, verify, then delete the *old* asset. The tag never stops resolving and there's no point where the release exists without an artifact. Slightly more API calls, but no destructive step before the thing that can fail.

.github/actions/updateVersion/action.yml
```diff
@@ -17,14 +24,18 @@ runs:
...
         cd packages/${{ inputs.path }}
         currentVersion=$(node -p "require('./package.json').version")
-        if [[ ${#currentVersion} -ge 8 ]]; then
-          currentVersionWOPreId=${currentVersion%-*}
+        currentVersionWOPreId=${currentVersion%-*}
+        # Only strip when the base version is long enough to actually carry an
+        # 8-char timestamp. A plain or '-SNAPSHOT' version (e.g. 5.13.0-SNAPSHOT
+        # -> base 5.13.0, 6 chars) has nothing to strip, and ':0:-8' on a
+        # shorter string is a fatal bash expansion error.
+        if [[ ${#currentVersionWOPreId} -gt 8 ]]; then
```
NipunaRanasinghe: this correctly handles the `-SNAPSHOT` crash and the exactly-8-char case described in the PR, but a 9+ char base that isn't a timestamped version is still mangled. `5.13.1000` (9 chars) passes `-gt 8`, `${currentVersionWOPreId:0:-8}` leaves `5`, which ends in a digit so the padding branch doesn't fire, and it runs `npm version 5`..

NipunaRanasinghe (commented):
LGTM except for some minor suggestions

.github/actions/updateVersion/action.yml
samithkavishke: Ack, addressed

.github/workflows/daily-build.yml
samithkavishke: This is the intended behavior. The pre-release tag is not tied to GitHub pre-releases (or any other release mechanism). It is only used for packaging the Ballerina Extension that is published to the Extension Marketplace. Since nightly builds are not published to the Marketplace, this is no longer a concern.

samithkavishke: Added a comment explaining in the workflow file

.github/actions/release/action.yml
samithkavishke: Addressed

.github/actions/release/action.yml
```diff
@@ -23,25 +31,151 @@ runs:
...
+            -d "$(jq -n --arg name "${RELEASE_NAME}" \
+              '{name: $name, draft: false, prerelease: true}')" > /dev/null
+          "${CURL_API[@]}" -X PATCH \
+            -H "Accept: application/vnd.github.v3+json" \
+            -H "Authorization:token ${GH_TOKEN}" \
+            "${API}/git/refs/tags/${TAG}" \
+            -d "$(jq -n --arg sha "${GITHUB_SHA}" \
+              '{sha: $sha, force: true}')" > /dev/null
+          echo "Moved tag ${TAG} to ${GITHUB_SHA}"
+        fi
```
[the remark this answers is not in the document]
samithkavishke: Addressed

NipunaRanasinghe:
@coderabbitai review

.github/actions/release/action.yml
```diff
@@ -23,25 +31,164 @@ runs:
...
+          echo "Failed to rename asset ${assetId} to ${VSIX}:" >&2
+          echo "${renameResponse}" >&2
+          exit 1
+        fi
+        echo "Uploaded ${VSIX} as asset ${assetId}"
+
+        # Refresh the reused release's title, and move the tag to this commit so
+        # the rolling tag tracks the newest build. Both happen after a verified
+        # upload, so tag and artifact never disagree on a failed run.
+        if [ -n "${existingId}" ]; then
```
NipunaRanasinghe, line 164: Minor, non-blocking: this guards on `existingId`, but the tag move should be conditioned on `ROLLING`.

If the `nightly` release is deleted while the tag survives — deleting a release in the UI leaves the tag behind — `existingId` is empty, so we take the create branch. GitHub ignores `target_commitish` when the tag already exists, so the release attaches to the old commit, and this block is then skipped, so nothing moves the tag.

Impact is small: the VSIX and the download URL are both correct, only the tag/release target is stale, and the next run repairs it. Suggest `if [ "${ROLLING}" = "true" ]; then` — the release-name PATCH inside just becomes redundant on the create path.

NipunaRanasinghe (approved):

.github/actions/release/action.yml
samithkavishke: Address in another PR

PR #723, opened 2026-07-28
Add a usage re-check to the header when the limit is reached

RNViththagan:
Fixes https://github.com/wso2/product-integrator/issues/1936

Once the usage limit was reached, the chat input stayed disabled for the rest of the session even after the quota reset. Nothing re-checked usage in the meantime — the input is disabled, so no turn ran and no completion event fired — so the only way to recover was to close and reopen the window.

- Show a refresh action in the usage header while the limit is reached, so usage can be re-checked on demand
- Keep the previous usage state when a manual re-check fails, instead of reporting that the quota is back and sending the next request into another rejection

image

## Summary

- Added a refresh control to the usage header when the Copilot quota is reached, enabling users to re-check availability without reopening the chat.
- Preserved the existing usage state when a manual re-check fails, keeping the chat disabled until quota status is confirmed.
- Added loading and alignment styles for the new refresh control.

### DIFF
--- a/packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/index.tsx
+++ b/packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/index.tsx
@@ -581,7 +581,7 @@ const AIChat: React.FC = () => {
         return parts.length > 0 ? parts.join(' ') : 'less than a minute';
     };
 
-    const fetchUsage = async () => {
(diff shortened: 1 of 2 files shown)


RNViththagan:
@coderabbitai review

SasinduDilshara (approved):

SasinduDilshara (approved):

PR #724, opened 2026-07-28
Suggest follow-up actions after a Copilot turn

RNViththagan:
Fixes https://github.com/wso2/product-integrator/issues/1934

After a turn ends, Copilot now offers a few short follow-up actions as clickable chips. Clicking one fills it into the chat input so it can be edited before sending. Suggestions are generated from the last exchange plus the recent conversation, so they follow what was actually built rather than being generic starters.

- Suggest follow-up actions after a completed turn, generated from the turn and the last few exchanges
- Offer a Continue action plus contextual alternatives when a turn is stopped part-way
- Offer a retry plus failure-aware suggestions when a turn fails, based on what went wrong
- Offer a Continue action when a turn is cut short by the usage limit, shown once the limit resets
- Keep suggestions in scope and free of technical jargon, and never phrased as questions to the user
- Show the feedback bar on responses that include code changes, which previously hid it
- Add `ballerina.copilot.followupSuggestions` to turn the feature off

Screenshot 2026-07-28 at 1 47 48 PM

### DIFF
--- a/packages/ballerina-core/src/rpc-types/ai-panel/index.ts
+++ b/packages/ballerina-core/src/rpc-types/ai-panel/index.ts
@@ -15,7 +15,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import { LoginMethod } from "../../state-machine-types";
+import { LoginMethod, FollowupSuggestion } from "../../state-machine-types";
 import {
     TestGenerationMentions,
     RequirementSpecification,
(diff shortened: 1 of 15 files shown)


RNViththagan:
@coderabbitai review

SasinduDilshara (approved):

SasinduDilshara (approved):

PR #728, opened 2026-07-29
Agent nodes, creation, and tool forms in the visualizer (3/4)

dan-niles:
Part **3 of 4** of the agent definition designer work, split out of #656.

Related to https://github.com/wso2/product-integrator/issues/1946
Resolves https://github.com/wso2/product-integrator/issues/2023

https://github.com/user-attachments/assets/3471d325-df1a-401d-b7ef-5859bb38245d

Adds the visualizer side of the agent foundation: agent nodes in the flow and focus diagrams, agent
listing in the artifacts sidebar, the Add Agent popup and search, agent-as-tool UI, and memory
configuration. Consolidates `AgentTypeNode` into `AgentNodeWidget`, extracts `BaseAgentNodeModel`,
and shares agent editing between the flow and focus diagrams via `useAgentEditorController`.
Removes the Agents Overview page, which duplicated the project-explorer sidebar.

### Review focus
- Focused-agent rendering vs. regular flow-diagram behaviour; no regressions in function, service,
  or package views.
- The shared agent editor controller — both diagrams must drive the same editing flow.

### Independent of the LS PRs
Shares **zero files** with `feat/agent-defs-ls-*`, so this can be reviewed in parallel with them.
Note the UI calls LS RPC endpoints added in #PR1; because everything lands on `feat/agent-changes`
first and `main` only sees the completed feature, merge order does not matter.

## Summary

- Added agent nodes and agent-type nodes to flow and focus diagrams, including shared models, rendering, sizing, navigation, breakpoints, and inline chat support.
- Introduced agent creation, search, configuration, and project-explorer integration, replacing the redundant Agents Overview experience.
- Added agent-as-tool workflows, memory configuration, MCP/tool management, and shared editing through `useAgentEditorController`.
- Improved form capabilities with read-only fields, connection creation, dynamic searches, derived values, and agent-specific layouts.
- Consolidated agent metadata and flow-node utilities, and replaced legacy tool-generation RPCs with source-code generation using AGENT_TOOL nodes.
- Extended artifact, search, state-machine, and RPC types to support agents and refined agent navigation and project updates.

PR #656, opened 2026-07-01
Make agents first-class artifacts and add support for agent definitions

dan-niles:
## Summary

This PR makes AI agents first-class artifacts in WSO2 Integrator and introduces reusable agent definitions. Users can discover, create, configure, reuse, and compose agents across an integration and its workspace.

## Features

1. **Agents as first-class project artifacts**

   Agents now appear alongside other project artifacts, making existing agent instances easier to find, open, and manage.

2. **Discover and add reusable agents**

   The Add Agent experience lets users discover reusable agent definitions from Ballerina Central or the user's workspace project and add them to an integration as configured instances.

3. **Workspace agent-definition discovery**

   Reusable agent definitions in workspace projects can be discovered and used without first publishing them to Central.

4. **Create dependencies while configuring an agent**

   Forms can select an existing model provider, memory, or client connection, or create the required dependency in place and immediately use it.

5. **Focused agent diagrams**

   Opening an agent shows a focused view of its configuration, including its model provider, memory, and tools.

6. **Clear separation between agent calls and configuration**

   Agent call nodes present execution details and provide navigation back to the configured agent, while configuration remains at the agent instance.

7. **Unified agent-tool authoring**

   Custom tools, function tools, connection tools, MCP tools, and class-owned tools follow one consistent authoring experience.

8. **Use one agent as another agent’s tool**

   An agent can delegate work to another configured agent, enabling multi-agent compositions.

9. **Agent Definition Designer**

   A dedicated designer supports authoring reusable agent classes.

   - **Create definitions** in the current integration or in a new library package.
   - **Edit the definition identity** through its name and description.
   - **Configure the inner agent**, including its prompt, model, memory, and settings.
   - **Manage class-owned tools**, including custom, function, connection, MCP, and agent tools.
   - **Manage initialization parameters** required when an integration creates an instance.
   - **Configure the response type** exposed by the definition.

PR #730, opened 2026-07-29
Agent definition designer (4/4)

dan-niles:
Part **4 of 4** of the agent definition designer work, split out of #656.

https://github.com/user-attachments/assets/7ac10677-8c8d-4e30-b352-3bbca91a4799

Related to https://github.com/wso2/product-integrator/issues/1946

**Layer: frontend (TypeScript only).** Touches no Java.

⚠️ **Depends on #728 — review after it merges.** Based on `feat/agent-defs-ui-1-base`, which it
overlaps on 72 files.

The agent definition designer itself: the `AgentDefinitionDesigner` view for editing an agent class
(model provider, memory, tools), the agent definition and agent tool forms, and the node-reference
selector generalised from connections to model providers, memories, agents, and clients.

### Review focus
- The largest UI surface in this series. Watch for duplicated agent-tool forms/flows, unnecessary
  state, class-owned dependency wiring, and refresh ordering after source edits.
- `AgentDefinitionDesigner.tsx` is ~1.6k lines and is the main thing to read.
- Agent-tool form logic moved out of `FunctionForm` into `AgentToolForm` — confirm nothing was lost.


dan-niles:
@coderabbit review

dan-niles:
@coderabbit review

dan-niles:
@coderabbit review

dan-niles:
@coderabbit review

packages/ballerina-visualizer/src/views/BI/AgentDefinitionEditor/AgentDefinitionDesigner.tsx
```diff
@@ -0,0 +1,1583 @@
...
+
+function promptText(raw?: string): string {
+    if (!raw) return "";
+    let s = raw.trim();
+    s = s.replace(/^string\s*`/, "");
+    if (s.endsWith("`")) s = s.slice(0, -1);
+    if (s.startsWith("`") && s.endsWith("`")) s = s.slice(1, -1);
+    if (s.startsWith('"') && s.endsWith('"')) s = s.slice(1, -1);
+    return s;
+}
```
NipunaRanasinghe: Line 328 strips the closing backtick before the paired check on line 329 can run, so that check is unreachable and the opening backtick leaks into the output:

```
'`Support Agent`'     -> '`Support Agent'     <-- leading backtick survives
'"Support Agent"'     -> 'Support Agent'      ok
'string `Be helpful`' -> 'Be helpful'         ok
```

Only the bare-backtick form breaks, and it renders with a stray `` ` `` in the designer.

Worth zooming out, because this is now the third implementation of the same operation and the second carrying this exact ordering bug:

1. `unwrap` in `FocusFlowDiagram/agent.ts` — same defect. This PR *deletes* its `if (v.startsWith("`") && v.endsWith("`"))` branch. That branch was genuinely dead code, but only because the unconditional `.replace(/`$/, "")` on the line above pre-empted it. Removing it leaves the bug in place and removes the signal that backtick values were meant to be handled — `parseSystemPrompt` right below still matches the `` `[^`]*` `` alternative for both `role` and `instructions`, so they are an intended input.
2. `promptText` here — new in this PR, same defect.
3. `stripWrappingQuotes` in `bi-diagram/src/components/nodes/agentNodeUtils.ts` — handles neither bare backticks nor concatenations (`"a" + name + "b"` -> `a" + name + "b`). This PR also drops the comment that documented its contract.

Suggest collapsing all three into one shared helper and fixing the ordering once: drop the unconditional trailing-backtick strip and let the paired check do the work.

```ts
export function unwrapBallerinaString(raw?: string): string {
    if (!raw) return "";
    let s = raw.trim();
    if (s.startsWith("string `") && s.endsWith("`")) return s.slice("string `".length, -1);
    if (s.startsWith("`") && s.endsWith("`")) return s.slice(1, -1);
    if (s.startsWith('"') && s.endsWith('"')) return s.slice(1, -1);
    return s;
}
```

Note I raised the `agent.ts` half of this on #728 as well; since that code is touched again here, it may be simpler to fix it once on this branch.

MohamedSabthar (approved):

PR #731, opened 2026-07-29
Add schema-driven trigger/service builder framework (1/4)

LakshanWeerasinghe:
## Summary

This is **PR 1 of 4** that together deliver the schema-driven trigger/service builder feature. This PR adds the generic engine
only — no per-connector data — and the two later PRs stack on top of it, targeting this branch
(`feature/schema-driven-triggers`):

1. **`phase1/schema-driven-framework`** (this PR) — the engine
2. `phase2/connector-onboarding` — per-connector registry + fixtures (stacked on 1)
3. `phase3/architecture-model-generator-tie-in` — `Artifact.icon` → `IconDescriptor` (stacked on 2)
4. `phase4/frontend-ui` — the visualizer form that renders all this (stacked on 1, independent of 2/3)

## Why

Today, adding a new trigger/connector to the service designer means hand-writing a dedicated
`XxxServiceBuilder`/`XxxFunctionBuilder` Java class per connector (17 of them, for kafka, rabbitmq,
ftp, mcp, github, hubspot, shopify, and three CDC variants) — each one re-implementing the same
"read config, generate a listener + service skeleton" logic with copy-pasted variation. Onboarding a
connector meant writing and maintaining Java code in this repo for every one of them.

This PR replaces that with a **generic, JSON-schema-driven builder**: a connector ships a
`trigger-metadata.json` describing its listeners/service-types/annotations, and the LS derives
everything else (the UI form, the source generation) from that JSON plus semantic-model introspection
of the connector's actual Ballerina API. Onboarding a new connector becomes "add a JSON file" instead
of "write a Java class."

## Scope of this PR

- A new shared read API, **`LibraryMetadataReader`**, in `model-generator-commons` — the reusable entry
  point any LS extension (today service-model-generator, in future e.g. a copilot-facing extension) uses
  to resolve a connector's trigger models without knowing about `.bala` resolution or classpath
  registries.
- The three JSON-backed model types those APIs return.
- The generic **`SchemaDrivenServiceBuilder`** / **`SchemaDrivenFunctionBuilder`**, which replace all 17
  old hardcoded per-connector builder classes (deleted in this PR).
- **`ConnectorModelReader`** (service-model-generator-ls-extension) — the orchestrator that resolves a
  connector's UI schema via a 3-tier fallback (bundled → connector-shipped → synthesized from metadata),
  with version-gating support for connectors whose schema changed between releases (e.g. `mcp`).
[...]

PR #732, opened 2026-07-29
Onboard connectors onto the schema-driven builder framework (2/4)

LakshanWeerasinghe:
## Summary

**PR 2 of 4** (see #731 for the full split and the overall architecture write-up). This PR is
stacked on `phase1/schema-driven-framework` and adds the actual per-connector data the generic
`SchemaDrivenServiceBuilder`/`SchemaDrivenFunctionBuilder` from PR 1 need to do anything useful.

PR 1 alone deletes the 17 hand-written per-connector builder classes and replaces them with the
generic engine, but ships with an **empty** connector registry — every migrated connector falls back
to a bare default until this PR lands. This PR is what actually restores (and improves) generated-code
correctness for those connectors.

## Scope

- **`bundled_trigger_models.json`** — the flat registry `ConnectorModelReader` (from PR 1) reads to
  resolve a connector's trigger model by name. Newly onboarded here: kafka, rabbitmq, ftp, mcp,
  cdc-mysql, cdc-postgresql, cdc-mssql, github, hubspot, shopify (plus a few more already-migrated
  entries carried over).
- **Version-gated entries** for connectors whose schema changed shape between releases — e.g. `mcp`
  deprecated `Service` in favor of `StreamableHttpService` starting at 1.2.0. An entry can be a bare
  resource path, or an ordered list of `{minVersion, resource}` variants (newest first):

  ```mermaid
  flowchart LR
      REQ["Resolve mcp trigger model<br/>for a project pinned to mcp 1.1.0"] --> REG["bundled_trigger_models.json<br/>mcp: [\n  {minVersion: 1.2.0, resource: mcp.json},\n  {resource: mcp_1.0.3.json}\n]"]
      REG --> PICK["Pick first variant whose<br/>minVersion &lt;= 1.1.0"]
      PICK --> V1["mcp_1.0.3.json<br/>(Service-based)"]
  ```

  The trailing variant omits `minVersion` entirely (rather than pinning it to `0.0.0`) so it acts as
  the unconstrained floor — both resolve identically, but the shipped data has no `minVersion` key on
  that entry.
- **`connector_models/<name>/resources/*.json`** test fixtures exercising each onboarded connector's
  service-creation, listener-creation, and function flows end to end.
- Remaining `src/test/java` coverage (`AddServiceAndListenerTest`, `GetServiceInitModelTest`,
  `GetServiceModelFromSourceTest`, `AddFunctionTest`, `UpdateFunctionTest`,
  `GetFunctionModelFromSourceTest`) for these connectors.

## Testing

- `./gradlew :service-model-generator:service-model-generator-ls-extension:test` — 404/409 pass.
  Restores all 38 tests that were expected-red on PR 1 alone (confirmed kafka/rabbitmq/ftp/mcp/the
  three CDC variants/github/hubspot/shopify no longer fall through to the generic default builder).
- **5 remaining failures are pre-existing and unrelated**: `GetListenerModelFromSourceTest`'s http
  listener fixtures were frozen at `http` 2.16.5, while this integration branch is now pinned to
  2.16.6 (`a82d50fc74`, "Pin http 2.16.6 for LS tests and align goldens") — needs its own golden
  refresh independent of this feature.
- Checkstyle clean.

## Merge order

Should merge directly after PR 1 (#731) — the two together are what restores the integration branch
to fully green for the LS side of this feature.


LakshanWeerasinghe:
@coderabbit review

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/bundled_trigger_models.json
```diff
@@ -0,0 +1,20 @@
...
+  "trigger.github": "trigger-models/trigger.github.json",
+  "asb": "trigger-models/asb.json",
+  "salesforce": "trigger-models/salesforce.json",
+  "trigger.twilio": "trigger-models/trigger.twilio.json",
+  "trigger.shopify": "trigger-models/trigger.shopify.json",
+  "mssql": "trigger-models/mssql.json",
+  "mysql": "trigger-models/mysql.json",
+  "postgresql": "trigger-models/postgresql.json",
+  "trigger.hubspot": "trigger-models/trigger.hubspot.json",
+  "mcp": [
```
NipunaRanasinghe, line 16: The data here is well-formed — newest-first ordering, with the trailing entry omitting `minVersion` so it acts as the unconstrained floor. That matches what `ConnectorModelReader` expects.

The behaviour worth confirming is the fallback. `resolveResource` returns the **first** variant when the version is null or blank, and `ModelVariant.matches` also returns `true` on an unparsable version, which given declaration order again lands on the first:

```java
if (version == null || version.isBlank()) {
    return Optional.of(variants.getFirst().resource());
}
```

So a project actually pinned to mcp 1.1.x whose version fails to resolve gets `mcp.json` — the `StreamableHttpService` shape — and the generated code references a type that does not exist in their version. The opposite default would misfire on new projects instead, so neither is free; it's a question of which direction fails more visibly. Since this PR adds the only version-gated entry, it's the first time the choice has practical consequences, and mcp is precisely the case where the two shapes are incompatible rather than merely different.

Worth a sentence in the registry or the reader's javadoc recording which way it was decided and why.

Two small related notes:

- The PR description's mermaid diagram shows the floor variant as `{minVersion: 0.0.0, resource: mcp-v1.json}`, but the shipped data omits `minVersion` entirely. Both resolve identically, so this is a description-only mismatch — just worth aligning so the next person doesn't think `0.0.0` is required.
- `trigger_properties.json` lists mcp at `"version": "1.2.0"`, exactly the gate boundary. If that value ever feeds the resolution path (rather than only the palette metadata), `mcp_1.0.3.json` becomes unreachable. I could not find a caller wiring it through, and the `connector_version` fixtures suggest the version comes from the project manifest — but it's a coincidence worth being deliberate about.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger_properties.json
```diff
@@ -9,7 +9,10 @@
       "messaging",
       "event"
     ],
-    "triggerName": "Kafka"
+    "triggerName": "Kafka",
+    "version": "4.5.0",
+    "icon": "https://bcentral-packageicons.azureedge.net/images/ballerinax_kafka_4.5.0.png",
```
NipunaRanasinghe: The version is now encoded twice per entry.

```json
"version": "4.5.0",
"icon": ".../ballerinax_kafka_4.5.0.png",
```

All 15 entries follow this shape, so every connector bump needs two edits in lockstep. Miss the URL and the icon 404s while `version` still reads correctly — a silent, cosmetic-looking failure that won't fail any test.

The URL is fully derivable from the fields already present (`orgName`, `name`/`packageName`, `version`), so a template applied at read time would remove the duplication:

```
https://bcentral-packageicons.azureedge.net/images/{orgName}_{name}_{version}.png
```

One wrinkle if you go that route: the prefix is not uniformly `ballerinax_`. Entries 3, 9, and 10 use `ballerina_` (`ballerina_mqtt_1.4.1.png`, `ballerina_file_1.13.0.png`, `ballerina_mcp_1.2.0.png`), which matches their `orgName`, so `orgName` is the correct source rather than a hardcoded `ballerinax`.

Not blocking — but 15 hand-maintained URLs carrying a version substring will drift.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger_properties.json
```diff
@@ -97,15 +122,23 @@
...
       "mcp"
-    ]
+    ],
+    "triggerName": "MCP Service",
+    "version": "1.2.0",
+    "icon": "https://bcentral-packageicons.azureedge.net/images/ballerina_mcp_1.2.0.png",
+    "kind": "mcp"
   },
   "11": {
     "name": "solace",
```
NipunaRanasinghe, line 134: `solace` is the only one of the 16 entries that gained none of the three new fields — no `version`, no `icon`, no `kind`.

It's also the only `name` here with no corresponding key in `bundled_trigger_models.json`, so "not onboarded yet, intentionally skipped" is the obvious reading. Just confirming that's deliberate rather than an entry missed while working through the list.

If it is deliberate, the thing to check is whether anything now reads `kind` unconditionally — every other entry has it, so a consumer written against this data could reasonably assume it's always present and trip over solace specifically.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/services/ftp_init.json
NipunaRanasinghe, line 1: Deleting this file leaves five prose references to it behind:

```
src/main/java/.../util/FTPListenerUtil.java:416   "...mirroring the structure in ftp_init.json."
src/main/java/.../util/FTPListenerUtil.java:479   "// Build choices mirroring ftp_init.json auth CHOICE structure"
src/main/java/.../util/FTPListenerUtil.java:630   "...mirroring the structure in ftp_init.json."
src/test/java/.../connector/SchemaDrivenSourceGeneratorTest.java:209   "(ftp_init.json's hardcoded-builder..."
src/test/java/.../connector/SchemaDrivenSourceGeneratorTest.java:230   "ftp_init.json's SFTP branch offered..."
```

All comments, no resource loads — I checked, nothing breaks. But they now describe a file that isn't in the tree, and `FTPListenerUtil.java`'s three are the ones that matter: they're the explanation for why that code builds the choice structure the way it does, and the reference that made them checkable is gone.

Worth repointing them at `trigger-models/ftp.json` (or at the schema-driven model generally) in this PR, while the correspondence is still fresh. `FTPListenerUtil.java` isn't otherwise in this diff, which is why this is filed here rather than on those lines.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/bundled_trigger_models.json
LakshanWeerasinghe: Confirmed the mechanics — `resolveResource` falls back to `variants.getFirst()` on null/blank version, and `ModelVariant.matches` also treats an unparsable version as a match, so both land on the newest variant (`mcp.json`) rather than the floor (`mcp_1.0.3.json`). Since the two mcp shapes are genuinely incompatible, agreed this is worth being deliberate about rather than an accident of ordering — added a note to `ModelVariant.matches` documenting the fallback direction and rationale (https://github.com/ballerina-platform/ballerina-vscode/pull/732/commits/9ce04e8bbeb5229cf6f8ea0d75caf06942546ca3, now on `TriggerModelReader` after the merge from `feature/schema-driven-triggers` renamed this class from `ConnectorModelReader`).

Also confirmed `trigger_properties.json`'s `version: 1.2.0` for mcp does not feed this resolution path — it's only used for display/palette metadata, so the 1.2.0 boundary coincidence isn't currently live. And fixed the PR description's mermaid diagram, which didn't match the shipped file names/floor structure (it showed `mcp-v1.json`/`minVersion: 0.0.0`; actual data is `mcp.json`/`mcp_1.0.3.json` with the floor entry omitting `minVersion` entirely).

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger_properties.json
LakshanWeerasinghe: Confirmed — all 15 icon URLs followed `{orgName}_{packageName}_{version}.png`, including the `ballerina_` vs `ballerinax_` split matching `orgName` exactly for mqtt/file/mcp. Switched to deriving the icon URL at read time via the existing `CommonUtils.generateIcon(orgName, packageName, version)` helper (already used for this exact pattern elsewhere in the codebase) instead of storing it, and dropped the `icon` field from all 15 `trigger_properties.json` entries (https://github.com/ballerina-platform/ballerina-vscode/pull/732/commits/9ce04e8bbeb5229cf6f8ea0d75caf06942546ca3).

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger_properties.json
LakshanWeerasinghe: Correct, deliberate — solace isn't onboarded in this PR (no `bundled_trigger_models.json` entry yet); will follow up separately. Confirmed there's no unconditional `kind` read that would trip on it: `getTriggerBasicInfoByName(TriggerProperty)` guards on `version`/`kind` both being non-null before using the new path (icon is now always derivable, so it's no longer part of that gate — see the reply on the icon-duplication comment), so solace safely falls through to the legacy sqlite-index lookup.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/services/ftp_init.json
LakshanWeerasinghe: Confirmed all 5 were stale and fixed them https://github.com/ballerina-platform/ballerina-vscode/pull/732/commits/9ce04e8bbeb5229cf6f8ea0d75caf06942546ca3 the 3 in `FTPListenerUtil.java` (present-tense "mirroring the structure in ftp_init.json") now point at `trigger-models/ftp.json`. The 2 in `SchemaDrivenSourceGeneratorTest.java` were describing historical/pre-migration behavior rather than current structure, so I reworded them to not depend on a file that no longer exists (e.g. "the pre-migration hardcoded builder supported it") instead of repointing them at the new file, since they weren't claiming a current correspondence to verify against.

NipunaRanasinghe (approved):

PR #744, opened 2026-07-30
 Unified Create flow with a pre-project Create Integration wizard

madushajg:
## Purpose

Replaces the old project/integration creation forms with a single Create flow: pick (or create) a project, pick a starting point, then name and configure the first artifact — all in one shell, before any project exists on disk.

Previously you created an empty integration first and configured its first artifact afterwards, in a separate screen, after a window reload. Now the whole thing is one flow and the artifact is ready when the project opens.

Addresses: 
https://github.com/wso2-enterprise/integration-engineering/issues/2132
https://github.com/wso2-enterprise/integration-engineering/issues/2133

## High-level architecture

```mermaid
flowchart TD
    A["Create Integration<br/>welcome view · tree view · package overview"] --> B["CreateProjectChooser<br/>project + starting point"]
    B -->|Integration| C["Create Integration Wizard"]
    B -->|Library| D["Library form"]

    C --> C1["Step 1 — Type<br/>name · location · artifact card"]
    C1 --> C2["Step 2 — Configure<br/>artifact-specific form"]
    C2 --> E{"Target is an already<br/>activated workspace?"}

    E -->|Yes| F["Generate artifact in place<br/>no reload"]
    E -->|No| G["Schedule pending artifact<br/>+ vscode.openFolder"]
    G --> H["Window reload<br/>startup screen narrates 'Creating …'"]
    H --> I["checkAndRunPendingArtifact<br/>generate + land on the new integration"]
```

**The pre-project problem.** The Configure step renders the same `ArtifactForm` the in-project flows use, and the language server can only compute an artifact model against a real package. So the wizard scaffolds a throwaway **staging package in the OS temp dir**, resolves the model against that, and throws it away at submit. The user's chosen path is only ever touched at final submit.

**Spanning the reload.** Opening a new folder reloads the window, which would normally lose the in-flight create. A pointer in `globalState` plus a payload file in `target/` carries the configured artifact across; the extension injects the same "Creating &lt;name&gt;" copy into the first HTML frame so the wizard, the reload, and the post-reload wait read as one continuous screen.

**Always-workspace model.** A project is a Ballerina workspace; integrations and libraries are packages inside it. Creating into an existing proj the new package in the workspace `Ballerina.toml`.

### Where things live

| Area | Path |
| --- | --- |
| Wizard UI | `ballerina-visualizer/src/views/BI/CreateIntegrationWizard/` |
| Create chooser + shell | `.../BI/ProjectForm/embedded/integrat
| Shared artifact card registry | `.../BI/components/artifactCards.tsx` |
| Wizard orchestration | `ballerina-extension/src/features/bi/in
| Cross-reload handoff | `.../features/bi/pending-artifact.ts`, `startup-progress.ts` |
| Project/package creation | `ballerina-extension/src/utils/bi.t

The artifact card registry is the single source of truth for the
**both** the wizard's Type step and the in-project Add Artifact screen, so the
two can't drift.

## Also in this PR

- Convert a standalone integration into a project, and add an in enclosing project, from the tree view.
- Continue from an empty integration: run the wizard against an and generate the artifact in place.
- Search + category filtering on the Add Artifact screen.
- View prefetching and progress indicators, so a first navigation to a view no longer looks frozen while its chunk loads.

https://github.com/user-attachments/assets/767c094f-381f-4222-9093-fb87e36aa9cb

https://github.com/user-attachments/assets/ce6c723b-40dc-4ea1-9125-6f5e09ea586e

## Summary

- Replaced separate project and integration forms with a unified Create flow for projects, integrations, and libraries.
- Added staging-package management and artifact creation across active workspaces and VS Code reloads.
- Added project conversion, enclosing-project support, empty-integration continuation, and collision-safe naming.
- Added shared artifact catalogs with search, category filtering, keyboard navigation, and progress feedback.
- Improved workspace validation, component creation, startup handling, visualizer readiness, and view prefetching.
- Added wizard RPC contracts, capability detection, language-server readiness checks, and shared service-model utilities.
- Improved Markdown loading, keyboard accessibility, command handling, and asynchronous form state management.


madushajg:
@coderabbitai review

packages/ballerina-extension/src/features/bi/activator.ts
```diff
@@ -176,6 +180,61 @@ export function activate(context: BallerinaExtension) {
...
+                if (choice !== 'Add to Project') {
+                    return;
+                }
+                adoptOrphanedPackageIntoProject(currentProjectPath, enclosing.projectPath!);
+                openInVSCode(enclosing.projectPath!);
+                return;
+            }
+            if (enclosing.status === 'invalid') {
+                window.showErrorMessage(
+                    `This integration is located inside another Ballerina project (${enclosing.projectPath}), which isn't a supported layout. Move it out before converting it into a project.`
```
LakshanWeerasinghe:
```suggestion
                    `This integration is located inside another project (${enclosing.projectPath}), which isn't a supported layout. Move it out before converting it into a project.`
```
Is it correct to use `Ballerina` term?

madushajg: Fixed with 5308a892833fe5caa42a5fbaab033aa7399f7022

packages/ballerina-extension/src/features/bi/integration-wizard.ts
```diff
@@ -0,0 +1,154 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import * as crypto from "crypto";
```
LakshanWeerasinghe: better of we import only the necessary ones

packages/ballerina-visualizer/src/views/BI/CreateIntegrationWizard/index.tsx
```diff
@@ -0,0 +1,629 @@
...
+            const seedBaseDir = projectContext?.workspacePath
+                ? Promise.resolve(projectContext.workspacePath)
+                : wsClient.getWorkspaceRoot().then(async (res: { path: string }) => res.path || (await wsClient.getDefaultCreationPath()).path);
+
+            seedBaseDir
+                .then(async (seedPath: string) => {
+                    // Fetched once: picks a collision-free default and flags collisions live.
+                    let taken = emptyTakenNames();
+                    try {
+                        taken = toTakenNames(await wsClient.getProjectComponentNames({ projectPath: seedPath }));
```
gigara, line 274: The collision list is a one-time snapshot of the seed path, but the user can retarget the location afterwards. `handleBrowse`/`handlePathChange` update `basicInfo.baseDir` without refetching (this effect only depends on `projectContext?.workspacePath`), so `checkNameCollision` keeps validating every keystroke against the *original* folder's contents.

Repro (standalone wizard, where the path field is editable — the MainPanel, `StandaloneCreateChooser`, and direct `EmbeddedBIProjectForm` mounts):
1. Open the wizard while a workspace containing a package `api` is open (its path becomes the seed via `getWorkspaceRoot()`).
2. Browse the location to a fresh empty folder.
3. Type the name `api` → "already exists in the project", and both **Continue** and **Create Empty Integration** stay disabled, with no way past it except renaming.

The reverse also holds: collisions in the newly targeted workspace aren't caught client-side, only at submit.

Suggestion: refetch `getProjectComponentNames` (or at least clear `takenNames`) whenever `baseDir` changes — e.g. a small effect keyed on the debounced path — and re-run the name check. `LibraryCreationView.tsx` takes the same one-time snapshot and needs the same fix.

packages/ballerina-visualizer/src/views/BI/CreateIntegrationWizard/hooks/useServiceInitModel.ts
```diff
@@ -0,0 +1,118 @@
...
+ * immediately; a slow one flips to a "pulling the package…" status first.
+ */
+export function useServiceInitModel({ wsClient, projectRoot, orgName, packageName, moduleName, cachedModel }: UseServiceInitModelOptions) {
+    const [model, setModel] = useState<ServiceInitModel | null>(cachedModel ?? null);
+    const [pullingStatus, setPullingStatus] = useState<PullingStatus | undefined>(
+        cachedModel ? undefined : PullingStatus.FETCHING
+    );
+    // Re-entering the Configure step with the same selection must reuse the model (with the
+    // user's values already written into it), so key the fetch by module identity.
+    const fetchedForRef = useRef<string | null>(
```
gigara: This claim only holds for CHOICE fields. `ServiceConfigureForm.handleOnChange` persists edits via `updateChoiceInModel`, which writes back **only** `fieldType === "CHOICE"` selections; every other field's value lives in `ArtifactForm`'s internal state and is applied to the model only at submit (`applyFormValuesToModel` in `handleOnSubmit`).

So: select HTTP Service → fill base path/port on Configure → back arrow to tweak the integration name → Continue: the step remounts from the cached model and all non-choice values silently reset to their defaults. `FunctionConfigureForm` and `AIAgentConfigureForm` have no caching at all, so automation/workflow/agent inputs are always lost on back-navigation.

`ArtifactForm` already reports every edit via `onChange(fieldKey, value, allValues)` — capturing `allValues` alongside the cached model (and rehydrating from it on re-entry), or writing all values back into the model on unmount, would make the cache do what this comment says.

packages/ballerina-visualizer/src/views/BI/CreateIntegrationWizard/index.tsx
madushajg: Fixed with 488e2c7f433771bc273d9591ba5ea6195cd3c4be

packages/ballerina-extension/src/features/bi/integration-wizard.ts
madushajg: Fixed with 488e2c7f433771bc273d9591ba5ea6195cd3c4be

packages/ballerina-visualizer/src/views/BI/CreateIntegrationWizard/hooks/useServiceInitModel.ts
madushajg: Addressed with f06da74104d1385ae0ecf7b4ceb39bcceacec8c2

madushajg:
This PR affects many E2E tests since this chantes the artifact creation flow. Will address them in a separate PR

gigara (approved):
LGTM

Issue #2, opened 2021-05-05
Add support for running/creating Ballerina Tests in VSCode


prabushi:
<a href="https://github.com/kaviththiranga"><img src="https://avatars.githubusercontent.com/u/1505855?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [kaviththiranga](https://github.com/kaviththiranga)**
_Wednesday Sep 12, 2018 at 08:31 GMT_
_Originally opened as https://github.com/ballerina-platform/plugin-vscode/issues/43_

----

- [ ]  Run tests directly from VSCode editor 
- [x]  Generate tests for a particular function/service
- [ ] Show test results (figure out a way to properly show)

prabushi:
<a href="https://github.com/kaviththiranga"><img src="https://avatars.githubusercontent.com/u/1505855?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [kaviththiranga](https://github.com/kaviththiranga)**
_Tuesday Oct 23, 2018 at 05:58 GMT_

----

We are starting the work on running tests and displaying results. Test generation work is already in progress and @rasika is working on it.

hevayo:
This is already available in the current VSCode plugin via VSCode test view.

PR #745, opened 2026-07-30
Support Ballerina submodules in type helper, function search, and fix calling function for functions form submodules 

### LINKED ISSUE #1747: Integrator low code mode cannot use EDI typed modules (from cli) — submodule functions not supported
### Current Limitation

EDI integrations depend on typed modules generated by the `edi` CLI (`bal edi codegen` and `bal edi libgen`) from EDI schemas. The generated API — typed records plus `fromEdiString` / `headersFromEdiString` / `interchangeFromEdiString` / `interchangeToEdiString` — is placed in **submodules**:

- `codegen` → a submodule inside the user's project (`modules/<name>/`).
- `libgen` → one package with the typed per-message API in `m<MESSAGE>` submodules; only the generic aggregators (`fromEdiString`/`toEdiString`/`getEDINames`) sit at the package root.

The Integrator low-code editor only supports a package's **root module** functions in the Call Function palette and only resolves root-module symbols in the expression editor. Submodule functions are never offered, and a hand-typed submodule-qualified call is rejected as `undefined function`.

Image

**Verified:**

- A project-local `codegen` submodule (`<pkg>.orders`) never appears in the palette; `check orders:interchangeFromEdiString(payload)` in a Declare Variable expression → `undefined function 'interchangeFromEdiString'`, Save disabled.
- The **published** package `ballerinax/edifact.d03a.supplychain` (root module + ~29 `m<MSG>` submodules incl. `mORDERS`) shows only its root `fromEdiString`/`toEdiString`/`getEDINames` in the palette; the typed per-message submodule functions (e.g. `edifact.d03a.supplychain.mORDERS`'s `interchangeFromEdiString`) do not appear.

So the boundary is root-module vs submodule, not local-vs-published.

**Impact:** a user building an EDI integration in Integrator cannot call the typed envelope parser/serializer the CLI generated — the main value of the typed module is unreachable in the low-code experience. Both default (`codegen`) and CLI-library (`libgen`) outputs are affected.

### Suggested Improvement

Two complementary directions:

**(A) Integrator low-code editor (general fix).** Support public functions of a package's **submodules** — not just the root module — in the Call Function palette, and resolve submodule-qualified calls (`mod.sub:fn(...)`) in the expression editor. Benefits any multi-module package, not only EDI.

**(B) edi-tools "library" generation mode (possible solution available sooner).** Generate the typed API at the package **root module** so it is Integrator-visible today:

- `bal edi codegen -p <org/name>` → single schema → one publishable package with records + the full API at the **root**, clean type names (no `EDI_<name>_` prefix), `ballerina/edi >= 1.6.0` pinned.
- `bal edi libgen --split` → schema folder → one root-level package per schema (independent versioning), with an optional `[workspace]` manifest so they build together.
- The `EDI_<name>_` prefix is applied by a single line in `libgen` solely for multi-submodule collision avoidance + the `EDI_NAME` enum/REST connector; it is skipped (guarded) in these root-level modes, so names stay clean.

The edi-tools/runtime changes live in `ballerina-platform/edi-tools` and `ballerina-platform/module-ballerina-edi`; this issue tracks the Integrator-side usability gap and the recommended generation shape. A linked issue/PR can be filed there.

### Version

WSO2 Integrator (VS Code extension); reproduced with Ballerina `2201.13.x` and the unreleased `edi` tool `2.2.0` / `ballerina/edi 1.6.0`.

dulajdilshan:
## Purpose
The Type Helper, Function Search, and Function Call flows currently do not fully support functions and types declared in Ballerina submodules.

This causes the following issues:
- The Type Helper does not consistently show types from manual or generated submodules.
- Function Search does not consistently show functions from the current integration’s submodules or from submodules of other packages within the project.
- Private declarations are not always filtered according to the module containing the current file.
- Package and module names can appear as duplicate sections in Function Search.
- Selecting a function from another submodule may not generate the required import and qualified call.
- The Function Call form may fail with `Function symbol not found` when the function belongs to a submodule of another package within the project.
- Flow-model generation may fail to resolve the definition location of a selected submodule function.

This change is limited to the Type Helper, Function Search, and Function Call flows. It does not change Artifact Search, `getTypes`, Type Diagram, or other main-content functionality.

Fixes https://github.com/wso2/product-integrator/issues/1747

image

image

image

## Goals

- Allow the Type Helper to discover types from the current default module, current submodule, other manual and generated submodules in the current integration, and submodules of an integration within the same workspace project 
- Allow Function Search to discover functions from the same module locations.
- Load the Function Call form for functions declared in submodules of other packages within the project.
- Resolve the correct source definition when a Function Call targets a submodule function.

## Approach
- Added shared module-resolution utilities to determine:

  - Which module contains the current file
  - The full owner module name
  - Whether a module is the default module or a submodule
  - Whether a submodule is generated

- Updated the Type Helper search path to inspect eligible modules in the current integration and other packages within
  the project.
- Updated Function Search with a module-aware search path while keeping unrelated search flows unchanged.
- Added cross-module visibility filtering so private types and functions are returned only for the current module.
- Added `moduleRelation`, `moduleKind`, and `isGenerated` to result `codedata`. These values are not duplicated in
  metadata.
- Updated Function Search UI adapters to:

  - Preserve the project/package/module hierarchy
  - Remove duplicate package and default-module labels
  - Display full submodule names
  - Treat current-module functions as unqualified calls
  - Treat functions from other modules as imported calls

- Updated Function Call template generation to create the correct module prefix, import, and qualified call.
- Updated Function Call form generation to resolve the exact owner submodule from the live project package.
- Updated flow-model definition resolution for Function Calls to use the function's owner module instead of always
  using the default module.
- Added regression tests covering:

  - Type Helper and Function Search requests from default modules and submodules
  - Manual and generated submodules
  - Private and public visibility across modules
  - Same-package and project-package module relations
  - Qualified Function Calls and imports
  - Function Call form loading for a sibling package's submodule
  - Function Call definition-view resolution
  - Removal of duplicate package/module labels

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary

- Added module-aware discovery for Ballerina submodule functions and types across workspace packages, generated modules, and dependencies.
- Improved visibility filtering, module categorization, hierarchy preservation, duplicate-label handling, and qualified function call and import resolution.
- Updated form generation, document lookup, semantic-model resolution, and flow-model definition resolution for submodule ownership.
- Added shared package and module utilities and improved frontend handling of nested module categories.
- Disabled flow visualization for submodule function calls.
- Added regression tests and fixtures for submodule search, visibility, generated modules, qualified calls, imports, form loading, and definition resolution.


packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/FunctionDataBuilder.java
```diff
@@ -300,6 +299,18 @@ private void resolvePackageAndSemanticModel() {
...
+        // package can also exist in the cache, but that copy may be stale and resolvedPackage() initially selects
+        // its default module. In particular, this would make a function in a sibling package's submodule invisible.
+        if (project != null) {
+            Optional<SemanticModel> workspaceSemanticModel = PackageUtil.getSemanticModelFromWorkspace(project,
+                    moduleInfo.org(), moduleInfo.packageName(), moduleInfo.moduleName());
+            if (workspaceSemanticModel.isPresent()) {
+                semanticModel(workspaceSemanticModel.get());
+                return;
+            }
+        }
```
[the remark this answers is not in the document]
dulajdilshan: What is the impact here?

packages/ballerina-visualizer/src/views/BI/TypeEditor/utils.tsx
```diff
@@ -235,43 +147,38 @@ export const getTypeBrowserTypes = (types: Category[]) => {
...
+                    description: item.codedata.node,
+                    detail: "",
+                },
+            } : {}),
+        }),
+        (categoryLabel, items) => ({ category: categoryLabel, items }),
+        (categoryLabel, items, subCategory) => ({
+            category: categoryLabel,
+            items,
+            subCategory,
```
LakshanWeerasinghe, line 181:
```suggestion
            items: items ?? [],
            subCategory: subCategory ?? [],
```

dulajdilshan: This is handled in the LS side and in the `buildHelperFunction` in `packages/ballerina-visualizer/src/utils/function-category.ts`. Here, we intentionally pass whatever the values we get (i.e., undefined or list)

LakshanWeerasinghe:
Lets remove the `View function flow` icon from the flow diagrams sub module function call node.
Screenshot 2026-08-03 at 14 43 45

dulajdilshan:
> Lets remove the `View function flow` icon from the flow diagrams sub module function call node.

Addressed via https://github.com/ballerina-platform/ballerina-vscode/pull/745/commits/4643ae1b55e4f577c386f4aa4731d483667eeec4

samithkavishke:
1. WorkspaceFunctionNodeBuilder.java:164 — functionsDoc is scoped to the wrong module. It's always the default module's functions.bal, but it's passed as context.current() ? functionsDoc : null. With the cursor in a submodule: the default module gets null (natural-expression functions no longer filtered from "Within Project"), and the submodule gets functionsDoc, so submodule TextRanges are resolved against the wrong syntax tree → misclassification or out-of-bounds. Resolve the doc per-module instead.
2. FunctionDataBuilder.java:302 — workspace-first resolution ignores moduleInfo.version(). getSemanticModelFromWorkspace matches org+name only, so a local myorg/common@0.1.0 shadows a depended-on published @2.0.0, and updateModuleInfo() rewrites the version — generated imports then disagree with Ballerina.toml.
3. PackageUtil.java:389 — dropping the default-module fallback turns a quiet degradation into a "Pulling the module…"/failure notification plus a Central round-trip for purely local workspace packages (compounds with #2).
4. categoryConfig.ts:52 — normalizeCategoryTitle now calls title.toLowerCase() unguarded. NodeList/index.tsx:719 passes group.title and only null-checks group at line 728, so a nullish title crashes the whole side-panel render (previously safe via Set.has(undefined)).

pasindufernando1:
### Heads-up: this will expose downstream paths that don't handle submodules

Not a blocker on this PR's own code — the module-awareness here looks right. But it's worth recording that making submodule types and functions *reachable* will light up consumers that were never exercised with submodule-qualified data before. Two I've already hit while testing this branch:

**1. `typesManager/findMatchingType` returns an empty result for a submodule type**

```json
{
  "filePath": ".../testxom2/automation.bal",
  "typeMembers": [{
    "kind": "RECORD_TYPE",
    "type": "Person",
    "packageInfo": "pasindufernando:testxom2.utils:0.1.0",
    "packageName": "testxom2.utils"
  }],
  "expr": "{name: \"Pasindu\"}"
}
```

`FindTypeRequest.TypePackageInfo.from(packageInfo, packageName)` only splits the qualified module name into package + module when the caller omits `packageName`. Here `packageName` is supplied and is itself the qualified module name, so the split is skipped and `packageName` ends up as `testxom2.utils`. Every resolver keys on the package, so all three tiers miss — `getSemanticModelIfMatched` fails `descriptor.name().equals(packageName)`, `getSemanticModelFromWorkspace` matches no child project, and the Central lookup searches for a package literally named `testxom2.utils`. The resulting `IllegalArgumentException` is then swallowed by the per-member `catch (Throwable ignored) {}` in `findMatchingType`, which is why the symptom is an empty response rather than a diagnostic.

Worth noting the resolvers themselves are fine — with `packageName = "testxom2"` the existing `ModuleName.from(PackageName.from(...), ...)` path resolves `testxom2.utils` correctly. The submodule plumbing works; this API just never reaches it.

**2. `dataMapper/visualizable` fails for a same-package submodule type**

```json
{
  "codedata": {
    "node": "RECORD", "org": "pasindufernando",
    "module": "testxom2.utils", "packageName": "testxom2",
    "symbol": "Person", "version": "0.1.0",
    "data": { "moduleRelation": "SAME_PACKAGE_MODULE", "moduleKind": "SUBMODULE" }
  }
}
```

```
errorMsg: Semantic model cannot be found for the module:
  ModuleInfo[org=pasindufernando, packageName=testxom2, moduleName=testxom2.utils, version=0.1.0]
  at DataMapManager.getVisualizableProperties(DataMapManager.java:2727)
```

The codedata is correct here. `getVisualizableProperties` branches on `org != null` and then tries only two tiers — workspace siblings, then local cache/Central. There's no tier for the current project's own package, even though `sm` (the current file's semantic model) is passed into the method and `project.currentPackage()` is in scope. For a type whose `moduleRelation` is literally `SAME_PACKAGE_MODULE`, both tiers legitimately miss: the package isn't a workspace sibling, and it isn't published.

`PackageUtil.getSemanticModelIfMatched(...)` already does exactly the right thing and is used as the first tier by `TypesManagerService.getCachedSemanticModel`. `DataMapManager` just doesn't call it.

### Why these surface now

Both are pre-existing gaps rather than defects introduced here, but this change is what makes them reachable, for two reasons:

- **New traffic.** Before this PR the Type Helper didn't consistently surface types from `modules/*`, so requests carrying a qualified module name were rare. Now they're routine.
- **New data shape.** `codedata.module` now carries the fully-qualified module name (`testxom2.utils`) where it previously carried the package name, plus the new `moduleRelation` / `moduleKind` fields. Any consumer that assumed `module == packageName`, or that treats the middle segment of an `org:module:version` string as a package name, now gets something it has never seen.

One related note: `getSemanticModelFromWorkspace` no longer falls back to the default module when a submodule isn't found — it returns `Optional.empty()`. That's the more correct behaviour and we'd keep it, but it does mean a missing current-package tier now produces a hard error rather than silently resolving against the wrong module. Case 2 may be an instance of that.

### Suggestion

Neither fix belongs in this PR — they're separate surfaces with their own tests, and this is already 58 files. But it's worth a tracking issue for the sweep before the release that carries submodule support. The pattern to grep for:

- any `getSemanticModelFromWorkspace(...)` or `getSemanticModel(ModuleInfo)` call not preceded by a `getSemanticModelIfMatched(...)` current-package check;
- any code splitting `org:module:version` and treating the middle segment as a package name.

The scope note in the description ("limited to the Type Helper, Function Search, and Function Call flows") is accurate about which code this PR edits, but the codedata shape change reaches every consumer of codedata — worth reflecting that in the description so reviewers of dependent areas know to look.

dulajdilshan:
> 1. WorkspaceFunctionNodeBuilder.java:164 — functionsDoc is scoped to the wrong module. It's always the default module's functions.bal, but it's passed as context.current() ? functionsDoc : null. With the cursor in a submodule: the default module gets null (natural-expression functions no longer filtered from "Within Project"), and the submodule gets functionsDoc, so submodule TextRanges are resolved against the wrong syntax tree → misclassification or out-of-bounds. Resolve the doc per-module instead.
> 2. FunctionDataBuilder.java:302 — workspace-first resolution ignores moduleInfo.version(). getSemanticModelFromWorkspace matches org+name only, so a local myorg/common@0.1.0 shadows a depended-on published @2.0.0, and updateModuleInfo() rewrites the version — generated imports then disagree with Ballerina.toml.
> 3. PackageUtil.java:389 — dropping the default-module fallback turns a quiet degradation into a "Pulling the module…"/failure notification plus a Central round-trip for purely local workspace packages (compounds with [Add support for running/creating Ballerina Tests in VSCode #2](https://github.com/ballerina-platform/ballerina-vscode/issues/2)).
> 4. categoryConfig.ts:52 — normalizeCategoryTitle now calls title.toLowerCase() unguarded. NodeList/index.tsx:719 passes group.title and only null-checks group at line 728, so a nullish title crashes the whole side-panel render (previously safe via Set.has(undefined)).

@samithkavishke 

1,2, and 4 are addressed via https://github.com/ballerina-platform/ballerina-vscode/pull/745/commits/3ae953f8c3f6d94bdd0ed152c7dbc64a682302b8

For 3:

This is intentionally not addressed. Restoring the default-module fallback would hide an invalid or stale submodule identity by resolving the package’s default module instead of the requested submodule. Valid local workspace packages already resolve without a Central lookup when the requested module identity is correct. If the module cannot be found, returning an empty result preserves exact module resolution and surfaces the underlying mismatch instead of silently using an incorrect semantic model.

dulajdilshan:
> ### Heads-up: this will expose downstream paths that don't handle submodules
> Not a blocker on this PR's own code — the module-awareness here looks right. But it's worth recording that making submodule types and functions _reachable_ will light up consumers that were never exercised with submodule-qualified data before. Two I've already hit while testing this branch:


These will be addressed as a separate implementation. All the LS APIs and front-end misalignment need to be checked. The task is tracked via https://github.com/wso2/product-integrator/issues/2036

NipunaRanasinghe (approved):
LGTM.

@dulajdilshan Please make sure to create an issue to track what's mentioned in https://github.com/ballerina-platform/ballerina-vscode/pull/745#issuecomment-5200031002

dulajdilshan:
> LGTM.
> 
> @dulajdilshan Please make sure to create an issue to track what's mentioned in [#745 (comment)](https://github.com/ballerina-platform/ballerina-vscode/pull/745#issuecomment-5200031002)

Already created https://github.com/wso2/product-integrator/issues/2036 and mentioned in here

PR #703, opened 2026-07-22
Durable Agentic Workflow designer (object model)

hasithaa:
## Summary

Introduces the **Durable Agentic Workflow** designer targeting the object-model `workflow:DurableAgent` API (module PRs ballerina-platform/module-ballerina-workflow#69 and #70, both merged; the module 0.8.0 build is unreleased — install it into the local repository):

- **Creation flow**: Add Artifact → Durable Agentic Workflow asks only for a workspace-unique name (identifier-validated, auto-numbered default) and generates the module-level agent declaration (`final workflow:DurableAgent <name> = check new ({systemPrompt..., model...});`) — the agent *is* the workflow. The submit button reads **Create Agent**.
- **Agent diagram**: opening the agent renders the declaration as the agent box — Role/Instructions in the body, model circle, and capability circles for activities, tools, peers, events and human tasks parsed from the config literal. The canvas is not a flow, so there is **no end node** after the box.
- **Capability editing on the node**: always-visible `+` affordances at fixed anchors (Human Task, Events, Activities, Agent Tools, Model when unset) append entries to the declaration's config lists; capability circles open pre-filled forms that rewrite their exact entry, with ✕ badges for comma-aware removal. Gated activities carry Reviewer Roles and the shared retry-policy form (No Automatic Retry / Auto Retry / Human Review with roles).
- **Agents as first-class artifacts**: `workflow:DurableAgent` declarations appear in the artifacts explorer right after the durable workflows (before activities) and clicking one opens the agent model. Deleting an agent from the design view removes the whole declaration.
- **Run-site references** (mirrors the AI module's agent-call pattern): outside the declaration canvas the agent box is a **read-only reference** — editing affordances are disabled and the header shows a **Go to Agent** button; box/capability clicks navigate to the agent's own model.
- **Overview integration**: durable agents join the workflow column (agent icon, declared events/human tasks shown, declared activities linked); `run()` draws the trigger edge, `sendEvent` connects to its event box, and read-only calls render as dashed edges.
- **Workflow palette regroup**: inside a workflow function the palette nests **Steps**, **Child Workflows** (Run/Call/Wait-for/Send-Data-to Child Workflow nodes for the 0.8.0 composition API) and a collapsed **Workflow Functions** group.
- **UX polish**: Add Artifact texts (no 'static'; short card tooltips), the receiver icon (data arriving from outside, with a waiting-state badge on capability circles) replaces the open-in-link arrow / sand clock for data-event waits — including the workflow diagram's Await Data node.
[...]

PR #747, opened 2026-07-31
Add Durable Agent support to the workflow designer

hasithaa:
## Purpose

Adds Durable Agent support to the workflow designer, along with general durable-workflow designer improvements. This is the agent-focused subset of #703; child-workflow nodes will follow in a separate PR.

Requires `ballerina/workflow` 0.8.0 (released).

## Durable Agent support

- **Agent flow diagram**: `workflow:DurableAgent` declarations render as an agent view — model provider, tools, data events, human tasks, and peer agents shown as connected capabilities, with add/edit/delete affordances (affordance buttons pinned to the diagram's bottom corners).
- **Agent creation**: "Durable Agent" artifact creation form with loader-on-submit (no flashing name-uniqueness errors), generating a `final workflow:DurableAgent` declaration plus its config literal.
- **Capability editing on the declaration**: capability forms (tools, events, human tasks, peers, model) edit the agent declaration's config literal directly, including for object-model agents.
- **Tools**: reuses the existing AI tool-creation flow unchanged; empty tool list shows a "No tools available to use. Create a Tool" CTA that opens the tool form as a popup and refreshes on submit. `ToolDecl` fields (`requiresApproval`, `userRoles`) are editable in the add/edit tool forms.
- **Agent driver API in flows**: palette items and forms for the full driver surface — Run Durable Agent (`run`), Send Agent Data (`sendData`, with per-agent Data Event dropdown and check-error flag), Get Agent Result / Data Result, Wait for Agent Result / Data Result. Durable agents are also listed in the Run Workflow search (distinguished by icon) so they can be started from non-agent contexts.
- **Design view**: agent → model-provider edge rendered in the architecture/design model.

## Durable workflow designer improvements

- Workflow functions and driver functions surfaced in non-workflow contexts (e.g. starting workflows/agents from services), with a Workflow palette category placed after the AI category.
- `workflow:Duration` fields render as typed duration forms; module-prefixed titles (e.g. "workflow : Run Workflow") cleaned up.
- Search index updated for the released `workflow` 0.8.0 (central + search sqlites).
- Plus-button placement fixes around workflow/agent boxes and palette icon fixes (Send Agent Data et al.).

## Out of scope

Child-workflow nodes (Run/Call/Wait/Send Data to Child Workflow) are intentionally excluded; they remain on #703 and will be raised separately.

## Samples

```ballerina
final workflow:DurableAgent supportAgent = new (
    name = "supportAgent",
    systemPrompt = {role: "Support Assistant", instructions: "Help users with their orders."},
    model = model,
    tools: [lookupOrder, {tool: cancelOrder, requiresApproval: true, userRoles: ["manager"]}],
    dataEvents: {userReply: UserReply},
    resultType: Summary
);
```

## Summary

- Added Durable Agent support across the workflow designer, language server, diagrams, and artifact explorer.
- Added Durable Agent creation, editing, configuration, capability management, tool integration, and execution forms.
- Added diagram nodes and interactions for activities, human tasks, events, tools, results, data exchange, and workflow context operations.
- Improved workflow analysis, search, connection visibility, retry-policy handling, duration types, error reporting, refresh behavior, and navigation.
- Added support for object-model and explicit `new` Durable Agent declarations.
- Added source-update coordination to improve flow-model consistency after edits and formatting.
- Updated `ballerina/workflow` to `0.8.0` and added `ballerina/ai`.
- Improved icons, form labels, palette grouping, and workflow connection rendering.
- Child-workflow nodes remain out of scope.


hasithaa:
Please find the Updated UX for this PR. @madushajg 

https://github.com/user-attachments/assets/d1b6477b-951c-400d-a7a4-9e085b8b9e7d

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -415,6 +416,43 @@ public void visit(MethodCallExpressionNode methodCallExpressionNode) {
...
+            return;
+        }
+        Workflow agent = intermediateModel.workflowMap.get(targetSymbol.get().getName().get());
+        if (agent == null || !Workflow.KIND_DURABLE_AGENT.equals(agent.getKind())) {
+            return;
+        }
+        // agent.sendData(id, "channel", data): correlate with the declared event channel so the
+        // overview draws the edge into the channel's in-port, like workflow:sendData does.
+        String methodName = methodCallExpressionNode.methodName().toSourceCode().trim();
+        if ("sendData".equals(methodName)) {
```
samithkavishke: Shall we extract this "sendData", "eventName", "run" as  Constants

hasithaa: Fixed.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentRunBuilder.java
```diff
@@ -0,0 +1,419 @@
...
+        // itself is not rewritten here.
+        if (WorkflowUtil.isDurableAgentObjectTarget(sourceBuilder)) {
+            String agentVarName = sourceBuilder.flowNode.codedata().parentSymbol();
+            String role = promptFieldSource(sourceBuilder.getProperty(ROLE_KEY).orElse(null));
+            String instructions = promptFieldSource(sourceBuilder.getProperty(INSTRUCTIONS_KEY).orElse(null));
+            String modelValue = sourceBuilder.getProperty(MODEL_KEY)
+                    .map(p -> p.value() == null ? "" : p.value().toString().trim()).orElse("");
+            String promptText = "{role: " + role + ", instructions: " + instructions + "}";
+            java.util.LinkedHashMap<String, String> fields = new java.util.LinkedHashMap<>();
+            fields.put(SYSTEM_PROMPT_KEY, promptText);
```
samithkavishke: **Saving the agent box destroys a `systemPrompt` that isn't an inline record.**

`systemPrompt` is put unconditionally, and `promptFieldSource(null)` (line 412) returns ``string ` ` `` when the property is absent. The property *is* absent whenever the form never loaded Role/Instructions — `CodeAnalyzer` only calls `applyAgentFormShape` when the declared value's kind is `MAPPING_CONSTRUCTOR` (CodeAnalyzer.java:1017).

So for:

```ballerina
final workflow:DurableAgent a = check new ({systemPrompt: SUPPORT_PROMPT, model: m});
```

opening the box's gear form and saving rewrites `SUPPORT_PROMPT` to ``{role: string ` `, instructions: string ` `}`` — the prompt is silently lost, and the user never edited those fields.

Suggest skipping the `systemPrompt` entry entirely when neither prompt property is present, so an untouched field round-trips unchanged.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -934,6 +938,237 @@ private void populateSleepNodeProperties(MethodCallExpressionNode callNode, Func
...
+        List<AgentCapabilityData> agentTools = new ArrayList<>();
+        List<AgentCapabilityData> updateEvents = new ArrayList<>();
+        for (MappingFieldNode field : ((MappingConstructorExpressionNode) configArg.expression()).fields()) {
+            if (!(field instanceof SpecificFieldNode specificField) || specificField.valueExpr().isEmpty()) {
+                continue;
+            }
+            String fieldName = specificField.fieldName().toSourceCode().trim();
+            ExpressionNode valueExpr = specificField.valueExpr().get();
+            switch (fieldName) {
+                case "systemPrompt" -> {
```
samithkavishke, line 1041: This is the guard behind the `systemPrompt` data loss noted on `DurableAgentRunBuilder.java:384`: the Role/Instructions properties are only created when the declared value is a mapping literal, but `DurableAgentRunBuilder.toSource` writes the field back unconditionally.

Either handle the non-literal shapes here (a const/variable/function-call `systemPrompt`), or make `toSource` treat "no prompt properties" as "leave the field alone".

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1491,23 +1726,31 @@ private void populateBuiltinActivityProperties(RemoteMethodCallActionNode callNo
...
                 Map<String, String> fields = parseSimpleRecord(rawValue);
                 maxRetries = fields.getOrDefault(ActivityCallBuilder.MAX_RETRIES_KEY, "");
                 retryDelay = fields.getOrDefault(ActivityCallBuilder.RETRY_DELAY_KEY, "");
                 retryBackoff = fields.getOrDefault(ActivityCallBuilder.RETRY_BACKOFF_KEY, "");
                 maxRetryDelay = fields.getOrDefault(ActivityCallBuilder.MAX_RETRY_DELAY_KEY, "");
+            } else if (!trimmed.equals("()") && !trimmed.contains("NoRetry")
+                    && !trimmed.contains("NoAutomaticRetry")) {
+                // HumanReview is the reviewer role(s): a string, a role list, or the legacy
+                // workflow:ManualRetry sentinel (any role).
+                dropdownValue = ActivityCallBuilder.MANUAL_RETRY_VALUE;
```
samithkavishke, line 1770: **Any retry-policy expression the analyzer doesn't recognise now becomes Human Review, and its source text is written back as a string literal.**

The new `else if` catches everything that isn't `{…}`, `()`, or a `NoRetry`/`NoAutomaticRetry` substring match — including a perfectly valid reference:

```ballerina
ctx->callActivity(chargeCard, retryPolicy = DEFAULT_RETRY);
```

`DEFAULT_RETRY` (a module const of the retry record type) sets `dropdownValue = MANUAL_RETRY` and `retryUserRoles = "DEFAULT_RETRY"`. Opening and saving that activity form then emits `quoteIfPlain("DEFAULT_RETRY")` (ActivityCallBuilder.java:796), rewriting the argument to the string literal `"DEFAULT_RETRY"` — a type error.

This affects plain workflow projects, not just durable agents. Suggest matching the Human Review sentinel forms explicitly and passing anything else through as an opaque expression instead of treating "unrecognised" as "reviewer roles".

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
```diff
@@ -744,14 +761,40 @@ private static void populateRetryPolicyArg(SourceBuilder sourceBuilder, Map<Stri
...
         }
         return switch (value) {
             // NO_RETRY is handled (and skipped) by populateRetryPolicyArg before reaching here.
-            case MANUAL_RETRY_VALUE -> "workflow:ManualRetry";
+            case MANUAL_RETRY_VALUE -> {
+                Property roles = properties.get(RETRY_USER_ROLES_KEY);
+                String rolesValue = roles == null || roles.value() == null
+                        ? "" : roles.value().toString().trim();
+                // ManualRetry is the reviewer role(s); an empty list means any role may decide.
+                yield rolesValue.isBlank() ? "[]" : WorkflowUtil.quoteIfPlain(rolesValue);
```
samithkavishke: This is the write-back half of the retry-policy issue flagged at `CodeAnalyzer.java:1744`. Because the analyzer now routes unrecognised policy expressions into `retryUserRoles`, `quoteIfPlain` turns a variable or const reference into a quoted string on save.

Worth guarding here too: only quote a value that is genuinely a bare role name, and leave anything that parses as an expression untouched.

Separately: the legacy `workflow:ManualRetry` sentinel now silently rewrites to `[]`. That looks intended for the new API, but it is a source rewrite on files the user didn't edit — worth calling out in the PR description.

packages/bi-diagram/src/visitors/NodeFactoryVisitor.ts
```diff
@@ -821,6 +822,40 @@ export class NodeFactoryVisitor implements BaseVisitor {
...
+        //
+        // Exception — the agent-only view: there the flow model is just
+        // [Start, agent box], so a start node has already been visited. Link it to the
+        // box with a non-editable edge (no add-button).
+        const isAgentBox = (node.metadata?.data as { agentBox?: boolean })?.agentBox === true;
+        // Only the synthetic declaration-canvas copy (agent-only view) gets the non-editable
+        // Start edge — an in-chain `agent.run(...)` statement also carries the agentBox marker
+        // but is a real statement, so its edges keep the add-button.
+        const isDeclarationCanvasBox =
+            node.id === "durable-agent-box" || node.id === "durable-agent-placeholder";
```
samithkavishke: **This branch is unreachable, so the behaviour the comment above describes never happens.**

Nothing in the repository ever assigns these ids — `grep -rn 'durable-agent-box' packages/` matches only this line. The declaration-canvas node ids come from the LS (`CodeAnalyzer.tryHandleDurableAgentDeclarationCanvas`), which emits generated ids.

`isDeclarationCanvasBox` is therefore always `false`, control falls through to `else if (this.lastNodeModel)`, and `updateNodeLinks` runs **without** `{showAddButton: false}`. The agent-only view renders an editable edge with an add-button between the Start pill and the agent box — exactly what the comment says must not happen.

Suggest keying off something the model actually carries (e.g. `metadata.data.agentBox` plus `lastNodeModel instanceof StartNodeModel`, or an explicit flag from the LS) rather than a node id.

packages/ballerina-visualizer/src/PopupPanel.tsx
```diff
@@ -186,6 +186,21 @@ const PopupPanel = (props: PopupPanelProps) => {
                         );
                     });
                     break;
+                case MACHINE_VIEW.BIDurableAgentForm:
+                    setIsFullScreen(true);
+                    rpcClient.getVisualizerLocation().then(async (location) => {
+                        const defaultFunctionsFile = (await rpcClient.getVisualizerRpcClient().joinProjectPath({ segments: ['functions.bal'] })).filePath;
```
samithkavishke: **`functions.bal` here, `workflow.bal` in `MainPanel` — same view, different target file.**

`MainPanel.tsx:766` resolves this view to `workflow.bal` with an explicit comment that "Durable agent declarations live in workflow.bal alongside the workflow artifacts". This popup path passes `functions.bal`, so an agent created from the popup lands in a different file than one created from the main panel.

Looks like a copy of the neighbouring cases; should this be the same `workflow.bal` lookup?

packages/ballerina-visualizer/src/views/BI/FlowDiagram/index.tsx
```diff
@@ -942,14 +1103,43 @@ export function BIFlowDiagram(props: BIFlowDiagramProps) {
...
+    // Holds the panel open with the loader while a capability write applies; the panel
+    // closes from getFlowModel() once the refreshed model has rendered.
+    const finishCapabilityOpAfterRefresh = () => {
+        pendingCapabilityCloseRef.current = true;
+        setShowProgressSpinner(true);
+        setProgressMessage("Applying changes...");
+        debouncedGetFlowModel();
+        setTimeout(() => debouncedGetFlowModel(), 1500);
+        setTimeout(() => debouncedGetFlowModel(), 4000);
+        // Failsafe: never leave the loader stuck if the refresh cannot land.
```
samithkavishke, line 1162: **This failsafe can close a *later* operation's panel, and none of the timers are cancellable.**

The timer fires unconditionally 10s after the op starts and acts on whatever `pendingCapabilityCloseRef.current` says at that moment:

1. t=0 — capability added, timer A armed for t=10s.
2. refresh lands, `pendingCapabilityCloseRef` clears.
3. t=8s — user adds a second capability, flag set again (line 1122).
4. t=10s — timer A sees `true`, clears the flag, hides the spinner and calls `closeSidePanelAndFetchUpdatedFlowModel`, closing the **second** op's panel mid-write.

The `1500/4000/8000ms` `debouncedGetFlowModel` ladders here and in `closeSidePanelAndFetchUpdatedFlowModel` have the same shape: they're never cleared, so they still fire (and `setState`) after the diagram has been navigated away from.

Suggest storing the timeout ids in a ref, clearing them at the start of each capability op and in an effect teardown.

packages/ballerina-extension/src/rpc-managers/bi-diagram/rpc-manager.ts
```diff
@@ -360,15 +361,17 @@ export class BiDiagramRpcManager implements BIDiagramAPI {
 
             if (model?.errorMsg) {
                 const errorMessage = model.errorMsg;
+                // The raw cause (often an internal exception) goes to the output channel;
+                // the toast stays actionable.
                 console.error(">>> error generating source code from ls", { errorMessage, stacktrace: model.stacktrace });
-                window.showErrorMessage(`Failed to save changes: ${errorMessage}`);
+                window.showErrorMessage("Failed to save changes: the operation could not be applied. Please try again.");
```
samithkavishke: **This drops the real error for every node kind, not just the agent ones.**

The new builders throw user-actionable messages — e.g. `IllegalStateException("An event name is required")` from `DurableAgentRegisterEventBuilder`, or "Cannot locate the durable agent declaration" — and those now surface as "the operation could not be applied. Please try again.", with the cause only in the output channel.

Since the goal is presumably to hide internal exception text, could the actionable validation messages still be shown (e.g. a distinguishable error field from the LS, or a prefix check) and the generic toast reserved for the unexpected cases? Same applies to line 374.

packages/ballerina-visualizer/src/utils/bi.tsx
```diff
@@ -441,11 +448,9 @@ export function getContainerTitle(view: SidePanelView, activeNode: FlowNode, cli
             ) {
                 return `${clientName || activeNode.properties.connection.value} → ${activeNode.metadata.label}`;
             } else if (activeNode.codedata?.node === "DATA_MAPPER_CALL") {
-                return `${activeNode.codedata?.module ? activeNode.codedata?.module + " :" : ""} ${activeNode.codedata.symbol
-                    }`;
+                return `${activeNode.codedata.symbol}`;
             }
-            return `${activeNode.codedata?.module ? activeNode.codedata?.module + " :" : ""} ${activeNode.metadata.label
-                }`;
+            return `${activeNode.metadata.label}`;
```
samithkavishke: **This drops the module prefix from every form title, not just the workflow ones.**

The PR describes cleaning up titles like "workflow : Run Workflow", but this fallback (and the `DATA_MAPPER_CALL` branch on line 451) is unconditional. A `regex:matches` `FUNCTION_CALL` form titled "regex : matches" now reads just "matches", losing the module that disambiguates same-named functions across modules.

If the intent is workflow-specific, suggest suppressing the prefix only for the workflow/agent node kinds and leaving library calls as they were.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -214,4 +333,593 @@ public static Optional<ClassSymbol> resolveConnectionClass(TypeSymbol typeSymbol
...
+            }
+        }
+    }
+
+    /**
+     * Asserts the node targets an object-model durable agent declaration.
+     *
+     * @param sourceBuilder the source builder
+     */
+    public static void requireDurableAgentObjectTarget(SourceBuilder sourceBuilder) {
```
samithkavishke, line 610: **This helper has no callers** — and neither do `addImportIfMissing` (638), `setAgentConfigField` (752, the singular one; only the plural `setAgentConfigFields` is used), `mergeTextEdits` (794), `resolveWrappableClass` (345), or the two-arg `declaredAgentEventOptions` overload. That's roughly 120 lines of new dead code.

Ironically this is the one the builders need: `DurableAgentRegisterToolBuilder`, `DurableAgentRegisterEventBuilder`, `DurableAgentHumanTaskBuilder` and friends each repeat

```java
if (WorkflowUtil.isDurableAgentObjectTarget(sourceBuilder)) { ... }
throw new IllegalStateException("Cannot generate the capability source: ...");
```

Either route them through `requireDurableAgentObjectTarget` or drop it, and drop the rest.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -214,4 +333,593 @@ public static Optional<ClassSymbol> resolveConnectionClass(TypeSymbol typeSymbol
...
+        return edits;
+    }
+
+    private record AgentDeclaration(Path filePath,
+            MappingConstructorExpressionNode config) {
+    }
+
+    // Scans the default module for `final workflow:DurableAgent <name> = check new ({...})`
+    // and returns the config mapping plus the declaring file.
+    private static AgentDeclaration findAgentDeclaration(SourceBuilder sourceBuilder, String agentVarName) {
```
samithkavishke, line 419: **Only the implicit-`new` shape is matched here, while `declaredAgentEventOptions` in this same file also accepts `ExplicitNewExpressionNode` — so the two disagree.**

```ballerina
final workflow:DurableAgent a = check new workflow:DurableAgent({events: [{name: "chat", request: string}]});
```

`declaredAgentEventOptions` lists `"chat"` in the Send Agent Data dropdown, but this method skips the declaration, so `insertAgentCapabilityEntry` / `replaceAgentCapabilityEntry` / `removeAgentCapabilityEntry` / `setAgentConfigFields` all throw "Cannot locate the durable agent declaration" — which (see the rpc-manager comment) reaches the user as the generic save-failed toast.

Also worth noting: the sample in the PR description uses named arguments (`new (name = …, systemPrompt = …, tools: […])`), which neither path supports — only `check new ({...})` with a positional mapping works. Either the description or the parser needs updating.

Suggest extracting one shared "resolve the agent config literal" helper and using it in both places.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
rdulmina: Fixed in 4715fab7.

`addNormalizedRetryPolicyProperties` now matches the shapes explicitly instead of treating "unrecognised" as reviewer roles:

- `()` / `NoRetry` / `NoAutomaticRetry` -> `NoRetry`
- the `ManualRetry` / `HumanReview` / `[]` sentinels -> Human Review, no roles
- a string or list literal -> Human Review with those roles
- anything else (a const or variable such as `DEFAULT_RETRY`) is carried through as the dropdown value itself

For that last case `addRetryPolicyFormProperties` adds the expression as its own option so it renders as the selection, and `retryPolicyExpression`'s `default` branch writes it back verbatim. `ctx->callActivity(chargeCard, retryPolicy = DEFAULT_RETRY)` now round-trips unchanged.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
rdulmina: Fixed in 4715fab7.

The roles value now goes through the new `WorkflowUtil.quoteIfBareRole`, which passes anything expression-shaped (module-qualified reference, call, member access, interpolation) through untouched and only quotes a bare word. The analyzer half is fixed as described on the `CodeAnalyzer.java` thread, so an unrecognised policy no longer reaches this field in the first place.

On the second point: the legacy `workflow:ManualRetry` -> `[]` rewrite is intended for the new API, and agreed it is a source rewrite on files the user did not edit — it should be called out in the PR description.

packages/bi-diagram/src/visitors/NodeFactoryVisitor.ts
rdulmina: Fixed in 4715fab7.

`tryHandleDurableAgentDeclarationCanvas` now stamps `metadata.data.agentDeclarationCanvas` on the synthetic box, and the visitor keys `isDeclarationCanvasBox` off that instead of the node ids (which, as you found, are never assigned). An in-chain `agent.run(...)` statement carries only `agentBox`, so its edges keep the add-button; the agent-only view gets the non-editable Start edge the comment describes.

packages/ballerina-visualizer/src/PopupPanel.tsx
rdulmina: Fixed in 4715fab7 — the popup path now resolves `workflow.bal`, matching `MainPanel.tsx:766`. It was a copy of the neighbouring cases.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/index.tsx
rdulmina: Fixed in 4715fab7.

The ladders are tracked in `refreshTimersRef` and the failsafe in `capabilityFailsafeTimerRef`:

- `finishCapabilityOpAfterRefresh` clears both before arming new ones, so a timer from operation A cannot act on operation B
- the failsafe is disarmed as soon as the refreshed model lands (in `getFlowModel`) and when the panel is closed
- both are cleared in the mount effect's teardown, so nothing `setState`s after navigation

The ladder is deliberately left running on a manual panel close — it is only a model fetch, and cancelling it there would leave the diagram stale after a write.

packages/ballerina-extension/src/rpc-managers/bi-diagram/rpc-manager.ts
rdulmina: Fixed in 4715fab7.

Added `UserFacingException` in flow-model-generator-core. `AbstractFlowModelResponse.setError` marks those responses with `userFacing` (now on `BISourceCodeResponse`), and `getSourceCode` shows `errorMsg` verbatim for them while keeping the generic toast for everything else. The durable agent validation throws use the new type — "An event name is required", "An activity function must be selected", "A human task name is required", the "Cannot locate the durable agent declaration" family and `requireDurableAgentObjectTarget`.

Line 374 (empty source update) stays generic: that one is an internal condition with nothing actionable in it.

packages/ballerina-visualizer/src/utils/bi.tsx
rdulmina: Fixed in 4715fab7.

Both branches go through a `modulePrefix(activeNode)` helper that returns the `module + " : "` prefix as before, suppressed only when `codedata.module` is `workflow` or the node kind starts with `DURABLE_AGENT`. `regex : matches` reads as it did.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
rdulmina: Fixed in 4715fab7.

The five durable agent builders (`DurableAgentRegisterToolBuilder`, `DurableAgentRegisterEventBuilder`, `DurableAgentHumanTaskBuilder`, `DurableAgentAddActivityBuilder`, `DurableAgentRunBuilder`) now call `WorkflowUtil.requireDurableAgentObjectTarget(sourceBuilder)` instead of repeating the if/throw, and `resolveWrappableClass`, `addImportIfMissing`, `setAgentConfigField`, `mergeTextEdits` and the two-arg `declaredAgentEventOptions` overload are deleted.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
rdulmina: Fixed in 4715fab7.

Both paths now go through one `agentConfigLiteral(ModuleVariableDeclarationNode)` helper that accepts the implicit `check new ({...})` and the explicit `check new workflow:DurableAgent({...})` shapes, so `findAgentDeclaration` and `declaredAgentEventOptions` no longer disagree.

Named arguments (`new (name = ..., systemPrompt = ...)`) are still unsupported — only a positional config mapping is handled — so the PR description sample needs updating.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentRunBuilder.java
rdulmina: Fixed in 4715fab7 — the `systemPrompt` entry is skipped entirely when neither prompt property is present on the form, so a `systemPrompt` that is not an inline record (`SUPPORT_PROMPT`) round-trips unchanged. `setAgentConfigFields` emits no edit for a field that is not in the map.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
rdulmina: Handled on the write side rather than here (see the `DurableAgentRunBuilder.java` thread): with no Role/Instructions properties on the form, `toSource` now leaves the declared `systemPrompt` alone instead of rewriting it. 4715fab7.

packages/ballerina-visualizer/src/views/BI/FunctionForm/index.tsx
```diff
@@ -767,8 +869,37 @@ export function FunctionForm(props: FunctionFormProps) {
...
+            const taken = (await rpcClient.getBIDiagramRpcClient().getFunctionNames()).mentions;
+            return taken.includes(name);
+        } catch (error) {
+            console.error("Failed to check function name availability:", error);
+            return false;
+        }
+    };
+
     const handleFormSubmit = async (data: FormValues, formImports?: FormImports) => {
         setSaving(true);
```
gigara, line 562: `handleFormSubmit` sets `saving = true`, but the `catch` around `await onSubmit(...)` below only shows the error notification and never resets it. Since the form is replaced by the full-screen loader while `saving && !functionName`, a failed creation leaves the user permanently stuck on "Creating the …" — and this handler serves every artifact type in this form, not just durable agents. Suggest resetting in a `finally { setSaving(false); }`.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/index.tsx
```diff
@@ -1940,6 +2296,51 @@ export function BIFlowDiagram(props: BIFlowDiagramProps) {
 
             })
             .then(async (response) => {
+                // A toolkit created from the durable agent's Add Tool/Activity list:
+                // register the new variable on the agent declaration's tools list.
+                const pendingMcpAgent = pendingDurableMcpAgentRef.current;
+                if (pendingMcpAgent && nodeToSubmit?.codedata?.node === "MCP_TOOL_KIT") {
```
gigara, line 2290: This branch registers the toolkit on the agent before checking whether the toolkit write actually succeeded. `getSourceCode` resolves with `{ artifacts: [], error }` on LS failure rather than rejecting, so a failed MCP toolkit creation still generates a `tools:` entry referencing a variable that was never declared — a compile error in the user's project. Guard on `response.artifacts.length > 0` (or absence of `response.error`) before registering.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/DesignModelGenerator.java
```diff
@@ -320,6 +364,163 @@ private void populateWorkflowEvents(Workflow workflow, FunctionSymbol functionSy
...
+                    || varDecl.lineRange().startLine().line() > lineRange.startLine().line()
+                    || varDecl.lineRange().endLine().line() < lineRange.startLine().line()
+                    || varDecl.initializer().isEmpty()) {
+                continue;
+            }
+            ExpressionNode initializer = varDecl.initializer().get();
+            if (initializer instanceof CheckExpressionNode checkExpr) {
+                initializer = checkExpr.expression();
+            }
+            if (!(initializer instanceof ImplicitNewExpressionNode newExpr)
```
gigara: This still accepts only the implicit-new + positional-config shape, and flow-model `CodeAnalyzer.populateAgentDeclarationMetadata` has the same restriction — while `WorkflowUtil.agentConfigLiteral` now accepts explicit `new workflow:DurableAgent({...})` too (its javadoc says "every path that reads or edits a declaration agrees", which these two paths no longer satisfy). Net effect: an explicit-new agent's capabilities are editable but invisible — the box renders with no capability circles.

Also worth noting: the named-argument form (`new (name = ..., systemPrompt = {...}, ...)`) is parsed by none of these paths, and the sample in the PR description uses exactly that form — either support it or fix the sample.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentHumanTaskBuilder.java
```diff
@@ -0,0 +1,249 @@
...
+        String title = sourceBuilder.getProperty(TITLE_KEY)
+                .map(p -> p.value() == null ? "" : p.value().toString().trim()).orElse("");
+        String taskDescription = sourceBuilder.getProperty(DESCRIPTION_KEY)
+                .map(p -> p.value() == null ? "" : p.value().toString().trim()).orElse("");
+        String resultType = sourceBuilder.getProperty(RESULT_TYPE_KEY)
+                .map(p -> p.value() == null ? "" : p.value().toString().trim()).orElse("");
+        String timeout = sourceBuilder.getProperty(TIMEOUT_KEY)
+                .map(p -> p.value() == null ? "" : p.value().toString().trim()).orElse("");
+        StringBuilder entry = new StringBuilder("{name: ").append(WorkflowUtil.constantNameLiteral(name))
+                .append(", roles: ").append(roles.isBlank() ? "\"manager\""
```
gigara: A blank User Roles field silently emits `roles: "manager"` — a privileged default the user never chose (and the form placeholder shows `"MANAGER"` while this emits lowercase). The non-agent `HumanTaskBuilder.toSource` rejects blank roles; suggest matching that behavior (or emitting an explicit any-role value if the config supports one) instead of defaulting to a named role.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -415,6 +421,43 @@ public void visit(MethodCallExpressionNode methodCallExpressionNode) {
...
+     * exactly like {@code workflow:run} does for workflow functions.
+     *
+     * @param methodCallExpressionNode the method call to inspect
+     */
+    private void handleDurableAgentCall(MethodCallExpressionNode methodCallExpressionNode) {
+        Optional<Symbol> targetSymbol = semanticModel.symbol(methodCallExpressionNode.expression());
+        if (targetSymbol.isEmpty() || targetSymbol.get().getName().isEmpty()) {
+            return;
+        }
+        Workflow agent = intermediateModel.workflowMap.get(targetSymbol.get().getName().get());
```
gigara, line 437: The agent is resolved by symbol *name* only — a local variable shadowing a module-level agent name (e.g. `http:Client supportAgent = ...; supportAgent.run(...)` inside a function) hits the map entry, and the kind check validates the map entry rather than the call target, so a false `run` edge is drawn to the agent. `WorkflowUtil.isDurableAgentVariable(targetSymbol.get())` already exists and would rule this out.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/ActivityWizardSteps.tsx
```diff
@@ -0,0 +1,79 @@
+/**
```
gigara: This component isn't imported anywhere in the PR — its name appears only in this file, and since the file is new, nothing pre-existing can reference it either. Dead code: either delete it or wire it into the create-activity-from-connection wizard it describes.

gigara (commented):
Reviewed the durable-agent changes — flagging the high-priority items inline (plus one dead file). Overall the object-model design and the `WorkflowUtil` declaration-edit machinery look solid.

packages/ballerina-visualizer/src/views/BI/FunctionForm/index.tsx
rdulmina: Fixed in 6f8c455 — `saving` now resets in a `finally`.

Worth flagging a second path with the same symptom that the `finally` also covers: in `onSubmit`, when `artifacts` is non-empty but neither `newArtifact` nor `updatedArtifact` matches, it returns without resetting `saving` and without navigating, stranding the user the same way.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/index.tsx
rdulmina: Fixed in 6f8c455.

Confirmed the mechanism: all three failure paths in `BiDiagramRpcManager.getSourceCode` return `{ artifacts: [], error }` and it never rejects, so the `.then` always ran.

Guarded on `response?.error` rather than `artifacts.length > 0` — the latter can legitimately be empty for the `NODES_TO_SKIP_ARTIFACT` kinds (`MODEL_PROVIDER`/`KNOWLEDGE_BASE`/`DATA_LOADER`), so `error` is the unambiguous signal.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentHumanTaskBuilder.java
rdulmina: Fixed in 6f8c455 — a blank User Roles now throws `UserFacingException` instead of defaulting, which also matches how this builder already rejects a blank task name.

`HumanTaskBuilder` turned out to state this explicitly ("in particular, never fall back to a privileged role for userRoles"), so this was contradicting a documented decision in its sibling. The `MANAGER`/`manager` casing mismatch you spotted disappears along with the default.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/CodeAnalyzer.java
rdulmina: Fixed in 6f8c455 — added the `WorkflowUtil.isDurableAgentVariable(targetSymbol.get())` guard, which was already imported in this file.

Placed it after the existing map/kind lookup so the type resolution only runs for calls whose name already matches a known agent, rather than on every method call.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/ActivityWizardSteps.tsx
rdulmina: Removed in 6f8c455 — confirmed nothing in the repo references it.

Deleted rather than wired in: hooking it into the create-activity flow would mean adding UI this PR doesn't otherwise have.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentHumanTaskBuilder.java
```diff
@@ -0,0 +1,252 @@
...
+            String title = sourceBuilder.getProperty(TITLE_KEY)
+                    .map(p -> p.value() == null ? "" : p.value().toString().trim()).orElse("");
+            String taskDescription = sourceBuilder.getProperty(DESCRIPTION_KEY)
+                    .map(p -> p.value() == null ? "" : p.value().toString().trim()).orElse("");
+            String resultType = sourceBuilder.getProperty(RESULT_TYPE_KEY)
+                    .map(p -> p.value() == null ? "" : p.value().toString().trim()).orElse("");
+            String timeout = sourceBuilder.getProperty(TIMEOUT_KEY)
+                    .map(p -> p.value() == null ? "" : p.value().toString().trim()).orElse("");
+            StringBuilder entry = new StringBuilder("{name: ").append(WorkflowUtil.constantNameLiteral(name))
+                    .append(", roles: ").append(roles.isBlank() ? "\"manager\""
```
gigara: Blank User Roles silently emits `roles: "manager"` — a privileged role the user never chose (and the form placeholder says `"MANAGER"` while this emits lowercase `"manager"`). The non-agent `HumanTaskBuilder` rejects blank roles; this should match (fail with a clear message) rather than defaulting to a privileged role.

packages/ballerina-visualizer/src/views/BI/FunctionForm/index.tsx
```diff
@@ -767,8 +869,37 @@ export function FunctionForm(props: FunctionFormProps) {
...
+        if (isDurableAgent && !functionName && await isDuplicateFunctionName(data.functionName)) {
```
gigara, line 563: With this PR the form is replaced by the full-screen loader while `saving && !functionName`, but `handleFormSubmit`'s `catch` around `await onSubmit(...)` shows the error notification without ever resetting `saving`. A failed creation now leaves the user stuck on "Creating the …" permanently, with no form to retry — and this handler is shared by every artifact type using this form, not just durable agents.

A `finally { setSaving(false); }` (or a reset in the catch) covers it; the early-return duplicate-name path above already resets it correctly.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/ActivityWizardSteps.tsx
gigara: This component isn't imported anywhere — `ActivityWizardSteps` only appears in this file. Either wire it into the create-activity-from-connection wizard it describes, or drop the file.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -214,4 +332,495 @@ public static Optional<ClassSymbol> resolveConnectionClass(TypeSymbol typeSymbol
...
+                }
+                Optional<MappingConstructorExpressionNode> config = agentConfigLiteral(varDecl);
+                if (config.isPresent()) {
+                    return new AgentDeclaration(project.documentPath(documentId).orElse(sourceBuilder.filePath),
+                            config.get());
+                }
+            }
+        }
+        return null;
+    }
```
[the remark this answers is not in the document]
rdulmina: https://github.com/wso2/product-integrator/issues/2041 to track this

rdulmina:
Closing this since these changes are included in #748

PR #749, opened 2026-07-31
Add Ballerina Central call to Library Search Tool

ravindu25:
## Summary

`LibrarySearchTool` now searches Ballerina Central instead of the SQLite index bundled in the LS jar, so library discovery reflects Central's current contents without needing an LS release.

The LSP method, request/response types, tool schema, event shape and `LibraryGetTool` are unchanged.

## What's new

**`CentralLibrarySearchAccessor`** — searches Central and narrows the results:

1. Sanitizes the model's keywords to alphanumeric tokens and joins them into one `q`.
2. Calls `GET /2.0/registry/search-packages?q=…&limit=30`.
3. Filters the 30 results, keeping Central's rank order:
   - **org filter** — `ballerina` and `ballerinax` only
   - **relevance gate** — keeps matches Central highlighted in `summary`/`keywords`, or whose package name carries a query term
   - dedupe by name
4. 30-minute in-memory cache keyed on the query.

**Fallback** — any failure on the Central path logs a warning and falls back to the existing SQLite search. `-Dballerina.copilot.librarySearch.useLocalIndex=true` forces the local path.

**Result cap** — top 10, with `exclusion.json` applied *before* truncation (previously an excluded library silently consumed a slot).

**Tool description** — removed the inaccurate "keywords are positionally weighted" claim, plus two pieces of guidance that were causing result dilution (the blanket `trigger` keyword advice, and encouraging generic terms like "issue"). Replaced with: prefer 1–3 distinctive keywords, avoid generic words, scope `trigger` to event-listening, one search per service.

## Testing

23 unit tests (`CentralLibrarySearchAccessorTest`), mocked via `RemoteCentral.setTestInstance` — no network in CI. Checkstyle and spotbugs clean.

Manually verified end to end: Central reached, org filter holds, search → select → fetch works on packages the model can't guess, generated code compiles, and the fallback returns results with Central blackholed.

### DIFF
--- a/packages/ballerina-extension/src/features/ai/agent/tools/library-search.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/tools/library-search.ts
@@ -35,7 +35,7 @@ const LibrarySearchToolSchema = jsonSchema<{
         keywords: {
             type: "array",
             items: { type: "string" },
-            description: `Array of search keywords to find relevant Ballerina libraries. Keywords are weighted by order - first keyword has highest weight, subsequent keywords have decreasing weight. Maximum ${MAX_SEARCH_KEYWORDS} keywords allowed. Examples: ["GitHub", "API", "integration"], ["Stripe", "payment", "gateway"], ["HTTP", "REST", "service"]`,
+            description: `Array of search keywords to find relevant Ballerina libraries. Keywords are combined with OR, so each additional keyword widens the search rather than narrowing it. Maximum ${MAX_SEARCH_KEYWORDS} keywords allowed. Examples: ["GitHub", "API", "integration"], ["Stripe", "payment", "gateway"], ["HTTP", "REST", "service"]`,
             minItems: 1,
             maxItems: MAX_SEARCH_KEYWORDS,
         },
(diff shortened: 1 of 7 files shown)


packages/ballerina-extension/src/features/ai/agent/tools/library-search.ts
```diff
@@ -132,23 +132,22 @@ export function getLibrarySearchTool(
...
 
 **Purpose:**
-This tool discovers relevant Ballerina libraries using keyword-based search. It searches against library names, descriptions, and function names. Keywords are weighted by order - the first keyword has the highest weight, with decreasing weight for subsequent keywords.
+This tool discovers relevant Ballerina libraries using keyword-based search. It searches against library names, descriptions, package keywords, and documentation. All keywords are combined with OR and the matches are ranked by overall relevance, so each additional keyword widens the search rather than narrowing it.
 
-**Scope - ALL Ballerina Libraries:**
+**Scope:**
 - **ballerina/*** - Standard/core libraries maintained by the Ballerina team
 - **ballerinax/*** - Extended connector packages for third-party services
-- **xlibb/*** - Experimental yet practical Ballerina packages
```
RNViththagan, line 143: we need xlibb support

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/database/LibraryDatabaseAccessor.java
```diff
@@ -145,8 +145,10 @@ public static Optional<String> getPackageDescription(String org, String packageN
         return Optional.empty();
     }
 
-    // Maximum number of libraries returned by keyword search.
-    private static final int MAX_SEARCH_RESULTS = 9;
+    // Number of libraries fetched by keyword search. This is an over-fetch: the caller applies
+    // exclusions and then truncates to the number of results it actually surfaces, so fetching more
+    // here keeps an excluded library from consuming a result slot.
+    private static final int SEARCH_FETCH_LIMIT = 30;
```
RNViththagan, line 151: Fetching 30 to surface 4 or 5 makes it hard for the agent. Better relevance should let this come down.

Non-blocking — we will improve the search accuracy later.

SasinduDilshara (approved):

PR #764, opened 2026-08-03
Add trigger search improvements and local-repository connector support (5/5)

LakshanWeerasinghe:
## Summary

Improves Ballerina Central trigger search, and adds **experimental** support for resolving a connector from the Ballerina local repository (`~/.ballerina/repositories/local`) — i.e. one pushed locally via `bal push --repository=local` during connector development, before it's published to Central — so it can be found and added through the same trigger search UI.

## Approach

- **`TriggerSearchUtil`** — Central search now scopes by the dedicated `org` query parameter (previously an unverified `org:` token embedded in `q`), runs per-org queries on a dedicated executor with independent per-org failure handling, and merges results round-robin across orgs so one org can't starve another out under result-limit truncation.
- **`LibraryMetadataReader` / `TriggerModelReader`** — new local-repository read paths reuse the same metadata + introspection synthesis pipeline already used for Central connectors, backed by a cached repository handle instead of re-resolving on every call.
- **`LocalDependencyEditUtil`** (new) — adds a `[[dependency]] ... repository = "local"` entry to `Ballerina.toml` when a local-repository connector is added, or bumps the version in place if the connector is already declared at a different version (the routine edit-and-repush loop).
- **Frontend** (`CentralSearchPanel`, `ServiceCreationView`) — surfaces local-repository results in their own section, distinct from Central results.

Both the local-repository search and the "More on Ballerina Central" panel are gated behind one new experimental setting, `ballerina.enableAdditionalTriggerSearch` (off by default) — the feature isn't fully built out yet.


pasindufernando1 (commented):
## Review

Reviewed the full diff (62 files, +1873/-153) and cross-checked the key call paths against the PR head and the base tree.

The architecture is sound: local and Central results are kept strictly separate, the feature is off by default behind `ballerina.enableLocalCentralSearch`, and the shared `synthesizeTriggerModel` extraction is a clean way to give the local tier both Central tiers. I also verified the `isLocalRepository` round-trip genuinely works — `ServiceCreationView.tsx:555` mutates the server's model in place rather than reconstructing it, so the flag survives to `createServiceAndListener`. The `ListenerUtil` overload fix and its explanation are good work.

Inline comments below. Summarising the ones I'd want resolved before merge:

**High**
1. `LibraryMetadataReader.localRepository()` calls `PackageUtil.getSampleProject()`, which creates a temp dir + `BuildProject.load` on *every* call and never cleans up — and `searchLocalRepository` hits it ~2x per local package, per debounced search.
2. `searchCentral`'s `future.join()` inside the blanket catch means one failing org call discards *both* orgs' results.
3. Results are concatenated `ballerina`-first then truncated to the limit, so `ballerinax` — where the connectors actually live — can be cut entirely.

**Medium**
4. `org:` embedded in `q` is only an established pattern for `searchSymbols` in this codebase; `searchPackages` has a dedicated `org` param. Currently only verified against a hand-written fake.
5. Dropping `hasListeners` regresses discovery for `smb`/`mqtt` (per the new javadoc) — worth confirming those now carry `Type/Trigger`.
6. `LocalDependencyEditUtil.alreadyDeclared` ignores version, so a re-pushed `0.2.0` silently keeps `0.1.0` pinned.
7. Tests write fixture packages into the real `~/.ballerina/repositories/local` with best-effort cleanup.

**Also verified as non-issues:** `TriggerListRequest` gaining a 5th record component is safe (Gson-only, no `new TriggerListRequest(` anywhere in Java, `boolean` defaults to `false` for older clients); `TriggerBasicInfo` field ordering in `toLocalRepositoryTriggerBasicInfo` matches `toTriggerBasicInfo` positionally.

**Security:** nothing concerning. The feature enumerates and compiles packages from the user's own local repository, behind a default-off experimental setting — right call.

**Test coverage:** good breadth (~20 new tests) across the routing flag, the synthesis tier, incomplete-`ModuleInfo` handling, the no-cached-miss property, and dependency-edit deduplication. Gaps: `searchCentral`'s partial-failure path and truncation behaviour are untested, and the local-repo tests aren't hermetic.

A few smaller notes: four new silent `catch (Throwable)` sites on the local path, the `ListenerUtil` overload pair now carrying opposite contracts for the same parameter, and several javadocs running to essay length with at least one claim ("cheap file-stat per candidate") contradicted by finding #1 — stale comments are worse than none.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReader.java
```diff
@@ -119,6 +129,97 @@ public Optional<TriggerMetadataModel> getPackagedTriggerMetadataModel(ModuleInfo
...
+        return getCompiledPackageFromLocalRepository(moduleInfo).map(pkg -> pkg.project().sourceRoot());
+    }
+
+    /**
+     * The Ballerina local repository ({@code ~/.ballerina/repositories/local}), resolved via a throwaway
+     * sample project purely to obtain an {@link io.ballerina.projects.environment.Environment} -- mirrors
+     * {@link PackageUtil#getSampleProject()}'s existing use of the same trick for Central resolution.
+     */
+    private PackageRepository localRepository() {
+        return BallerinaUserHome.from(PackageUtil.getSampleProject().projectEnvironmentContext().environment())
```
pasindufernando1: **High: this creates a fresh temp-dir project on every call, and `searchLocalRepository` calls it O(N) times per search.**

`PackageUtil.getSampleProject()` is not the cached `SAMPLE_PROJECT` field — it's the public method (`PackageUtil.java:139`) that creates a temp directory, writes `main.bal` + `Ballerina.toml`, and runs `BuildProject.load(tempDir)` on **every invocation**, with no cleanup or `deleteOnExit`.

`TriggerSearchUtil.searchLocalRepository` calls into this up to **2x per package** (UI-schema probe + metadata probe), plus once for the enumeration. A local repo with 20 packages means ~41 temp directories and ~41 `BuildProject.load` calls per debounced search, each leaked to `/tmp`.

This also makes the `searchLocalRepository` docstring's claim of "a cheap file-stat per candidate" inaccurate.

**Suggested fix, which preserves the design intent stated below:** cache the `PackageRepository` **handle** in a lazily-initialised static, not the resolution results. The `getCompiledPackageFromLocalRepository` javadoc's concern — that caching would "silently persist a stale not-found" — applies to caching `getPackage` *results*, which is a different thing. The repository object is a lookup surface over the live filesystem; reusing it doesn't cache a miss.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/TriggerSearchUtil.java
```diff
@@ -49,58 +51,108 @@ public final class TriggerSearchUtil {
...
+                        queryMap.put("offset", String.valueOf(effectiveOffset));
+                        PackageResponse response = central.searchPackages(queryMap);
+                        return toTriggerResults(response, existingKeys);
+                    }))
+                    .toList();
+
+            List<TriggerBasicInfo> results = new ArrayList<>();
+            Set<String> seen = new HashSet<>();
+            for (CompletableFuture<List<TriggerBasicInfo>> future : futures) {
+                for (TriggerBasicInfo info : future.join()) {
```
pasindufernando1: **High: one failing org call discards both orgs' results.**

`future.join()` throws `CompletionException` if either per-org call fails, and the surrounding `catch (Throwable) -> List.of()` then drops **everything**. If `ballerina` succeeds and `ballerinax` (where nearly all trigger connectors live) times out, the user sees no Central results at all.

```java
.map(org -> CompletableFuture.supplyAsync(() -> { ... })
        .exceptionally(e -> List.of()))
```

Separately: `supplyAsync` without an executor puts blocking Central HTTP on `ForkJoinPool.commonPool()`, nested inside `searchTriggers`' own common-pool task. Deadlock is unlikely thanks to join help-stealing, but under concurrent searches this starves a pool shared with every parallel stream in the LS. Worth passing an explicit executor.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/TriggerSearchUtil.java
```diff
@@ -49,58 +51,108 @@ public final class TriggerSearchUtil {
...
+                    if (seen.add(info.orgName() + "/" + info.packageName())) {
+                        results.add(info);
+                    }
                 }
-                budget[0]--;
-                return hasListeners(central, pkg.organization(), pkg.name(), pkg.version());
-            };
-            return toTriggerResults(response, existingKeys, exportsListener);
+            }
+            return results.size() > effectiveLimit ? results.subList(0, effectiveLimit) : results;
```
pasindufernando1: **High: truncation systematically favours `ballerina` over `ballerinax`.**

`ALLOWED_ORGS` is iterated in declaration order and results are appended org-by-org, so once `ballerina` alone yields `effectiveLimit` (30) post-filter matches, **every** `ballerinax` result is truncated away. Since `ballerinax` is where the trigger connectors actually live, this fails in the worst direction.

Interleave the two lists, or apply a per-org limit of `effectiveLimit / ALLOWED_ORGS.size()`.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/TriggerSearchUtil.java
```diff
@@ -49,58 +51,108 @@ public final class TriggerSearchUtil {
...
+            String effectiveQuery = (query == null || query.isBlank()) ? DEFAULT_QUERY : query.trim();
+            int effectiveLimit = limit == null || limit <= 0 ? DEFAULT_LIMIT : limit;
+            int effectiveOffset = offset == null || offset < 0 ? 0 : offset;
+
+            // One org can't be expressed per call (see ALLOWED_ORGS), so the per-org Central calls are
+            // independent - run them concurrently rather than paying their latency serially.
+            List<CompletableFuture<List<TriggerBasicInfo>>> futures = ALLOWED_ORGS.stream()
+                    .map(org -> CompletableFuture.supplyAsync(() -> {
+                        Map<String, String> queryMap = new HashMap<>();
+                        queryMap.put("q", effectiveQuery + " org:" + org);
```
pasindufernando1: **Medium: the `org:` token inside `q` is unverified for `searchPackages`.**

In this codebase the `org:<name>`-inside-`q` form is only established for **`searchSymbols`** (`CentralSearchUtil.java:221`). For `searchPackages`, `SearchListGenerator.java:81` uses a **dedicated param** instead:

```java
centralApi.searchPackages(Map.of("org", org, "limit", String.valueOf(LIMIT)));
```

`TriggerSearchUtilTest` only asserts against a hand-written `FakeCentralAPI` built to match this assumption, so it can't catch a mismatch. Recommend `queryMap.put("org", org)` alongside `q`, matching the existing `searchPackages` caller — or confirming against live Central before merge.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/TriggerSearchUtil.java
```diff
@@ -113,32 +165,20 @@ static List<TriggerBasicInfo> toTriggerResults(PackageResponse response, Set<Str
...
-        }
-    }
-
     /**
-     * Whether a Central package is an event-integration trigger, based on its keywords or the
-     * {@code trigger.*} module-name convention.
+     * Whether a Central package is an event-integration trigger, based on its {@code Type/Trigger}
+     * classification tag, a bare trigger/listener/event keyword, or the {@code trigger.*} module-name
+     * convention. {@code Type/Trigger} is Central's own curated classification (e.g. {@code aws.sqs},
+     * {@code kafka}, {@code rabbitmq}). A package that exports a Listener but carries none of these
```
pasindufernando1: **Medium: worth confirming this coverage loss is acceptable.**

Dropping the `hasListeners` fallback removes up to 30 Central calls per search, which is a real win — but this javadoc is candid about the cost: `smb` and `mqtt` are no longer detected.

Note this PR simultaneously adds an `smb` test fixture (`connector_models/smb/resources/service-creation-blank-listener.json`) while making `smb` undiscoverable through search. Could you confirm `smb`/`mqtt`/`activemq` now carry the `Type/Trigger` tag on Central? If not, the trade is a net coverage loss for exactly the connectors being tested here.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/LocalDependencyEditUtil.java
```diff
@@ -0,0 +1,123 @@
...
+        Optional<BallerinaToml> toml = currentPackage.ballerinaToml();
+        if (toml.isEmpty()) {
+            return;
+        }
+        TextEdit tomlEdit = createLocalDependencyEdit(toml.get(), org, name, version);
+        String tomlPath = project.sourceRoot().resolve(ProjectConstants.BALLERINA_TOML).toString();
+        edits.computeIfAbsent(tomlPath, ignored -> new ArrayList<>()).add(tomlEdit);
+    }
+
+    private static boolean alreadyDeclared(PackageManifest manifest, String org, String name) {
```
pasindufernando1: **Medium: ignoring version breaks the bump-and-repush loop.**

`alreadyDeclared` matches on `org`/`name` only. The target workflow here is a developer iterating on a connector, where bumping `0.1.0` -> `0.2.0` and re-pushing is routine. On the second add, `Ballerina.toml` keeps `version = "0.1.0"` pinned, no edit is emitted, and the user gets either a stale resolve or a build failure — with nothing surfaced to explain it.

Suggest: when the declared version differs from the requested one, emit a replacement edit for the version value rather than a silent no-op.

packages/ballerina-language-server/model-generator-commons/src/test/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReaderTest.java
```diff
@@ -18,25 +18,119 @@
...
 public class LibraryMetadataReaderTest {
 
     private static final LibraryMetadataReader READER = LibraryMetadataReader.getInstance();
 
+    private static final String LOCAL_ORG = "readertestlocalrepo";
+    private static final String LOCAL_PACKAGE = "probeconnector";
+    private static final String LOCAL_VERSION = "0.1.0";
+
+    private static Path localRepoPackageDir() {
+        return Path.of(System.getProperty("user.home"), ".ballerina", "repositories", "local", "bala", LOCAL_ORG);
```
pasindufernando1: **Medium: these tests write into the developer's real `~/.ballerina/repositories/local`.**

The class javadoc is honest about why ("no injectable/fake repository root in the current API"), but the consequences are real, and this pattern is now repeated in three test classes:

- Cleanup is best-effort (`catch (IOException ignored)`). A JVM crash or interrupted run leaves `readertestlocalrepo` / `connectormodelreadertest` / `triggersearchutiltest` behind in the real repo — and `searchLocalRepository` will then happily surface them in the developer's own IDE trigger search.
- Concurrent test runs, or CI with a shared home, collide on these fixed org names.

Suggest adding a seam rather than shipping tests that write to a machine-wide location: a package-private repo-root override on `LibraryMetadataReader`, or a test-only `user.home` system property in the Gradle `test {}` block.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/ListenerUtil.java
```diff
@@ -332,21 +332,39 @@ private static String getListenerProtocol(String packageName) {
 
     public static Optional<Listener> getListenerModelByName(Codedata codedata, SemanticModel semanticModel,
                                                             ModuleInfo moduleInfo) {
-        return getListenerModelByName(codedata, semanticModel, moduleInfo, true);
+        // `semanticModel` here is the connector's own package semantic model (see
+        // ConnectorModelReader#synthesizeTriggerModel), not a user file's -- passing it through to the
+        // builder is what lets a local-repository connector's "Listener" class resolve at all, since it
+        // isn't Central-indexed for FunctionDataBuilder's own moduleInfo-based derivation to find.
+        return getListenerModelByName(codedata, semanticModel, moduleInfo, true, semanticModel);
```
pasindufernando1, line 355: **Low, but a latent trap:** the 3-arg and 4-arg public overloads now assign *opposite* meanings to the same `semanticModel` parameter (the connector's own package model here vs. the current file's model below), distinguished only by the presence of an unrelated `removeDeprecated` boolean.

I checked, and the only 3-arg caller today is `ConnectorModelReader.java:578`, so this is correct as written. But the next caller who reaches for the shorter overload reintroduces exactly the `IllegalStateException: Parent symbol not found` regression described in the PR body.

Suggest renaming it — e.g. `getListenerModelFromConnectorPackage` — so the contract lives in the name rather than a comment.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerModelReader.java
```diff
@@ -390,6 +438,44 @@ public Optional<ServiceInitModel> getSchemaDrivenServiceInitModel(String orgName
...
+                    .getTriggerMetadataModelFromLocalRepository(moduleInfo);
+            if (metadata.isEmpty()) {
+                return Optional.empty();
+            }
+            Optional<Package> pkg = metadataReader.getCompiledPackageFromLocalRepository(moduleInfo);
+            if (pkg.isEmpty()) {
+                return Optional.empty();
+            }
+            return synthesizeTriggerModel(metadata.get(), pkg.get(), moduleName);
+        } catch (Throwable e) {
```
pasindufernando1, line 373: **Low: silent `catch (Throwable)`.**

This is one of four new blanket catches on the local-repository path (also `listLocalRepositoryModules`, `getCompiledPackageFromLocalRepository`, `searchLocalRepository`), none of which log. Every new failure mode becomes invisible, so "my local connector doesn't show up in the search" turns into an undebuggable report.

At minimum, log at debug via the existing `LSClientLogger`.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/model/ServiceInitModel.java
```diff
@@ -49,6 +49,12 @@ public class ServiceInitModel {
     private final String type;
     private final String icon;
     private final Map<String, Value> properties = new LinkedHashMap<>();
+    // Not constructor-set, like `properties` above: whether this connector was resolved via the
+    // Ballerina local repository rather than Central. Set explicitly by the schema-driven resolution
+    // path (ConnectorModelReader) after the model is built -- Gson (which deserializes this class
+    // directly on the addServiceAndListener round-trip) defaults it to false, which is exactly right for
+    // every other builder, so no constructor/Builder plumbing is needed for those.
+    private boolean isLocalRepository;
```
pasindufernando1: **Nit:** a primitive `boolean` forces `"isLocalRepository": false` into the wire payload for **every** connector, which is why 20 test fixtures needed updating in this PR. `Boolean` would let Gson omit it by default and leave those fixtures untouched.

packages/ballerina-visualizer/src/views/BI/ComponentListView/CentralSearchPanel.tsx
```diff
@@ -109,33 +125,61 @@ export function CentralSearchPanel(props: CentralSearchPanelProps) {
     // Central may echo a package already available locally; the local sections already show those.
     const localTriggerIds = new Set(props.triggers.local.map((t) => `${t.orgName}/${t.packageName}`));
     const visibleResults = results.filter((item) => !localTriggerIds.has(`${item.orgName}/${item.packageName}`));
+    // Never deduped against Central/bundled results: the same org/name resolved from the local
+    // repository is a distinct artifact (a different, in-development version) worth showing separately.
+    const visibleLocalRepositoryResults = localRepositoryResults;
```
pasindufernando1: **Nit:** this is a pure alias — `visibleLocalRepositoryResults` is never anything other than `localRepositoryResults`. Suggest inlining it and moving the (useful) comment up to the `useState` declaration.

packages/ballerina-visualizer/src/views/BI/ComponentListView/CentralSearchPanel.tsx
```diff
@@ -70,28 +84,29 @@ export function CentralSearchPanel(props: CentralSearchPanelProps) {
...
+                        setLocalRepositoryResults(res?.localRepositoryResults ?? []);
                     })
                     .finally(() => {
                         if (isMountedRef.current && latestQueryRef.current === searchQuery) {
                             setSearching(false);
                         }
                     });
             }, SEARCH_DEBOUNCE_MS),
-        [rpcClient]
+        [rpcClient, localCentralSearchEnabled]
```
pasindufernando1: **Low: double search on mount.** `localCentralSearchEnabled` starts `false` and flips when the RPC resolves, which recreates `runSearch`, which refires the effect below. So the first search always goes out without local results and is then immediately superseded.

Consider gating the initial search on the setting having loaded (e.g. a `settingLoaded` flag), so the panel issues one request instead of two.

packages/ballerina-language-server/model-generator-commons/build.gradle
```diff
@@ -47,6 +47,8 @@ dependencies {
     implementation "org.testng:testng:${testngVersion}"
 }
 
+rootProject.configureBallerinaDistribution(project)
```
pasindufernando1: **FYI / worth noting in the PR description:** this wires `model-generator-commons`' `test` task to `unpackJballerinaTools` + `buildAndCopyLangserverStdlib` (see `configureBallerinaDistribution` in `ballerina-language-server/build.gradle:203`).

That turns a previously-fast unit-test module into one that unpacks a full jballerina distribution and builds `langserver-stdlib`. It's genuinely required for the new tests to resolve real `.bala` packages, but it's a meaningful new CI cost that's easy to miss — only `langserver-core` and `bal-shell-service` did this before.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReader.java
LakshanWeerasinghe: Cached the `PackageRepository` handle in a lazily-initialized static instead of resolving a fresh temp-dir project on every call. Fixed in b678a82a.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/TriggerSearchUtil.java
LakshanWeerasinghe: Each per-org future now degrades to an empty list on its own failure via `.exceptionally`, and Central calls run on a dedicated executor instead of the shared common pool. Fixed in b678a82a.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/TriggerSearchUtil.java
LakshanWeerasinghe: Merge now interleaves round-robin across orgs instead of appending org-by-org, so ballerinax can no longer be truncated away just because ballerina alone fills the limit. Fixed in b678a82a.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/TriggerSearchUtil.java
LakshanWeerasinghe: Switched to the dedicated `org` query param instead of embedding `org:<name>` in `q`, matching SearchListGenerator/PackageListGenerator's established usage. Updated the test fake to match. Fixed in b678a82a.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/TriggerSearchUtil.java
LakshanWeerasinghe: Checked live against Central: ballerina/smb (keywords: smb, file, network shares) and ballerina/mqtt (keywords: mqtt, client, messaging, network, pubsub, iot) carry no Type/Trigger tag and no trigger/listener/event keyword. So this is a real coverage loss for connectors this PR tests, not just a theoretical one. Leaving it as documented for now rather than reintroducing the hasListeners fallback -- open to revisiting if we want it back.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/LocalDependencyEditUtil.java
LakshanWeerasinghe: A version mismatch now replaces the existing [[dependency]] stanza's version in place instead of silently no-oping. Added a regression test. Fixed in b678a82a.

packages/ballerina-language-server/model-generator-commons/src/test/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReaderTest.java
LakshanWeerasinghe: This is moot now -- the local-repo-writing tests in this class (and the other two) were removed entirely in an earlier commit since the underlying resolution wasn't working yet. Nothing writes to the real repo anymore.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/ListenerUtil.java
LakshanWeerasinghe: Renamed to getListenerModelFromConnectorPackage and updated its call site/javadoc. Fixed in b678a82a.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerModelReader.java
LakshanWeerasinghe: Added FINE-level logging to all four local-repository catch(Throwable) sites you listed. Fixed in b678a82a.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/model/ServiceInitModel.java
LakshanWeerasinghe: Changed to boxed Boolean so Gson omits it when unset, and cleaned up the now-redundant "isLocalRepository": false line from all 20 fixtures. This also fixed 2 stale fixtures that were missing the field entirely. Fixed in b678a82a.

packages/ballerina-visualizer/src/views/BI/ComponentListView/CentralSearchPanel.tsx
LakshanWeerasinghe: Inlined the alias and moved the comment up to the useState declaration. Fixed in b678a82a.

packages/ballerina-visualizer/src/views/BI/ComponentListView/CentralSearchPanel.tsx
LakshanWeerasinghe: Turned out to already be fixed as a side effect of gating the whole panel behind the experimental flag -- the search effect now no-ops until the flag resolves, so only one search fires total. Verified, no further change needed.

packages/ballerina-language-server/model-generator-commons/build.gradle
LakshanWeerasinghe: removed

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/LocalDependencyEditUtil.java
```diff
@@ -0,0 +1,154 @@
...
+        TextEdit tomlEdit = createLocalDependencyEdit(toml.get(), org, name, version);
+        edits.computeIfAbsent(tomlPath, ignored -> new ArrayList<>()).add(tomlEdit);
+    }
+
+    private static Optional<String> declaredVersion(PackageManifest manifest, String org, String name) {
+        if (manifest == null || manifest.dependencies() == null) {
+            return Optional.empty();
+        }
+        for (PackageManifest.Dependency dependency : manifest.dependencies()) {
+            if (org.equals(dependency.org().value()) && name.equals(dependency.name().value())) {
```
pasindufernando1: **Medium: the version-bump fix doesn't consider `repository`, so it can silently rewrite a Central dependency.**

Thanks for the in-place version replacement — that closes the original no-op. But the match here is still `org`/`name` only, and neither branch of `addIfMissing` inspects or writes the `repository` key. Two cases fall through:

- Declared as a **Central** dependency (no `repository` key) at a *different* version → `findVersionValueEdit` rewrites the version in place and leaves it resolving from Central, at a version that likely doesn't exist there.
- Declared at the **same** version without `repository = "local"` → the early return on line 72 fires, so `repository = "local"` is never added and the local connector silently resolves from Central.

The `already_declared` fixture carries `repository = "local"`, which is why the current tests don't surface either case.

Suggest matching on the repository field alongside org/name, so a Central-declared dependency of the same org/name is treated as "not declared locally" rather than as the same entry — and adding a fixture for the Central-declared case.

pasindufernando1 (commented):
Re-reviewed against the fixes — 11 of the 13 earlier findings are fully addressed, and the tests-per-fix discipline makes them verifiable. One gap left in the version-bump fix, inline.

pasindufernando1 (approved):
LGTM. Added a minor comment

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/LocalDependencyEditUtil.java
```diff
@@ -0,0 +1,154 @@
...
+                }
+            }
+            if (org.equals(declaredOrg) && name.equals(declaredName) && versionField != null) {
+                LineRange valueRange = versionField.value().lineRange();
+                Range range = new Range(
+                        new Position(valueRange.startLine().line(), valueRange.startLine().offset()),
+                        new Position(valueRange.endLine().line(), valueRange.endLine().offset()));
+                return Optional.of(new TextEdit(range, "\"" + newVersion + "\""));
+            }
+        }
```
NipunaRanasinghe: This matches a `[[dependency]]` stanza on **org + name only**, ignoring whether it declares `repository = "local"`, and then rewrites just the version value.

Failure case: the project already depends on `ballerinax/foo` from Central at `1.0.0`. The developer pushes `1.1.0` with `bal push --repository=local` and adds it through this UI. `declaredVersion` finds the existing stanza, sees `1.0.0 != 1.1.0`, and this method rewrites the version to `1.1.0` — but never adds `repository = "local"`. The dependency still resolves against Central, where `1.1.0` doesn't exist, so the build fails with nothing in the flow explaining why.

The test fixture doesn't cover this: `local_dependency/already_declared/Ballerina.toml` already contains `repository = "local"`, so `testExistingDependencyVersionBump` only exercises local -> local.

Two options depending on intent: include `repository` in the stanza match so a Central-declared dependency falls through to `createLocalDependencyEdit` (adding a correct new stanza), or keep matching on org/name and emit a second edit adding `repository = "local"` when the key is absent. Either way a fixture without the key would pin the behavior.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReader.java
```diff
@@ -130,12 +209,7 @@ private Optional<TriggerUISchemaModel> readTriggerUISchemaModel(Path packageRoot
         });
     }
 
-    /**
-     * The local {@code .bala} root of {@code moduleInfo}. Only a hit is memoized: a miss means the
-     * package is not in the local repository <i>yet</i>, and caching that would hide it for the rest of
-     * the session once the user pulls it (the pull itself is left to the LS's explicit, user-notified
-     * flow -- see {@link PackageUtil#getModulePackageOffline}).
-     */
+    /** The local {@code .bala} root of {@code moduleInfo}. Only a hit is memoized. */
```
NipunaRanasinghe, line 215: This cache has no expiry and no invalidation, so an entry lives for the whole language-server process. `resolvePackageRoot` resolves whichever version is currently available and the result is pinned under `org/moduleName`, which means after a `bal push --repository=local` of a **new version** the cache keeps handing back the previous version's `.bala` root and the UI reads stale trigger metadata. Recovering needs a language-server restart.

The "Only a hit is memoized" design is deliberate and does handle one half of this: a connector that didn't exist yet is picked up, because misses aren't cached. It's the update case that goes stale — and that's the loop the PR description targets ("one pushed locally via `bal push --repository=local` during connector development", and `LocalDependencyEditUtil` explicitly handling "bumps the version in place if the connector is already declared at a different version"). So version changes are the expected, routine case here.

`maximumSize(2)` won't rescue it either — someone iterating on one or two connectors will never evict.

A short `expireAfterWrite` (seconds) would keep the resolution cost down while letting a repush be seen without a restart. Keying on `org/moduleName/version` wouldn't help on its own, since `resolvePackageRoot` doesn't resolve by version.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/LocalDependencyEditUtil.java
```diff
@@ -0,0 +1,154 @@
...
+    private static int getDependencyStartLine(BallerinaToml toml) {
+        DocumentNode tomlSyntaxTree = toml.tomlDocument().syntaxTree().rootNode();
+        return tomlSyntaxTree.members().stream()
+                .filter(member -> member.kind().equals(SyntaxKind.TABLE)
+                        && TomlSyntaxTreeUtil.toQualifiedName(((TableNode) member).identifier().value())
+                                .equals("package"))
+                .findFirst()
+                .map(member -> member.lineRange().endLine().line() + 2)
+                .orElse(0);
+    }
```
NipunaRanasinghe, line 161: `endLine().line() + 2` assumes a blank line follows the `[package]` table. The `no_dependency` fixture is exactly the case where that doesn't hold:

```toml
[package]          <- line 0
org = "wso2"
name = "no_dependency"
version = "0.1.0"  <- line 3, last line
```

`[package]` ends at line 3, so this returns 5 and `createLocalDependencyEdit` builds `Position(5, 0)` for a document whose last valid line is 3. `testNoDependencyDeclared` only asserts `edits.size() == 1`, never the range, so it passes regardless. VS Code will most likely clamp to the document end — which happens to be what's wanted — but that's relying on client leniency for correctness.

Clamping explicitly to the document's line count would make the intent hold without depending on the client.

Separately, `.orElse(0)` inserts the `[[dependency]]` block at line 0 when no `[package]` table is found, putting it above `[package]` in the file. Valid TOML, just unusual output for a generated edit.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/TriggerSearchUtil.java
```diff
@@ -19,78 +19,144 @@
...
+            int effectiveLimit = limit == null || limit <= 0 ? DEFAULT_LIMIT : limit;
+            int effectiveOffset = offset == null || offset < 0 ? 0 : offset;
+
+            List<CompletableFuture<List<TriggerBasicInfo>>> futures = ALLOWED_ORGS.stream()
+                    .map(org -> CompletableFuture.supplyAsync(() -> {
+                        Map<String, String> queryMap = new HashMap<>();
+                        queryMap.put("q", effectiveQuery);
+                        queryMap.put("org", org);
+                        queryMap.put("limit", String.valueOf(effectiveLimit));
+                        queryMap.put("offset", String.valueOf(effectiveOffset));
```
NipunaRanasinghe: With the per-org fan-out, the same `effectiveOffset` is sent to **every** org query, and the merged results are then truncated to `effectiveLimit`. Offset is therefore per-org rather than global, and pagination no longer composes: with 2 orgs and `limit=30`, page 1 shows roughly 15 from each org, but page 2 (`offset=30`) starts at each org's item 30 — so items 15-29 from each org can never be reached.

Not live today: the only production caller (`ServiceModelGeneratorService.searchTriggers`) passes `null, null` for limit and offset, so offset is always 0 and limit always `DEFAULT_LIMIT`. Only tests pass a limit.

Since `limit`/`offset` remain in the signature, a future caller would reasonably assume global paging semantics and get silently skipped results. Either compute per-org offsets from a global cursor, or drop the parameters until real paging is needed.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/LocalDependencyEditUtil.java
LakshanWeerasinghe: Matching now also requires `repository = "local"` in both `declaredVersion()` and `findVersionValueEdit()`, so a Central-declared dependency of the same org/name is treated as not-locally-declared and gets a new local `[[dependency]]` stanza added rather than having its version rewritten in place. Added a `central_declared` fixture covering both the same-version and different-version cases. Fixed in a0e8e665.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/LocalDependencyEditUtil.java
LakshanWeerasinghe: Same underlying gap as the thread above — the version-rewrite match on `org`/`name` alone. Fixed the same way: both `declaredVersion()` and `findVersionValueEdit()` now also require `repository = "local"`, so a Central-declared dependency falls through to `createLocalDependencyEdit` instead of having its version rewritten in place. Added a `central_declared` fixture (no `repository` key) with tests covering both the same-version and different-version cases. Fixed in a0e8e665.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReader.java
LakshanWeerasinghe: Added an `expireAfterWrite` TTL to `packageRootCache` so a repushed local version is picked up without an LS restart, while `maximumSize(2)` still keeps the miss-path cost down for the common one/two-connector iteration loop. Fixed in a0e8e665.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/LocalDependencyEditUtil.java
LakshanWeerasinghe: `getDependencyStartLine` now clamps the candidate line to the TOML document's own `lineRange().endLine()` instead of assuming a blank line follows `[package]`, and falls back to end-of-document (rather than line 0) when no `[package]` table is found. Added `testDependencyStartLineIsClampedToDocumentBounds` against the `no_dependency` fixture (no trailing blank line) to pin it. Fixed in a0e8e665.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/TriggerSearchUtil.java
LakshanWeerasinghe: Dropped the `offset` parameter from `searchCentral` — the only production caller (`ServiceModelGeneratorService.searchTriggers`) always passed `null`, and computing a real per-org offset from a global merged cursor isn't needed yet. `limit` stays, since each per-org call already fetches up to the full limit and the merge/truncation step composes correctly. Fixed in a0e8e665.

NipunaRanasinghe (approved):
LGTM

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/TriggerSearchUtil.java
```diff
@@ -19,78 +19,146 @@
...
-        return toTriggerResults(response, existingKeys, pkg -> false);
+    /**
+     * Searches the Ballerina local repository for packages shipping trigger metadata/schema files.
+     */
+    public static List<TriggerBasicInfo> searchLocalRepository(Set<String> existingKeys) {
+        try {
+            LibraryMetadataReader reader = LibraryMetadataReader.getInstance();
+            Set<String> known = existingKeys == null ? Set.of() : existingKeys;
+            List<TriggerBasicInfo> results = new ArrayList<>();
+            for (ModuleInfo moduleInfo : reader.listLocalRepositoryModules()) {
```
NipunaRanasinghe, line 124: Unlike `searchCentral` (which keeps a `seen` set keyed by `org/name`), this loop doesn't dedupe across the module list it gets from `listLocalRepositoryModules()`. Since that lists one `ModuleInfo` per `org/name/version` triple, a connector with several versions pushed to the local repo (`bal push --repository=local` bumped a few times during the exact "edit → repush → search" loop this PR targets) will surface as multiple cards here.

That compounds on the frontend: `CentralSearchPanel.tsx`'s local-repo card list keys each item by `` `local/${item.orgName}/${item.packageName}` `` — no version — so duplicate versions of the same connector collide on key.

Worth either deduping to the latest version per `org/name` here, or including version in both the result and the React key if showing every version is intentional.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReader.java
```diff
@@ -96,6 +107,77 @@ public Optional<TriggerMetadataModel> getPackagedTriggerMetadataModel(ModuleInfo
...
+    private Optional<Path> localPackageRoot(ModuleInfo moduleInfo) {
+        return getCompiledPackageFromLocalRepository(moduleInfo).map(pkg -> pkg.project().sourceRoot());
+    }
+
+    /** The Ballerina local repository handle, resolved once and cached. */
+    private PackageRepository localRepository() {
+        return LocalRepositoryHolder.INSTANCE;
+    }
+
+    private static final class LocalRepositoryHolder {
```
NipunaRanasinghe, line 176: This handle is resolved once and cached for the process lifetime (deliberately, per the PR description). What isn't clear from the diff or the added tests is whether `PackageRepository#getPackages()`/`#getPackage()` re-scan `~/.ballerina/repositories/local` live on each call, or whether the repository object itself snapshots the package listing at construction.

If it's the latter, a developer who runs `bal push --repository=local` for a new connector *after* the trigger-search panel has already been opened once in the current LS session won't see it without reloading the window — which would undercut the exact local-connector-development workflow this feature is for. None of the new tests (`LocalDependencyEditUtilTest`, `TriggerSearchUtilTest`) exercise this singleton's live-ness, since they go through `BuildProject.load`/fakes rather than `LibraryMetadataReader.localRepository()`.

Worth confirming this manually (push a new local package mid-session, search again without reloading) before merge, or adding a short-TTL re-resolve like `packageRootCache` got in this same PR if it turns out to be stale.

LakshanWeerasinghe: this is experimantal feature only apply for developer testing so this won't be an issue.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/TriggerSearchUtil.java
LakshanWeerasinghe: This is one is experimental feature only for the use of connector developer will address this in a seperate PR.

NipunaRanasinghe (approved):

PR #768, opened 2026-08-03
Add CodeRabbit Configurations

NipunaRanasinghe:
## Purpose

CodeRabbit already reviews PRs on this repo, but it is configured entirely through the CodeRabbit dashboard. That configuration is not versioned, is invisible to contributors, and carries no knowledge of this repo, so the bot reviews a Rush monorepo that is roughly half Java language server and half React webviews using generic advice.

This adds a root `.coderabbit.yaml` so review rules live next to the code and change through PRs.

It also deletes `packages/ballerina-language-server/.coderabbit.yaml`, inherited when the LS was extracted into this repo. CodeRabbit reads only the config at the repository root, so that file has never had any effect, and its `base_branches: ['main', '1\.\d+\.x']` predates the current release lines. Everything it declared is either carried forward or intentionally dropped, per the inheritance note below.

## Goals

- Review guidance is versioned and reviewable instead of held in dashboard state.
- Instructions are scoped per path, so Java, TypeScript, React webviews, and tests each get rules that apply to them.
- A short list of hard rules is flagged loudly; everything else stays advisory. The bot never gates the merge button.
- The bot stops commenting on generated files.
- Existing dashboard configuration keeps working for everything this file does not speak to.

## Approach

No UI change and no product code change: one new config file, one deletion.

### `inheritance: true` — the one setting worth reviewing carefully

Precedence is: workspace global overrides → org global overrides → **this file** → central YAML → repo UI → org UI → workspace UI → schema defaults.

`inheritance` defaults to `false`, and with it off **only the highest-priority source applies**. That would make this file the entire configuration and stop every existing dashboard setting for this repo from taking effect — including settings in use before this file existed. Since CodeRabbit has been dashboard-configured here for a while, discarding that was not the intent.

With inheritance on, values set here win (repository YAML outranks all the UI levels) and values omitted here still resolve from the dashboard. Arrays merge child-first, so an org-level `path_instructions` entry is appended after ours rather than replacing them.

### The file is deliberately narrow

Because an omitted setting means "keep whatever the dashboard says", a value is stated only where there is a reason: a verified overlap with CI, a decision recorded in a tracked issue, or something the review design depends on. That comes to nine explicit settings:

| Setting | Why it is stated |
| --- | --- |
| `inheritance: true` | The decision above. |
| `profile: chill` | Matches the schema default. Pinned because all seven instruction blocks assume a chill baseline; an org-level switch to `assertive` would quietly change how every one of them is applied. |
| `request_changes_workflow: false` | Matches the schema default. Pinned for the same reason — the rules are written for a bot that comments rather than blocks. |
| `pmd: false` | 2,156 `.java` files, and Gradle already runs checkstyle + spotbugs on every PR in `Build_Stage`. Enabling it reports the same findings twice. |
| `trivy: false` | Already runs in `reusable-build.yml` and `schedule.yml`, governed by `.trivyignore`. |
| `osvScanner: false` | Same dependency-CVE ground as Trivy. |
| `eslint: false` | By decision — see below. |
| `suggested_labels: true` | Matches the schema default, but `labeling_instructions` reaches the author only through the walkthrough. If the dashboard has this off, the label suggestion below produces nothing. |
| `auto_apply_labels: false` | Matches the schema default. Pinned the other way for the same reason: if the dashboard has auto-apply on, adding `labeling_instructions` would start applying a label that triggers a full Playwright run. |

Every other tool is left unlisted on purpose. Tools default to enabled, so anything omitted keeps whatever the dashboard already has, which is preferable to asserting a value we cannot point at a reason for.

`eslint` is off despite being the only lint that would run on a PR — CI runs none. The configs are too uneven to act on: six packages still use tslint, the rest are split across `.eslintrc` and flat config, `data-mapper` and `overview-view` have conflicting configs, and four packages have a `lint` script that cannot run. The rules worth keeping are in `path_instructions` instead. Tracked in wso2-enterprise/integration-engineering#2167, referenced from a `TODO` on the key.

### Review rules

**Posture.** `chill` with a four-item hard-flag list, stated in `tone_instructions` and in the first `path_instructions` block: leaked secrets, a missing WSO2 license header, a broken published API or RPC contract, and undisposed VS Code listeners. The `assertive` profile was considered and rejected, since it raises nitpick volume globally, which is the opposite of a short list.

**The E2E label.** The Playwright suite runs only when a PR carries `Checks/Run Ballerina UI Tests`, or when the base branch ends in `.x` (`.github/workflows/pull-request.yml:59`). Nothing in the PR template says so, so UI changes routinely merge without it. `labeling_instructions` makes CodeRabbit suggest that label on PRs touching React components, webview and side panels, diagram rendering, or the form and expression editors — and explicitly *not* on changes limited to the language server, build config, tests, or docs, so it stays worth reading.

The label string was checked byte-for-byte against the workflow condition and the repo's label list; all three match. Note that listing a label here restricts CodeRabbit to suggesting only from this list, which is intended — this is the one label that changes what CI runs. `Runner/AWS` is referenced by the workflow but does not exist as a repo label, so it is deliberately not mentioned.

**Seven `path_instructions` blocks.** Each rule traces to something already written down in `AGENTS.md`, `CONTRIBUTING.md`, `docs/TEST_GUIDE.md`, or enforced by a hook. The ones carrying the most weight:

- The three-layer RPC contract (`ballerina-core/src/rpc-types` → `ballerina-rpc-client/src/rpc-clients` → `ballerina-extension/src/rpc-managers`) must move together. A descriptor changed without its client and handler compiles fine in both packages, because they build independently, and breaks at runtime.
- Java review is told explicitly **not** to comment on formatting, naming, or import order, because Gradle already gates those with checkstyle and spotbugs on every PR. Without that instruction the bot duplicates CI on 2,156 files.
- Never rename `jest.ls-integration.config.js`, `jest.realdata.config.js`, `jest.headless-view.config.js`, or `jest.perf.config.js` to `jest.config.js` — only the exact name is PR-gated, and those four need Java, a Ballerina distribution, or VS Code.
- Webview review covers the things nothing else catches here: hardcoded colors instead of `ThemeColors` / `var(--vscode-*)`, which break the light, dark, and high-contrast themes; `React.lazy` panels outside `WebviewErrorBoundary`; and host-to-webview notification churn that isn't debounced and gated on `viewActive === "viewReady"`. The same block encourages an E2E test when a new UI flow isn't covered by the existing suites.

**30 `path_filters`**, covering build output, the submodule, and generated content that is tracked anyway — chiefly the ~10k expected-output blobs under the LS `src/test/resources/`.

### Pre-merge check

Org- and workspace-level **global overrides** rank above repository YAML regardless of the inheritance setting. If any are set for this org, nothing in this file applies.

CodeRabbit's own review of this PR reports `Configuration used: Organization UI` and `Review profile: CHILL`. Organization UI is precedence level 6, below repository YAML, which is good evidence that no global overrides are in play — if there were, they would be the reported source. Worth one confirming look at the dashboard before merging, since that is inference rather than proof.

After merge, commenting `@coderabbitai configuration` on any PR prints the effective config with a per-setting source, which is how to confirm these values are attributed to the repository YAML.

## UI Component Development

N/A — no UI change. No new components, no `ui-toolkit` usage, no submodule change.

## User stories

N/A — contributor tooling.

## Release note

N/A — no user-facing change.

## Documentation

N/A — the file documents itself in comments, and no product documentation describes the review bot.

## Training

N/A

## Certification

N/A — no product behavior change.

## Marketing

N/A

## Automation tests

- Unit tests

  N/A — there is no test harness for a CodeRabbit config, and adding one is not warranted for a single declarative file. It was validated manually instead:

  - Schema-validated against `https://coderabbit.ai/integrations/schema.v2.json` with `ajv`. Note that `reviews` does not set `additionalProperties: false`, so a misspelled key validates clean and is then silently ignored — confirmed with a negative control, where an injected `pathh_filters` key passed validation. Every key was therefore also cross-checked against the schema's property lists, which caught one real bug: `osv-scanner` should be `osvScanner`.
  - `path_filters` tested with `minimatch` over all 16,493 tracked files: 10,453 excluded, 6,040 reviewable, and zero `.ts`/`.tsx`/`.java` source files excluded by accident. This pass found two wrong exclusions before commit — a bare `**/build/**` also matched `.github/actions/build/action.yml`, and a bare `**/lib/**` matched `packages/ballerina-extension/test/lib/index.ts`, so both are now scoped to the package root. It also caught that excluding the LS `src/main/resources/**` wholesale would have hidden the hand-written `META-INF/services` SPI registrations, where a missing entry makes the language server silently fail to load an extension; that exclusion is now scoped to `*.json`.
  - `base_branches` entries checked as anchored regexes against representative branch names.
  - All seven `path_instructions` globs confirmed to match at least one real file, so none is dead config.
  - The `Checks/Run Ballerina UI Tests` string compared byte-for-byte against the workflow condition and the repo's label list — 29 characters, identical in all three. This also showed `Runner/AWS` is read by the workflow but is not a defined repo label, so the config no longer points authors at it.
  - After the final comment pass, the parsed YAML was diffed against the previous revision to confirm the rewording changed no values.

- Integration tests

  Not run, and not possible before merge: the config only takes effect once it is on the default branch. CodeRabbit's review of this PR still uses the current dashboard configuration.

  After merge, the behavioral check is one PR touching a `.tsx` webview file, a `.java` LS file, and an RPC descriptor, confirming that Java comments are about logic rather than formatting, that the RPC instruction fires on a descriptor changed without its client and handler, that the E2E label prompt appears, and that no comments land on filtered paths.

## Security checks

- Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? N/A — declarative configuration, no code.
- Ran FindSecurityBugs plugin and verified report? N/A — no Java change.
- Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? Yes. Leaked secrets are also the first item on the review blocklist this file defines.

## Samples

N/A

## Related PRs

None.

## Migrations (if applicable)

N/A. The deleted LS config was already inert, so removing it changes nothing. `inheritance` is the only setting that affects how existing configuration resolves, and it is covered above.

## Test environment

N/A — no build or runtime change. Validation was done locally on macOS with Node 22 using `ajv-cli`, `js-yaml`, and `minimatch`.

## Learning

CodeRabbit's configuration inheritance is the non-obvious part and is worth reading before editing this file: https://docs.coderabbit.ai/configuration/configuration-inheritance. Two things there are easy to get wrong:

1. Only a root `.coderabbit.yaml` is read. A copy inside a package is silently ignored, which is why the LS file never did anything.
2. A repository YAML does **not** merge with dashboard settings by default. `inheritance` defaults to `false`, and with it off only the highest-priority source applies, so a repo file silently replaces dashboard configuration rather than layering on top of it.

The `@coderabbitai configuration` command (https://docs.coderabbit.ai/guides/commands) prints the effective configuration with a per-setting source, which is the only reliable way to see what actually applies.

## Summary

- Added a root `.coderabbit.yaml` for consistent CodeRabbit reviews across the Rush monorepo.
- Added repository-specific guidance for Java, TypeScript, React webviews, RPC contracts, and tests.
- Configured review behavior, branch handling, labels, inherited dashboard settings, and generated-file exclusions.
- Removed the unused package-level CodeRabbit configuration.
- Validated the configuration against repository paths and workflow requirements.
- No product, UI, or runtime changes are included.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,45 +1,20 @@
 # yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
 #
-# CodeRabbit configuration - https://docs.coderabbit.ai/getting-started/yaml-configuration
+# CodeRabbit config - https://docs.coderabbit.ai/getting-started/yaml-configuration
+# Only the root .coderabbit.yaml is read. A copy inside a package does nothing.
 #
-# Only a .coderabbit.yaml at the repo root is read; a copy inside a package has
-# no effect.
+# Precedence, high to low: workspace and org global overrides, THIS FILE, central
+# YAML, repo UI, org UI, workspace UI, schema defaults. Global overrides beat this
+# file, so if any are set, nothing here applies.
 #
-# Precedence, highest to lowest: workspace global overrides, org global
-# overrides, THIS FILE, central YAML, repository UI settings, org UI settings,
-# workspace UI settings, schema defaults.
+# `inheritance` is on, so anything left unset keeps resolving from the dashboard
+# instead of snapping back to a schema default. CodeRabbit ran on dashboard config
+# here for a while and that was worth keeping. So only set a value when there is a
+# reason: CI already covers it, a tracked issue says so, or the rules below depend
+# on it. Arrays merge child-first, so an org-level path_instructions entry lands
+# after ours rather than replacing it.
 #
-# `inheritance` is enabled below. It defaults to false, and leaving it off would
-# mean only the highest-priority source applies -- this file would become the
-# entire configuration and every dashboard setting for this repo would stop
-# taking effect, including the ones already in use before this file existed.
-# CodeRabbit was configured through the dashboard for a while, so discarding
-# that was not the intent. With inheritance on:
-#
-#   - Values set here win, since repository YAML outranks all the UI levels.
-#   - Values NOT set here fall through to central YAML, then the repo, org, and
-#     workspace UI settings, then schema defaults -- so existing dashboard
-#     configuration keeps working for everything this file is silent about.
-#   - Because of that, this file deliberately stays narrow. A setting is stated
-#     only where there is a concrete reason for it: a verified overlap with CI, a
-#     decision recorded in a tracked issue, or something the review design here
-#     depends on. Everything else is left out so it keeps resolving from the
-#     dashboard rather than being silently reset to a schema default.
-#   - Arrays merge as "child items first, then unique parent items appended", so
-#     an org-level path_instructions entry would be appended after ours rather
-#     than replaced.
-#   - Org- or workspace-level *global overrides* outrank this file either way. If
-#     any are set, settings here are silently ignored. Run
-#     `@coderabbitai configuration` on a PR to see the effective config with its
-#     per-setting source, and confirm these values are attributed to the
-#     repository YAML.
-#
-# Baseline is `chill`. The blocklist in the first path_instructions block is the
-# only thing meant to be raised loudly. request_changes_workflow is off, so the
-# bot never gates the merge button.
-
-# Fill values this file leaves undefined from the dashboard rather than from
-# schema defaults. See the precedence note above.
+# To see what actually applies, comment `@coderabbitai configuration` on a PR.
 inheritance: true
 
 tone_instructions: >-
@@ -48,11 +23,9 @@ tone_instructions: >-
   VS Code listeners. Everything else is a suggestion.
 
 reviews:
-  # These two match the schema defaults, but they are stated because the whole
-  # design depends on them: a chill baseline with a short hard-flag list, and a
-  # bot that comments rather than gating the merge button. If the dashboard ever
-  # moves to `assertive` or enables the request-changes workflow, omitting these
-  # would silently change how every path instruction below is applied.
+  # Both match the schema default, but every instruction below assumes a chill
+  # baseline that comments rather than blocks. Pinned so an org-level switch to
+  # `assertive` cannot quietly change how all of it is applied.
   profile: chill
   request_changes_workflow: false
 
@@ -66,13 +39,12 @@ reviews:
       - 'staging/.*'
 
   path_filters:
-    # --- Build output. Gitignored, listed anyway so a stray commit is filtered.
+    # Build output. Gitignored already; listed so a stray commit is still skipped.
     - '!**/node_modules/**'
     - '!**/dist/**'
-    # 'build' and 'lib' are scoped to the package root on purpose. A bare
-    # '**/build/**' also swallows .github/actions/build/action.yml, and a bare
-    # '**/lib/**' swallows packages/ballerina-extension/test/lib/index.ts --
-    # both hand-written.
+    # Scoped to the package root: a bare '**/build/**' also catches
+    # .github/actions/build/action.yml, and '**/lib/**' catches
+    # packages/ballerina-extension/test/lib/index.ts.
     - '!packages/*/build/**'
     - '!packages/*/build-app/**'
     - '!packages/*/lib/**'
@@ -86,20 +58,18 @@ reviews:
     # Build-time copy of packages/ballerina-grammar/syntaxes/.
     - '!packages/ballerina-extension/grammar/ballerina-grammar/**'
 
-    # --- Not this repo's code. Separate git repo; git tracks it as a single
-    # gitlink entry with no trailing path, so both forms are needed.
+    # Separate repo. Git tracks it as one gitlink with no trailing path, so both
+    # forms are needed.
     - '!submodules'
     - '!submodules/**'
 
-    # --- Machine-generated content that IS tracked.
-    # ~10k expected-output blobs for the language server. Reviewing them is noise.
+    # Generated content that is tracked anyway.
+    # ~10k expected-output blobs for the language server.
     - '!packages/ballerina-language-server/**/src/test/resources/**'
-    # main/resources is NOT excluded wholesale. It holds large generated data
-    # blobs (moduleInfo.json 452K, node_templates.json 316K, search_list.json,
-    # the trigger definitions) alongside hand-written files that matter --
-    # META-INF/services SPI registrations in particular. Miss one of those and
-    # the language server silently fails to load the extension. Exclude the
-    # data, keep the rest reviewable.
+    # Only the JSON under main/resources, not the whole tree. It mixes big
+    # generated blobs (moduleInfo.json is 452K) with hand-written files that
+    # matter -- miss a META-INF/services entry and the LS silently fails to load
+    # an extension.
     - '!packages/ballerina-language-server/**/src/main/resources/**/*.json'
     - '!packages/ballerina-extension/test/data/**'
     - '!packages/ballerina-extension/ui-test/data/**'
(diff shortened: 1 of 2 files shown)


.coderabbit.yaml
```diff
@@ -0,0 +1,328 @@
...
+
+reviews:
+  # Both match the schema default, but every instruction below assumes a chill
+  # baseline that comments rather than blocks. Pinned so an org-level switch to
+  # `assertive` cannot quietly change how all of it is applied.
+  profile: chill
+  request_changes_workflow: false
+  # Same reason: a dashboard `fail_commit_status` would let the bot gate merges
+  # through the check instead of the review.
+  commit_status: false
```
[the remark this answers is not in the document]
NipunaRanasinghe: Fixed in 0fb33a8be1, though the reasoning here is off in two places.

Correct: `commit_status` is documented as "only used when `review_progress` is disabled", and `review_progress` defaults to `true`, so `commit_status: false` was dead config. Good catch.

Two corrections:

- `fail_commit_status` defaults to **`false`**, not `true`. The exposure isn't an inherited `true` — it's that `inheritance: true` lets a dashboard value win over any key this file leaves unset. That's still a reason to pin it, just not the stated one.
- There's no "matching `review_progress` configuration" to nest under. Per `schema.v2.json`, `review_progress`, `commit_status`, and `fail_commit_status` are all flat booleans directly under `reviews`.

Since `fail_commit_status` applies to whichever surface is active — check run, or the legacy mirror when `review_progress` is off — it covers both cases on its own, so I replaced `commit_status: false` with `fail_commit_status: false` rather than keeping both.

Worth noting for anyone reading this later: it fires on review *errors*, not on findings. It's a CodeRabbit-side failure turning the check red, not the bot gating merges on what it reports.

.coderabbit.yaml
```diff
@@ -0,0 +1,330 @@
...
+  # which is the intent. `Checks/Run LS Tests` also changes what CI runs but is
+  # left out on purpose: the `ls` path filter turns it on automatically, so a
+  # suggestion adds nothing. This label has no such fallback.
+  labeling_instructions:
+    - label: 'Checks/Run Ballerina UI Tests'
+      instructions: >-
+        Suggest this label when the PR changes user-facing editor behavior:
+        React components under packages/*/src, webview panels, side panels,
+        diagram rendering, form or expression editors, or anything under
+        packages/ballerina-extension/src/views. Do not suggest it for changes
```
gigara: we don't need to run it for every ui change IMO?

NipunaRanasinghe: Fair point — narrowed in b31fb68203.

The old criteria ("React components under `packages/*/src` … or anything under `src/views`") matched **1,477 `.tsx` files across 17 packages**, i.e. effectively every UI change.

The new instruction leads with the cost and makes "no" the default, then points at `e2e-test/e2e-playwright-tests/test.list.ts` for what the suite actually covers rather than restating it here — a list copied into this file would go stale as specs are added or skipped, and a stale list fails silently in the worse direction (no suggestion on a flow that *is* covered).

```yaml
instructions: >-
  A run is four parallel Playwright jobs capped at 60 minutes each, so
  the default answer is no. Suggest this label only when the change
  could break an end-to-end flow the suite actually covers; read
  packages/ballerina-extension/e2e-test/e2e-playwright-tests/test.list.ts
  to see what that is. Do not suggest it for styling, copy, or
  layout-only changes, or for language server, build config, test, or
  documentation changes.
```

.coderabbit.yaml
```diff
@@ -0,0 +1,332 @@
...
+
+        1. SECRETS. Any token, API key, password, connection string, or
+           credential in source or in a test fixture.
+
+        2. LICENSE HEADER. Every new source file needs the WSO2 Apache-2.0
+           header, and the form differs by language. For .ts and .tsx, copy the
+           17-line form from the top of packages/ballerina-core/src/index.ts.
+           For .java, match the 15-line form used by sibling files in the same
+           package -- the Java headers use different wording from the
+           TypeScript one, so do not carry it across. New files use year 2026.
```
gigara: can we use the current year instead of a hard-coded one?

.coderabbit.yaml
```diff
@@ -0,0 +1,332 @@
...
+          submodules/wso2-vscode-extensions alongside packages/. Submodule edits
+          belong in their own commit against their own repo (AGENTS.md,
+          "Submodule changes are not monorepo changes").
+
+    # CVE suppressions. Its own entry because .trivyignore has no extension, so
+    # no `**/*.{ext}` glob above reaches it.
+    - path: '{.trivyignore,common/config/rush/pnpm-config.json}'
+      instructions: |
+        A fixable CVE goes to common/config/rush/pnpm-config.json under
+        globalOverrides, not to .trivyignore. .trivyignore must stay one plain
```
gigara, line 179: there should be a linked issue if we add an entry to trivyignore

.coderabbit.yaml
NipunaRanasinghe: Fixed with https://github.com/ballerina-platform/ballerina-vscode/pull/768/commits/2a15f0d848ab2fc66b4ec4b55ac5e947fbc35474

.coderabbit.yaml
NipunaRanasinghe: Agreed and better to enforce that. Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/768/commits/2a15f0d848ab2fc66b4ec4b55ac5e947fbc35474

gigara (approved):
LGTM

PR #772, opened 2026-08-03
Add "Import From OpenAPI Specification" option to MCP service creation

dan-niles:
## Purpose

Resolves https://github.com/wso2/product-integrator/issues/2462

Adds an **"Import From OpenAPI Specification"** path to MCP service creation, mirroring the existing HTTP OpenAPI import. The user picks an OpenAPI contract, selects which operations to expose, and the language server generates a **proxy** MCP service — one `@mcp:Tool remote function` per selected operation, forwarding to the upstream REST API via an `http:Client` — plus the record types.

https://github.com/user-attachments/assets/1a1e2e36-4646-4740-85a2-68c3ef17b47b

## How it works

1. `mcp_design_approach.json` injects a `designApproach` CHOICE into the MCP init model: *Design From Scratch* vs *Import From OpenAPI Specification* (the latter exposing a `FILE_SELECT` for the spec). This renders through the generic CHOICE editor — no bespoke form code.
2. On submit with a spec selected, the visualizer calls the new `serviceDesign/listOpenApiEndpoints` RPC. The LS parses the contract with `mcp-core` and returns the candidate operations plus CLI-equivalent defaults (service name, version, base path, port, listener name).
3. The client shows a selection step — search, HTTP-method filter chips, select-all, and an "Advanced Configurations" section for overriding the generated defaults. All operations are selected by default.
4. Confirming sends `ServiceInitModel.selectedTools` through the existing `createServiceAndListener` request.
5. `McpOpenApiSchemaDrivenServiceBuilder` flattens the enabled choice and, when a spec is present, delegates to `McpOpenApiServiceGenerator`; otherwise normal schema-driven MCP creation proceeds. The generator filters the parsed endpoints to the selected tool names, generates the service source and types, and returns LSP `TextEdit`s applied to the current project.

`ServiceBuilderRouter` selects the MCP-specific builder only for MCP; every other schema-driven connector keeps using `SchemaDrivenServiceBuilder`.

## Changes

**Language server** (`service-model-generator-ls-extension`)
- `core/McpOpenApiServiceGenerator.java` — parses the spec, filters endpoints to the selection, reconciles form values over generated ones, emits service + types files as `TextEdit`s. `mcp-core` writes to stdout, which would corrupt the LS's JSON-RPC stream, so all calls into it are wrapped in a `runSilently(...)` redirect.
- `serviceDesign/listOpenApiEndpoints` endpoint, with `McpToolEndpoint`, `McpServiceDefaults`, `OpenApiEndpointsRequest`/`Response` DTOs.
- `ServiceInitModel.selectedTools`.
- Depends on `io.ballerina:mcp-core` (GitHub Packages) instead of a vendored jar.

**Frontend** (`ballerina-core`, `ballerina-rpc-client`, `ballerina-extension`, `ballerina-visualizer`)
- RPC plumbing for `listOpenApiEndpoints` mirroring `getServiceInitModel`.
- `ServiceCreationView` gains the operation-selection step, gated to `moduleName === "mcp"` so the HTTP OpenAPI import (which also has a `designApproach` CHOICE) is not hijacked.
- Layout pass so both creation steps share one content inset, the selection step's body text is 13px, its action buttons are right-aligned like every other form, and method pills are uppercase and fixed-width so paths align.

## Summary

- Added an **Import From OpenAPI Specification** workflow for MCP service creation.
- Added RPC support to discover OpenAPI endpoints and service defaults.
- Added endpoint search, HTTP-method filters, bulk selection, and selected-tool handling.
- Added MCP-specific generation for `main.bal`, `types.bal`, and `@mcp:Tool remote function` tools.
- Added request forwarding through `http:Client`.
- Improved generation reliability through name handling, base-path processing, import deduplication, and error reporting.
- Updated `mcp-core` dependency handling.
- Preserved existing behavior for non-MCP schema-driven services.
- Added tests for endpoint selection, base paths, identifier handling, import deduplication, and generated type-name conflicts.


packages/ballerina-extension/src/rpc-managers/service-designer/rpc-handler.ts
```diff
@@ -95,5 +97,6 @@ export function registerServiceDesignerRpcHandlers(messenger: Messenger) {
     messenger.onRequest(createServiceAndListener, (args: ServiceInitSourceRequest) => rpcManger.createServiceAndListener(args));
     messenger.onRequest(generateExamplePayloadJson, (args: PayloadContext) => rpcManger.generateExamplePayloadJson(args));
     messenger.onRequest(getOASSpec, (args: GetOASSpecRequest) => rpcManger.getOASSpec(args));
+    messenger.onRequest(listOpenApiEndpoints, (args: OpenApiEndpointsRequest) => rpcManger.listOpenApiEndpoints(args));
```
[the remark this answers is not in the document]
dan-niles: Verified: `registerServiceDesignerRpcHandlers` is only ever called once (from `RPCLayer.init()`'s single call site in `extension.ts` activation), so there's no re-registration path and no duplicate-handler risk. None of the other 26 `.onRequest` registrations in this file dispose their handler either — adding disposal to just this one would be inconsistent with the file's own pattern. Not fixing.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/McpOpenApiSchemaDrivenServiceBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+            if (stream == null) {
+                return model;
+            }
+            Value designApproach = new Gson().fromJson(new JsonReader(
+                    new InputStreamReader(stream, StandardCharsets.UTF_8)), Value.class);
+            Map<String, Value> properties = new LinkedHashMap<>(model.getProperties());
+            model.getProperties().clear();
+            model.addProperty(PROPERTY_DESIGN_APPROACH, designApproach);
+            model.addProperties(properties);
+        } catch (IOException ignored) {
```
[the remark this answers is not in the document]
dan-niles: This resource is bundled in the extension's own JAR, not user input — an IOException here means a corrupted build artifact, not a real-world scenario. Checked the established pattern for this exact case (`Utils.getServiceByServiceType`, `HttpServiceBuilder.getModelTemplate`/`getServiceInitModel`): all of them silently swallow `IOException` on bundled resource reads with no logging. This matches that convention. Not fixing.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/ServiceModelGeneratorService.java
```diff
@@ -1285,6 +1290,24 @@ public CompletableFuture<CommonSourceResponse> addServiceAndListener(ServiceInit
...
+                        .map(endpoint -> new McpToolEndpoint(endpoint.getToolName(), endpoint.getMethod(),
+                                endpoint.getPath(), endpoint.getDescription()))
+                        .toList();
+                return new OpenApiEndpointsResponse(endpoints, McpOpenApiServiceGenerator.defaultsFor(specInfo));
+            } catch (Throwable error) {
+                return new OpenApiEndpointsResponse(error);
+            }
+        });
+    }
+
```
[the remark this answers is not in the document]
dan-niles: No other @JsonRequest method in this class (30+) has L3 RPC-contract test coverage — this would be a new testing tier not used anywhere else in the module, disproportionate to one endpoint. If coverage is wanted, an L1 fixture test directly on McpOpenApiServiceGenerator would match this module's actual convention better. Not fixing as proposed.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/ServiceCreationView.tsx
```diff
@@ -394,28 +697,166 @@ export function ServiceCreationView(props: ServiceCreationViewProps) {
...
+                                        <Codicon name="file-code" sx={{ display: "flex", alignItems: "center", justifyContent: "center", cursor: "default" }} iconSx={{ fontSize: 13 }} />
+                                        {specFileName}
+                                    </SpecFileBadge>
+                                    {loadingEndpoints ? (
+                                        <RelativeLoader message="Reading OpenAPI specification..." />
+                                    ) : endpointError ? (
+                                        <StatusCard>
+                                            <Icon name="bi-error" sx={{ color: ThemeColors.ERROR, fontSize: "18px" }} />
+                                            <StatusText variant="body2">{endpointError}</StatusText>
+                                        </StatusCard>
```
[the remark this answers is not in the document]
dan-niles: Double-checked this rather than trusting the auto-resolve tag — the gap was still there, just moved into McpOpenApiImportWizard.tsx during the extraction. Fixed in bd2143e: added a Back button wired to the wizard's existing onBack handler, matching the pattern already used on the success path.

packages/ballerina-language-server/gradle.properties
```diff
@@ -19,6 +19,7 @@ sqliteJdbcVersion=3.41.2.2
 jakartaPersistenceVersion=3.1.0
 swaggerParserVersion=2.1.22
 openAPICoreVersion=2.3.1
+mcpCoreVersion=1.0.0-20260915-153900-f81661b
```
NipunaRanasinghe: **Blocker for a release branch.** `1.0.0-20260915-153900-f81661b` is a timestamped dev build, and the base here is `staging/5.14.0`. Can we get a released `mcp-core` version cut before this merges? Right now the shipped extension's MCP-import path is pinned to a specific CI artifact from Sep 15.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/build.gradle
```diff
@@ -55,6 +55,13 @@ dependencies {
         transitive = false
     }
     implementation "io.ballerina.openapi:core:${openAPICoreVersion}"
+    implementation "io.ballerina:mcp-core:${mcpCoreVersion}"
+    dist "io.ballerina:mcp-core:${mcpCoreVersion}"
+    // mcp-core publishes without a POM (artifact jar only), so its own runtime deps aren't pulled in transitively.
```
NipunaRanasinghe, line 60: This comment names the real problem, and I don't think we should work around it here. Hand-mirroring `jackson-databind` and `jackson-dataformat-yaml` means the next dependency `mcp-core` picks up becomes a runtime `NoClassDefFoundError` in the LS with no build-time signal at all — and the failure will surface as "MCP OpenAPI import is broken" long after the version bump that caused it.

Worth fixing on the publishing side (`mcp-core` should publish a POM) rather than absorbing it as a permanent maintenance cost in this build file.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/module-info.java
```diff
@@ -22,6 +22,7 @@
     requires org.eclipse.lsp4j;
     requires io.ballerina.diagram.util;
     requires io.ballerina.openapi.core;
+    requires mcp.core;
```
NipunaRanasinghe, line 25: Follows from the POM-less publish: with no `Automatic-Module-Name` manifest entry, `mcp.core` is derived from the jar *filename*. A rename on the publishing side breaks this build with a fairly opaque JPMS error. Another argument for getting `mcp-core` published properly (it can set `Automatic-Module-Name` even before it has a `module-info`).

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
```diff
@@ -0,0 +1,272 @@
...
+        String path = title.toLowerCase(Locale.ROOT).replaceAll("[^a-z0-9]+", "_")
+                .replaceAll("^_|_$", "");
+        return path.isBlank() ? "mcp" : path;
+    }
+
+    // Serializes access since System.out/err are JVM-global, not per-call.
+    private static final Object STDIO_REDIRECT_LOCK = new Object();
+
+    static <T> T runSilently(SilentAction<T> action) throws McpGenerationException, IOException {
+        synchronized (STDIO_REDIRECT_LOCK) {
```
NipunaRanasinghe: **Blocker — I think this can be deleted outright.** The stated justification (the PR description and the comment on the `runSilently` call sites: "`mcp-core` writes to stdout, which would corrupt the LS's JSON-RPC stream") doesn't hold in this codebase. The launcher already does exactly this, permanently, before lsp4j ever starts — `packages/ballerina-language-server/launcher/src/main/java/org/ballerinalang/langserver/launchers/stdio/Main.java:71`:

```java
static void protectJsonRpcStream() {
    System.setProperty("enableOutputStream", "false");
    System.setOut(new PrintStream(System.err, true));
}
```

and its javadoc: *"lsp4j keeps writing to the original stdout stream captured before this method is invoked."* So mcp-core's `System.out` writes already land on stderr and cannot reach the protocol stream.

What this block does instead, as a side effect:

1. Swallows **`System.err`** — which, after the launcher's redirect, is the LS's only diagnostic channel — for *every other thread* for the duration of a spec parse plus a full `McpProjectGenerator` run. Unrelated concurrent requests lose their stack traces.
2. Serializes every mcp-core call in the JVM behind one lock, so two concurrent `listOpenApiEndpoints` calls queue behind each other's temp-project generation.

If the actual goal is suppressing mcp-core's console noise from the extension's output channel, that's a fine goal — but the comment should say so, and it'd be worth weighing against losing everyone else's stderr for the duration.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/McpOpenApiImportWizard.tsx
```diff
@@ -0,0 +1,460 @@
...
+        let isMounted = true;
+        rpcClient.getServiceDesignerRpcClient().listOpenApiEndpoints({ specPath })
+            .then((res) => {
+                if (!isMounted) return;
+                if (res.errorMsg) {
+                    setEndpointError(res.errorMsg);
+                    return;
+                }
+                const modelWithDefaults = res.defaults
+                    ? applyMcpImportConfiguration(initialModel, toMcpImportConfiguration(res.defaults))
```
NipunaRanasinghe: **Reachable bug: this mutates the parent's state object.**

`applyMcpImportConfiguration` assigns straight into `serviceModel.properties.<x>.value` (L200-205), and it's called here with the `initialModel` prop. That prop is the same object `ServiceCreationView` holds as its `model` state — `applyFormValuesToModel` in `serviceInitModelUtils.ts` mutates its input and returns it, so `updatedModel === model` in `handleOnSubmit`. The spec-derived `serviceName` / `version` / `basePath` / `listenTo` / `listenerVarName` therefore get written into the main creation form's model.

Repro:
1. Pick **Import From OpenAPI Specification**, select a spec, Next.
2. Wizard loads and applies the spec defaults.
3. **Back**.
4. Switch to **Design From Scratch**, Create.

The from-scratch service comes out carrying the OpenAPI spec's service name, base path and port.

A structured clone of the model before applying (and in `handleConfirmSelection` at L301, which also assigns `model.selectedTools` directly rather than through `setModel`) would fix both.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
```diff
@@ -0,0 +1,272 @@
...
+            String normalized = basePath.startsWith("/") ? basePath : "/" + basePath;
+            serviceSource = SERVICE_PATH_PATTERN.matcher(serviceSource)
+                    .replaceFirst("$1" + Matcher.quoteReplacement(normalized) + "$2");
+        }
+        ModulePartNode mainModulePart = mainDocument.syntaxTree().rootNode();
+        // Uniquified like refreshListenerName, so a second import doesn't redeclare apiClient.
+        String clientName = Utils.generateVariableIdentifier(semanticModel, mainDocument,
+                mainModulePart.lineRange().endLine(), DEFAULT_CLIENT_NAME);
+        serviceSource = serviceSource.replace(DEFAULT_CLIENT_NAME, clientName)
+                .replace(DEFAULT_LISTENER_NAME, listenerName)
```
NipunaRanasinghe: These are unanchored substring replacements over generated source that embeds arbitrary OpenAPI text — operation descriptions end up in the `@mcp:Tool` annotations. Two ways this bites:

- A spec whose description or summary mentions `apiClient` or `mcpListener` gets silently rewritten inside a string literal.
- If mcp-core ever emits a derived identifier such as `apiClientConfig`, it becomes `apiClient2Config` (the uniquified name is spliced mid-token).

Can the names be passed in through `SpecInfo` / `GeneratorOptions` so mcp-core emits them directly? If not, at minimum these need word-boundary regexes rather than `String.replace`.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
```diff
@@ -0,0 +1,272 @@
...
+                .replace("new (" + DEFAULT_PORT + ")", "new (" + port + ")");
```
NipunaRanasinghe, line 125: The port is configured twice by two different mechanisms: `filteredSpec` is built as `new SpecInfo(fullSpec.getBaseUrl(), port, ...)` on L104, and then this replace patches the emitted text.

Either the generator honors the `SpecInfo` port — in which case this line is dead, and can only do harm by rewriting an unrelated `new (9090)` literal that happens to appear in the generated source — or it ignores it, in which case the two paths can disagree. Worth establishing which and dropping the other.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
```diff
@@ -0,0 +1,272 @@
...
+                defaults.listenerName());
+
+        SpecInfo filteredSpec = new SpecInfo(fullSpec.getBaseUrl(), port, serviceName, version, endpoints);
+        String serviceSource = runSilently(() -> new MainBalGenerator().generate(filteredSpec));
+        String basePath = resolveValue(model, PROPERTY_BASE_PATH, ARG_TYPE_SERVICE_BASE_PATH, null);
+        if (basePath != null && !basePath.isBlank()) {
+            basePath = basePath.trim();
+            String normalized = basePath.startsWith("/") ? basePath : "/" + basePath;
+            serviceSource = SERVICE_PATH_PATTERN.matcher(serviceSource)
+                    .replaceFirst("$1" + Matcher.quoteReplacement(normalized) + "$2");
```
NipunaRanasinghe: This fails silently. `SERVICE_PATH_PATTERN` requires the literal `on mcpListener` with that exact spacing in mcp-core's output; if mcp-core changes the emitted listener name or reformats, `replaceFirst` matches nothing and returns the source unchanged — so the user's base path is dropped with no error anywhere. Given we're pinned to a dev build of mcp-core, that's a realistic future regression.

At minimum, check `matcher.find()` and throw if the anchor is gone, so a mcp-core bump fails loudly instead of quietly ignoring a user-entered base path.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
```diff
@@ -0,0 +1,272 @@
...
+                    }
+                });
+            }
+        }
+    }
+
+    private static List<EndpointInfo> selectedEndpoints(List<EndpointInfo> endpoints, List<String> selectedTools) {
+        if (selectedTools == null || selectedTools.isEmpty()) {
+            return endpoints;
+        }
```
NipunaRanasinghe: Two things here:

1. Empty/null `selectedTools` means "generate every tool". The wizard disables Create at zero selection so it isn't reachable through the UI today, but that's an odd default for the server side to pick — "the user selected nothing" reading all-in seems more likely to be a bug than an intent.
2. Nothing validates that each selected name actually resolved. If the spec is edited between `listOpenApiEndpoints` and create, the unmatched names are dropped silently and the user gets a service with fewer tools than they asked for (or zero, which then generates an empty service with no error).

Failing with a clear message on an unresolvable tool name would be better than either.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/ServiceCreationView.tsx
```diff
@@ -311,74 +381,35 @@ export function ServiceCreationView(props: ServiceCreationViewProps) {
         }
         setServerValidationErrors([]);
 
-        const newArtifact = res.artifacts.find(res => res.isNew && model.moduleName === res.moduleName);
+        const newArtifact = res.artifacts.find((artifact) => artifact.isNew && model.moduleName === artifact.moduleName)
+            || res.artifacts.find((artifact) => artifact.isNew);
```
NipunaRanasinghe: This fallback leaks into every other creation flow. `createService` is shared by all schema-driven service creation, so any connector where the expected module's artifact happens to be missing from the response now navigates to an arbitrary new artifact instead of staying on the form — previously it fell through to the "release the button rather than hanging" path below.

I assume this exists because the MCP import writes `main.bal`/`types.bal` and the artifact moduleName doesn't match. If so, gate it on `isMcpOpenApiImport` (or pass a flag into `createService`) rather than loosening it for everyone.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/ServiceCreationView.tsx
```diff
@@ -311,74 +381,35 @@ export function ServiceCreationView(props: ServiceCreationViewProps) {
...
             setIsSaving(false);
             return;
         }
         // No artifact came back and nothing was rejected — release the button rather than hanging.
         setIsSaving(false);
-    }
+    };
+
+    const enabledDesignApproachProperties = getEnabledDesignApproachProperties(model);
+    const isMcpOpenApiImport = moduleName === "mcp" && Boolean(enabledDesignApproachProperties?.spec);
```
NipunaRanasinghe, line 402: This predicate and the one in `handleOnSubmit` disagree: here it's `Boolean(...spec)` (the field *exists*, i.e. the import choice is selected), there it's `if (moduleName === "mcp" && specPath)` (the field *has a value*).

Consequence today: the submit button reads "Next" the moment the import choice is picked, before any spec is chosen. It's currently held together by `spec` being `optional: false` so required-validation blocks the submit — but if that ever loosens, pressing a button labelled "Next" creates a plain MCP service and navigates away. Worth deriving both from one value.

packages/ballerina-side-panel/src/components/editors/FileSelect.tsx
```diff
@@ -64,7 +65,7 @@ export function FileSelect(props: DropdownEditorProps) {
                     btnText="Select File"
                     selectedFile={value}
                     required={!field.optional}
-                    errorMsg={error?.message}
+                    errorMsg={isSubmitted ? error?.message : undefined}
```
NipunaRanasinghe, line 68: This changes error display for **every** `FILE_SELECT` field in the product — including the existing HTTP OpenAPI import — to fix a cosmetic issue in this one new flow. It may well be the right behavior everywhere (don't show a required-error before the first submit), but it isn't mentioned in the PR description and nothing else in the diff tests or exercises the other call sites.

Either call it out explicitly as an intentional product-wide change, or scope it to this flow.

packages/ballerina-side-panel/src/components/Form/index.tsx
```diff
@@ -1673,7 +1673,7 @@ export const Form = forwardRef((props: FormProps, _ref) => {
                 <Provider {...contextValue}>
                     <S.Container nestedForm={nestedForm} compact={compact} footerActionButton={footerActionButton} className="side-panel-body">
                         {isFormLoading && (
-                            <div style={{ display: "flex", flexDirection: "column", justifyContent: "center", alignItems: "center", gap: "12px", flex: 1 }}>
+                            <div style={{ display: "flex", flexDirection: "column", justifyContent: "center", alignItems: "center", gap: "12px", flex: 1, minHeight: "300px" }}>
```
NipunaRanasinghe, line 1676: Same concern as the `FileSelect` change: a hard 300px floor on the shared form loading state affects every form in the product, for a layout issue in the new wizard. Fine if intended globally — just worth saying so, since the PR description frames the layout work as local to the two creation steps.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/McpOpenApiSchemaDrivenServiceBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+        try (InputStream stream = getClass().getClassLoader().getResourceAsStream(DESIGN_APPROACH_RESOURCE)) {
+            if (stream == null) {
+                return model;
+            }
+            Value designApproach = new Gson().fromJson(new JsonReader(
+                    new InputStreamReader(stream, StandardCharsets.UTF_8)), Value.class);
+            Map<String, Value> properties = new LinkedHashMap<>(model.getProperties());
+            model.getProperties().clear();
+            model.addProperty(PROPERTY_DESIGN_APPROACH, designApproach);
+            model.addProperties(properties);
```
NipunaRanasinghe: This copy → `clear()` → re-add dance works only because `ServiceInitModel.getProperties()` hands back the internal `LinkedHashMap` (the field is `private final`, so the getter leaking it mutable is the only reason `clear()` is legal here). That's a fragile thing to depend on from a subclass.

An `addPropertyFirst(key, value)` on `ServiceInitModel` would express the actual intent (prepend `designApproach`) without the temporary-clear, and wouldn't break if the getter is ever wrapped in `Collections.unmodifiableMap`.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/McpOpenApiSchemaDrivenServiceBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+        } catch (IOException ignored) {
+            // The schema-driven creation form remains available if the optional import model is absent.
+        }
```
NipunaRanasinghe, line 65: Two inconsistencies in this error handling:

- A missing/unreadable resource silently drops the entire OpenAPI-import feature — the form still renders, just without the choice, and nothing is logged. That's very hard to diagnose from a user report.
- Malformed JSON in the resource throws `JsonSyntaxException`, which is unchecked and *not* caught here, so it propagates. So the two ways this resource can be broken behave in opposite ways.

Since the resource is bundled in this jar, its absence is a build error rather than an expected condition — I'd rather it fail loudly (or at least log) than degrade silently.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/McpOpenApiSchemaDrivenServiceBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+        Value spec = model.getOpenAPISpec();
+        if (spec == null || spec.getValue() == null || spec.getValue().isBlank()) {
+            return super.addServiceInitSource(context);
+        }
+        try {
+            return new McpOpenApiServiceGenerator(Path.of(spec.getValue()), context.project().sourceRoot())
+                    .generateService(model, context.document(), context.workspaceManager(), context.semanticModel());
+        } catch (McpGenerationException | IOException error) {
+            throw new RuntimeException("Failed to generate MCP service from OpenAPI specification: "
+                    + error.getMessage(), error);
```
NipunaRanasinghe, line 82: Worth confirming what the RPC layer does with a `RuntimeException` out of `addServiceInitSource`. Other failures on this path surface through `CommonSourceResponse`'s error fields; if this one escapes as an lsp4j internal error, the user sees a generic failure (or nothing) rather than "couldn't parse your OpenAPI spec", which is the one message that actually helps here.

The `listOpenApiEndpoints` endpoint already models this properly via `OpenApiEndpointsResponse(Throwable)` — same treatment would be good here.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
```diff
@@ -0,0 +1,272 @@
...
+                paths.sorted(Comparator.reverseOrder()).forEach(path -> {
+                    try {
+                        Files.deleteIfExists(path);
+                    } catch (IOException ignored) {
+                        // Best-effort cleanup of generated temporary files.
+                    }
+                });
+            }
+        }
+    }
```
NipunaRanasinghe, line 248: Three notes on this method:

1. The `finally` block calls `Files.walk(tempDir)` a second time, and that call itself throws `IOException` — so a cleanup failure can replace (mask) the real `McpGenerationException` the caller needs to see. Wrapping the cleanup so it can't throw, or using `Files.walkFileTree` with a deleting visitor, avoids that.
2. `findFirst()` over an unordered `Files.walk` picks a nondeterministic `types.bal` if mcp-core ever emits more than one module.
3. Bigger picture: the spec is now parsed three times per import — once in `listOpenApiEndpoints`, once at L100 here, and then a **full project generation into a temp directory** just to read one file back off disk. If mcp-core can hand back the types source (or the parsed model) directly, that's worth asking for; if not, a short comment noting it's the only available API would save the next reader the same question.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/model/response/OpenApiEndpointsResponse.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Response carrying candidate MCP tools and the generated-service defaults.
+ *
+ * @param endpoints the selectable operations
+ * @param defaults the generated service defaults
+ * @param errorMsg the failure message, if parsing failed
+ * @param stacktrace the failure stack trace, if available
+ */
+public record OpenApiEndpointsResponse(List<McpToolEndpoint> endpoints, McpServiceDefaults defaults,
+                                       String errorMsg, String stacktrace) {
```
NipunaRanasinghe, line 36: Nit: `stacktrace` is populated (`Arrays.toString(error.getStackTrace())`) but never read — the wizard only renders `errorMsg`. If it's for parity with the other response records, fine; otherwise it's dead payload on every failure.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/McpOpenApiImportWizard.tsx
```diff
@@ -0,0 +1,460 @@
...
+            const next = new Set(previous);
+            checked ? next.add(toolName) : next.delete(toolName);
+            return next;
+        });
+    };
+
+    const allSelected = endpoints.length > 0 && selectedTools.size === endpoints.length;
+    const toggleAll = (checked: boolean) => {
+        setSelectedTools(checked ? new Set(endpoints.map((endpoint) => endpoint.toolName)) : new Set());
+    };
```
NipunaRanasinghe: Nit / UX: "Select all" and "Deselect all" act on *all* endpoints while the list below shows only `filteredEndpoints`, and `allSelected` likewise ignores the filter. So with a `GET` filter active, "Deselect all" silently clears the POST/PUT selections the user can't see, and the checkbox reads unchecked even when every visible row is checked.

Acting on the filtered set (and computing `allSelected` from it) is usually what people expect from a select-all next to a filter.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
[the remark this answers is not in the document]
dan-niles: No test coverage exists for this generator yet, agreed. Tracking that as a separate follow-up rather than adding it in this PR.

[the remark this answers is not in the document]
dan-niles: Fixed by matching on word boundaries instead of a plain substring replace, so an operation name like apiClientStatus no longer gets touched.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/McpOpenApiImportWizard.tsx
```diff
@@ -0,0 +1,460 @@
...
+                            <ArtifactForm
+                                fileName={filePath}
+                                targetLineRange={targetLineRange}
+                                fields={formFields.filter((field) => field.key !== "designApproach")}
+                                isSaving={false}
+                                nestedForm={true}
+                                onSubmit={handleConfigSubmit}
+                                onBack={onBack}
+                                cancelText="Back"
+                                serverValidationErrors={[]}
```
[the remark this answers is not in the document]
dan-niles: Fixed. The Configure step now receives the actual serverValidationErrors instead of an empty array.

packages/ballerina-language-server/gradle.properties
dan-niles: Bumped to the released 1.1.0.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/build.gradle
dan-niles: Agreed, the real fix belongs in mcp-core publishing a POM. Leaving the mirrored deps as the necessary workaround until then.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/module-info.java
dan-niles: Same root cause as the jackson comment above. No safe fix on our side until mcp-core publishes a real module name.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
dan-niles: Confirmed via the launcher: System.out is already redirected to System.err at startup using the real captured stream, so this never protected the JSON-RPC channel. Removed the System.err redirect and the incorrect justification, kept only the System.out suppression and the lock (needed for the swap/restore itself).

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/McpOpenApiImportWizard.tsx
dan-niles: Fixed. The wizard now clones the model on mount instead of holding the same reference as ServiceCreationView's state, and stops mutating it directly on confirm.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
dan-niles: The word-boundary fix already applied covers the identifier-collision case. A description that literally contains the word "apiClient" as text is still a residual gap, left as an accepted low-probability risk rather than adding string-literal-aware parsing.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
dan-niles: Decompiled mcp-core to check: it ignores SpecInfo's port entirely and always emits 9090, so this replace is the only thing applying the user's port, not dead code. Added a comment, didn't remove anything.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
dan-niles: Fixed. Throws now if the pattern doesn't match instead of silently returning the source unchanged.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
dan-niles: Fixed the unresolved-name case, it now throws listing the tool names it couldn't match. Left the null/empty-means-all default alone since nothing else calls this today.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/ServiceCreationView.tsx
dan-niles: Fixed. The fallback is now gated on isMcpOpenApiImport instead of applying to every creation flow.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/ServiceCreationView.tsx
dan-niles: Tried unifying these, but checking spec.value instead of existence makes the Configure form flash unfiltered fields before a spec is chosen, since isMcpOpenApiImport also drives which fields render. Leaving as is since it's inert while spec stays required.

packages/ballerina-side-panel/src/components/editors/FileSelect.tsx
dan-niles: This is intentionally product-wide, not scoped to the wizard. Flagging that it's a deliberate call, not an oversight.

packages/ballerina-side-panel/src/components/Form/index.tsx
dan-niles: Same as the FileSelect comment, left as a global change.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/McpOpenApiSchemaDrivenServiceBuilder.java
dan-niles: Fixed. Added addPropertyFirst() to ServiceInitModel and used it here instead.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/McpOpenApiSchemaDrivenServiceBuilder.java
dan-niles: Checked. This matches the identical pattern used in every other resource loader in this module, so leaving it consistent rather than changing just this one.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/McpOpenApiSchemaDrivenServiceBuilder.java
dan-niles: Traced it. CommonSourceResponse(Throwable) already catches this at the RPC layer and carries errorMsg/stacktrace, same shape as OpenApiEndpointsResponse. No change needed.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
dan-niles: Fixed the cleanup masking the real exception. Left the findFirst nondeterminism and the triple-parsing alone, the first is speculative against today's flat output layout, the second needs an mcp-core API change. Added a comment noting why the temp project exists.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/model/response/OpenApiEndpointsResponse.java
dan-niles: Checked. No frontend code reads .stacktrace off any response type in this repo, this one included. It's for parity with the rest of the module's error responses, so leaving it.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/McpOpenApiImportWizard.tsx
dan-niles: Fixed. Both now scope to the filtered list instead of every endpoint.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/McpOpenApiSchemaDrivenServiceBuilder.java
```diff
@@ -0,0 +1,85 @@
...
```
NipunaRanasinghe, line 82: Reopening this one — your reply was right about the LS half, so I traced the client half, and nothing consumes `errorMsg`. Full chain for a generation failure:

1. `selectedEndpoints` throws `McpGenerationException("Selected tool(s) no longer exist…")`
2. this line wraps it in `RuntimeException`
3. `ServiceModelGeneratorService.addServiceAndListener` → `catch (Throwable e) → new CommonSourceResponse(e)` — so `textEdits = Map.of()`, `errorMsg` set, `validationErrors = List.of()` ✅ (this part is exactly as you described)
4. `rpc-manager.ts#createServiceAndListener` — `getBlockingValidationErrors(res.validationErrors)` is empty so it passes the gate, **`res.errorMsg` is never read**, and it goes on to call `updateSourceCode` with an empty edit set
5. `ServiceCreationView#createService` — `hasBlockingValidationErrors` false, no `strictMatch`, no fallback artifact → falls through to the `// No artifact came back and nothing was rejected — release the button rather than hanging.` branch

Net effect: the user presses **Create with N tools**, the button un-spins, and nothing happens. No service, no error, no explanation.

This was low-stakes before — the MCP builder is the first thing on this path that throws domain exceptions as a matter of course. It isn't now: this round added two throw sites, and the base-path anchor check fires on *every* create if a future mcp-core bump changes the emitted service declaration. That's the same silent failure the anchor check was added to eliminate, just moved one layer up. Both new messages are good, diagnostic messages — they're worth getting in front of the user.

Smallest fix: have `createServiceAndListener` in `rpc-manager.ts` reject (or return) when `res.errorMsg` is set, and have `createService` render it in the existing `StatusCard` the way the wizard already renders `endpointError`. (`rpc-manager.ts` isn't in this diff, hence the comment here at the head of the chain.)

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
```diff
@@ -0,0 +1,290 @@
...
+        for (EndpointInfo endpoint : endpoints) {
+            if (selectedTools.contains(endpoint.getToolName())) {
+                selected.add(endpoint);
+                matchedTools.add(endpoint.getToolName());
+            }
+        }
+        if (matchedTools.size() < selectedTools.size()) {
+            List<String> unresolved = selectedTools.stream().filter(tool -> !matchedTools.contains(tool)).toList();
+            throw new McpGenerationException(
+                    "Selected tool(s) no longer exist in the OpenAPI specification: " + String.join(", ", unresolved));
```
NipunaRanasinghe, line 273: Still no tests, and this round is the strongest argument for adding some — everything written here is behaviour that a fixture test would pin and that nothing currently guards.

You told CodeRabbit that "an L1 fixture test directly on `McpOpenApiServiceGenerator` would match this module's actual convention better" than an RPC-contract test. I agree — that's exactly the shape I'd want, and none was added. Three cases, all cheap against a small checked-in spec:

1. A selection naming a tool that isn't in the spec → throws, message lists the missing names (this line).
2. A source that doesn't match `SERVICE_PATH_PATTERN` → throws rather than silently dropping the base path (L112-116).
3. A fixture spec + a partial selection → expected `main.bal` / `types.bal` edits.

(3) is the one that earns its keep long-term: it pins the `\b`-bounded identifier replacement, the `new (9090)` port substitution, and the leading-import dedup in `appendGeneratedSource` — all three of which are string surgery over mcp-core's output, and all three of which break silently on an mcp-core bump. We're pinned to `1.1.0` today with no compile-time coupling to that output format, so a golden-file test is the only thing that would catch the next version doing something slightly different.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/build.gradle
NipunaRanasinghe, line 60: Worth re-confirming this against `1.1.0` specifically rather than carrying the conclusion over from the dev build — a proper release is exactly the point at which the publishing side tends to start emitting a POM. I couldn't check myself (it's not on Maven Central, 404, so it's GitHub Packages only).

If 1.1.0 does publish one, the mirrored jackson deps here and the automatic-module-name fragility in `module-info.java` both go away with it. If it doesn't, no action — the workaround stands and the comment is accurate.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/ServiceCreationView.tsx
```diff
@@ -311,74 +381,37 @@ export function ServiceCreationView(props: ServiceCreationViewProps) {
...
```
NipunaRanasinghe, line 402: Nit, and I accept the reason you gave for keeping the split (checking `spec.value` here would flash the unfiltered fields, since this also drives `visibleFormFields`). That reasoning isn't visible in the code though, and "these two predicates should obviously be one" is exactly the kind of thing someone unifies six months from now and then has to debug the flash.

One line — e.g. `// Existence, not value: this also drives field filtering, so it must be true before a spec is picked.` — would prevent that.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/McpOpenApiSchemaDrivenServiceBuilder.java
dan-niles: Fixed in rpc-manager.ts: createServiceAndListener now checks res.errorMsg and mirrors the exact pattern already used in addFunctionSourceCode in the same file — logs it, shows a VS Code error toast, and returns { artifacts: [], error: res.errorMsg } instead of silently proceeding with an empty edit set.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
dan-niles: Added all three. Cases 1 and 2 are in McpOpenApiServiceGeneratorTest: selectedEndpoints throwing with the unresolved names, and applyBasePath (extracted so it's testable on its own) throwing when the pattern doesn't match. Case 3 is McpOpenApiServiceGeneratorFixtureTest: a checked-in spec plus a main.bal fixture that already declares apiClient/mcpListener, asserting the generated source uniquifies both, keeps only the selected tool, and dedupes imports correctly.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/ServiceCreationView.tsx
dan-niles: Added the comment.

packages/ballerina-core/src/interfaces/extended-lang-client.ts
```diff
@@ -2386,6 +2387,7 @@ export interface BIInterface extends BaseLangClientInterface {
     addFunctionSourceCode: (params: FunctionSourceCodeRequest) => Promise<ResourceSourceCodeResponse>;
     getResourceReturnTypes: (params: ResourceReturnTypesRequest) => Promise<VisibleTypesResponse>;
     getServiceInitModel: (params: ServiceModelRequest) => Promise<ServiceModelInitResponse>;
+    listOpenApiEndpoints: (params: OpenApiEndpointsRequest) => Promise<OpenApiEndpointsResponse>;
```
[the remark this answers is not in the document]
dan-niles: These are internally used RPC calls in the extension.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/ServiceCreationView.tsx
```diff
@@ -311,74 +381,38 @@ export function ServiceCreationView(props: ServiceCreationViewProps) {
         }
         setServerValidationErrors([]);
 
-        const newArtifact = res.artifacts.find(res => res.isNew && model.moduleName === res.moduleName);
+        const strictMatch = res.artifacts.find((artifact) => artifact.isNew && model.moduleName === artifact.moduleName);
+        // Only the MCP OpenAPI import falls back to any new artifact; its edits may not carry the "mcp" moduleName.
+        const newArtifact = strictMatch
+            || (isMcpOpenApiImport ? res.artifacts.find((artifact) => artifact.isNew) : undefined);
```
samithkavishke: **[Medium] Wrong artifact can be opened after a successful MCP-from-OpenAPI creation**

The fallback (used because the generated service's `moduleName` may not be `"mcp"`, so `strictMatch` is expected to miss on this path) takes the *first* `isNew` artifact with no filter on `artifact.type`. The backend generator can emit edits to both `main.bal` (the new service) and `types.bal` (generated types) as separate map entries — both come back `isNew: true`. Nothing guarantees `main.bal` is listed before `types.bal` in the `ArtifactsUpdated` notification; if the order is reversed, `openView()` navigates the user to the generated type instead of the new service right after import.

Suggestion: filter the fallback by `artifact.type === "SERVICE"` instead of taking the first `isNew` entry.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
```diff
@@ -0,0 +1,299 @@
...
+            Path typesPath = projectPath.resolve(TYPES_BAL).toAbsolutePath();
+            Document typesDocument = FileSystemUtils.getDocument(workspaceManager, typesPath);
+            edits.put(typesPath.toString(),
+                    appendGeneratedSource(typesDocument.syntaxTree().rootNode(), typesSource));
+        }
+        return edits;
+    }
+
+    /** Appends {@code generatedSource}, deduping its leading imports against ones already in {@code modulePart}. */
+    private static List<TextEdit> appendGeneratedSource(ModulePartNode modulePart, String generatedSource) {
```
samithkavishke, line 173: **[Medium] Duplicate type declarations can be silently appended to `types.bal`**

`appendGeneratedSource` only dedups leading import statements against what's already in the target file; it never checks whether a generated type/record name already exists in the project's `types.bal`. If a prior MCP-from-OpenAPI import (or a hand-written type) already defines a record with the same name that gets scaffolded again (e.g. a common name like `ErrorResponse`), a second import appends a duplicate declaration with no collision check — producing a compile error in the merged file with no warning to the user. The current fixture test only exercises `main.bal` merging, not a `types.bal` collision case.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
```diff
@@ -0,0 +1,299 @@
...
+        }
+        List<EndpointInfo> selected = new ArrayList<>();
+        Set<String> matchedTools = new HashSet<>();
+        for (EndpointInfo endpoint : endpoints) {
+            if (selectedTools.contains(endpoint.getToolName())) {
+                selected.add(endpoint);
+                matchedTools.add(endpoint.getToolName());
+            }
+        }
+        if (matchedTools.size() < selectedTools.size()) {
```
samithkavishke: **[Medium, unverified — depends on mcp-core] `selectedEndpoints()` can silently include two operations under one requested tool name**

If the OpenAPI contract yields two `EndpointInfo` entries with the same `toolName` (e.g. colliding/sanitized operationIds), selecting that one name emits **both** endpoints — `matchedTools.size() == selectedTools.size()` still holds so no exception fires — and the main.bal generator then emits two functions with the same name, breaking compilation. Whether this is reachable depends on whether `OpenApiSpecParser` already guarantees unique tool names; worth a quick check against mcp-core.

**[Low] False-positive exception on a duplicate-name request list**

Separately, `matchedTools` is a `Set` compared against `selectedTools.size()` (a `List`, not deduped): a request like `["getPets","getPets"]` makes `matchedTools.size() (1) < selectedTools.size() (2)`, throwing "tool no longer exists" even though it does. Low severity since the current UI builds `selectedTools` from a `Set` before sending (`Array.from(selectedTools)`), so normal UI usage can't trigger this — only a malformed/replayed RPC call could.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/McpOpenApiImportWizard.tsx
```diff
@@ -0,0 +1,466 @@
...
+    const { initialModel, specPath, filePath, targetLineRange, recordTypeFields, isSaving, serverValidationErrors,
+        onBack, onCreate } = props;
+    const { rpcClient } = useRpcContext();
+
+    const [step, setStep] = useState<WizardStep>("configure");
+    // Cloned so edits here don't leak into ServiceCreationView's own model state.
+    const [model, setModel] = useState<ServiceInitModel>(() => structuredClone(initialModel));
+    const [formFields, setFormFields] = useState<FormField[]>([]);
+    const [endpoints, setEndpoints] = useState<McpToolEndpoint[]>([]);
+    const [selectedTools, setSelectedTools] = useState<Set<string>>(new Set());
```
samithkavishke, line 236: **[Low] Reimplements an existing reusable tool-selection component**

This file's endpoint list/checkbox rows, `selectedTools` Set, `toggleTool`/`toggleAll`, search+filter, and "Selected X of Y" summary duplicate the exported, reusable `ToolsList` component in `packages/ballerina-visualizer/src/views/BI/AIChatAgent/Mcp/McpToolsSelection.tsx` (commented "Reusable Tools List Component"), with different styling and a subset of features (missing the description read-more / scope-tag handling the existing component already has). Two independent implementations of the same selection UI will drift over time.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/ServiceModelGeneratorService.java
```diff
@@ -1285,6 +1290,24 @@ public CompletableFuture<CommonSourceResponse> addServiceAndListener(ServiceInit
         });
     }
 
+    /** Lists the operations in an OpenAPI contract as candidate MCP tools. */
+    @JsonRequest
+    public CompletableFuture<OpenApiEndpointsResponse> listOpenApiEndpoints(OpenApiEndpointsRequest request) {
```
samithkavishke, line 1295: **[Low] OpenAPI spec parsed twice per creation flow (minor perf + TOCTOU)**

The spec is parsed once here (`listOpenApiEndpoints`, to list endpoints for the wizard) and again in `McpOpenApiServiceGenerator.generateService` when the user clicks Create — the first parse's `SpecInfo` is never passed forward. This doubles parse/I/O cost for large specs, and in principle the two parses could observe a changed file between them (TOCTOU), silently creating a service from a different endpoint set than the one reviewed. Low likelihood, but free to fix by threading the already-parsed spec through, or caching by path+mtime.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/build.gradle
dan-niles: The OpenAPI tool also does not emit a POM file (only a JAR) and its dependencies are also hardcoded in LS currently. So, we will follow the same approach for now.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/ServiceCreationView.tsx
dan-niles: Fixed. Filters the fallback to artifact.type === DIRECTORY_MAP.SERVICE. Generation can emit a new TYPE artifact (types.bal) alongside the service in the same response, so "any new artifact" wasn't safe.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
dan-niles: Fixed, but not by renaming. A rename would need to touch serviceSource too (its function signatures reference the type), and generated @mcp:Tool descriptions often repeat the schema name in prose — a text-level rename can't tell a type reference from a mention in a description apart, so it risked corrupting description text on the exact path it was meant to fix. Detects the collision and fails with a clear message instead of silently appending a duplicate. Added a fixture test that reproduces the collision.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/McpOpenApiServiceGenerator.java
dan-niles: Fixed both. Confirmed the Medium case is reachable: decompiled mcp-core's OpenApiSpecParser and it dedupes some other identifiers but not toolName, so two colliding operationIds really can produce two same-named EndpointInfo. selectedEndpoints now throws a clear error on that instead of silently emitting two same-named functions. Also fixed the Low case by comparing against a deduped Set of the requested names instead of the raw List size. Added unit tests for both.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/McpOpenApiImportWizard.tsx
dan-niles: Agreed this is duplication, but didn't fix it here. ToolsList takes a plain {name, description} shape with no method/path rendering, so reusing it means extending a component also used by the AI-agent tool picker — more risk than I wanted to take in this pass. Left as a follow-up.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/ServiceModelGeneratorService.java
dan-niles: Agreed, left as-is for now. Fixing it means threading the parsed SpecInfo through the client round-trip or adding a path+mtime cache, and the actual cost is low (small spec, narrow window). Can revisit if it turns out to matter in practice.

samithkavishke (commented):
Only the fix for the duplication seems inconsistent.

The setup: Say the OpenAPI spec has two endpoints:
- GET /products → uses a schema named Product
- GET /orders → uses a schema named Order

 The whole point of the tool-selection wizard is "only generate what I picked." The main.bal side (selectedEndpoints()) already honors that — it filters down to just the endpoints the user checked. The types.bal side doesn't apply that same filter before checking for collisions. So you get two halves of one feature disagreeing with each other: one says "scope to selection," the other says "scope to the whole spec." That's not a deliberate fail-safe design, it's a filter that got left out on one path.

The failure is unexplainable from the user's perspective. If the user picked "just the Product endpoint" and gets an error mentioning Order — a name they've never seen, tied to an endpoint they explicitly didn't check — there's no way for them to connect cause and effect. A fail-safe is only useful if the user can act on it; here they can't even tell why it's blocking them, because the diagnostic references something outside what they asked for.

dan-niles:
Re: [review](https://github.com/ballerina-platform/ballerina-vscode/pull/772#pullrequestreview-5248349148) — valid, but the actual fix is narrower than it looks.

The filter isn't "left out" of the types.bal side by accident — it can't be applied there at all: mcp-core has no API to generate types.bal from a subset of `components.schemas`, so every schema in the spec is always written to types.bal regardless of which tools were selected (this is a pre-existing limitation). Scoping the *collision check* to the selection (which I tried first) would make the error message friendlier, but it would also let a real collision on an unselected schema through silently, since that schema's type still gets physically written either way. That's worse than the confusing message.

Instead I reworded the error to say this explicitly: it now states that every schema in the spec is generated regardless of selection, so an unexpected name is no longer unexplainable, just a fact about how types.bal generation currently works. Properly scoping *generation itself* to the selection is worth doing (it would also cut unused-record bloat), but it means pruning declarations out of the generated Ballerina source in an AST-aware way, not just a text filter — bigger than I want to take on in this pass. Happy to do it as a follow-up if wanted.

samithkavishke (approved):
LGTM

NipunaRanasinghe (approved):
LGTM

PR #779, opened 2026-08-04
Sync hotfix-5.12.4 extension fixes

pasindufernando1:
## Purpose

Sync of two fixes already merged into the old extensions repo, so that this repo carries them as well:

- wso2/vscode-extensions#2438 — Remove ICP for ballerina-ext users (wso2/product-integrator#1927)
- wso2/vscode-extensions#2447 — Fix side panel crash for a list element with a comma (wso2/product-integrator#1979)

One commit per upstream PR.

## Goals

Carry both fixes over with the same behaviour as upstream, adapted where this repo's layout or test tooling differs.

### Commit 1 — Remove ICP for ballerina-ext users

ICP ships as part of the WSO2 Integrator installation, but the Ballerina extension activated it unconditionally, so standalone users were shown ICP commands, the ICP status bar item and ICP sections in the overview webviews.

`isICPSupported()` in `packages/ballerina-extension/src/utils/config.ts` is the single source of truth (delegating to `isInWI()`), applied at four layers:

1. **Activation** — `extension.ts` skips `activateICP()`, and `features/bi/activator.ts` skips the per-project `isIcpEnabled` probe, avoiding an LS round trip on every visualizer open.
2. **Commands** — `activateICP()` sets the `ballerina.icpSupported` context key, which the `enablement` clauses of `ballerina.icp.start` / `ballerina.icp.stop` now require. The key is only ever set inside `activateICP()`, so it evaluates as false for standalone users and no teardown path is needed.
3. **RPC** — `ICPServiceRpcManager` guards `addICP`, `disableICP`, `viewInICP`, `isIcpEnabled` and `isICPServerRunning`. Defence in depth: `addICP` writes ICP imports and `Config.toml` entries, so a restored or stale webview must not be able to mutate a project with an unsupported feature.
4. **Webviews** — `isICPSupported` is threaded through `VisualizerMetadata` → `MainPanel` → `PackageOverview` / `WorkspaceOverview`, which hide the ICP panel and its divider and skip the ICP status fetch. `WorkspaceOverview` also suppresses the per-package ICP badge and short-circuits `syncProjectICPStatus`.

The flag is optional (`isICPSupported?: boolean`) and read via `value.metadata?.isICPSupported`, so a context payload without it falls through to hiding ICP.

Behaviour under the Integrator is unchanged.

### Commit 2 — Fix side panel crash for a list element with a comma

The form of a node such as the following could not be opened:

```ballerina
io:println(string `Customer details: ID ${customerIdElement.data()}, Name ${firstNameElement.data()}`);
```

The rest parameter is a `REPEATABLE_LIST`, which `FormArrayEditor` rendered by building a string out of the value array and splitting it back apart. `stringToRawArrayElements()` tracked only a double quoted string and the `[` / `{` nesting; a backtick template was tracked by neither, and the braces of the `${...}` interpolation preceding the comma close before it is reached. The comma was therefore read as an element separator and the single argument was split into two. The editor then held one key per element of the value array and two values, and the length check threw inside an effect, so the side panel crashed instead of showing the form. `io:println(foo(a, b))` failed the same way, since parentheses were not counted either.

The node template returned by the language server is correct, so this is a fix on the editor alone.

- `FormArrayEditor` reads the values directly from the value array, so the keys and the values can no longer disagree. The string path is unchanged, since the mode switcher hands over a raw string.
- `stringToRawArrayElements()` tracks a backtick template and treats the whole span as opaque — a comma within a template is never a separator, not even within an interpolation such as `${f(1, 2)}` — and counts parentheses along with `[` and `{`. Moved to a new `rawValueUtils.ts` that holds no import of its own, and re-exported from `utils.ts` so existing imports keep working. This path is still used by the mode switcher and by `FlowNodeForm`, where a wrong split silently attaches the diagnostics of one element to another.
- The length check no longer throws. The keys only identify the elements of the editor, so they are adjusted to the values and a warning is logged.

## Deviations from the upstream PRs

Both are consequences of this repo's layout, not behaviour changes:

- **`PackageOverview` effect** — upstream also dropped a redundant second `getReadmeContent` call in the same effect. This repo has an `isWorkflowManagementEnabled` call between it and the ICP call; that call is kept, and only the redundant readme fetch is dropped.
- **Side panel tests** — upstream added `test/rawValueUtils.test.mjs`, a `node:test` suite run against the compiled `lib/`, plus a `test` script change in `package.json`. This repo runs the side panel tests through jest from `src/test`, so the cases are ported to `src/test/rawValueUtils.test.ts` and the upstream `package.json` change is dropped (jest discovers the file).

## UI Component Development

- [x] Matches with the native VSCode look and feel — no new components; the ICP change only removes existing sections from the two overview views.
- No new reusable components, so nothing to add to the ui-toolkit.

## Release note

- ICP is no longer offered in the standalone Ballerina extension. It remains available as before when the WSO2 Integrator extension is installed.
- Fixed a side panel crash when opening the form of a node whose list argument contains a comma, such as a string template or a nested call.

## Documentation

N/A — no documented behaviour changes for Integrator users, and the second change is a crash fix.

## Automation tests

- Unit tests — 10 cases ported to `packages/ballerina-side-panel/src/test/rawValueUtils.test.ts`, covering the failing template element, a comma within parentheses, within double quotes and within an interpolation, a template holding a backtick inside an interpolation, plus the inputs that must still split (plain separator, mixed nesting, a template followed by a separator, an empty array, a trailing empty element).
- The ICP change is a set of activation and render guards with no new logic to assert against; the existing ICP behaviour under the Integrator is unchanged.

Verified locally:

- `rush build --to @wso2/ballerina-visualizer` — success (includes `ballerina-core` and `ballerina-side-panel`)
- `npx tsc -p ./ --noEmit` in `packages/ballerina-extension` — clean
- `npx jest` in `packages/ballerina-side-panel` — 24 suites, 138 tests passed, including the pre-existing `editorUtils` and `ArrayEditor` suites that cover these helpers

## Security checks

## Related PRs

- wso2/vscode-extensions#2438 (merged)
- wso2/vscode-extensions#2447 (merged)

## Test environment

macOS, Node.js 22.17.0. Behaviour verified upstream in the extension per the original PRs; this sync was verified through the build, typecheck and unit test runs listed above.

## Summary

- Disabled ICP activation, commands, RPC operations, and UI controls for standalone Ballerina extension users.
- Preserved ICP behavior when the WSO2 Integrator extension is installed.
- Fixed repeatable-list parsing for values that contain commas or nested structures.
- Prevented side panel crashes caused by mismatched array keys and values.
- Added raw-value parsing utilities and Jest tests for the updated behavior.

### DIFF
--- a/packages/ballerina-core/src/state-machine-types.ts
+++ b/packages/ballerina-core/src/state-machine-types.ts
@@ -204,6 +204,7 @@ export interface ApprovalOverlayState {
 export interface VisualizerMetadata {
     haveLS?: boolean;
     isBISupported?: boolean;
+    isICPSupported?: boolean; // ICP is only available alongside the WSO2 Integrator extension
     recordFilePath?: string;
     enableSequenceDiagram?: boolean; // Enable sequence diagram view
     target?: LinePosition;
--- a/packages/ballerina-extension/package.json
+++ b/packages/ballerina-extension/package.json
@@ -813,13 +813,13 @@
                 "command": "ballerina.icp.start",
                 "title": "Start ICP",
                 "category": "Ballerina",
-                "enablement": "isSupportedProject && !ballerina.icpRunning && !devant.editor"
+                "enablement": "isSupportedProject && ballerina.icpSupported && !ballerina.icpRunning && !devant.editor"
             },
             {
                 "command": "ballerina.icp.stop",
                 "title": "Stop ICP",
                 "category": "Ballerina",
-                "enablement": "isSupportedProject && ballerina.icpRunning && !devant.editor"
+                "enablement": "isSupportedProject && ballerina.icpSupported && ballerina.icpRunning && !devant.editor"
             },
             {
                 "command": "ballerina.showTraceWindow",
--- a/packages/ballerina-extension/src/RPCLayer.ts
+++ b/packages/ballerina-extension/src/RPCLayer.ts
@@ -44,6 +44,7 @@ import { registerTestManagerRpcHandlers } from './rpc-managers/test-manager/rpc-
 import { registerIcpServiceRpcHandlers } from './rpc-managers/icp-service/rpc-handler';
 import { registerWorkflowManagementServiceRpcHandlers } from './rpc-managers/workflow-management-service/rpc-handler';
 import { extension } from './BalExtensionContext';
+import { isICPSupported } from './utils/config';
 import { registerAgentChatRpcHandlers } from './rpc-managers/agent-chat/rpc-handler';
 import { ChatPanel } from './views/agent-chat/webview';
 import { activeAgentChanged, tracingStatusChanged, TraceStatus } from '@wso2/ballerina-core';
@@ -182,6 +183,7 @@ async function getContext(): Promise<VisualizerLocation> {
             rootDiagramId: context.rootDiagramId,
             metadata: {
                 isBISupported: context.isBISupported,
+                isICPSupported: isICPSupported(),
                 haveLS: StateMachine.langClient() && true,
                 recordFilePath: context.projectPath ? path.join(context.projectPath, "types.bal") : undefined,
                 enableSequenceDiagram: extension.ballerinaExtInstance.enableSequenceDiagramView(),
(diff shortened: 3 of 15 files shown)


NipunaRanasinghe (approved):
Approving as the 4 individual fixes are independently reviewed and already approved

PR #785, opened 2026-08-04
Show the ambient Copilot surfaces only where they help

RNViththagan:
Follow-up to #769. Each ambient Copilot surface now appears only where it helps.

- Remove the prompt bar from the project (workspace) overview — it stays on the package/integration overview
- Move that prompt bar into the design panel, below the component diagram, so it sits with the artifacts it acts on rather than above the page
- Vary its placeholder by state — "What would you like to build?" on an empty integration, "What would you like to change?" once artifacts exist — matching the wording the chat panel already uses
- Show the floating orb only on the hubs and design canvases — overview, artifacts, the flow/service/service-class/agent designers, the ER/type/GraphQL diagrams and both data mappers. Forms, wizards, list and settings pages, setup/welcome pages and Copilot's own views go without it, and a view added later has to opt in
- Stop the orb rendering over loading and error states: it moves inside the `viewReady` branch (so the language-server loading and dependency-pulling views no longer carry it), and it stays hidden while no view is resolved, on an unknown view, and on the view-load error state
- Generalise the hero-presence store into orb suppression (`useSuppressAgentStatusOrb()` / `subscribeOrbSuppressed()`), so a view opts out whether or not it carries a prompt bar
- Hide the Copilot status bar item unless it has something to report: it now shows only for a live run, a pending input or an error, and only while the panel is not on screen. The AI panel gains an `onDidChangeViewState` listener so a panel hidden behind another tab still gets the status bar
- Stop the follow-up chips being clipped: a negative bottom margin put the last ~14px of the chip row past the transcript's scroll end, so auto-scroll could never reveal the full row
- Fix a pre-existing e2e bug in the HTTP upload test: it invoked "Run Integration" by command ID through the palette, which never matches the palette's label text, so the service never started

The view declares the orb opt-out because the orb cannot learn the current view from `onStateChanged` — one listener per notification, owned by the panel roots — and mount/unmount already follows navigation. Panel visibility is tracked separately from `aiPanelOpen`, which keeps its "panel is alive" meaning for the orb and the mini-chat handoff.

Screenshot 2026-08-04 at 11 40 57 PM

Screenshot 2026-08-04 at 11 40 49 PM

## Summary

- Removed the Copilot prompt bar from workspace overviews.
- Moved the prompt bar into the package and integration design panel.
- Updated prompt text for empty and populated integrations.
- Limited the floating agent status orb to supported design views.
- Suppressed the orb on unsupported, loading, error, unresolved, and Copilot views.
- Added shared orb-suppression state and removed hero-presence APIs.
- Updated AI panel visibility tracking and status bar behavior.
- Suppressed the orb in embedded Create flows.

### DIFF
--- a/packages/ballerina-extension/src/features/ai/state/AgentStatusManager.ts
+++ b/packages/ballerina-extension/src/features/ai/state/AgentStatusManager.ts
@@ -127,9 +127,6 @@ class AgentStatusManager {
             return;
         }
         this.status = { ...this.status, aiPanelOpen: open, timestamp: Date.now() };
-        if (open) {
-            this.acknowledgeTerminalState();
-        }
         this.render();
         this.broadcast();
     }
@@ -139,7 +136,9 @@ class AgentStatusManager {
             return;
         }
         this.aiPanelVisible = visible;
-        const acknowledged = visible && this.acknowledgeTerminalState();
+        // Either direction means the panel has been on screen: becoming visible
+        // shows the outcome, and going hidden means it was visible until now.
+        const acknowledged = this.acknowledgeTerminalState();
         this.render();
         if (acknowledged) {
             this.broadcast();
--- a/packages/ballerina-extension/src/features/ai/state/AgentStatusManager.ts
+++ b/packages/ballerina-extension/src/features/ai/state/AgentStatusManager.ts
@@ -46,6 +46,8 @@ class AgentStatusManager {
     private status: AgentRunStatus = { state: 'idle', aiPanelOpen: false, timestamp: Date.now() };
     private runActive = false;
     private resetTimer: NodeJS.Timeout | undefined;
+    /** Panel on screen right now, as opposed to `status.aiPanelOpen`, which is merely alive. */
+    private aiPanelVisible = false;
 
     init(context: vscode.ExtensionContext): void {
         if (this.statusBarItem) {
@@ -56,7 +58,6 @@ class AgentStatusManager {
         this.statusBarItem.command = SHARED_COMMANDS.OPEN_AI_PANEL;
         context.subscriptions.push(this.statusBarItem, new vscode.Disposable(() => this.clearResetTimer()));
         this.render();
-        this.statusBarItem.show();
     }
 
     runStarted(generationId: string): void {
@@ -126,15 +127,37 @@ class AgentStatusManager {
             return;
         }
         this.status = { ...this.status, aiPanelOpen: open, timestamp: Date.now() };
-        // Opening the panel acknowledges a finished/failed run — reset to idle
-        // so the 'Done — click to open' nudge doesn't reappear when the panel
-        // is closed again within the terminal-state window.
-        if (open && !this.runActive && (this.status.state === 'completed' || this.status.state === 'error')) {
-            this.clearResetTimer();
-            this.status = { ...this.status, state: 'idle', label: undefined };
+        if (open) {
+            this.acknowledgeTerminalState();
         }
         this.render();
         this.broadcast();
+    }
+
+    setAiPanelVisible(visible: boolean): void {
+        if (this.aiPanelVisible === visible) {
+            return;
+        }
+        this.aiPanelVisible = visible;
+        const acknowledged = visible && this.acknowledgeTerminalState();
+        this.render();
+        if (acknowledged) {
+            this.broadcast();
+        }
+    }
+
+    /**
+     * Seeing the panel acknowledges a finished/failed run — reset to idle so the
+     * 'Done — click to open' nudge doesn't reappear once the panel is closed or
+     * hidden again within the terminal-state window.
+     */
+    private acknowledgeTerminalState(): boolean {
+        if (this.runActive || (this.status.state !== 'completed' && this.status.state !== 'error')) {
+            return false;
+        }
+        this.clearResetTimer();
+        this.status = { ...this.status, state: 'idle', label: undefined };
+        return true;
     }
 
     getStatus(): AgentRunStatus {
@@ -164,6 +187,13 @@ class AgentStatusManager {
         if (!this.statusBarItem) {
             return;
         }
+        // Only worth a slot in the status bar when there is live status to report
+        // and no panel on screen already reporting it. A panel that is open but
+        // hidden behind another tab still needs the status bar.
+        if (this.status.state === 'idle' || this.aiPanelVisible) {
+            this.statusBarItem.hide();
+            return;
+        }
         const label = truncate(this.status.label, STATUS_BAR_LABEL_MAX);
         switch (this.status.state) {
             case 'running':
@@ -182,7 +212,6 @@ class AgentStatusManager {
                 this.statusBarItem.text = `$(error) Copilot error`;
                 this.statusBarItem.backgroundColor = new vscode.ThemeColor('statusBarItem.errorBackground');
                 break;
-            case 'idle':
             default:
                 this.statusBarItem.text = `$(bi-ai-chat) WSO2 Integrator Copilot`;
                 this.statusBarItem.backgroundColor = undefined;
@@ -192,6 +221,7 @@ class AgentStatusManager {
         tooltip.appendMarkdown(`**WSO2 Integrator Copilot**${this.status.label ? ` — ${this.status.label}` : ''}\n\n`);
         tooltip.appendMarkdown('Click to open the Copilot chat.');
         this.statusBarItem.tooltip = tooltip;
+        this.statusBarItem.show();
     }
 
     private broadcast(): void {
--- a/packages/ballerina-extension/src/features/ai/state/AgentStatusManager.ts
+++ b/packages/ballerina-extension/src/features/ai/state/AgentStatusManager.ts
@@ -211,10 +211,6 @@ class AgentStatusManager {
                 this.statusBarItem.text = `$(error) Copilot error`;
                 this.statusBarItem.backgroundColor = new vscode.ThemeColor('statusBarItem.errorBackground');
                 break;
-            default:
-                this.statusBarItem.text = `$(bi-ai-chat) WSO2 Integrator Copilot`;
-                this.statusBarItem.backgroundColor = undefined;
-                break;
         }
         const tooltip = new vscode.MarkdownString();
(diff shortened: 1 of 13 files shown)


packages/ballerina-extension/src/features/ai/state/AgentStatusManager.ts
```diff
@@ -126,17 +127,39 @@ class AgentStatusManager {
...
+     */
+    private acknowledgeTerminalState(): boolean {
+        if (this.runActive || (this.status.state !== 'completed' && this.status.state !== 'error')) {
+            return false;
+        }
+        this.clearResetTimer();
+        this.status = { ...this.status, state: 'idle', label: undefined };
+        return true;
+    }
+
```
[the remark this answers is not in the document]
RNViththagan: Addressed in ea6c542a23.

packages/ballerina-visualizer/src/MainPanel.tsx
```diff
@@ -203,6 +205,7 @@ const MainPanel = () => {
     const remountKeyRef = useRef<number>(0);
     const previousNavTargetRef = useRef<string | undefined>(undefined);
 
+    useSuppressAgentStatusOrb(viewHidesAgentStatusOrb(activeView) || !!viewError);
```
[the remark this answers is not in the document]
RNViththagan: This one was already fixed before the review landed — `setActiveView` now runs after the view component has been set, so suppression only lifts once the new view is actually on screen. It sits at `MainPanel.tsx:864` on the current head.

Same symptom was reported separately: the orb flashed for about a second mid-switch while the lazy chunk for the target view was still loading.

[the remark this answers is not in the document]
RNViththagan: You're right about `BIDiagram` — that case fires `getSTByRange` without awaiting it, so the switch finished and the active view settled while the diagram was still being fetched. It was the only un-awaited case in the switch. Addressed in 0dede96dc2.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
```diff
@@ -1220,30 +1178,26 @@ export function PackageOverview(props: PackageOverviewProps) {
...
                                     )}
                                 </DiagramContent>
                             )}
+                            {showHero && (
+                                <HeroRow>
+                                    <CopilotHeroBox
+                                        placeholder={isEmptyIntegration() ? "What would you like to build?" : "What would you like to change?"}
+                                    />
+                                </HeroRow>
+                            )}
```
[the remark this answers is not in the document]
RNViththagan: The direction described here is already covered: the orb doesn't rely on the hero's suppression while the panel is open, it checks `aiPanelOpen` itself (`AgentStatusOrb/index.tsx:362`). Both values come from the same status store, so one notification updates the hero and the orb in a single batch — in the render where the hero unmounts, the orb is already hidden.

Checking the other direction did turn up a real gap though: suppression was applied in a passive effect, so closing the panel (hero remounts) or navigating to a view that opts out painted one frame with the orb still visible. Fixed in 74bfdc40cd by making the suppression hook a layout effect, so it lands in the same frame as the render that caused it.

SasinduDilshara (dismissed):

SasinduDilshara:
@NipunaRanasinghe Can you please review this

packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
```diff
@@ -359,33 +359,66 @@ export function subscribeCopilotChatNotify(
...
+    MACHINE_VIEW.BIDiagram,
+    MACHINE_VIEW.ServiceDesigner,
+    MACHINE_VIEW.BIServiceClassDesigner,
+    MACHINE_VIEW.AIAgentDesigner,
+    MACHINE_VIEW.ERDiagram,
+    MACHINE_VIEW.TypeDiagram,
+    MACHINE_VIEW.GraphQLDiagram,
+    MACHINE_VIEW.DataMapper,
+    MACHINE_VIEW.InlineDataMapper,
+]);
```
NipunaRanasinghe, line 412: `MACHINE_VIEW.WorkspaceOverview` isn't in this set, so `viewHidesAgentStatusOrb` returns `true` for it and `MainPanel` suppresses the floating orb there. Combined with this PR removing `CopilotHeroBox` (plus `HeroRow` and the `useAiPanelOpen` call) from `WorkspaceOverview/index.tsx`, the workspace overview ends up with **no Copilot surface at all** — where before it showed a hero box whenever the AI panel was closed.

The two halves live in different files, so the combined effect isn't visible from either diff on its own, which is why I'm flagging it rather than assuming.

The doc comment above doesn't settle it either: WorkspaceOverview is a "hub" (argues for the orb) and a package list page (argues against). If dropping it is deliberate, worth a line in the PR description — it's the multi-root workspace landing page, so it's a fairly prominent place to have no way into the Copilot.

For what it's worth, `SequenceDiagram` is the only other diagram-ish view missing from this set, and that one is correct — `MACHINE_VIEW.SequenceDiagram` has zero references anywhere in the repo, so it's a dead enum member.

packages/ballerina-extension/src/features/ai/state/AgentStatusManager.ts
```diff
@@ -164,6 +186,13 @@ class AgentStatusManager {
         if (!this.statusBarItem) {
             return;
         }
+        // Only worth a slot in the status bar when there is live status to report
+        // and no panel on screen already reporting it. A panel that is open but
+        // hidden behind another tab still needs the status bar.
+        if (this.status.state === 'idle' || this.aiPanelVisible) {
+            this.statusBarItem.hide();
+            return;
```
NipunaRanasinghe, line 194: Minor, but this early return combined with dropping `case 'idle':` from the switch below leaves the `default:` branch unreachable.

`AgentRunState` has four states. `idle` returns here, and `running` / `completed` / `error` each have their own case, so `default:` can now only be hit by a state that doesn't exist — meaning its `$(bi-ai-chat) WSO2 Integrator Copilot` text and `backgroundColor = undefined` reset can never render.

Not a functional problem, but it reads as though idle still shows that neutral label, which is now precisely the opposite of what happens. Either drop the branch or relabel it as the unknown-state fallback so the next reader doesn't have to re-derive this.

The visible/open split itself looks right, incidentally — I traced run-active-with-panel-hidden (status bar shows), reveal via the status bar (hides, and `runActive` correctly blocks a premature acknowledge), completion while visible (acknowledged on the hide transition), and the `setAiPanelOpen(false)` -> `setAiPanelVisible(false)` ordering in `dispose`. I couldn't find a path where the "Done — click to open" nudge reappears after the panel had been seen.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
```diff
@@ -116,18 +116,20 @@ const HeaderControls = styled.div`
...
-    max-height: ${(props: { withHero?: boolean }) => props.withHero ? 'calc(100vh - 177px)' : 'calc(100vh - 90px)'};
+    // Adjust based on header and any margins.
+    max-height: calc(100vh - 90px);
 `;
 
+// Sits at the foot of the design panel, which carries no padding of its own.
 const HeroRow = styled.div`
-    margin: 16px 16px 0 16px;
+    flex: none;
+    padding: 16px;
```
NipunaRanasinghe: this is a question rather than a finding:

`DiagramPanel` is `display: flex; flex-direction: column`, so `flex: none` does what's intended and the hero lands at the end of the column. But `DiagramPanel` is also `overflow: auto`, so the hero is inside the scroll container rather than pinned to the panel — it's the foot of the *content*, not a sticky footer. With `min-height: calc(60vh)` and a long artifact list, it should scroll out of view.

That matters because the empty-state copy was changed in the same diff to "describe what you want to build in the Copilot box **below**", which only reads correctly while the panel isn't scrolled. The empty-integration case has little content so it's likely fine there; it's the populated case where the hero becomes hard to find, having previously been in a fixed row above `MainContent`.

If it should stay put, `position: sticky; bottom: 0` on `HeroRow` with a background would do it. If scrolling with content is the intent, no change needed — just confirming it was a choice.

RNViththagan:
Went through all four e2e groups locally to chase down the failures. Fixed two pre-existing bugs unrelated to this PR:

- `http-upload.spec.ts` was invoking "Run Integration" by command ID through the command palette, which never matches the palette's label text, so the service never started. Reverted it to type the label instead, matching how the sibling test right next to it already does it.
- The same spec's "add return node" diagram step and `project-explorer.spec.ts`'s artifact-picker check both had timeouts that were fine early in a run but too tight once the suite had been going a while and the extension host had more state to push through. Gave both more time rather than changing how they interact.

All three are e2e-only fixes, nothing in `main` or in this PR's actual diff — none of it touches the prompt bar, the orb, the chips, or the status bar. Confirmed by running everything locally afterward: all four groups pass now.

packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
RNViththagan: Yeah that's on purpose — we pulled both the prompt bar and the orb off the project overview intentionally, it's a follow-up to the panel-open change in #769.

packages/ballerina-extension/src/features/ai/state/AgentStatusManager.ts
RNViththagan: Good catch, that's dead code — fixed in ac68302565. idle bails out early and the other four states each have their own case, so default could never actually run. Just deleted it instead of relabeling. And thanks for tracing through the visible/open logic, good to have a second set of eyes confirm it holds up.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
RNViththagan: Went with sticky — pushed a fix in bc077cc314 so the hero bar stays put at the bottom of the panel instead of scrolling away on a long artifact list.

SasinduDilshara (approved):

NipunaRanasinghe (approved):
LGTM

PR #790, opened 2026-08-05
Fix infinite loading when editing or deleting a type

### LINKED ISSUE #1988: Editing and Deleting an existing type result in infinite loading screen
#### Editing

https://github.com/user-attachments/assets/ba0b6164-815e-4d21-a0db-a0c68ad60633

#### Deleting

https://github.com/user-attachments/assets/e682ef9c-7140-48fb-b12b-4db3f865cbeb

```
[Trace - 11:21:14 PM] Received response 'flowDesignService/getFlowModel - (503)' in 3ms.
Result: {
    "errorMsg": "Line number: '3', Size: '2'",
    "stacktrace": "[io.ballerina.tools.text.LineMap.lineRangeCheck(LineMap.java:78), io.ballerina.tools.text.LineMap.textPositionFrom(LineMap.java:50), io.ballerina.tools.text.TextDocument.textPositionFrom(TextDocument.java:52), io.ballerina.flowmodelgenerator.core.ModelGenerator.getFlowModel(ModelGenerator.java:129), io.ballerina.flowmodelgenerator.extension.FlowModelGeneratorService.lambda$getFlowModel$0(FlowModelGeneratorService.java:150), java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768), java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1760), java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387), java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312), java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1843), java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1808), java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188)]"
}
```

### Steps to Reproduce

Edit any type and save

### Version

v5.0.1

_No response_

dulajdilshan:
## Purpose

Fixes https://github.com/wso2/product-integrator/issues/1988

Recording 2026-08-05 at 10 51 53

## Goals

- Editing or deleting a type no longer leaves the Type Diagram stuck on a spinner.
- Deleting a type completes instead of timing out after 10s.
- A failed save shows an error instead of spinning forever.

## Approach

- Fixed a stale closure in `TypeDiagram` — the debounced refresh was frozen at first render and always bailed out before clearing the loading flag.
- `getComponentModel` now owns that flag and always clears it in a `finally`.
- Added `skipPayloadCheck: true` to `deleteType`; deletions publish an empty payload, which the caller was waiting on forever.
- Wrapped `onTypeSave` (and the JSON/XML import forms) in `try/finally` so the saving flag is always cleared.
- Memoised the `EditorContext` value and stopped `handleFieldValidation` churning state, to cut needless re-renders.

Not included: the `fieldType is null` NPE, the workspace artifact routing, and the relative `filePath` — all tracked separately.

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary

- Fixed infinite loading when editing or deleting types in the Type Diagram.
- Prevented stale model refreshes and handled overlapping requests.
- Reset loading and saving states after successful or failed operations.
- Improved error reporting for type save failures.
- Added `skipPayloadCheck: true` to type deletion updates.
- Memoized the `EditorContext` value and reduced unnecessary state updates.
- Ensured JSON and XML imports always clear their saving state.
- Added unit and end-to-end coverage for type editing, deletion, stale responses, current-location refreshes, and missing type files.


packages/ballerina-visualizer/src/views/TypeDiagram/index.tsx
```diff
@@ -274,23 +290,54 @@ export function TypeDiagram(props: TypeDiagramProps) {
     }
 
     const getComponentModel = async () => {
-        if (!rpcClient || !visualizerLocation?.metadata?.recordFilePath) {
+        // The visualizer location resolves asynchronously. Until it does there is nothing to
+        // fetch and the progress ring is the correct state, so return without clearing it.
+        if (!rpcClient || !visualizerLocation) {
             return;
```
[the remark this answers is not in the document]
dulajdilshan: Addressed via https://github.com/ballerina-platform/ballerina-vscode/pull/790/commits/849bc399f7b625f5598e2f4de9611cec20ba84bf

packages/type-editor/src/TypeEditor/ContextBasedTypeEditor/ContextTypeEditor.tsx
```diff
@@ -160,31 +160,40 @@ export function ContextTypeEditor(props: ContextTypeEditorProps) {
...
+                    .updateType({ filePath: type.codedata?.lineRange?.fileName || 'types.bal', type, description: "" });
+                if (!isPopupTypeForm) {
+                    await props.rpcClient
+                        .getVisualizerRpcClient()
+                        .openView({ type: EVENT_TYPE.UPDATE_PROJECT_LOCATION, location: { identifier: response.name, addType: false } });
+                }
             }
+            props.onTypeChange(type);
+            props.onSaveType(type)
+        } catch (error) {
```
NipunaRanasinghe, line 198: This `catch` covers both the actual save call (`createGraphqlClassType`/`updateType`) and the follow-up `openView(UPDATE_PROJECT_LOCATION)` navigation. If only the navigation throws — the save having already succeeded — the user sees "Failed to save the type '...'" for an operation that actually completed. Worth narrowing the catch to the save call specifically (or catching the `openView` call separately and not surfacing it as a save failure), so the error message stays accurate about which step actually failed.

Same shape exists in `TypeEditor.tsx:132` (identical `onTypeSave`, duplicated across these two files) — worth applying the same fix there too if you go this route.

packages/type-editor/src/TypeEditor/RecordEditor.tsx
```diff
@@ -62,16 +62,19 @@ export const RecordEditor = forwardRef<{ addMember: () => void }, RecordEditorPr
 
     const handleFieldValidation = (functionIndex: number, isIdentifier: boolean, hasError: boolean) => {
```
NipunaRanasinghe, line 63: Same bail-out-on-unchanged-value fix is duplicated in `SchemaRecordEditor.tsx` (and the `onTypeSave` try/finally fix is likewise duplicated between this file's `TypeEditor.tsx`/`ContextTypeEditor.tsx` pair). Not asking for a dedup in this PR, but since a bug fix already had to be pasted into two near-identical files twice, it's worth a follow-up ticket to unify these pairs — otherwise every future fix in this logic has to remember to touch both copies, and it's easy to miss one.

packages/ballerina-visualizer/src/views/TypeDiagram/index.tsx
```diff
@@ -274,23 +297,54 @@ export function TypeDiagram(props: TypeDiagramProps) {
     }
 
     const getComponentModel = async () => {
-        if (!rpcClient || !visualizerLocation?.metadata?.recordFilePath) {
+        // The visualizer location resolves asynchronously. Until it does there is nothing to
```
NipunaRanasinghe, line 300: This fix is entirely built around ordering guarantees — the monotonic `componentModelRequestId` guard against a stale in-flight fetch clobbering a newer result, and the ref-indirected debounce so it never re-invokes a first-render closure. Both are exactly the kind of logic that regresses silently under a future refactor without anyone noticing, since the failure mode (spinner stuck, or a stale response briefly flashing) is easy to miss in manual testing. No test changes are included in this PR for either of these. Worth at least one test that simulates an older request resolving after a newer one has already completed, to lock in the request-id guard's behavior.

NipunaRanasinghe (dismissed):

packages/type-editor/src/TypeEditor/RecordEditor.tsx
dulajdilshan: Agreed. We need to unify the `TypeEditor`/`ContextTypeEditor` and `RecordEditor`/`SchemaRecordEditor` pairs — this PR did have to paste the same fix twice into each, which is exactly the argument for it. Kept out of scope here to keep the fix reviewable.

packages/type-editor/src/TypeEditor/ContextBasedTypeEditor/ContextTypeEditor.tsx
dulajdilshan: Narrowed the save call, the navigation, and the post-save callbacks now have separate handlers, and only an actual save failure surfaces to the user. Applied to both files

packages/ballerina-visualizer/src/views/TypeDiagram/index.tsx
dulajdilshan: Added three tests covering both guards and verified that each fails when its guard is removed.

packages/type-editor/src/TypeEditor/RecordEditor.tsx
dulajdilshan: Created [issue](https://github.com/wso2-enterprise/integration-engineering/issues/2257) to track this

dulajdilshan:
@coderabbitai

Fixed the Nipick 2.

**Nitpick 1 — `renderWithRpc`**

`renderWithRpc` doesn't exist in this package's harness — it only exports `TestRpcContext` and `useRpcContext`. The render helper needs `@testing-library/react`, which is a dep of `ballerina-side-panel`, not `ballerina-visualizer`; the manual Provider wrap avoids adding a test-only dependency here.

**Nitpick 3 — `issue-1988.json` fixture**

 The `issue-<n>.json` policy covers recorded LS/RPC traffic under `src/test/fixtures/<method>/`. This regression is temporal — an older promise resolving after a newer one — which no JSON payload can express, and `ballerina-visualizer` has no fixtures dir or invariant test to consume one, so the file would be unused. Covered by the bespoke ordering test instead.

NipunaRanasinghe:
@coderabbitai review

dulajdilshan:
@coderabbitai

About the Nitpick.

Fair, and it matches the e2e skill's guidance. None of these five elements has a data-testid today — the Rename button, the editable type-name field and the rename Save in TypeCreatorTab.tsx, the node menu item in EntityHead.tsx, and the confirm button inside ui-toolkit's Confirm (which lives in submodules/, off-limits per the skill). Doing this properly means adding testids across two packages plus a prop on a shared component, then a VSIX rebuild and re-run. These selectors are pre-existing — this PR only removes test.skip — so I'd rather land the re-enable and do the selector hardening as a follow-up once the tests have a green baseline.

packages/type-editor/src/TypeEditor/ContextBasedTypeEditor/ContextTypeEditor.tsx
```diff
@@ -159,32 +159,46 @@ export function ContextTypeEditor(props: ContextTypeEditorProps) {
...
+            try {
+                // IF type nodeKind is CLASS then we call graphqlEndpoint
+                // TODO: for TypeDiagram we need to give a generic class creation
+                response = type.codedata.node === "CLASS"
+                    ? await props.rpcClient.getBIDiagramRpcClient().createGraphqlClassType(request)
+                    : await props.rpcClient.getBIDiagramRpcClient().updateType(request);
+            } catch (error) {
+                // Only a failure of the save itself is reported as a save failure.
+                console.error(">>> error saving the type", error);
+                props.rpcClient.getCommonRpcClient().showErrorMessage({
```
NipunaRanasinghe, line 178: this now reports only genuine save failures, and the `return` inside this catch still runs the `finally`, so the saving flag can't leak.

One thing about the message itself: `(error as Error)?.message || …` makes the raw error the preferred text and the friendly one a fallback for when `message` is empty. So in the normal failure case the user gets whatever string the LS/RPC layer produced in a VS Code error toast, and the version naming the type — the one written for this — effectively never renders.

Given "improved error reporting for type save failures" is one of the goals here, I'd either lead with the friendly message and keep the raw detail out of the toast (it's already on the line above via `console.error`, which is where it's useful), or combine them so the toast still says which type failed:

```ts
message: `Failed to save the type '${name}'.${(error as Error)?.message ? ` ${(error as Error).message}` : ""}`
```

Same expression exists in `TypeEditor.tsx` (the twin `onTypeSave`), so it'd need the same edit — flagging here rather than opening a second thread, since it's one decision.

packages/ballerina-visualizer/src/views/TypeDiagram/index.test.tsx
```diff
@@ -0,0 +1,206 @@
...
+    it("clears the loading state when the location has no types file", async () => {
+        const { rpcClient, getTypes } = makeRpc(null);
+
+        await renderDiagram("/a", rpcClient);
+        await settle();
+
+        // Nothing can be fetched, but the diagram must still leave the progress ring.
+        expect(getTypes).not.toHaveBeenCalled();
+        expect(renderedModel()).not.toBeNull();
+    });
```
NipunaRanasinghe, line 210: The one guard left without coverage is the `catch`/`finally` in `getComponentModel`: a `getTypes` rejection clearing the loading state. That's the third of the failure modes in the PR description ("a failed save shows an error instead of spinning forever" has an equivalent on the fetch side), and it's the branch most likely to be quietly dropped in a future refactor, since nothing observable breaks until someone hits a failing fetch. The harness already supports it — `pending` holds the resolver, so a `reject` variant is a few lines:

```tsx
it("clears the loading state when the fetch fails", async () => {
    // reject the in-flight getTypes, then assert the ring is gone
});
```

Small related point on this test: `expect(renderedModel()).not.toBeNull()` asserts the ring is gone only indirectly, via the model stub having rendered in its place. Since `ProgressRing` is already stubbed with a testid, asserting it directly states the intent and stays true if the render branches ever change shape:

```tsx
expect(container.querySelector('[data-testid="progress-ring"]')).toBeNull();
```

NipunaRanasinghe (dismissed):
Approved with some non-blocking suggestions

packages/ballerina-visualizer/src/views/TypeDiagram/index.test.tsx
dulajdilshan: Addressed via https://github.com/ballerina-platform/ballerina-vscode/pull/790/commits/c89c67d338e351b1f1e1be4e195ee7ad65b4f749

packages/type-editor/src/TypeEditor/ContextBasedTypeEditor/ContextTypeEditor.tsx
dulajdilshan: Addressed via https://github.com/ballerina-platform/ballerina-vscode/pull/790/commits/c89c67d338e351b1f1e1be4e195ee7ad65b4f749

NipunaRanasinghe (approved):

PR #798, opened 2026-08-05
Rework the Copilot entry point on the overview page

RNViththagan:
Related to https://github.com/wso2/product-integrator/issues/1998

Follow-up to #785 on where the Copilot prompt bar lives in the integration (package) overview.

- Move the prompt bar into the empty-integration state — centred, above the Add Integration button — instead of pinned to the foot of the design panel
- Show the bar only while the integration is empty; once artifacts exist the floating orb takes over, so there is one ambient Copilot surface rather than two competing ones
- Make the empty-state copy follow the bar, so it stops inviting the user to describe what they want to build when the bar is hidden
- While a Copilot turn is in flight, show a loader with "Copilot is working…" and hide the Add Integration button — the turn may not be creating artifacts at all, so neither an add invitation nor a generation claim is right until it settles
- Restore the project overview's empty-state copy, which still pointed at a Copilot box that #785 removed from that page

https://github.com/user-attachments/assets/6bb8b1b3-1cc4-4c22-8ece-4c52da65a634

Reworked the Copilot entry point on the integration overview page.

Moved the Copilot prompt bar into the centered empty-integration state above the Add Integration button. Show the prompt bar only when the integration has no artifacts. Use the floating orb after artifacts exist.

Added shared Copilot run-status tracking. Show the “Copilot is working…” state during active turns and hide the Add Integration button.

Updated the package overview and restored the project overview empty-state messages.

### DIFF
--- a/packages/ballerina-extension/e2e-test/e2e-playwright-tests/connections/connections.spec.ts
+++ b/packages/ballerina-extension/e2e-test/e2e-playwright-tests/connections/connections.spec.ts
@@ -17,7 +17,7 @@
  */
 import { Frame, Locator, test } from '@playwright/test';
 import * as path from 'path';
-import { BI_INTEGRATOR_LABEL, BI_WEBVIEW_NOT_FOUND_ERROR, initTest, page, logStep, newProjectPath } from '../utils/helpers';
+import { BI_INTEGRATOR_LABEL, BI_WEBVIEW_NOT_FOUND_ERROR, initTest, page, logStep, newProjectPath, domClick } from '../utils/helpers';
 import { switchToIFrame, Form } from '@wso2/playwright-vscode-tester';
 import { ProjectExplorer, Diagram, SidePanel } from '../utils/pages';
 import { DEFAULT_PROJECT_NAME } from '../utils/helpers/constants';
@@ -326,10 +326,17 @@ export default function createTests() {
             artifactWebView = await switchToIFrame(BI_INTEGRATOR_LABEL, page.page, 30000);
 
             await artifactWebView.getByRole('button', { name: /Add Artifact/i }).click({ force: true });
-            await artifactWebView.locator('[data-testid="function-card-Connection"], #connection').first().click({ force: true });
+            // The floating Copilot orb can dock exactly over this card/link, so a
+            // coordinate click (even with `force`) can silently land on the orb
+            // instead — same failure class `domClick` exists to avoid (see its doc).
+            const connectionCard = artifactWebView.locator('[data-testid="function-card-Connection"], #connection').first();
+            await connectionCard.waitFor({ state: 'visible', timeout: 30000 });
+            await domClick(connectionCard);
             await page.page.waitForTimeout(1500);
 
-            await artifactWebView.getByText('Connect via API Specification', { exact: false }).first().click({ force: true });
+            const apiSpecOption = artifactWebView.getByText('Connect via API Specification', { exact: false }).first();
+            await apiSpecOption.waitFor({ state: 'visible', timeout: 30000 });
+            await domClick(apiSpecOption);
             await page.page.waitForTimeout(1000);
 
             const connectorNameInput = artifactWebView.locator('#connector-name');
--- a/packages/ballerina-extension/e2e-test/e2e-playwright-tests/connections/connections.spec.ts
+++ b/packages/ballerina-extension/e2e-test/e2e-playwright-tests/connections/connections.spec.ts
@@ -325,12 +325,12 @@ export default function createTests() {
             // diagram — no explicit "Open View" navigation is needed.
             artifactWebView = await switchToIFrame(BI_INTEGRATOR_LABEL, page.page, 30000);
 
-            await artifactWebView.getByRole('button', { name: /Add Artifact/i }).click({ force: true });
-            // The floating Copilot orb can dock exactly over this card/link, so a
-            // coordinate click (even with `force`) can silently land on the orb
-            // instead — same failure class `domClick` exists to avoid (see its doc).
+            // The diagram can still be settling right after the previous test's save
+            // (same drag-tolerance trap as elsewhere in this file) — retry the click
+            // rather than assume it registered.
             const connectionCard = artifactWebView.locator('[data-testid="function-card-Connection"], #connection').first();
-            await connectionCard.waitFor({ state: 'visible', timeout: 30000 });
+            await clickUntil(artifactWebView.getByRole('button', { name: /Add Artifact/i }), connectionCard, 'Add Artifact button');
+            // domClick avoids the floating Copilot orb intercepting a coordinate click.
             await domClick(connectionCard);
             await page.page.waitForTimeout(1500);
 
--- a/packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
+++ b/packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
@@ -303,6 +303,27 @@ export function useAiPanelOpen(): boolean {
     return open;
 }
 
+/**
+ * True while a Copilot turn is in flight — it says nothing about what the turn
+ * will produce, since the prompt may not be asking for artifacts at all.
+ * `awaiting-input` counts: the turn has not settled yet.
+ */
+export function useAgentWorking(): boolean {
+    const { rpcClient } = useRpcContext();
+    const isWorking = (status: AgentRunStatus | null) =>
+        status?.state === "running" || status?.state === "awaiting-input";
+    const [working, setWorking] = useState(() => isWorking(currentStatus));
+
+    useEffect(() => {
+        if (!rpcClient) {
+            return;
+        }
+        return subscribeAgentRunStatus(rpcClient, (status) => setWorking(isWorking(status)));
+    }, [rpcClient]);
+
+    return working;
+}
+
 // ---------------------------------------------------------------------------
 // Contextual mini-chat launch requests.
 //
--- a/packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
+++ b/packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
@@ -45,9 +45,10 @@ export function loadAnchor(): Anchor {
     } catch {
         stored = null;
     }
-    // Default to bottom-center so the copilot invitation is front and center
-    // when BI opens; users can drag the orb to any of the six anchors.
-    return stored && (ANCHORS as readonly string[]).includes(stored) ? (stored as Anchor) : "bottom-center";
+    // bottom-center sits on top of whatever the active view docks at its own
+    // bottom-center (form submit buttons, artifact-picker cards, …), so default
+    // to bottom-right instead; users can still drag the orb to any anchor.
+    return stored && (ANCHORS as readonly string[]).includes(stored) ? (stored as Anchor) : "bottom-right";
 }
 
 export const ORB_COLORS: Record<AgentRunState, [string, string, string]> = {
--- a/packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
+++ b/packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
@@ -286,6 +286,14 @@ export function subscribeAgentRunStatus(
     };
 }
 
+/** Test-only: clears the module-level status cache and listeners between test cases. */
+export function __resetAgentRunStatusStoreForTests(): void {
+    currentStatus = null;
+    statusWired = false;
+    receivedStatusNotification = false;
+    statusListeners.clear();
+}
+
 /**
  * True while the Copilot panel is open — inline copilot surfaces stand down so
  * the panel is the only chat entry point. Seeded from the cached status so a
--- a/packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
+++ b/packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
@@ -303,21 +303,19 @@ export function useAiPanelOpen(): boolean {
     return open;
 }
 
-/** True while a turn is in flight — says nothing about what it will produce. */
-export function useAgentWorking(): boolean {
+/** The live run status, for views that need the state and its label, not just a flag. */
+export function useAgentRunStatus(): AgentRunStatus | null {
     const { rpcClient } = useRpcContext();
-    const isWorking = (status: AgentRunStatus | null) =>
-        status?.state === "running" || status?.state === "awaiting-input";
-    const [working, setWorking] = useState(() => isWorking(currentStatus));
+    const [status, setStatus] = useState(() => currentStatus);
 
     useEffect(() => {
         if (!rpcClient) {
             return;
         }
-        return subscribeAgentRunStatus(rpcClient, (status) => setWorking(isWorking(status)));
+        return subscribeAgentRunStatus(rpcClient, setStatus);
     }, [rpcClient]);
 
-    return working;
+    return status;
 }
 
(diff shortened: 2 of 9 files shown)


packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
```diff
@@ -1208,25 +1215,45 @@ export function PackageOverview(props: PackageOverviewProps) {
...
+                                                >
+                                                    {showHero
+                                                        ? "Describe what you want to build, or add an artifact to get started"
+                                                        : "Add an artifact to get started"}
+                                                </Typography>
+                                            )}
+                                            {showHero && (
+                                                <HeroRow>
+                                                    <CopilotHeroBox placeholder="What would you like to build?" />
+                                                </HeroRow>
```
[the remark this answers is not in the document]
RNViththagan: Fixed in 3c17cc5c83 — guidance now sits below the bar, so the bar comes right after the heading. Also folded the guidance and the Add Integration button under one `!agentWorking` guard, since the reorder would otherwise have left the same condition checked twice back to back. Working row is unchanged, still above the bar.

SasinduDilshara (dismissed):

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
```diff
@@ -1208,25 +1215,46 @@ export function PackageOverview(props: PackageOverviewProps) {
...
+                                                <WorkingRow>
+                                                    <ProgressRing color={ThemeColors.PRIMARY} sx={{ width: 16, height: 16 }} />
+                                                    <Typography
+                                                        variant="body1"
+                                                        sx={{ color: "var(--vscode-descriptionForeground)" }}
+                                                    >
+                                                        Copilot is working…
+                                                    </Typography>
+                                                </WorkingRow>
+                                            )}
```
NipunaRanasinghe: This row and `HeroRow` below are gated independently — `agentWorking` vs `showHero` — so with the AI panel closed during a run both render at once. `CopilotHeroBox` subscribes to the same status feed and morphs into orb + `activeStateLabel(status)` + "Open Copilot →" for any non-idle state, so you end up with two stacked working indicators:

- spinner + "Copilot is working…"
- the hero box reading "Working on it…" (or the run's own `status.label`)

Reachable by closing the panel mid-run, or by starting a turn from the orb / mini-chat while the panel is closed. The recording doesn't hit it because `submit()` fires `OPEN_AI_PANEL`, which flips `aiPanelOpen` true and unmounts the hero.

Since the hero's own active form carries strictly more information (specific label plus a way back into the panel), gating this row on the hero's absence gives exactly one status surface in both configurations:

```tsx
{agentWorking && !showHero && (
```

packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
```diff
@@ -303,6 +303,23 @@ export function useAiPanelOpen(): boolean {
     return open;
 }
 
+/** True while a turn is in flight — says nothing about what it will produce. */
+export function useAgentWorking(): boolean {
+    const { rpcClient } = useRpcContext();
+    const isWorking = (status: AgentRunStatus | null) =>
+        status?.state === "running" || status?.state === "awaiting-input";
```
NipunaRanasinghe: Folding `awaiting-input` in with `running` is right for hiding the Add Integration button — the turn genuinely hasn't settled — but the call site renders it as a spinner plus "Copilot is working…", and in that state Copilot is blocked on the *user*, not working. So the page claims progress, removes the only CTA, and (with the panel closed) says nothing about the agent needing an answer.

`activeStateLabel` a few functions up already has the copy for this: "Copilot needs your input". Either drop `awaiting-input` from this predicate, or return the status/state instead of a boolean so `WorkingRow` can take its text from `activeStateLabel` and skip the spinner for that case.

Separately: `isWorking` is declared in the hook body but isn't in the effect's dep list. It's pure so there's no bug, but it trips `react-hooks/exhaustive-deps`, and `useAiPanelOpen` right above inlines its expression rather than naming it. Module scope would settle both.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
```diff
@@ -1208,25 +1215,46 @@ export function PackageOverview(props: PackageOverviewProps) {
...
+                                            {showHero && (
+                                                <HeroRow>
+                                                    <CopilotHeroBox placeholder="What would you like to build?" />
+                                                </HeroRow>
+                                            )}
+                                            {!agentWorking && (
```
NipunaRanasinghe: The reasoning in the description holds — an in-flight turn may not be producing artifacts, so an add invitation is a guess. But the consequence is that for the whole turn (43s in the recording, potentially minutes) an empty integration has no manual add affordance anywhere on the page: the Design header only grows its Add Artifact button once artifacts exist, so nothing takes over while this is hidden. A user who asked Copilot something unrelated is locked out of adding an artifact until the run settles.

Disabling the button, or leaving it and letting the loader carry the status on its own, both keep that door open. Also worth noting the swap shifts the remaining block ~15px vertically (heading moves 593 → 608 between the idle and working frames), because the container centres a shorter child while `padding-bottom` still only accounts for the hero.

packages/ballerina-visualizer/src/views/BI/WorkspaceOverview/index.tsx
```diff
@@ -1091,8 +1091,7 @@ export function WorkspaceOverview({ isInDevant, isICPSupported }: WorkspaceOverv
                                         variant="body1"
                                         sx={{ marginBottom: "24px", color: "var(--vscode-descriptionForeground)" }}
                                     >
-                                        Add an integration or library to get started, or describe what you want to
-                                        build in the Copilot box above
+                                        Start by adding integrations and libraries to your project
```
NipunaRanasinghe, line 1094: Dropping the stale "Copilot box above" clause is the right call, but the replacement changes voice as well: the sibling empty state on the package overview says "Add an artifact to get started", and this now says "Start by adding…". The plural also invites adding several where the button adds one.

"Add an integration or library to get started" is the original string minus the Copilot clause, and stays consistent with the other page.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
RNViththagan: fixed in b0a7a95abe. The hero carries the status on its own now, so the status row is skipped while it's on screen. With the panel open there's no hero, so the row carries it instead. One either way.

packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
RNViththagan: Went with your second option in 3552165f0d — the hook returns the state now, so waiting shows "Copilot needs your input" with no spinner, while still counting for the button gate.

Dep-list nit went away with it, no derived predicate left in the hook.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
RNViththagan: Disabled instead of hidden — 737ae8201c, with an "Available once Copilot finishes" tooltip.

Kept it blocked though. Revert restores the pre-turn checkpoint and drops anything not in it, so an artifact added mid-run can just vanish. Rather hold the door shut and say why than let someone lose work.

The shift is still partly there — the status row does go away once the hero takes over. Preferred that over two things saying the same thing, but easy to revisit.

packages/ballerina-visualizer/src/views/BI/WorkspaceOverview/index.tsx
RNViththagan: That's the original wording, from before the Copilot box was on this page. "Add an integration or library to get started" came in with the box, as the first half of the sentence pointing at it. Box is gone, so I went back to what was there before.

Can switch if you'd rather keep it consistent with the other page.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
```diff
@@ -1208,22 +1217,49 @@ export function PackageOverview(props: PackageOverviewProps) {
...
+                                                        sx={{ color: "var(--vscode-descriptionForeground)" }}
+                                                    >
+                                                        {agentWorking
+                                                            ? (awaitingInput ? "Copilot needs your input" : "Copilot is working…")
+                                                            : showHero
+                                                                ? "Describe what you want to build, or add an artifact to get started"
+                                                                : "Add an artifact to get started"}
+                                                    </Typography>
+                                                </StatusRow>
+                                            )}
```
NipunaRanasinghe, line 1249: Folding the guidance into `StatusRow` is a good consolidation — one row that carries either the run status or the idle copy, skipped when the hero speaks for itself. The invariant is structural now rather than two conditions that have to be kept in agreement, which is what I was asking for.

One side effect worth making deliberate though: the guidance now renders *above* the prompt bar, where at 3c17cc5 you had deliberately moved it below (commit "Render the prompt bar above the empty-state guidance", in response to the earlier thread on this file). So the empty state has gone from

    heading → bar → copy → button

back to

    heading → copy → bar → button

Both are defensible, and for what it's worth I think the earlier thread was reading "the empty-state copy follows the bar" in the description as positional when it meant conditional — so the reorder it asked for may never have been needed. But right now that thread is marked resolved while describing the opposite of the code, so whichever order you want, it's worth being explicit rather than leaving it as a consequence of the refactor.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
```diff
@@ -92,6 +99,8 @@ const EmptyStateContainer = styled.div`
     flex-direction: column;
     align-items: center;
     justify-content: center;
+    // Offsets the prompt bar's height so the block keeps its former position.
+    padding-bottom: ${(props: { withHero?: boolean }) => (props.withHero ? "96px" : "0")};
```
NipunaRanasinghe, line 103: The offset still does what the comment says — I worked it through for the current child set and the button lands at ~`H/2 + 27` both with and without the hero, so the reorder didn't break it.

What the 96px doesn't cover is `StatusRow` disappearing. Once a turn starts with the hero on screen, that row is skipped entirely rather than swapped for a working row, so the block loses its height plus margin (~44px) and everything below moves up ~22px:

- idle + hero: heading + StatusRow + HeroRow + button → button at `(H-305)/2 + 179`
- working + hero: heading + HeroRow + button → button at `(H-261)/2 + 135`

That's a bigger jump than the ~15px in the previous revision, and it now fires exactly when the user hits Enter, which is the moment they're looking at this area. Reserving the row's height while the hero carries the status (or giving `StatusRow` a `min-height` and rendering it empty) would hold the layout still.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
```diff
@@ -1208,22 +1217,49 @@ export function PackageOverview(props: PackageOverviewProps) {
...
+                                            {/* Skipped only while the hero carries the status itself, so
+                                                the two never state it at once. */}
+                                            {!(agentWorking && showHero) && (
+                                                <StatusRow>
+                                                    {agentWorking && (
+                                                        awaitingInput
+                                                            ? <Codicon name="comment-discussion" />
+                                                            : <ProgressRing color={ThemeColors.PRIMARY} sx={{ width: 16, height: 16 }} />
+                                                    )}
+                                                    <Typography
```
NipunaRanasinghe, line 1238: Giving `awaiting-input` its own icon rather than a spinner is the right treatment — a progress ring while the agent is blocked on the user was the misleading part.

Small thing on the string: `activeStateLabel` in `shared.ts` already returns `status.label ?? "Copilot needs your input"` for this state, so the literal now exists in two places. They can't contradict each other on screen — this row is skipped whenever the hero is up — but two consequences follow from the duplication:

- the two surfaces can drift as one gets reworded
- the hero honours a run's own `status.label` where this row always shows the generic string, so the same run reads differently depending on whether the panel is open

Reusing `activeStateLabel` here (or lifting both strings to shared constants) keeps them in step.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
RNViththagan: You were on an older commit — already put it back, so it's bar then copy now. Fair point that the refactor undid it quietly though.

Agreed the earlier thread read that as a placement thing when it wasn't. Keeping bar-first anyway, since that's the whole point here.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
RNViththagan: Tried it — kept the row's space and left it blank. The empty gap looked worse than the jump it fixes, so I dropped it again.

So the shift stays. Happy to revisit if you've got a better idea.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
RNViththagan: moved it to one shared string, both use it now.

Left the other bit alone though. This row is just the simple one, and the panel's already open whenever it shows, so the detailed text lives there instead.

NipunaRanasinghe (dismissed):
LGTM. 

@SasinduDilshara @IsuruMaduranga PRAM

SasinduDilshara (dismissed):

SasinduDilshara:
Will merge after tests run

packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
```diff
@@ -303,6 +305,21 @@ export function useAiPanelOpen(): boolean {
...
+    useEffect(() => {
+        if (!rpcClient) {
+            return;
+        }
+        return subscribeAgentRunStatus(rpcClient, (status) => setState(status?.state));
+    }, [rpcClient]);
+
+    return state;
+}
+
```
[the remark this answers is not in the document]
RNViththagan: Added `shared.test.ts` covering all four cases you listed — cached initialization from `currentStatus`, live updates via `subscribeAgentRunStatus`, the no-`rpcClient` path, and cleanup after unmount. All 4 tests pass.

packages/ballerina-extension/e2e-test/e2e-playwright-tests/connections/connections.spec.ts
```diff
@@ -325,11 +325,18 @@ export default function createTests() {
             // diagram — no explicit "Open View" navigation is needed.
             artifactWebView = await switchToIFrame(BI_INTEGRATOR_LABEL, page.page, 30000);
 
-            await artifactWebView.getByRole('button', { name: /Add Artifact/i }).click({ force: true });
-            await artifactWebView.locator('[data-testid="function-card-Connection"], #connection').first().click({ force: true });
+            // The diagram can still be settling right after the previous test's save
```
samithkavishke, line 328: Are you sure this is the correct user experience we have now. Ideally even after create a connection we should move to the architecture diagram view

RNViththagan: That's existing behavior, not something this PR changed — we only touched test flakiness around this step (retrying a click, avoiding the floating orb).

SasinduDilshara (approved):

NipunaRanasinghe (approved):

SasinduDilshara (approved):

PR #800, opened 2026-08-06
Make the nightly patch version a readable UTC timestamp

samithkavishke:
## Summary
- The nightly/pre-release patch number was whole minutes since 2020-01-01 UTC (e.g. `5.13.3458370`) — correct, but opaque to read.
- Switches it to `YYMMDDHH` (e.g. `5.13.26080512` for 2026-08-05 12:00 UTC), which stays comfortably under the Marketplace's int32 version-component limit while actually decoding to a date at a glance.
- Dropping minute granularity means two builds cut within the same UTC hour (e.g. a manual re-run) get the same patch number — accepted, since nightlies only run once a day in practice.

## Test plan
- [x] `node common/scripts/nightly-version.js --timestamp 26080512` prints `5.13.26080512`
- [x] `date -u +'%y%m%d%H'` matches the JS `timestamp()` output for the current moment
- [x] Out-of-range stamps are still rejected via the existing int32 guard

- Updated nightly and pre-release patch versions to use the UTC `YYMMDDHH` format.
- Preserved Marketplace `int32` compatibility and documented same-hour version behavior.
- Updated timestamp validation, CLI guidance, and related error messages.
- Added tests for timestamp conversion, current UTC output, and range validation.

### DIFF
--- a/.github/actions/updateVersion/action.yml
+++ b/.github/actions/updateVersion/action.yml
@@ -22,7 +22,7 @@ inputs:
   nightly:
     description: >-
       Nightly build. Uses the same derivation as a pre-release —
-      major.(minor-1).<minutes since 2020-01-01 UTC> from the extension '-SNAPSHOT' version, see
+      major.(minor-1).<YYMMDDHH, UTC> from the extension '-SNAPSHOT' version, see
       common/scripts/nightly-version.js — but is a separate input because it selects
       only the version. Packaging is isPreRelease's job. Build callers must pass
       both for a nightly; version-only callers such as schedule.yml may stamp with
@@ -53,21 +53,22 @@ runs:
       id: timestamp
       run: |
         set -euo pipefail
-        # Whole minutes since 2020-01-01 UTC. Must match nightly-version.js's timestamp()
-        # exactly — it is the fallback when no stamp is passed, and the two disagreeing
-        # would give the same build two different versions.
+        # YYMMDDHH, UTC. Must match nightly-version.js's timestamp() exactly — it is
+        # the fallback when no stamp is passed, and the two disagreeing would give
+        # the same build two different versions.
         #
-        # Not a readable 'yymmddHHmm': that is 2607291530 today, over the Marketplace's
-        # int32 component limit (2147483647), so every pre-release would be rejected at
-        # publish time. See nightly-version.js for the decode one-liner.
-        # 1577836800 = 2020-01-01T00:00:00Z. Hardcoded rather than computed with
-        # 'date -d', which is GNU-only and would not run on a macOS runner.
-        timestamp=$(( ( $(date -u '+%s') - 1577836800 ) / 60 ))
+        # Hour-granular rather than minute-granular: a full 'yymmddHHmm' is 10 digits
+        # and exceeds the Marketplace's int32 component limit (2147483647) from 2022
+        # onward, so every pre-release would be rejected at publish time. Dropping
+        # minutes keeps it to 8 digits, safely under the limit. Two builds cut in the
+        # same UTC hour do collide, which is accepted since nightlies only run once a
+        # day in practice. See nightly-version.js for the decode one-liner.
+        timestamp=$(date -u +'%y%m%d%H')
         echo "timestamp=$timestamp" >> $GITHUB_OUTPUT
 
     # Nightlies and pre-releases share one derivation: the extension version names the
     # NEXT release as a snapshot, and both of these are unreleased work heading
-    # toward it, so they are published as major.(minor-1).<minutes since 2020-01-01 UTC> — above every
+    # toward it, so they are published as major.(minor-1).<YYMMDDHH, UTC> — above every
     # release of the previous line, below the release main is heading for. See
     # common/scripts/nightly-version.js for the full rationale.
     #
--- a/common/scripts/nightly-version.js
+++ b/common/scripts/nightly-version.js
@@ -7,30 +7,32 @@
  * (e.g. 5.14.0-SNAPSHOT). A nightly is unreleased work heading toward that version,
  * so it is published as:
  *
- *     major.(minor - 1).<minutes since 2020-01-01 UTC>    5.14.0-SNAPSHOT -> 5.13.3458370
+ *     major.(minor - 1).<YYMMDDHH, UTC>    5.14.0-SNAPSHOT -> 5.13.26080512
+ *                                          (Aug 5 2026, 12:00 UTC)
  *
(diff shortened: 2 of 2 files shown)


common/scripts/nightly-version.js
```diff
@@ -7,30 +7,32 @@
...
- * scheme can produce from 2022 onward exceeds it, so 'vsce publish' would reject
- * every pre-release. Minutes since 2020-01-01 is 7 digits (~3.4M), three orders of
- * magnitude clear of the limit, still monotonic and still minute-granular. It stays
- * under int32 until the year 6099, which is long enough.
+ * WHY HOUR-GRANULAR AND NOT MINUTE-GRANULAR:
+ * VS Code Marketplace version components are int32, max 2147483647. A readable
+ * yymmddHHmm stamp (10 digits, e.g. 2608051230) exceeds that limit from 2022
+ * onward, so 'vsce publish' would reject every pre-release. Dropping minutes gives
+ * an 8-digit yymmddHH stamp (max ~99123123), comfortably under the limit until the
+ * year 2099. The tradeoff: two builds cut within the same UTC hour (e.g. a manual
```
NipunaRanasinghe: The collision is a fair tradeoff to take, but it is worth naming what actually happens, because "get an identical patch number" undersells it: the second build does not publish a duplicate-but-harmless version, `vsce publish` rejects the existing version and the job fails red with a duplicate-version error that does not obviously point back to this scheme.

Also, the trigger is not really the once-a-day schedule — it is a **manual re-run after a transient failure**, which is common. So the case this scheme collides on is the case someone is most likely to hit while already debugging something else.

Suggest one more sentence here, and ideally in the workflow README, saying the recovery is to wait for the next UTC hour. That turns a confusing publish error into a known behaviour.

common/scripts/nightly-version.js
```diff
@@ -47,15 +49,18 @@ const SNAPSHOT_VERSION = /^(\d+)\.(\d+)\.\d+-SNAPSHOT$/;
...
- * Whole minutes since 2020-01-01 UTC. Mirrors the shell in updateVersion/action.yml,
- * which computes the same value from `date -u +%s` — the two must agree.
+ * YYMMDDHH, UTC. Mirrors the shell in updateVersion/action.yml, which computes the
+ * same value via `date -u +'%y%m%d%H'` — the two must agree.
  */
 function timestamp(now) {
-  return String(Math.floor((now.getTime() - STAMP_EPOCH_UTC) / 60000));
+  const pad = (n) => String(n).padStart(2, '0');
+  return (
+    pad(now.getUTCFullYear() % 100) +
```
NipunaRanasinghe, line 62: `pad(2100 % 100)` is `"00"`, so the stamp becomes `00MMDDHH`, and the guard on L70 (`/^(0|[1-9]\d*)$/`) rejects any leading-zero string — `/^(0|[1-9]\d*)$/.test("00010100")` is `false`.

Failing loudly is the right direction; the value would otherwise decode to ~51208 and silently regress below every prior nightly. The only nit is attribution: the comment says the scheme is safe "until the year 2099" because of the int32 limit, but the binding constraint is this regex, not int32. A clause noting the leading-zero rejection is what stops it would save the next reader working that out.

common/scripts/nightly-version.js
```diff
@@ -64,8 +69,7 @@ function timestamp(now) {
 function deriveNightlyVersion(extensionVersion, stamp) {
   if (typeof stamp !== 'string' || !/^(0|[1-9]\d*)$/.test(stamp)) {
     throw new Error(
-      `Invalid nightly timestamp "${stamp}": expected decimal digits ` +
-      `(minutes since 2020-01-01 UTC).`
+      `Invalid nightly timestamp "${stamp}": expected decimal digits (YYMMDDHH, UTC).`
```
NipunaRanasinghe: This guard is unchanged from the old scheme, where it fit: any integer was a semantically valid minute count, and a typo produced a wrong-but-nearby value.

Under `YYMMDDHH` that is no longer true. `--timestamp 96080512` — one slipped leading digit — passes both checks, publishes `5.13.96080512`, and outranks every genuine nightly until 2096. Marketplace versions cannot be withdrawn, so the pre-release channel is poisoned permanently and the only fix is publishing something higher still. The flag now documents an exact 8-digit format, so the validation can match it:

```js
if (!/^\d{8}$/.test(stamp)) throw ...
```

Optionally also reject a stamp more than a day ahead of the current UTC hour, which catches the year-typo case specifically.

Related: the int32 check just below is now only reachable through `--timestamp`. Max `YYMMDDHH` is `99123123`, so the computed path can never trip it. Worth keeping — it is exactly the check that catches the typo above — but it no longer guards the automated path it was written for.

.github/actions/updateVersion/action.yml
```diff
@@ -53,21 +53,22 @@ runs:
...
-        # 1577836800 = 2020-01-01T00:00:00Z. Hardcoded rather than computed with
-        # 'date -d', which is GNU-only and would not run on a macOS runner.
-        timestamp=$(( ( $(date -u '+%s') - 1577836800 ) / 60 ))
+        # Hour-granular rather than minute-granular: a full 'yymmddHHmm' is 10 digits
+        # and exceeds the Marketplace's int32 component limit (2147483647) from 2022
+        # onward, so every pre-release would be rejected at publish time. Dropping
+        # minutes keeps it to 8 digits, safely under the limit. Two builds cut in the
+        # same UTC hour do collide, which is accepted since nightlies only run once a
+        # day in practice. See nightly-version.js for the decode one-liner.
+        timestamp=$(date -u +'%y%m%d%H')
```
NipunaRanasinghe, line 66: Confirmed this matches `timestamp()` in `nightly-version.js` exactly — `%y%m%d%H` and `pad(y%100)+pad(M+1)+pad(d)+pad(H)` produce identical strings for the same instant, including the zero-padding on single-digit months, days and hours.

Worth noting the change also removes the portability caveat the old comment carried: this no longer needs `date -d`, so the GNU-only concern about macOS runners is gone. `%y%m%d%H` is POSIX and behaves the same on both.

common/scripts/nightly-version.js
samithkavishke: Addressed

.github/actions/updateVersion/action.yml
samithkavishke: Addressed

common/scripts/nightly-version.js
samithkavishke: Addressed

NipunaRanasinghe (approved):

PR #806, opened 2026-08-06
Create the integration from the project chooser, skipping the wizard steps

gigara:
## Purpose

The unified Create flow asked for a project, then handed off to the three-step integration wizard (Name → Type → Configure). That is three screens to get to an empty integration, and the wizard's first step asks for a name the user has already been thinking about on the previous screen.

No linked issue.

## Goals

Collapse creating an integration into a single screen: the project chooser collects the integration name inline and creates the empty integration itself. The type/configure steps are skipped **for now** — the wizard component is untouched and the `Screen` type carries a comment describing exactly what to restore to bring it back.

Two related fixes to the same screen came out of testing it:
- Browse was overwriting the project name with the picked folder's name.
- Landing on an existing project was styled as neutral info, though it silently changes what Create does.

## Approach

All changes are in `CreateProjectChooser.tsx`.

**Single-screen create**
- Dropped the `"integration"` screen that mounted `CreateIntegrationWizard`. The wizard itself is unchanged and still reachable from `MACHINE_VIEW.BIProjectForm`, `AddProjectForm` and `AddIntegrationPanel`.
- The wizard's Name-step field now sits at the bottom of the chooser, rendered only when the starting point is "integration". **Create Integration** submits `createIntegration` with no artifact — the same payload the wizard's "Create Empty Integration" skip sent. While it is in flight the screen swaps to the same `CreatingIntegrationView`, so the hand-off across the window reload still reads as one continuous screen.
- The library route is unchanged: picking "library" hides the field and the button reverts to **Next** → the existing library form.
- Carried over the wizard's name handling so the inline field is not a downgrade — live validation, collision checks against the target project (re-listed when the project name or location retargets), and default-name indexing (`Untitled` → `Untitled_2` when taken). Create is gated on a synchronously computed validity check rather than the debounced diagnostic, so it never stays clickable for a beat after a bad edit.
- The collision is **re-checked at submit time** against a fresh listing. Because the project name and location are edited on the same screen as Create, `resolvedPath` moves while the user types and the debounced list can still describe the previous target. That window is not cosmetic: `createBIProjectPure` writes `Ballerina.toml`, `main.bal`, `connections.bal` and the rest unconditionally into the resolved path with no guard for a package already being there, so a colliding name that slips through would blank an existing integration's sources. Fails open when the path cannot be listed (a project that does not exist yet). Follows the library form's `isValidating` pattern.

**Browse no longer overwrites the project name**

The picked folder is now the parent *location*, not the project folder, so the name is appended as `<picked>/<name>` instead of replacing what the user typed. Dropped the `getExistingProjectInfo` lookup that was overwriting the name with the picked folder's `Ballerina.toml` title, and the `setDirTouched(true)` that decoupled the folder from the name afterwards — renaming after a browse now retargets correctly. Typing directly in the field is unchanged; its last segment is still the editable folder name. Existing-project detection is unaffected: it comes from `useRealtimeProjectPathValidation` watching the resolved target.

**Existing-project status reads as a warning**

Hitting an existing project is a legal target, but it silently changes what Create does, so the status strip is toned as a warning instead of neutral info — warning-tinted background, warning-colored lead, and the codicon switched `info` → `warning` with an explicit color (it would otherwise inherit the strip's muted description foreground). The "New project" state is untouched.

## UI Component Development

- [ ] Added reusable UI components to the ui-toolkit — **N/A**. No new reusable component; this rearranges an existing screen.
- [x] Use ui-toolkit components wherever possible — `TextField`, `DirectorySelector`, `Button` and `Icon` all come from `@wso2/ui-toolkit`. The two styled elements added (the form-error row and the progress slot) mirror the ones already in `LibraryCreationView`.
- [x] Matches with the native VSCode look and feel — all colors are VS Code theme variables (`--vscode-editorWarning-foreground`, `--vscode-inputValidation-warningBackground`, `--vscode-errorForeground`), with a non-`color-mix()` fallback for Chromium < 111.

## User stories

As a developer creating an integration, I give the project a name and location, pick "integration", name it, and press Create — without stepping through a wizard to reach an empty integration.

## Release note

Creating an integration from the Create flow is now a single screen: the integration is named alongside the project and created directly, instead of stepping through the Name/Type/Configure wizard. Browsing for a location no longer overwrites the project name, and targeting an existing project is now flagged as a warning.

## Documentation

N/A — no documented behavior changes. This rearranges fields within an existing screen and does not add or remove a user-facing capability; the create/library flows and their outcomes are unchanged.

## Training

N/A

## Certification

N/A — no impact on certification exams. The change is a form layout and validation change with no new concepts or APIs.

## Marketing

N/A

## Automation tests

- Unit tests
  > None added. The chooser's direct-create path (payload shape for new vs existing projects, and the rejection path restoring the form with an error) is not covered, and both reviewers asked for it.
  >
  > An earlier revision of this description claimed the package had no way to test this. That was wrong: it was written against a base that predated #790, which added `src/test/rpcHarness.tsx` and `views/TypeDiagram/index.test.tsx` — component tests rendering via `react-dom/client` + `act`, explicitly with no extra test-only dependency. This branch is now rebased onto current `main`, so that pattern is available and the test is writable as-is.
  >
  > Leaving it as a follow-up to keep this PR to the behaviour change; happy to add it here instead if reviewers would rather it not merge without one.
- Integration tests
  > None added, and none affected. Every create-flow locator in `e2e-test` ("Integration Name", "Select Path", "Next", "Create Empty Integration") targets the standalone `CreateIntegrationWizard`, which this PR does not modify. No spec references "Project name" or "Choose your starting point". All four e2e groups passed on the previous push.

## Security checks

- Ran FindSecurityBugs plugin and verified report? **N/A** — FindSecurityBugs is a Java/SpotBugs plugin; this is a TypeScript/React webview change with no Java code.
- Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? **yes**

## Samples

N/A

## Related PRs

None.

## Migrations (if applicable)

N/A

## Test environment

macOS 15 (Darwin 25.5.0), Node 16.15.1 for the isolated typecheck. Not exercised end to end in a running VS Code — see the note below on reachability. Rebased onto current `main` (was ~176 commits behind); nothing upstream touches this file and the rebase was content-identical.

## Learning

Worth flagging for reviewers: `CreateProjectChooser` has no mount point in this repo. Its only call site is `EmbeddedBIProjectForm.tsx` under `mode="create"`, and that component is exposed solely as the `ballerinaBiForm` Module Federation remote — built and shipped here, consumed by the WSO2 Integrator extension. So this change is only visible through that host; the in-repo entry points (BI Welcome card, `ballerina.open.bi.new`, `BI.project-explorer.add`) all still route to the wizard or `AddProjectForm`.

Unrelated, but it bites anyone testing this: `EmbeddedBIProjectForm`'s doc comment calls `create` "the primary mode going forward," while the prop defaults to `mode = "integration"`. A host that mounts the remote without explicitly passing `mode="create"` silently gets the old wizard.

## Summary

- Simplifies integration creation by collecting the name and creating an empty integration on the project chooser screen.
- Preserves the existing wizard through its current entry points.
- Keeps library creation unchanged.
- Adds name validation, default-name indexing, debounced feedback, and collision checks.
- Rechecks collisions during submission before creation.
- Allows listing failures for nonexistent target projects.
- Updates folder browsing to select the parent location without replacing the project name.
- Uses warning styling for existing-project targets.

### DIFF
--- a/packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
+++ b/packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
@@ -499,22 +499,27 @@ export function CreateProjectChooser({
             // clicked — and the extension scaffolds over whatever sits at the resolved path
             // (`createBIProjectPure` writes Ballerina.toml/main.bal/... unconditionally), so
             // a stale pass here would blank an existing package's sources.
-            let taken: TakenNames | null = null;
+            let taken: TakenNames;
             try {
                 taken = toTakenNames(await wsClient.getProjectComponentNames({ projectPath: resolvedPath }));
             } catch (error) {
-                // Unlistable target — in practice a project that does not exist yet, so
-                // there is nothing to collide with. Fail open, as the live check does.
+                // Deliberately NOT failing open. A missing or unreadable target is already
+                // reported as an EMPTY listing rather than a rejection — the extension
+                // swallows the readdir failure and `readBallerinaProject` is existsSync-
+                // guarded — so a rejection here is a genuine failure (transport, or an
+                // unexpected throw), never "this project does not exist yet". Continuing
+                // would scaffold over whatever is at the path without having verified it,
+                // which is the exact outcome this whole block exists to prevent.
                 console.error("Failed to re-check existing component names:", error);
+                setCreateError("Could not verify the contents of the target project. Please try again.");
+                return;
             }
-            if (taken) {
-                takenNamesPathRef.current = resolvedPath;
-                setTakenNames(taken);
-                const collision = resolveNameCollisionMessage(effectiveIntegrationName, taken, sanitizePackageName);
-                if (collision) {
-                    setIntegrationNameError(collision);
-                    return;
-                }
+            takenNamesPathRef.current = resolvedPath;
+            setTakenNames(taken);
+            const collision = resolveNameCollisionMessage(effectiveIntegrationName, taken, sanitizePackageName);
+            if (collision) {
+                setIntegrationNameError(collision);
+                return;
             }
 
             setIsCreating(true);
@@ -531,7 +536,11 @@ export function CreateProjectChooser({
         } catch (error) {
             console.error("Failed to create the integration:", error);
             setIsCreating(false);
-            setCreateError(error instanceof Error ? error.message : "Failed to create the integration");
+            // Lead with the operation, append the reason only when there is one — a bare
+            // transport/filesystem message leaves the user to infer what failed. The raw
+            // error is on the console above, which is where the detail is useful.
+            const reason = error instanceof Error ? error.message : "";
+            setCreateError(`Failed to create the integration.${reason ? ` ${reason}` : ""}`);
         } finally {
             setIsValidating(false);
         }
--- a/packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
+++ b/packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
@@ -216,6 +216,8 @@ export function CreateProjectChooser({
     const [integrationNameError, setIntegrationNameError] = useState<string | null>(null);
     // Folders/titles already used in the target project, for live collision flagging.
     const [takenNames, setTakenNames] = useState<TakenNames>(emptyTakenNames());
+    // Submit-time re-check of the collision list, before anything is created.
+    const [isValidating, setIsValidating] = useState(false);
     const [isCreating, setIsCreating] = useState(false);
     const [createError, setCreateError] = useState<string | null>(null);
 
@@ -400,7 +402,7 @@ export function CreateProjectChooser({
         }
         debouncedSetIntegrationNameError(integrationNameIssue);
         return () => debouncedSetIntegrationNameError.cancel();
-    }, [integrationNameIssue]);
+    }, [debouncedSetIntegrationNameError, integrationNameIssue]);
 
     useRealtimeProjectPathValidation({
         wsClient,
@@ -487,10 +489,35 @@ export function CreateProjectChooser({
      * the creating state until it is torn down.
      */
     const handleCreateIntegration = async () => {
-        if (!canCreateIntegration || workspaceSupportPending || isCreating) return;
+        if (!canCreateIntegration || workspaceSupportPending || isValidating || isCreating) return;
         setCreateError(null);
-        setIsCreating(true);
+        setIsValidating(true);
         try {
+            // Re-check the collision against a FRESH listing rather than trusting the live
+            // one. The project name and location are edited on this same screen, so the
+            // debounced `takenNames` can still describe the previous target when Create is
+            // clicked — and the extension scaffolds over whatever sits at the resolved path
+            // (`createBIProjectPure` writes Ballerina.toml/main.bal/... unconditionally), so
+            // a stale pass here would blank an existing package's sources.
+            let taken: TakenNames | null = null;
+            try {
+                taken = toTakenNames(await wsClient.getProjectComponentNames({ projectPath: resolvedPath }));
+            } catch (error) {
+                // Unlistable target — in practice a project that does not exist yet, so
+                // there is nothing to collide with. Fail open, as the live check does.
+                console.error("Failed to re-check existing component names:", error);
+            }
+            if (taken) {
+                takenNamesPathRef.current = resolvedPath;
+                setTakenNames(taken);
+                const collision = resolveNameCollisionMessage(effectiveIntegrationName, taken, sanitizePackageName);
+                if (collision) {
+                    setIntegrationNameError(collision);
+                    return;
+                }
+            }
+
+            setIsCreating(true);
             await biWsClient.createIntegration({
                 project: {
                     integrationName: effectiveIntegrationName,
@@ -505,6 +532,8 @@ export function CreateProjectChooser({
             console.error("Failed to create the integration:", error);
             setIsCreating(false);
             setCreateError(error instanceof Error ? error.message : "Failed to create the integration");
+        } finally {
+            setIsValidating(false);
         }
     };
 
@@ -642,12 +671,13 @@ export function CreateProjectChooser({
                         disabled={
                             ballerinaUnavailable ||
                             workspaceSupportPending ||
+                            isValidating ||
                             (isLibrary ? !canProceed : !canCreateIntegration)
                         }
                         onClick={isLibrary ? handleNext : handleCreateIntegration}
                         appearance="primary"
                     >
-                        {isLibrary ? "Next" : "Create Integration"}
+                        {isLibrary ? "Next" : isValidating ? "Validating..." : "Create Integration"}
                     </Button>
                 </span>
             </FormFooter>
--- a/packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
+++ b/packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
@@ -25,20 +25,27 @@ import {
     joinPath,
     splitPath,
     sanitizePackageName,
+    validateComponentName,
     validateProjectName,
 } from "./utils";
 import { useRealtimeProjectPathValidation } from "./useRealtimeProjectPathValidation";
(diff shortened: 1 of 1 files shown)


gigara:
Thanks — addressed in b6d5984, plus a note on the test one.

**Collision race (nitpick 1) — fixed, but differently.** This turned out not to be trivial. `createBIProjectPure` writes `Ballerina.toml`, `main.bal`, `connections.bal` and the rest unconditionally into the resolved path, with no guard for a package already being there, so a colliding name that slips through blanks an existing integration's sources rather than failing.

I didn't take the proposed `hasCurrentTakenNames` gate though. `resolvedPath` changes on every keystroke in the project-name field, so gating the button on the lookup having settled would disable Create for 300 ms after each keystroke — and if `getProjectComponentNames` ever hangs, `takenNamesPathRef` is only assigned after the await, so Create would stay disabled indefinitely.

Instead the collision is re-checked at submit time against a fresh listing, before anything is created, failing open when the path can't be listed (a project that doesn't exist yet). No latency while typing, no way to wedge the button, and it closes the window completely rather than narrowing it. Follows the `isValidating` pattern already in `LibraryCreationView`.

**Dependency array (nitpick 2) — done.**

**L1 contract test (major) — not in this PR, with a reason.** There's no chooser test setup to extend: `ballerina-visualizer` has no React component tests at all and doesn't depend on `@testing-library/react`, only `@testing-library/jest-dom` (matchers, no renderer) via `@wso2/test-config`. All 14 test files in the package are pure-logic unit tests over `.ts` modules. Adding this test means standing up component testing in the package — a reasonable thing to do, but a bigger change than this PR, and it'd be the first of its kind here. Flagged under Automation tests in the description; happy to do it as a follow-up if reviewers want it blocking.

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
```diff
@@ -335,23 +468,86 @@ export function CreateProjectChooser({
...
+            // a stale pass here would blank an existing package's sources.
+            let taken: TakenNames | null = null;
+            try {
+                taken = toTakenNames(await wsClient.getProjectComponentNames({ projectPath: resolvedPath }));
+            } catch (error) {
+                // Unlistable target — in practice a project that does not exist yet, so
+                // there is nothing to collide with. Fail open, as the live check does.
+                console.error("Failed to re-check existing component names:", error);
+            }
+            if (taken) {
```
NipunaRanasinghe: The submit-time re-check is the right call, and the reasoning above it is convincing — the debounced list really can describe the previous target, and `createBIProjectPure` really does scaffold unconditionally. My concern is that the guard switches itself off in the one case it exists for.

A throw here doesn't mean "the path isn't a project yet". A WS transport failure, a permissions error, or any extension-side exception raised against an *existing* project produces the same `null`, `if (taken)` is skipped, and `createIntegration` proceeds to write `Ballerina.toml` / `main.bal` / `connections.bal` over whatever is there. So the destructive scenario this block was written to prevent is exactly the scenario in which it's bypassed — and the failure is silent, since the only trace is a `console.error` in a webview.

The component already holds the discriminator: `existingWorkspace` comes from `useRealtimeProjectPathValidation` and is used a few lines down for `projectContext.isNewProject`. Failing closed only on the dangerous combination keeps the fail-open convenience for genuinely new projects:

```ts
if (!taken && existingWorkspace) {
    setCreateError("Could not verify the contents of the target project. Please try again.");
    return;
}
```

A new project still lists as unlistable and still sails through; an existing project that can't be read stops instead of being scaffolded over. Given the consequence is overwriting a user's sources with no undo, I'd rather refuse and ask them to retry than guess.

Worth saying: if `getProjectComponentNames` can distinguish "no such path" from a genuine failure, discriminating on that directly would be better still — `existingWorkspace` is just the signal that's already in hand.

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
```diff
@@ -335,23 +468,86 @@ export function CreateProjectChooser({
...
+                    projectPath: resolvedPath,
+                    directoryName: integrationPackageName,
+                    newProject: projectContext.isNewProject,
+                    workspaceName: projectContext.workspaceName,
+                },
+            });
+        } catch (error) {
+            console.error("Failed to create the integration:", error);
+            setIsCreating(false);
+            setCreateError(error instanceof Error ? error.message : "Failed to create the integration");
```
NipunaRanasinghe: When the error carries a `message`, this shows it alone — so the `FormError` strip can render a bare RPC or filesystem string with nothing saying which operation produced it. The fallback text naming the action only appears for the rarer non-`Error` throw, which is the opposite of when the context is most useful.

The wizard this screen replaces prefixes it (`CreateIntegrationWizard/index.tsx:554`):

```ts
setSubmitError(`Failed to create the integration: ${message}`);
```

Same shape as the fix that just landed in #790 for the type editors — lead with what failed, append the reason when there is one:

```ts
const reason = error instanceof Error ? error.message : "";
setCreateError(`Failed to create the integration.${reason ? ` ${reason}` : ""}`);
```

Keeps the raw detail (it's already on the `console.error` above, which is where it's genuinely useful) without making the user infer the operation from a transport message.

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
gigara: You're right, and it's worse than the conditional fix covers — so I took the stronger option you pointed at in your last paragraph.

`getProjectComponentNames` **can** already distinguish the two cases, implicitly: a missing or unreadable target comes back as an *empty listing*, not a rejection. `bi.ts:251-258` swallows the `readdir` failure ("Project directory doesn't exist yet — nothing taken"), `readBallerinaProject` is `existsSync`-guarded with its own catch, and the per-package toml reads each catch to `undefined`. So nothing about a nonexistent path produces a throw.

Which means the premise of my comment was just wrong. Every rejection that reaches the webview is a genuine failure — transport, unregistered handler, unexpected throw — and none of them mean "no project yet". So there was never a fail-open case to preserve, and it now fails closed unconditionally:

```ts
} catch (error) {
    console.error("Failed to re-check existing component names:", error);
    setCreateError("Could not verify the contents of the target project. Please try again.");
    return;
}
```

That also drops the `TakenNames | null` and the `if (taken)` wrapper, so the collision check is no longer conditional on anything.

I went this way rather than `!taken && existingWorkspace` because `existingWorkspace` has its own settling delay — it comes from the same debounced hook — so a failure arriving while it was still `false` for a project that does in fact exist would have sailed through the conditional too. Unconditional has no such window.

ac208b3.

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
gigara: Agreed — fixed in ac208b3, using your shape:

```ts
const reason = error instanceof Error ? error.message : "";
setCreateError(`Failed to create the integration.${reason ? ` ${reason}` : ""}`);
```

For what it's worth on provenance: I lifted the inverted version from `LibraryCreationView.tsx:447`, which does the same thing (`error instanceof Error ? error.message : "Failed to create the library"`). So that one has the same problem — I've left it alone to keep this PR scoped, but it's worth a follow-up if #790 set the direction for the codebase.

gigara:
Correction to [my earlier reply](https://github.com/ballerina-platform/ballerina-vscode/pull/806#issuecomment-5203167438) on the contract test.

I said there was no chooser test setup to extend and that adding one would mean standing up component testing in `ballerina-visualizer`. That was wrong. It was true of the base this branch was cut from, but that base was ~176 commits behind — and #790 has since added exactly that: `src/test/rpcHarness.tsx` and `views/TypeDiagram/index.test.tsx`, rendering with `react-dom/client` + `act`, with the harness explicitly noting it needs no extra test-only dependency. CI also runs a `Fast tests (unit + component + contract)` job.

So the objection doesn't hold: the test is writable today, no new dependency, following an established in-package pattern.

I've rebased onto current `main` (clean — nothing upstream touches this file, and the rebased content is byte-identical). I've left the test itself as a follow-up to keep this PR to the behaviour change, and corrected the Automation tests section of the description to say so rather than repeat the stale reasoning. If you'd rather this didn't merge without the test, say so and I'll add it here.

NipunaRanasinghe (approved):
LGTM

PR #809, opened 2026-08-07
Revamp `Use Connection` agent tool creation UX

dan-niles:
## Summary

Resolves https://github.com/wso2/product-integrator/issues/2152

Revamps the `Use Connection` agent tool creation flow, so users can discover connector capabilities before creating a connection.

https://github.com/user-attachments/assets/dae11089-d29a-4e85-8489-801f0ecf1cf7

- Replace the connection-first flow with connector → action → tool form.
- Add a browsable 514-connector catalogue, category navigation, sticky headers, curated Popular entries, and instant local filtering.
- Search Ballerina Central for connectors outside the shipped catalogue, without duplicating catalogue results.
- Discover remote and resource actions from Central documentation, including OpenAPI-generated resource actions.
- Prefill and organize tool forms: suggested names/descriptions, type-filtered connection selection, collapsible Inputs, OAuth, and Result Type sections.
- Improve related action labels, connection-select behavior, form loading, navigation, and deletion behavior.
- Note two intentional product-wide relabels, not scoped to the new browser: `formatMethodName` now preserves acronyms (`getUserId` → "Get User ID"), which also affects its existing `GroupList` caller, and the expanded `connector_name_correction.json` renames connectors across all search results (13 golden files updated).

## Summary

- Replaced the connection-first tool flow with connector and action discovery.
- Added a categorized connector catalogue with search, popular entries, existing connections, and Central-backed results.
- Added remote and resource action discovery with filtering, deduplication, and improved labels.
- Enhanced tool forms with suggested names, typed connection selectors, collapsible groups, OAuth support, result-type handling, and validation indicators.
- Improved agent tool return-type handling and agent-call metadata.
- Improved loading, error, navigation, deletion, and connection-selection behavior.
- Added shared acronym-preserving method formatting and expanded connector-name mappings.
- Added connector categorization, metadata indexing, and related unit and integration tests.
- Improved library request error handling and connector browser styling.


dan-niles:
@coderabbit review

packages/ballerina-visualizer/src/components/ConnectionSelector/useCreateNode.tsx
```diff
@@ -103,7 +105,7 @@ export function useCreateNode(
         try {
             const flowNode = await fetchTemplate();
             addModal(renderCreator(flowNode, () => closeModal(modalId)), modalId, title, 600, 520);
-        } catch {}
+        } catch { }
```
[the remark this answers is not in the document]
dan-niles: Fixed in https://github.com/dan-niles/ballerina-vscode/commit/8fdda16fe93ff4f84ef0309e1e46c258de975f6b — the catch block now logs the error and shows a user-facing message via `rpcClient.getCommonRpcClient().showErrorMessage(...)`, matching the pattern used elsewhere in this module.

packages/ballerina-visualizer/src/views/BI/Connection/ConnectorBrowser/ConnectorList.tsx
```diff
@@ -0,0 +1,526 @@
...
+        background-clip: content-box;
+    }
+    &:hover::-webkit-scrollbar-thumb {
+        background: ${ThemeColors.OUTLINE_VARIANT};
+        background-clip: content-box;
+    }
+`;
+
+const Section = styled.div`
+    background-color: rgba(255, 255, 255, 0.02);
```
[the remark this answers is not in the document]
dan-niles: Fixed in https://github.com/dan-niles/ballerina-vscode/commit/8fdda16fe93ff4f84ef0309e1e46c258de975f6b — both `Section` and `SectionHeader` now use `ThemeColors.SURFACE_DIM` instead of the hardcoded rgba() wash, matching the convention already used by the sibling `ConnectorActionList.tsx` in this same PR.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/utils.ts
```diff
@@ -41,6 +43,63 @@ const OPENAI_PROVIDER_CODEDATA: CodeData = {
...
+    const camel = toCamelCase(name);
+    // Strip known suffixes to avoid e.g. "salesAgentAgent"
+    const lower = camel.toLowerCase();
+    for (const suffix of KNOWN_AGENT_NAME_SUFFIXES) {
+        if (lower.endsWith(suffix) && lower.length > suffix.length) {
+            return camel.slice(0, -suffix.length);
+        }
+    }
+    return camel;
+}
```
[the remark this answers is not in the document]
dan-niles: Not fixing this one — `toBaseName` (and its only caller, `createBuiltInAgent`) has no callers anywhere in the codebase right now, so the exact-match "Agent"/"Model" edge case is unreachable in practice. It's also not a regression: the pre-PR implementation (`toCamelCase(name).replace(/^(.+?)(?:agent|model)$/i, "$1")`) had the same blind spot on an exact match, since the non-capturing group requires at least one preceding character. Happy to revisit with the `>=` fix and a test once this function actually gets wired up to a UI entry point.

packages/ballerina-visualizer/src/views/BI/Connection/ConnectorBrowser/ConnectorBrowser.tsx
```diff
@@ -0,0 +1,305 @@
...
+        }, 300);
+    };
+
+    useEffect(() => () => {
+        if (searchDebounce.current) {
+            clearTimeout(searchDebounce.current);
+        }
+    }, []);
+
+    const handleSelectConnection = (connectionName: string, connectionActions: PanelNode[]) => {
```
gigara, line 186: **P0 — the connection the user picked is discarded.**

`connectionName` is used only as a display label (line 198); it is never stored. The action-list callback below (line 297) then fires `onSelect({ action, connector: selectedConnector })` with no `connectionName`, so `loadConnectionCallFields` runs with `options.connectionName === undefined` and builds the connection select with `value: ""`.

Flow that breaks: pick **myHttpClient** -> pick an action -> land on a form whose Connection field is empty. And because `NodeReferenceSelectEditor` now auto-selects the first item when the field has no value, it silently fills in whichever type-compatible connection resolves first — which may not be the one that was clicked. The generated tool then binds to the wrong connection with no error shown.

Suggest holding the name in state next to `selectedConnector` and threading it through. Belt-and-braces in `loadConnectionCallFields`, since the LS template may already carry it:

```ts
options.connectionName ?? String(connectionProperty?.value ?? "")
```

Worth a second look at the new auto-select-first in `NodeReferenceSelectEditor` too — the out-of-order guard there is correct, but silently writing a value the user never chose is what turns this bug from "empty field" into "wrong code".

gigara (commented):
Review focused on logic, reuse, and cross-cutting risk. The flow itself reads well and the "why" comments are genuinely useful — the `acceptTypeImports` null guard, "hidden, not unmounted" on the group body, and the `isInferredTypedescReturn` rationale all saved me a round trip. The two new test files are thorough; the `resourceToolNameSeed` edge-case table in particular.

Comments are tagged **P0** / **P1** / **P3**:

- **P0 (4)** — correctness in the new flow. The first one matters most: picking an existing connection discards it, and combined with the new auto-select-first behaviour the tool can silently bind to the *wrong* connection with no error surfaced. That one plus the `codedata.object` fallback are both small fixes.
- **P1 (6)** — changes that reach outside this feature. Three shared `Form`/`ExpressionEditor` behaviour changes now affect every form in the product while one caller benefits; `formatMethodName` and `connector_name_correction.json` relabel things product-wide. None is wrong, but they are the highest-regression-risk part of the diff and read like good candidates for prop-gating or separate PRs.
- **P3 (12)** — cleanup, all cheap.

Two things I checked and found clean, for the record: `connector_name_correction.json` has no duplicate keys and re-adds all 21 entries the rewrite removed; and the out-of-order guard in `NodeReferenceSelectEditor.fetchItems` is correctly placed at the top of `.then`.

Not covered here, happy to add if useful: a handful of lower-priority items (Java `catch (Throwable)`, trailing whitespace that may trip checkstyle, an uncached 2000-row connector sweep per landing request, the `jest` `moduleNameMapper` that aliases a whole package to one file) and some copy nits (`Search 1 actions`, resource-action labels rendering as sentences with trailing periods, a no-actions message that offers an action the screen has no button for).

packages/ballerina-visualizer/src/views/BI/Connection/ConnectorBrowser/connectorActions.ts
```diff
@@ -0,0 +1,277 @@
...
+    // Fall back to any module declaring the client: dotted modules (aws.s3) come as one.
+    const clientName = codedata.object || "Client";
+    const clients: DocsClient[] = modules
+        .filter((module) => !codedata.module || !module.id || module.id === codedata.module)
+        .flatMap((module) => (module.clients ?? []) as DocsClient[]);
+    const allClients: DocsClient[] = clients.length
+        ? clients
+        : modules.flatMap((module) => (module.clients ?? []) as DocsClient[]);
+
+    const client = allClients.find((candidate) => candidate.name === clientName) ?? allClients[0];
```
gigara, line 122: **P0 — `codedata.object` can name a class that doesn't own these methods.**

When `find` misses and the `?? allClients[0]` fallback kicks in, `client` and `clientName` are different classes — but line 183 emits `object: clientName`, i.e. the *requested* name rather than the resolved one. Every action then carries codedata pointing at a class that doesn't declare it, and `getNodeTemplate` fails downstream. The user sees a populated action list and a dead end on selection.

```ts
object: client.name ?? clientName,
```

This is exactly the multi-client case the new `ConnectorUtil` exact-key lookup is adding support for, so it is reachable.

packages/ballerina-visualizer/src/views/BI/Connection/ConnectorBrowser/connectorActions.ts
```diff
@@ -0,0 +1,277 @@
...
+        orgName: codedata.org,
+        moduleName: codedata.module,
+        version: codedata.version,
+    });
+
+    const modules = response?.docsData?.modules ?? [];
+    // Fall back to any module declaring the client: dotted modules (aws.s3) come as one.
+    const clientName = codedata.object || "Client";
+    const clients: DocsClient[] = modules
+        .filter((module) => !codedata.module || !module.id || module.id === codedata.module)
```
gigara: Minor, same function: `!codedata.module ||` is unreachable — `codedata.module` is guarded non-empty by the throw above. The condition reduces to `!module.id || module.id === codedata.module`.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/toolForm.ts
```diff
@@ -0,0 +1,168 @@
...
+        documentation: "Adds ai:Context ctx as the first parameter so this tool can access the "
+            + "invoking agent's context.",
+        optional: true,
+        editable: true,
+        enabled: true,
+        hidden: false,
+        advanced: false,
+        value: false,
+        types: [{ fieldType: "FLAG", selected: true }],
+        ...(group ? { group } : {}),
```
gigara, line 41: **P0 (hardening) — a collapsed card can hide the error that is blocking Save.**

`buildToolFormGroups` correctly expands Inputs when a required field is empty *at build time*, which covers the blanked-SQL case the comment describes. It does not cover a field that becomes invalid later — the user clears it, or an LS diagnostic lands on it.

In `Form`, `expandedGroups` only ever reads `defaultCollapsed` (index.tsx:1547); nothing reconciles it against `errors`. So a validation failure inside a collapsed group leaves Save unclickable with no visible reason anywhere on screen.

Cheapest fix is in `Form`: on submit failure, expand any group that contains a key present in `errors`.

packages/ballerina-side-panel/src/components/Form/index.tsx
```diff
@@ -537,7 +636,37 @@ export const Form = forwardRef((props: FormProps, _ref) => {
...
+            // busy -> idle: done
+            setInitialLoadSettled(true);
+            return;
+        }
+        // Nothing registered; settle next tick so editors mounting this commit still count.
+        const timer = setTimeout(() => setInitialLoadSettled(true), 0);
+        return () => clearTimeout(timer);
+    }, [loadingFields.size, initialLoadSettled]);
+
+    const isFormLoading =
```
gigara: **P1 — this changes loading behaviour for every form in the product.**

`isFormLoading` is now latched off once `initialLoadSettled` flips, so any *later* async field resolution no longer shows a loading state — anywhere, not just on the tool form. The comment above explains why the gate is needed for prefilled editors in collapsible cards, and that reasoning is sound, but the blast radius is every `Form` consumer while exactly one caller benefits today.

Suggest gating on an opt-in prop (`gateLoadingToInitialResolve` or similar) so existing forms keep their current behaviour.

packages/ballerina-side-panel/src/components/Form/index.tsx
```diff
@@ -537,7 +636,37 @@ export const Form = forwardRef((props: FormProps, _ref) => {
...
         });
     }, []);
-    const isFormLoading = loadingFields.size > 0;
+    // Initial resolution only. Prefilled editors register on mount, so without this gate
+    // any later mount (expanding a card) blanks the form and drops focus.
+    const [initialLoadSettled, setInitialLoadSettled] = useState(false);
+    // Editors register one commit after first paint, so assume busy up front when there is
+    // something to resolve. Otherwise the form paints, blanks, then paints again.
+    const expectsInitialLoadRef = useRef<boolean | null>(null);
+    if (expectsInitialLoadRef.current === null) {
```
gigara, line 632: **P1 — side effect during render.**

Line 645-649 is the sanctioned lazy-ref-init pattern, fine. Line 651-653 is not: it mutates a ref in the render phase on every render where `loadingFields` is non-empty. Not safe under StrictMode double-render or concurrent rendering.

Since there is already an effect keyed on `loadingFields.size` directly below, the `sawInitialLoading` transition can move into it.

Related fragility: `expectsInitialLoadRef` snapshots the *first* render's `formFields`. Hosts that populate fields asynchronously after mount (which `AIAgentSidePanel` does — `INITIAL_FIELDS` then `setFields`) will read `false` here and get the paint/blank/paint flicker this gate was added to remove.

packages/ballerina-side-panel/src/components/Form/index.tsx
```diff
@@ -902,6 +1037,12 @@ export const Form = forwardRef((props: FormProps, _ref) => {
         (field) => field.editable !== false && getPrimaryInputType(field.types)?.fieldType === "IDENTIFIER"
     );
 
+    // Autofocus selects the text, so skip it when prefilled.
+    const firstEditableFieldValue = formFields[firstEditableFieldIndex]?.value;
+    const autoFocusFirstField =
```
gigara, line 1027: **P1 — also global.** Every form whose first editable field arrives prefilled now loses autofocus, not just the tool form. Reasonable behaviour, but it should probably ride the same opt-in prop as the loading gate rather than change silently for all callers.

packages/ballerina-side-panel/src/components/editors/ExpressionEditor.tsx
```diff
@@ -139,8 +139,7 @@ export namespace S {
     export const HeaderContainer = styled.div({
         display: 'flex',
         alignItems: 'flex-end',
-        justifyContent: 'space-between',
-        minHeight: '26px'
+        justifyContent: 'space-between'
```
gigara: **P1 — global.** Dropping `minHeight: '26px'` changes the vertical rhythm of the header for every expression editor in the product. If this is here to tighten up the new grouped cards, it would be safer as a variant than as a change to the shared default.

packages/ballerina-side-panel/src/utils/formatMethodName.ts
```diff
@@ -16,6 +16,15 @@
  * under the License.
  */
 
+const ACRONYMS = new Set([
```
gigara, line 19: **P1 — silent product-wide relabel.**

`GroupList/index.tsx:342` is an existing caller, so this list changes labels well outside the connector browser: `getUserId` -> "Get User **ID**", `deleteAcl` -> "Delete **ACL**", `sendSms` -> "Send **SMS**".

Mostly an improvement, but (a) it is worth calling out in the PR description as an intentional global change, and (b) there is no test for it in the owning package — the new coverage lives in `ballerina-visualizer` and only reaches this file through a `moduleNameMapper` that aliases the whole `@wso2/ballerina-side-panel` package to this single file. A test alongside the source would be more durable.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/main/resources/connector_name_correction.json
```diff
@@ -1,29 +1,436 @@
...
-  "grpc": "gRPC",
-  "mysql": "MySQL",
-  "mssql": "MS SQL",
-  "mongodb": "MongoDB",
-  "aws.dynamodb": "DynamoDB",
-  "oracledb": "OracleDB",
-  "postgresql": "PostgreSQL",
-  "sql": "SQL",
+  "aayu.mftg.as2": "Aayu Mftg AS2",
+  "activecampaign": "ActiveCampaign",
```
gigara, line 3: **P1 — suggest splitting this out into its own PR.**

428 entries here relabel connectors **product-wide**, not only in the new browser — the 13 updated `search/config/**` golden files are the visible fallout. That is a meaningful UX change that deserves its own review pass rather than being read inside a 4.5k-line diff.

I did sanity-check it: no duplicate keys, and all 21 entries removed by the rewrite are re-added, so nothing is dropped. The lowercase values (`figshare`, `logoraisr`, `commercetools *`) look like deliberate brand casing.

Two that read oddly and may want a second opinion:
- `commercetools.cartsordershoppinglists` -> "commercetools Carts" (drops Orders and ShoppingLists)
- `commercetools.configuration` / `customizebehavior` / `customizedata` -> "Config" / "Behavior" / "Data", which are vague on their own in a flat list.

packages/ballerina-side-panel/src/components/Form/index.tsx
```diff
@@ -143,6 +143,94 @@ namespace S {
...
+        font-family: var(--vscode-font-family);
+
+        &:focus-visible {
+            outline: 1px solid ${ThemeColors.PRIMARY};
+            outline-offset: -1px;
+            border-radius: 4px;
+        }
+    `;
+
+    export const GroupHeaderText = styled.div`
```
gigara: P3 — `GroupHeaderText` and `GroupDescription` are declared but never rendered, and `FieldGroup` has no `description` field to feed them. Either wire up a group description or drop both.

packages/ballerina-side-panel/src/components/Form/index.tsx
```diff
@@ -143,6 +143,94 @@ namespace S {
         border-top: ${({ topBorder }) => (topBorder ? `1px solid ${ThemeColors.OUTLINE_VARIANT}` : "none")};
     `;
 
+    /** Collapsible group card: a labelled container for fields the user rarely edits. */
+    export const GroupCard = styled.div<{ expanded?: boolean }>`
```
gigara: P3 — `GroupCard` and `GroupHeader` both take an `expanded` prop that neither template body uses (the chevron handles the rotation). Safe to drop from both signatures.

packages/ballerina-side-panel/src/components/Form/index.tsx
```diff
@@ -143,6 +143,94 @@ namespace S {
...
+    `;
+
+    export const GroupBody = styled.div`
+        display: flex;
+        flex-direction: column;
+        gap: 20px;
+        padding: 4px 14px 16px;
+        border-top: 1px solid ${ThemeColors.OUTLINE_VARIANT};
+        margin-top: -1px;
+        padding-top: 16px;
```
gigara: P3 — `GroupBody` sets `padding: 4px 14px 16px` on line 228 and then overrides the top with `padding-top: 16px` here. Collapse into one declaration.

packages/ballerina-visualizer/src/views/BI/Connection/ConnectorBrowser/ConnectorList.tsx
```diff
@@ -0,0 +1,525 @@
...
+ */
+
+import { useEffect, useMemo, useRef, useState } from "react";
+import styled from "@emotion/styled";
+import { Category as PanelCategory, Node as PanelNode } from "@wso2/ballerina-side-panel";
+import { Codicon, ProgressRing, SearchBox, ThemeColors, Typography } from "@wso2/ui-toolkit";
+
+const POPULAR_CATEGORY = "Popular";
+
+const Container = styled.div`
```
gigara: P3 — reuse: `Container`, `HeaderArea`, `ScrollArea`, `Row`, `RowIcon`, `RowText`, `RowLabel`, `RowDescription`, `RowChevron` and `EmptyState` are near-identical to the set in `ConnectorActionList.tsx` — roughly 200 duplicated lines. `height: calc(100vh - 56px)` is hardcoded in three places (both lists plus `ConnectorBrowser`'s `LoaderWrapper`).

A shared `ConnectorBrowser/styles.ts` would cover both, with the magic 56px as a named constant.

Broader question for the PR description: these two components reimplement search + categories + rows, which `NodeList` already does — and this PR *does* extend `NodeList` with `description` and `alwaysExpandedCategories`. Worth a line on why `NodeList` couldn't carry the wizard, otherwise there are now two list implementations to keep in sync.

packages/ballerina-visualizer/src/views/BI/Connection/ConnectorBrowser/ConnectorList.tsx
```diff
@@ -0,0 +1,525 @@
...
+    gap: 8px;
+    align-items: center;
+    padding: 12px;
+    border: 1px solid ${ThemeColors.OUTLINE_VARIANT};
+    border-top: none;
+    border-radius: 0 0 5px 5px;
+    color: ${ThemeColors.ON_SURFACE_VARIANT};
+`;
+
+interface Section {
```
gigara: P3 — `interface Section` shadows the `Section` styled component declared at line 137. Legal (one is type-only) but confusing in a 525-line file. `ListSection` would read better.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AIAgentSidePanel.tsx
```diff
@@ -1265,14 +1394,17 @@ export function AIAgentSidePanel(props: BIFlowDiagramProps) {
...
-            includeContext,
+            includeContext: data[INCLUDE_CONTEXT_KEY] === true,
             selectedCodeData: selectedNodeCodeData,
             toolParameters: toolParameters,
             functionNode: clonedFunctionNode,
             flowNode: clonedFlowNode,
             parameterImports: paramImports,
+            connectionName: chosenConnection || undefined,
         };
         console.log("New Agent Tool:", toolModel);
```
gigara: P3 — leftover debug logging on the submit path. Pre-existing, but this PR touches the surrounding lines, so it is a good moment to drop it (the whole tool model, including OAuth config, goes to the console today). There are also two stray blank lines immediately below.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/NewTool.tsx
```diff
@@ -149,12 +164,13 @@ export function NewTool(props: NewToolProps): JSX.Element {
                 return;
             }
             flowNode = data.flowNode;
-            connection = data.selectedCodeData.parentSymbol || "";
+            connection = data.connectionName || data.selectedCodeData.parentSymbol || "";
         }
 
         setSavingForm(true);
 
         try {
```
gigara, line 161: P3 — stray blank line immediately after `try {`.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AddMcpServer.tsx
```diff
@@ -578,20 +561,13 @@ export function AddMcpServer(props: AddMcpServerProps): JSX.Element {
...
-                        />
-                        <div>
-                            Pass agent context
-                            <ContextHint>
-                                Adds ai:Context ctx as the first parameter so this tool can access the invoking
-                                agent's context.
-                            </ContextHint>
-                        </div>
-                    </ContextOption>
+                    <RequiresAuthCheckbox
```
gigara, line 564: P3 — good reuse, but the name no longer fits: `RequiresAuthCheckbox` is doing duty as a generic labelled checkbox for "Pass agent context". Worth renaming to something neutral (`LabeledCheckbox`) and lifting it out of `./Mcp/`.

Also, the description string here is a near-copy of the one in `buildIncludeContextField` (`toolForm.ts:33`) and has drifted — no trailing period, "the generated tools" vs "this tool". Reusing that field's `documentation` would keep them in sync.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/UseAgentToolForm.tsx
```diff
@@ -73,27 +59,86 @@ export function UseAgentToolForm(props: UseAgentToolFormProps): JSX.Element {
...
+            .map(({ key, property }) => ({
+                key,
+                property,
+                recordTypeMembers: getPrimaryInputType(property?.types)?.typeMembers
+                    .filter((member) => member.kind === "RECORD_TYPE"),
+            })),
+        [oauthProperties]
     );
 
+    const fields: FormField[] = [
```
gigara: P3 — `groups` and `recordTypeFields` are memoized just above, but `fields` is rebuilt inline on every render. That hands `ArtifactForm` a fresh array identity each pass, which is the thing the neighbouring memos are there to avoid.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/UseAgentToolForm.tsx
```diff
@@ -73,27 +59,86 @@ export function UseAgentToolForm(props: UseAgentToolFormProps): JSX.Element {
...
+                : (await rpcClient.getVisualizerRpcClient().joinProjectPath({
+                    segments: [agentNode?.codedata?.lineRange?.fileName ?? "agents.bal"],
+                })).filePath;
             setAgentFilePath(filePath);
+            setOauthProperties(await fetchOAuthConfigProperties(rpcClient, filePath));
+            setDefaultReturnType(await fetchAgentRunReturnType(rpcClient, filePath, agentVarName,
+                hostClass?.className));
             setReady(true);
         })();
     }, [agentNode]);
```
gigara, line 77: P3 — the effect reads `hostClass`, `agentVarName` and `rpcClient` but declares only `[agentNode]`. Worth completing the deps (or documenting why it should run once).

packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/index.tsx
```diff
@@ -572,11 +572,11 @@ export function BIFocusFlowDiagram(props: BIFocusFlowDiagramProps) {
                 setShowProgressIndicator(false);
                 showEditForm.current = false;
                 return;
-                }
+            }
```
gigara: P3 — indentation-only change, unrelated to this PR. Not a problem, just noting it as reviewer noise; easier to land separately.

packages/ballerina-visualizer/src/views/BI/Connection/ConnectorBrowser/ConnectorBrowser.tsx
dan-niles: Fixed in [2b190fb](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/2b190fbfacbc09a7941f827e9ec00111a0e60436): the name is held in state next to `selectedConnector` and threaded through the action list, plus the `connectionProperty.value` fallback you suggested — the template does carry it (`CallBuilder` sets `.value(codedata.parentSymbol())`).

packages/ballerina-visualizer/src/views/BI/Connection/ConnectorBrowser/connectorActions.ts
dan-niles: Applied in [2b190fb](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/2b190fbfacbc09a7941f827e9ec00111a0e60436) as `object: client.name ?? clientName`.

One note: it's the identical string whenever `find` succeeds, and I couldn't reach the fallback through a supported path — `codedata.object` is always `searchResult.name()` from the index and docs client names are the same class names, including for the multi-client packages the exact-key lookup adds. So I've treated it as hardening rather than a live bug.

packages/ballerina-visualizer/src/views/BI/Connection/ConnectorBrowser/connectorActions.ts
dan-niles: Removed in [2b190fb](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/2b190fbfacbc09a7941f827e9ec00111a0e60436).

packages/ballerina-side-panel/src/components/Form/index.tsx
dan-niles: Moved the latch into the effect in [2b190fb](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/2b190fbfacbc09a7941f827e9ec00111a0e60436), before the early return so it still sets.

On the second half: `expectsInitialLoadRef` reads `true` for the tool form. It's mounted only at `sidePanelView === TOOL_FORM` (`AIAgentSidePanel:1708`), and `setFields` runs before `setSidePanelView` in both `handleWizardSelect` and `handleOnSelectNode` — so the first render already sees populated fields.

packages/ballerina-side-panel/src/components/Form/index.tsx
dan-niles: Both dropped in [2b190fb](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/2b190fbfacbc09a7941f827e9ec00111a0e60436).

packages/ballerina-side-panel/src/components/Form/index.tsx
dan-niles: Dropped from both signatures in [2b190fb](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/2b190fbfacbc09a7941f827e9ec00111a0e60436) — kept `aria-expanded` and `GroupChevron`'s.

packages/ballerina-side-panel/src/components/Form/index.tsx
dan-niles: Collapsed in [2b190fb](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/2b190fbfacbc09a7941f827e9ec00111a0e60436) — to `padding: 16px 14px`, since the `padding-top` override was the value actually in effect.

packages/ballerina-visualizer/src/views/BI/Connection/ConnectorBrowser/ConnectorList.tsx
dan-niles: Renamed to `ListSection` in [2b190fb](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/2b190fbfacbc09a7941f827e9ec00111a0e60436).

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AIAgentSidePanel.tsx
dan-niles: Dropped the log and the blank lines in [2b190fb](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/2b190fbfacbc09a7941f827e9ec00111a0e60436) — the OAuth config going to the console was reason enough. Left the other nine `console.log`s in this file alone.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/NewTool.tsx
dan-niles: Removed in [2b190fb](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/2b190fbfacbc09a7941f827e9ec00111a0e60436).

packages/ballerina-visualizer/src/views/BI/AIChatAgent/UseAgentToolForm.tsx
dan-niles: Memoized in [2b190fb](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/2b190fbfacbc09a7941f827e9ec00111a0e60436).

packages/ballerina-visualizer/src/views/BI/AIChatAgent/UseAgentToolForm.tsx
dan-niles: Leaving as is: `hostClass` is an object prop, so adding it risks refetching the OAuth properties on every parent render that recreates it, and `rpcClient` is a stable context value. Keying on `agentNode` identity is intentional.

packages/ballerina-side-panel/src/components/Form/index.tsx
dan-niles: Fixed in [96f8cac](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/96f8cac02204502209e45dd9da12f5aed600a37d) — this and the autofocus change now ride an `opensPrefilled` prop, off by default, passed only by the two tool-form entry points (`UseAgentToolForm` and `AIAgentSidePanel`), both of which mount once their fields are ready. Every other consumer is back to `loadingFields.size > 0`.

packages/ballerina-side-panel/src/components/Form/index.tsx
dan-niles: Fixed in [96f8cac](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/96f8cac02204502209e45dd9da12f5aed600a37d)

packages/ballerina-side-panel/src/components/editors/ExpressionEditor.tsx
dan-niles: Reverted in [96f8cac](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/96f8cac02204502209e45dd9da12f5aed600a37d)

packages/ballerina-side-panel/src/utils/formatMethodName.ts
dan-niles: (b) fixed in [96f8cac](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/96f8cac02204502209e45dd9da12f5aed600a37d): added `ballerina-side-panel/src/test/formatMethodName.test.ts` next to the source, matching the 27 jest suites already in that directory. (a) going into the PR description.

The alias has to stay — `connectorActions.test.ts` needs `formatMethodName` transitively and the real barrel pulls in the whole editor tree — but it now carries a comment saying what widening it requires. It is currently harmless: the only other visualizer tests touching the package use `import type` (`function-category.test.ts`) or `jest.mock` (`TypeDiagram/index.test.tsx`), so neither resolves through it.

Worth noting the old implementation disagreed with two of its own docstring examples: `HTTPRequest` returned `"Http Request"` and `method2Call` returned `"Method2call"`. The rewrite is what makes those comments true.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/main/resources/connector_name_correction.json
dan-niles: Taking the `cartsordershoppinglists` one — now "commercetools Carts & Orders" in [96f8cac](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/96f8cac02204502209e45dd9da12f5aed600a37d), with the one golden file that asserted the old label updated.

On the other three: the values carry the prefix, so they render as "commercetools Config" / "commercetools Behavior" / "commercetools Data" in a flat list, not bare "Config"/"Behavior"/"Data".

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AddMcpServer.tsx
dan-niles: Added the trailing period in [96f8cac](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/96f8cac02204502209e45dd9da12f5aed600a37d). Leaving the rest.

The file is untouched by this PR — the generic `label`/`description` props with defaults already existed — and both call sites are in `AddMcpServer`, i.e. entirely within the MCP flow, so there is nothing to lift out of `./Mcp/`.

On the strings, the divergence is deliberate rather than drift: the MCP toolkit generates several tools ("the generated tools"), while `buildIncludeContextField` describes one ("this tool"). Sharing a single `documentation` would make one of them wrong.

packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/index.tsx
dan-niles: Dropped in [39ecfd6](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/39ecfd6ebb77e1f0e8979fff397ea7221548536d) — reverted to the base indentation, and since that hunk was the file's only change in this PR, `FocusFlowDiagram/index.tsx` is out of the diff entirely now. The mis-indentation it was correcting is pre-existing, so it can land on its own.

packages/ballerina-visualizer/src/views/BI/Connection/ConnectorBrowser/ConnectorList.tsx
dan-niles: Done in [d952258](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/d952258978fed1b30edd1a8d37bd9ebe9c7f76e3) — added `ConnectorBrowser/styles.ts` with `Container`, `RowText`, `RowChevron`, `EmptyState` and a `CONTENT_HEIGHT` constant, which also covers the `calc(100vh - 56px)` triplication.

Less was actually shared than it looks, though: of 20 styled components per file only 10 have the same name, and just `Container` and `RowText` were identical. `Row`, `RowIcon`, `ScrollArea`, `HeaderArea` and `RowDescription` stay per-file — the badge geometry, border strategy, custom scrollbar, header contents and description clamp all differ deliberately, so unifying them would mean a variant prop each.

The comparison did turn up real drift, now fixed: the two steps rendered row labels at different font weights (500 vs 600), plus a 1px chevron offset and differing empty-state padding.

Same story for `NodeList`: each file has 10 styled components the other doesn't, so the anatomy genuinely differs. I'll add the rationale to the PR description.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/toolForm.ts
dan-niles: Fixed in [b7297cf](https://github.com/ballerina-platform/ballerina-vscode/pull/809/commits/b7297cf7e34be30dae5270de4d3f6e8d3d6ccaac) — a collapsed group header now shows a warning icon when the card is what is blocking Save. Chose the indicator over auto-expanding: the `Advanced options` collapse has the same gap and does not auto-expand, and an effect on `errors` would pop cards open mid-keystroke while a field is transiently empty.

Two things I hit that are worth recording, since they mean the suggested fix would not have worked:

Expanding on submit failure has no hook — Save is *disabled* (`index.tsx:1159`), so `onSubmit` never fires to fail.

And keying on `errors` would not have caught the blanked-SQL case you describe. `ArtifactForm` never passes `onFormValidation`, so the form-level `hasIncompleteRequiredFields` is inert here, and the form runs in react-hook-form's default `onSubmit` mode — a missing required value shows up only in `formStateIsValid`, which is a single boolean with no field attribution, while `errors` stays empty. The predicate therefore checks emptiness directly (ungated), plus `errors` and ERROR-severity live diagnostics.

`groupHasBlockingIssue` is a pure helper in `Form/utils.ts` with 6 cases in `formValidationUtils.test.ts`. It is only reached when `groups` is passed, so forms without cards are untouched.

dan-niles:
@coderabbit review

dan-niles:
@coderabbit full review

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/NodeReferenceSelectEditor/NodeReferenceSelectEditor.tsx
```diff
@@ -115,9 +116,14 @@ export const NodeReferenceSelectEditor: React.FC<NodeReferenceSelectEditorProps>
...
-            setSelectItems(ensureValueInItems(
+            const resolved = ensureValueInItems(
                 applyNodeReferenceFilter([...staticItems, ...items]), value, searchNodesKind
-            ));
+            );
+            setSelectItems(resolved);
+            // Fetched items arrive after the mount-time staticItems preselect.
+            if (!value && resolved.length > 0) {
+                onChange(resolved[0].value, resolved[0].value.length);
+            }
```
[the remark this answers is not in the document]
dan-niles: Not fixing — I don't think this is reachable.

For the captured `value` to go stale it has to be `""` when the request starts and non-empty when it resolves, with at least two pickable items. Two gates block that:

- `showLoading = loading && !selectedItem` makes the select `inert` (`NodeReferenceSelect.tsx:220-221`) exactly when `value` is empty, so the selection can't be changed during that window.
- The only thing that fills `value` mid-flight is the mount effect, and it sets `staticItems[0]` — which is also `resolved[0]`, so both branches end on the same value.

That leaves needing `staticItems.length >= 2` **and** an empty initial value. All three producers are mutually exclusive on that, and the language server never sends `staticItems`:

| Producer | Static items | Initial value |
|---|---|---|
| `node-property-utils.ts:187` (model provider) | 1 | empty |
| `ConnectionSelector/utils.tsx:68` | 1 | empty |
| `KnowledgeBaseForm/index.tsx:219` (chunker) | 2 | preset — `if (!field.value) field.value = "ai:AUTO"` |

Skipping the regression test for the same reason. Happy to revisit if a field ever ships two static items with no default.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/UseAgentToolForm.tsx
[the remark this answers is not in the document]
dan-niles: Not fixing. `joinProjectPath` has no reject path — `visualizer/rpc-manager.ts:242` is `new Promise((resolve) => ...)` with no `reject`, and the missing-project case resolves `{ filePath: "", projectPath: "" }`.

Your own supporting point covers the rest: `fetchOAuthConfigProperties` and `fetchAgentRunReturnType` both catch internally and return safe defaults, so the line this PR added can't throw either. The effect itself predates this PR — `cd8a4dae0c` only inserted `setDefaultReturnType(...)` inside it.

The proposed patch would also set `ready` after swallowing the error, so a genuine failure would render a form with an empty file path and no OAuth fields rather than a visible loader.

That said, you're right that this module has a convention I'm not following: the sibling `AgentToolForm.tsx:319` wraps its init effect in `try/finally` with a `cancelled` flag. `UseAgentToolForm` has neither. Tracking that separately rather than in this PR.

packages/ballerina-visualizer/src/views/BI/Connection/ConnectorBrowser/ConnectorBrowser.tsx
```diff
@@ -0,0 +1,314 @@
...
+            },
+            codedata: first?.codedata,
+            enabled: true,
+        } as AvailableNode);
+        setSelectedConnectionName(connectionName);
+        setSelectedCategory("Connections");
+        setActionError("");
+        setActions(actionNodes);
+        setStep(WizardStep.ACTION_LIST);
+    };
```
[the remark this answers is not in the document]
dan-niles: Not fixing — the empty case can't be reached.

`connectionRowsOf` (`ConnectorList.tsx:261`) filters `(sub.items ?? []).length > 0` before building any row, and only rows with `id === CONNECTION_ROW_ID` reach `onSelectConnection` (`ConnectorList.tsx:388`). The other branch returns real node ids, which route to `onSelect` instead.

So a connection with no actions never renders a clickable row, and `connectionActions` is non-empty whenever `handleSelectConnection` fires.

MohamedSabthar (dismissed):

packages/ballerina-visualizer/src/views/BI/AIChatAgent/toolForm.ts
```diff
@@ -0,0 +1,168 @@
...
+    let base = `${cleaned.charAt(0).toLowerCase()}${cleaned.slice(1)}`;
+    if (!/^[a-zA-Z_]/.test(base)) {
+        base = `tool${base}`;
+    }
+    if (!base.toLowerCase().endsWith("tool")) {
+        base = `${base}Tool`;
+    }
+    if (BALLERINA_RESERVED_TOOL_NAMES.has(base)) {
+        base = `${base}Tool`;
+    }
```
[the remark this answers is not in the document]
dan-niles: Good catch — confirmed unreachable. The block above appends `Tool` whenever `base` doesn't already end in `tool`, so no reserved word can reach this check (verified across all seven words plus edge cases: identical output with and without the guard).

Removed the guard and the now-unused set in 9a97608 (and 128bec82 on the split-out client PR #877).

dan-niles:
Closing this PR since I split it into the following 2 PRs:
- https://github.com/ballerina-platform/ballerina-vscode/pull/876
- https://github.com/ballerina-platform/ballerina-vscode/pull/877

PR #810, opened 2026-08-07
Open a single-integration workspace on startup

### LINKED ISSUE #2052: Single-integration projects should open on the integration overview, not the workspace overview
### Description

A project holding exactly one integration opens on the workspace overview: a list with a single item that the user has to click through before reaching anything useful. The landing view should be that integration's package overview.

### Steps to Reproduce

N/A

### Version

5.1.0-m2

_No response_

gigara:
Fixes: https://github.com/wso2/product-integrator/issues/2052

https://github.com/user-attachments/assets/e2db167c-04ab-4a77-ac1e-6d10b7c1e5b5

Open such a project directly on that integration's overview. Projects with more than one package are unchanged, and the workspace overview stays reachable — this removes a click on the way in, it does not remove the screen.

The redirect lives in `openView` rather than at each call site. That is the only place that can put `projectPath` into the machine context — which is where the webview reads the package an overview is for — and it means a navigation route added later inherits the behaviour instead of quietly reintroducing the one-item list. It applies to every navigation, not only the first: the project explorer re-navigates once its tree finishes loading, seconds after startup, and a first-navigation gate let that second one land back on the workspace overview.

Two shapes reach the workspace overview and both redirect: an explicit `view: WorkspaceOverview` (the project explorer's Open Overview), and a bare navigation with no view at all, which `findView` would otherwise resolve to it. Navigations that already name a package, or that carry an artifact position to resolve, are left alone.

The rule itself is `resolveSingleIntegrationOverride`, a pure function in `state-machine-utils`. It returns the whole location and `openView` swaps it in, so anything the redirected navigation carried — `documentUri`, `identifier`, `artifactType`, a `groupId` position — is dropped. All of it describes a target that stopped applying once the destination became the package overview, and `identifier` with `artifactType` would otherwise survive onto the overview's history entry and send `updateView` hunting for an artifact this view does not have.

**What does not redirect, and why.** The workspace overview is a real workspace-level surface: the workspace README, renaming the project, Add Integration or Library, deploy-all, ICP enable-all. None of that exists anywhere else in the webview. So the one-item list is an unhelpful thing to land *on*, not an unhelpful thing to *have*, and two cases keep it deliberately:

| Workspace | Lands on |
|---|---|
| `orders` | `orders` overview |
| `orders` + `shipping` | workspace overview |
| `orders` + `utils` (library) | workspace overview |
| `utils` (library) only | workspace overview |

- **A library alongside an integration counts.** Cardinality counts packages, not integrations — two packages means two things worth showing, and only the workspace overview shows both. A library-only workspace is excluded for the same reason: that overview is the sole surface offering to add the first integration.
- **Home opts out**, via a new `exactView` flag on `openView`, and is the only caller passing it. Home is the way back up to the workspace; redirecting it would land on the page the user is already looking at, and since `MainPanel` keys its remount on view/identifier/documentUri/projectPath — all unchanged — the button rendered nothing and looked dead. Keeping the escape hatch was preferred over hiding Home, which would cost access to everything listed above in exactly the workspaces this change targets.
- **A package with no `projectPath` does not qualify.** The field is optional on `ProjectStructure` and the package overview resolves its contents by path, so redirecting to a pathless package would replace the list with a view that never finishes loading.

Back is unaffected: `goBack` replays history through `updateView` and never reaches `openView`.

## UI Component Development
- [x] Added reusable UI components to the ui-toolkit — N/A, no new UI. This changes which existing view the window opens on; no component was added or altered.

As a developer whose project contains one integration, I want the window to open on that integration so I can start working without clicking through a list of one.

A project containing a single integration now opens directly on that integration's overview instead of the project overview. Projects with more than one integration are unchanged, and the project overview remains available from the Home button.

N/A — no documented behaviour changes. The screens, their contents and the navigation controls are all as before; only the view the window opens on differs, for a single-integration project.

## Training
N/A

## Certification
N/A — no user-facing concept or workflow is added or removed; this changes a default landing view.

## Marketing
N/A

## Automation tests
 - Unit tests
   > `packages/ballerina-extension/src/test-support/singleIntegrationLanding.test.ts` — 14 cases over `getSoleIntegration` and `resolveSingleIntegrationOverride`, covering the rule from both sides: every navigation shape that must redirect (explicit workspace overview, bare, bare carrying a document, bare carrying a `groupId` position, bare carrying artifact fields) and every shape that must be left alone (another view, one already naming a package, one resolving an artifact position). Also the workspace shapes that disqualify — multi-package, library alongside integration, library-only, empty, absent structure, and a sole integration with no `projectPath` — plus the invariant that a redirect always names a package, since the overview renders nothing without one. Full host suite: 112 passing.
 - Integration tests
   > None added. The behaviour is a navigation default in the extension host, exercised by the unit tests above. Labelled `Checks/Run Ballerina UI Tests` so the existing UI suite runs against it.

## Security checks

## Samples
N/A

## Related PRs
None

## Migrations (if applicable)
N/A — no persisted state. Adding a second integration makes the redirect stop firing on its own; nothing to clear.

macOS 15 (darwin 25.5.0), VS Code, Node 20. Verified by hand in a single-integration workspace: window opens on the integration overview and stays there (an earlier revision was overridden a few seconds in by the explorer's second navigation), Home reaches the workspace overview, Back behaves as before.

Three producers of the workspace overview had to be found before the change would hold: an explicit `view: WorkspaceOverview` from callers, `findView`'s fallback when no package is selected, and `showView`'s no-view fallbacks. An earlier revision missed the first — the project explorer's Open Overview names the view outright — so nothing changed at startup. A later one scoped the redirect to the first navigation and was overridden seconds later by the explorer re-navigating once its tree loaded. Handling the shared choke point in `openView` covers all of them.

Worth recording for the next person: `goBack` does not route through `openView` at all, so navigation changes made there cannot affect Back — which is what identified a stale-`projectPath` write in the `webViewLoaded` assign, since that broke Home and Back together, as the cause of a regression rather than the redirect.

### DIFF
--- a/packages/ballerina-extension/src/rpc-managers/visualizer/rpc-manager.ts
+++ b/packages/ballerina-extension/src/rpc-managers/visualizer/rpc-manager.ts
@@ -100,7 +100,11 @@ export class VisualizerRpcManager implements VisualizerAPI {
                         ? MACHINE_VIEW.WorkspaceOverview
                         : MACHINE_VIEW.PackageOverview
                 },
-                true
+                true,
+                // Home is the way back up to the workspace, so it keeps the workspace overview
+                // even when a single integration would otherwise collapse onto itself — the
+                // landing view is already that integration, so redirecting here does nothing.
+                { exactView: true }
             );
         });
     }
--- a/packages/ballerina-extension/src/stateMachine.ts
+++ b/packages/ballerina-extension/src/stateMachine.ts
@@ -35,7 +35,8 @@ import {
     getNodeByIndex,
     getNodeByName,
     getNodeByUid,
-    getView
+    getView,
+    resolveSingleIntegrationOverride
 } from './utils/state-machine-utils';
 import * as path from 'path';
 import { extension } from './BalExtensionContext';
@@ -1003,13 +1004,35 @@ export const StateMachine = {
     },
 };
 
-export function openView(type: EVENT_TYPE, viewLocation: VisualizerLocation, resetHistory = false) {
+export interface OpenViewOptions {
+    /**
+     * Take `viewLocation.view` literally, skipping the single-integration redirect in
+     * {@link openView}. For callers whose whole purpose is to reach the workspace overview —
+     * Home being the one — since redirecting them to the integration overview would land on
+     * the page the user is already looking at and make the control appear dead.
+     */
+    exactView?: boolean;
+}
+
+export function openView(
+    type: EVENT_TYPE,
+    viewLocation: VisualizerLocation,
+    resetHistory = false,
+    options?: OpenViewOptions
+) {
     if (resetHistory) {
         StateMachine.setReadyMode();
         history?.clear();
     }
     extension.hasPullModuleResolved = false;
     extension.hasPullModuleNotification = false;
+    // A workspace holding a single integration opens on that integration rather than a one-item
+    // workspace overview. Applied to every navigation, not just the first: the project explorer
+    // re-navigates once its tree finishes loading, seconds after startup, and a first-navigation
+    // gate would let that second one land back on the workspace overview.
+    if (!options?.exactView) {
+        Object.assign(viewLocation, resolveSingleIntegrationOverride(viewLocation, StateMachine.context()));
+    }
     const projectPath = viewLocation.projectPath || StateMachine.context().projectPath;
     const { orgName, packageName } = getOrgAndPackageName(StateMachine.context().projectInfo, projectPath);
     viewLocation.org = orgName;
--- a/packages/ballerina-extension/src/stateMachine.ts
+++ b/packages/ballerina-extension/src/stateMachine.ts
@@ -1030,14 +1030,20 @@ export function openView(
     // workspace overview. Applied to every navigation, not just the first: the project explorer
     // re-navigates once its tree finishes loading, seconds after startup, and a first-navigation
     // gate would let that second one land back on the workspace overview.
-    if (!options?.exactView) {
-        Object.assign(viewLocation, resolveSingleIntegrationOverride(viewLocation, StateMachine.context()));
-    }
-    const projectPath = viewLocation.projectPath || StateMachine.context().projectPath;
+    //
+    // The override REPLACES the location rather than merging into it. The fields a redirected
+    // navigation carried describe a target that no longer applies, and `identifier` with
+    // `artifactType` would survive onto the overview's history entry and send `updateView`
+    // hunting for an artifact that this view does not have.
+    const singleIntegrationOverride = options?.exactView
+        ? undefined
+        : resolveSingleIntegrationOverride(viewLocation, StateMachine.context());
+    const location = singleIntegrationOverride ?? viewLocation;
+    const projectPath = location.projectPath || StateMachine.context().projectPath;
     const { orgName, packageName } = getOrgAndPackageName(StateMachine.context().projectInfo, projectPath);
-    viewLocation.org = orgName;
-    viewLocation.package = packageName;
-    stateService.send({ type: type, viewLocation: viewLocation });
+    location.org = orgName;
+    location.package = packageName;
+    stateService.send({ type: type, viewLocation: location });
 }
 
 export function updateView(refreshTreeView?: boolean, updatedIdentifier?: string) {
--- a/packages/ballerina-extension/src/test-support/singleIntegrationLanding.test.ts
+++ b/packages/ballerina-extension/src/test-support/singleIntegrationLanding.test.ts
@@ -75,6 +75,13 @@ describe("getSoleIntegration", () => {
         expect(getSoleIntegration(workspaceOf(pkg("utils", true)))).toBeUndefined();
         expect(getSoleIntegration(undefined)).toBeUndefined();
     });
+
+    // `projectPath` is optional on ProjectStructure. The package overview resolves its contents
+    // by path, so redirecting to a pathless package would hang on a spinner forever — worse than
+    // the one-item list it replaced.
+    it("returns nothing for a sole integration with no project path", () => {
+        expect(getSoleIntegration(workspaceOf({ ...pkg("orders"), projectPath: undefined }))).toBeUndefined();
+    });
 });
 
 describe("resolveSingleIntegrationOverride", () => {
--- a/packages/ballerina-extension/src/test-support/singleIntegrationLanding.test.ts
+++ b/packages/ballerina-extension/src/test-support/singleIntegrationLanding.test.ts
@@ -17,11 +17,13 @@
  */
 
 import type {
+    DIRECTORY_MAP as DirectoryMap,
     MACHINE_VIEW as MachineView,
     ProjectStructure,
     ProjectStructureResponse,
     VisualizerLocation,
 } from "@wso2/ballerina-core";
+import type { NodePosition } from "@wso2/syntax-tree";
 
 // The real barrel pulls in a WebSocket LS client that jest cannot load, so stub it. The view
 // names must match the real enum values — the assertions below compare against them.
@@ -30,9 +32,10 @@ const MACHINE_VIEW = {
     WorkspaceOverview: "Workspace Overview",
     ServiceDesigner: "Service Designer",
 } as unknown as typeof MachineView;
+const DIRECTORY_MAP = { FUNCTION: "FUNCTION" } as unknown as typeof DirectoryMap;
 jest.mock("@wso2/ballerina-core", () => ({
     MACHINE_VIEW,
-    DIRECTORY_MAP: {},
+    DIRECTORY_MAP,
     EVENT_TYPE: {},
     FOCUS_FLOW_DIAGRAM_VIEW: {},
     isSamePath: (a: string, b: string) => a === b,
@@ -91,11 +94,22 @@ describe("resolveSingleIntegrationOverride", () => {
     // Both shapes a navigation can take, since the caller cannot tell them apart: the project
     // explorer names the workspace overview outright, other entry points send a bare location
     // and let `findView` resolve it.
+    //
+    // `toEqual(expected)` is also asserting what the redirect DROPS, and that is load-bearing:
+    // `openView` replaces the caller's location with this one, so a field missing here is a
+    // field that does not reach the state machine. The `groupId` position mirrors the clause in
+    // `findView` (stateMachine.ts) that sends such a navigation to the workspace overview —
+    // `groupId` is not on `NodePosition`, so nothing but this test pins the two together.
     it.each<[string, VisualizerLocation]>([
         ["an explicit workspace overview", { view: MACHINE_VIEW.WorkspaceOverview }],
         ["a bare navigation with no view", {}],
         ["a bare navigation carrying only a document", { documentUri: "/workspace/orders/main.bal" }],
-    ])("redirects %s to the sole integration", (_label, viewLocation) => {
+        ["a bare navigation carrying a groupId position", { position: { groupId: "g1" } as unknown as NodePosition }],
+        [
+            "a bare navigation carrying artifact fields",
+            { documentUri: "/workspace/orders/main.bal", identifier: "greet", artifactType: DIRECTORY_MAP.FUNCTION },
+        ],
+    ])("redirects %s to the sole integration, dropping the rest", (_label, viewLocation) => {
         expect(resolveSingleIntegrationOverride(viewLocation, soleIntegration)).toEqual(expected);
     });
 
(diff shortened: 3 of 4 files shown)


packages/ballerina-extension/src/rpc-managers/visualizer/rpc-manager.ts
```diff
@@ -100,7 +100,11 @@ export class VisualizerRpcManager implements VisualizerAPI {
                         ? MACHINE_VIEW.WorkspaceOverview
                         : MACHINE_VIEW.PackageOverview
                 },
-                true
+                true,
+                // Home is the way back up to the workspace, so it keeps the workspace overview
+                // even when a single integration would otherwise collapse onto itself — the
+                // landing view is already that integration, so redirecting here does nothing.
+                { exactView: true }
```
NipunaRanasinghe, line 107: This is the one thing I would settle before merge, and it is a product call rather than a code fix.

`goHome` passes `view: WorkspaceOverview` whenever `workspacePath` is set, and `exactView: true` suppresses the redirect. So in a single-integration workspace **Home always lands on the one-item workspace overview** — the screen this PR opens by describing as "a list with exactly one item to click through before reaching anything useful."

The diagnosis in the comment is right: redirecting Home would return the user to the page they are already on, and since `MainPanel` keys its remount on view/identifier/documentUri/projectPath — all unchanged — the control would render nothing and look dead. But the fix trades one bad outcome for another. Home is no longer dead; it now navigates to the list this change exists to remove.

Worth deciding explicitly rather than by omission:

- If the one-item overview is genuinely useless, then Home has nowhere to go in this workspace and arguably should not be offered — hiding or disabling it is more honest than landing there.
- If it is useful as a workspace-level surface (adding a second integration, workspace settings), then the premise is narrower than "nothing to choose from", and that is worth saying in the PR description, because it is the justification for the whole change.

Either reading is defensible. What I would not want is for the answer to be settled by which call site happened to get the flag.

packages/ballerina-extension/src/stateMachine.ts
```diff
@@ -1003,13 +1004,35 @@ export const StateMachine = {
...
         history?.clear();
     }
     extension.hasPullModuleResolved = false;
     extension.hasPullModuleNotification = false;
+    // A workspace holding a single integration opens on that integration rather than a one-item
+    // workspace overview. Applied to every navigation, not just the first: the project explorer
+    // re-navigates once its tree finishes loading, seconds after startup, and a first-navigation
+    // gate would let that second one land back on the workspace overview.
+    if (!options?.exactView) {
+        Object.assign(viewLocation, resolveSingleIntegrationOverride(viewLocation, StateMachine.context()));
```
NipunaRanasinghe: `Object.assign` is a merge, not a replace, so every field the caller supplied that the rule does not return survives the redirect — `documentUri`, `identifier`, `artifactType`, and a `position` in the `groupId` case.

That is probably harmless for `PackageOverview`, but nothing pins it, and the unit test actively suggests otherwise: the bare-navigation-with-a-document case asserts the rule returns exactly `{ view, projectPath }`, which reads as proof that `documentUri` is dropped. At runtime it is still on the object when `stateService.send` fires.

The seam that ships is this line, not the pure function. One test over the applied result — call `openView` with a mocked `stateService` and assert the location it sends — would cover the merge and make the retained fields a decision rather than an accident.

(Worth noting `Object.assign(viewLocation, undefined)` is a correct no-op on the non-redirect path, so that part is fine.)

packages/ballerina-extension/src/utils/state-machine-utils.ts
```diff
@@ -28,6 +28,58 @@ import { getConstructBodyString } from "./history/util";
...
+    viewLocation: VisualizerLocation,
+    context: Pick<VisualizerLocation, "workspacePath" | "projectPath" | "projectStructure">
+): VisualizerLocation | undefined {
+    if (viewLocation.projectPath || !context.workspacePath) {
+        return undefined;
+    }
+    const isBareNavigation =
+        !viewLocation.view &&
+        !context.projectPath &&
+        (!viewLocation.position || "groupId" in viewLocation.position);
```
NipunaRanasinghe, line 78: This is the subtlest line in the change and the only one with no test behind it. The suite exercises a plain `NodePosition` (left alone) but never a position carrying `groupId`, which should still redirect.

It matters more than a typical missing case because `position` is typed `NodePosition` and `groupId` is not on that type, so the branch depends on an untyped shape reaching here at runtime. Nothing in the type system will tell you when that stops being true — a test is the only thing that would.

For what it is worth I checked this against `stateMachine.ts:677` and the condition matches exactly, including this clause. Mirroring the predicate that produces the workspace overview is the right approach; it just means the mirror needs pinning on both sides.

packages/ballerina-extension/src/test-support/singleIntegrationLanding.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    const soleIntegration = { workspacePath: WORKSPACE_ROOT, projectStructure: workspaceOf(pkg("orders")) };
+    const expected = { view: MACHINE_VIEW.PackageOverview, projectPath: `${WORKSPACE_ROOT}/orders` };
+
+    // Both shapes a navigation can take, since the caller cannot tell them apart: the project
+    // explorer names the workspace overview outright, other entry points send a bare location
+    // and let `findView` resolve it.
+    it.each<[string, VisualizerLocation]>([
+        ["an explicit workspace overview", { view: MACHINE_VIEW.WorkspaceOverview }],
+        ["a bare navigation with no view", {}],
+        ["a bare navigation carrying only a document", { documentUri: "/workspace/orders/main.bal" }],
```
NipunaRanasinghe, line 106: This case is the one most likely to mislead a later reader. It asserts the rule returns exactly `{ view, projectPath }` for a navigation carrying a `documentUri`, which reads as "the document is dropped on redirect".

It is not — `openView` applies the result with `Object.assign`, so `documentUri` is still on the location that reaches `stateService.send`. The assertion is correct about the function and wrong about the behaviour, which is the more expensive kind of green test.

Either assert the merged result somewhere, or add a line here noting the caller merges, so the next person changing this does not assume the returned object is the whole location.

packages/ballerina-extension/src/test-support/singleIntegrationLanding.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    it.each<[string, VisualizerLocation]>([
+        ["a view other than the workspace overview", { view: MACHINE_VIEW.ServiceDesigner }],
+        ["a navigation that already names a package", { projectPath: `${WORKSPACE_ROOT}/orders` }],
+        ["a navigation resolving an artifact position", { position: { startLine: 1, startColumn: 0, endLine: 2, endColumn: 0 } }],
+    ])("leaves %s alone", (_label, viewLocation) => {
+        expect(resolveSingleIntegrationOverride(viewLocation, soleIntegration)).toBeUndefined();
+    });
+
+    it("names the package it redirects to — the overview renders nothing without one", () => {
+        expect(resolveSingleIntegrationOverride({}, soleIntegration).projectPath).toBeTruthy();
```
NipunaRanasinghe: Minor: this compiles only because the ts-jest tsconfig in `jest.config.js` sets `strictNullChecks: false` — the return type is `VisualizerLocation | undefined`. If that ever tightens, this line breaks in a way unrelated to what it is testing.

It also fails as a `TypeError` rather than a clean assertion if the rule regresses to returning `undefined`, which is a worse failure message for the invariant this test exists to protect. `?.projectPath` with `toBeTruthy()` keeps both properties.

packages/ballerina-extension/src/utils/state-machine-utils.ts
```diff
@@ -28,6 +28,53 @@ import { getConstructBodyString } from "./history/util";
...
+ * Libraries deliberately do not qualify: a workspace whose only package is a library still
+ * needs the workspace overview, which is the only surface offering to add the first
+ * integration.
+ */
+export function getSoleIntegration(projectStructure?: ProjectStructureResponse): ProjectStructure | undefined {
+    const projects = projectStructure?.projects ?? [];
+    if (projects.length !== 1 || projects[0].isLibrary) {
+        return undefined;
+    }
+    return projects[0];
```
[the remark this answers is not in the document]
gigara: Skipping this one — the behaviour is intentional, though the doc comment gave you every reason to read it as a bug, so I've reworded it.

Cardinality here counts **packages, not integrations**. A workspace pairing one integration with a shared library has two things worth showing, and the workspace overview is the only surface that shows both; collapsing onto the integration would hide the library from the landing page. The narrower rule the description meant is that a *library-only* workspace does not qualify, because that overview is the sole place offering to add the first integration.

| Workspace | Lands on |
|---|---|
| `orders` | `orders` overview |
| `orders` + `shipping` | workspace overview |
| `orders` + `utils` (library) | workspace overview |
| `utils` (library) only | workspace overview |

The second finding on this file was valid and is fixed in 1c2ba5db86 — `projectPath` is optional on `ProjectStructure`, and redirecting to a pathless package would have replaced the list with a view that never finishes loading.

packages/ballerina-extension/src/stateMachine.ts
gigara: Correct, and the consequence is worse than the misleading test. `documentUri` surviving is mostly cosmetic, but `identifier` and `artifactType` would ride onto the overview's history entry, and `updateView` branches on exactly that pair to relocate an artifact — so it would go hunting for one on a view that has none.

Fixed in b09bb1f398 by making the override a replacement rather than a patch. `resolveSingleIntegrationOverride` now returns the whole location and `openView` swaps it in, so everything the redirected navigation carried is dropped: `documentUri`, `identifier`, `artifactType`, and a `groupId` position. All of it describes a target that stopped applying the moment the destination became the package overview.

That makes the `toEqual` assertion describe real behaviour instead of contradicting it, and I added cases that carry those fields so the drop is pinned rather than incidental.

On the seam test: I did not add one. `stateMachine.ts` pulls in the extension host, `vscode` and xstate, so it will not load under the current jest config without real scaffolding. What I did instead is remove the divergence — with a replacement there is no merge behaviour left for the pure function's contract to be wrong about. That closes this specific gap, not the general one, so if you want the applied result covered properly it is worth doing as its own change.

packages/ballerina-extension/src/utils/state-machine-utils.ts
gigara: Added in b09bb1f398 — a bare navigation carrying `{ position: { groupId } }` now has a case asserting it redirects.

Your reason for it being more than a missing case is the right one: the clause rests on a shape that is not on `NodePosition`, so nothing in the type system pins it to the `findView` predicate it mirrors. The test is the only thing that will notice if the two drift apart, and I've said so in a comment next to it pointing at `stateMachine.ts`.

packages/ballerina-extension/src/test-support/singleIntegrationLanding.test.ts
gigara: Fixed in b09bb1f398 — now `?.projectPath` with `toBeTruthy()`.

Both of your reasons hold: it was leaning on `strictNullChecks: false` in the ts-jest config for a `VisualizerLocation | undefined`, and a regression to `undefined` would have surfaced as a `TypeError` rather than as the invariant this case exists to state.

packages/ballerina-extension/src/test-support/singleIntegrationLanding.test.ts
gigara: This one is resolved by the change to the caller rather than to the test. As of b09bb1f398 the override replaces the location instead of merging into it, so `documentUri` genuinely is dropped and the assertion is now accurate about both the function and the behaviour.

I took the first of your two options — assert the merged result, or note that the caller merges — because the merge itself was the defect. The comment above the block now says what `toEqual` is really doing: it is asserting what the redirect drops, and a field missing from `expected` is a field that does not reach the state machine.

packages/ballerina-extension/src/rpc-managers/visualizer/rpc-manager.ts
gigara: Fair, and you've identified the actual defect: the flag placement settled a product question by omission, and the description states a premise the code does not hold to.

Taking your second reading. The workspace overview is a real workspace-level surface — the workspace README, renaming the project, Add Integration or Library, deploy-all, ICP enable-all — and none of that exists anywhere else in the webview. That is also why `1 integration + 1 library` deliberately stays on it, and why Home has somewhere worth going. So the honest claim is narrower than what I wrote: the one-item list is an unhelpful thing to land *on*, not an unhelpful thing to *have*.

Which makes "a list with exactly one item to click through before reaching anything useful" wrong as written, and the description needs to say the workspace overview stays one click away on Home and what it is for. Left as is, the justification reads as stronger than the change actually warrants.

Hiding or disabling Home is the other consistent option, but it costs access to all of the above in exactly the workspaces this change targets, so I would rather keep the escape hatch and fix the framing.

NipunaRanasinghe (approved):

NipunaRanasinghe (commented):
@gigara Better to attach a screen recording in the PR description since this is a UX change (hence better to verify the overall UX with the changes).

PR #811, opened 2026-08-07
Improve file trigger form descriptions

niveathika:
## Purpose

Reword the FTP, SMB and Azure Files trigger model descriptions so the handler form reads in terms of what the user does, rather than how data binding is implemented, and apply the trigger-model side of the On Create form review.

The On Create descriptions previously narrated the widgets sitting directly beneath them. For CSV, every clause was already on screen:

| Old text | Already shown by |
|---|---|
| "Invoked when a file is created…" | dialog title "New On Create Configuration" |
| "Comma-separated rows" | Format dropdown reading "CSV" |
| "emits T[]" / "stream<T, error?>" | Define Row Schema action + Stream checkbox |

This also brings the file triggers in line with the event triggers, none of whose 25 `DATA_BINDING` descriptions name a Ballerina type.

### Descriptions

- **`Monitored Path`** replaces `Monitoring Path` on the path field, with one shared description across the three connectors. `Monitoring Path` reads as the path that *does* the monitoring.
- **Handler descriptions** drop `binds`, `T`, `T[]`, `stream<T, error?>` and `byte[]`, and instead point at the schema action with concrete example types:

  > Invoked when a file is created in the monitored directory. Define a row schema below to receive the rows as custom types (e.g., User[], Student[]).

  `User[]` for CSV, `User` for JSON/XML, since those bind a single value.
- **Azure's extension routing** is folded into the prefix ("Invoked when a .csv file is created…") rather than carried as a separate clause.
- **Azure `documentation` fields** now carry their own values. They previously duplicated the form description verbatim, so rewording alone would have put "Define a row schema below…" into the generated Ballerina doc comments.

### Form simplification

- **Removed the CSV `rows` marker.** It rendered as a permanently checked, non-editable checkbox above the live Stream toggle, which reads as a disabled option rather than a statement about the payload. The array-ness it signalled is now in the handler description and carried by the payload template.
- **"File Handling Options"** is now the after-processing group name in all three connectors; FTP and Azure Files called it "After File Processing".
- **Dropped the `@ftp:` / `@smb:` / `@files:FunctionConfig` prefix** from that group's description — the annotation type is implementation detail. SMB's text also advertised a per-handler file-name filter that the group does not expose.
- **File Metadata** descriptions now share a sentence shape across the three connectors, keeping each protocol's own field list.

### Naming kept deliberately

`Row Schema` is retained for CSV rather than unified with the `Content Schema` label used by JSON and XML. The CSV payload template is `{{type}}[]`: the schema editor edits one row and the handler receives an array of it, whereas JSON and XML templates are `{{type}}`, where the edited type is the content itself. Naming all three "Content Schema" misdescribed CSV in both the form action and the type editor's modal title, which derives from the same label while the editor shows the bare row element.

## Notes

Removing the `rows` flag does not affect generated code. It was presentational only — `composePayloadType` reads the `PAYLOAD_MODIFIER` roles and the parameter template, and the `[]` comes from `payload.codedata.template`. Two `TriggerFunctionExpansionTest` assertions that pinned the marker now assert its absence.

Open question: it is unconfirmed whether these strings also render in the "Add On Create Handler" picker. The `Invoked when a file is created…` prefix is retained on that assumption; if the picker does not use them, it could be dropped as redundant with the dialog title.

## Testing

`:service-model-generator:service-model-generator-ls-extension:test` — **439 tests, 1 failure**.

The failure is `get_service_init_model/config/mqtt_service_model_2.json`, which fails identically on `main` (verified in a clean worktree) and is unrelated to these changes.

## Checklist

- [x] JSON validated for all three files
- [x] Test suite run locally; only a pre-existing unrelated failure remains

## Related

- wso2/product-integrator#2059 — the remaining form-renderer items from the same review (heading wording, Stream checkbox position, schema action tooltip)

### DIFF
--- a/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger-models/azure.storage.files.json
+++ b/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger-models/azure.storage.files.json
@@ -1290,7 +1290,7 @@
                 "properties": {
                   "payload": {
                     "metadata": {
-                      "label": "Content Schema",
+                      "label": "Row Schema",
                       "description": "Create a record type describing one row of the file. Each row arrives as one record."
                     },
                     "value": "string[][]",
--- a/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger-models/azure.storage.files.json
+++ b/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger-models/azure.storage.files.json
@@ -1290,7 +1290,7 @@
                 "properties": {
                   "payload": {
                     "metadata": {
-                      "label": "Row Schema",
+                      "label": "Content Schema",
                       "description": "Create a record type describing one row of the file. Each row arrives as one record."
                     },
                     "value": "string[][]",
@@ -1348,7 +1348,7 @@
             {
               "metadata": {
                 "label": "File Metadata (fileInfo)",
-                "description": "File properties: path, name, size, eTag and last-modified time."
+                "description": "File properties such as path, name, size, eTag and last-modified time."
               },
               "kind": "OPTIONAL",
               "type": {
@@ -1373,7 +1373,7 @@
               "name": {
                 "metadata": {
                   "label": "fileInfo",
-                  "description": "File properties: path, name, size, eTag and last-modified time."
+                  "description": "File properties such as path, name, size, eTag and last-modified time."
                 },
                 "placeholder": "fileInfo",
                 "value": "fileInfo",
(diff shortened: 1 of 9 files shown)


packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger-models/azure.storage.files.json
```diff
@@ -1239,7 +1239,7 @@
         {
           "metadata": {
             "label": "On Create",
-            "description": "Handles a file whose extension routes to the CSV format (.csv), binding its content as string[][] or as rows of a record type (T[]).",
+            "description": "Invoked when a .csv file is created in the monitored directory. Define a row schema below to receive the rows as custom types (e.g., User[], Student[]).",
```
niveathika, line 1242: @gayaldassanayake FYI

gayaldassanayake: This is a good description. Explains users the importance of data binding. +1

LakshanWeerasinghe (approved):

PR #815, opened 2026-08-07
Add language server unit test writer skill

pasindufernando1:
Resolves wso2-enterprise/integration-engineering#2292

There was no documented procedure for writing a language-server unit test. The JVM (Gradle/TestNG) suite in `packages/ballerina-language-server` is data-driven in a way that is not obvious from reading one test — a new case is a `.json` file in a config directory, not a test method — and the expected output is generated by temporarily uncommenting a line, which nothing tells you.

The immediate trigger was narrower. The `stdlib*Version` pins in `packages/ballerina-language-server/gradle.properties` exist to inject packages that ship in the full Ballerina distribution but not in the minimal `jballerina-tools` distribution the tests run on. They are **not** where fixture dependency versions are owned — that is `build-config/ballerina_dependencies/`. Raising `stdlibAiVersion` to the locked `ai` version broke language-server tests: raised in 23d85849ea, reverted in 008f4cc370. The interesting part is that the version number was never wrong — the pin and the lock held the *same* version, and resolution preferring the distribution copy is what broke it.

That turned out to be one instance of a general problem, so this PR documents the whole task rather than just the trap.

- Give a contributor — human or agent — one place to learn how to write, extend, regenerate and debug a language-server test.
- Stop `gradle.properties` from leaving contributors to guess that `stdlib*Version` is not where fixture versions live.
- Document the follow-up step that is easy to miss: when the lock moves a package you did not name, tests referring to that package need updating too.

**`docs/LS_TEST_GUIDE.md`** (new) — the guide itself: how a test is shaped (data-driven over a `config/` directory; the two `AbstractLSTest` base classes), writing a new one (the four hooks, with `FindMatchingTypeTest` as the worked example), regenerating a golden, adding and skipping cases, debugging a failure, and why tests never reach Ballerina Central.

**`docs/LS_FIXTURE_DEPENDENCIES.md`** (new) — the fixture dependency-version topic, which is self-contained: the two layers and why resolution prefers the distribution copy, the bump procedure, the `import <org>/<pkg> as _;` requirement when adding a package, and reading the `Dependencies.toml` diff.

**`.claude/skills/ballerina-ls-test-writer/SKILL.md`** — reduced from 296 lines to a thin skill: pointers to both docs plus the non-negotiables that are cheap to state and expensive to catch in review. The sibling `ballerina-e2e-writer` skill is not a precedent for this, as review pointed out — it is 154 lines and links to no docs at all. Moving the content to `docs/` is still the right call, just not an established pattern being followed.

**Discoverability** — linked from `docs/TEST_GUIDE.md` (which is the canonical test guide but whose levels are all Jest/Playwright), `CONTRIBUTING.md` § Testing, and `AGENTS.md` (both the tests section and the "Files that contain the truth" table).

**`packages/ballerina-language-server/gradle.properties`** — a note under `# Standard Library Dependencies` saying what these pins do and that they are not where fixture versions are changed, pointing at `build-config/ballerina_dependencies/` and `docs/LS_FIXTURE_DEPENDENCIES.md`. Comment-only; no pin value is touched.

No production code changes.

### Review feedback applied

The first revision of this PR was a single 296-line `SKILL.md`, and review found both a placement problem and a set of factual errors. What changed:

| Issue | Change |
|---|---|
| Guide lived only under `.claude/`, which no human opens | Moved to `docs/`, linked from the three human entry points; skill is now a pointer |
| `gradle.properties` edit was missing | Restored. The staging merge in 2be6b2d4fd had resolved it away, and upstream 01ff87ece8 had already removed the line the original diff replaced, so it is rewritten for the current grouped-by-level layout and points at `docs/` rather than the renamed-away `ballerina-ls-test-deps` path |
| Counts rot | Every count removed in favour of the command that produces it. Confirmed during review: the config-file count was already stale, and the live `updateConfig` count moved twice during review |
| `testng.xml` registration described as class-only | Both mechanisms documented. `langserver-core` is almost entirely `<packages>`-based, so "a class not listed there silently never runs" was wrong for a third of the tree — and a redundant `<class>` entry in a package-covered suite runs the class twice |
| `updateConfig` behaviour imprecise | Now states the comparison has already run against the old in-memory `testConfig.output()` and `Assert.fail` follows the write, so the current run still fails; only the next run passes |
| `-i` needed to see the diff | Dropped. `gradle/javaProject.gradle` sets `showStandardStreams = true`, so the stderr diff already prints. The results-XML `grep` is kept for a run that has already finished |
| `extracted-distribution` stated as universal | All three layouts named — `extracted-distribution`, `extracted-distributions` (`misc/ballerinalang-data-mapper`), and bare `build/jballerina-tools-<ver>` (`misc/diagram-util`) |
| Concurrent-Gradle warning unsupported | Softened to the mechanism that actually collides: these tasks write shared build state, so one build can extract over a distribution another is reading |
| `PackageUtil.FORCE_OFFLINE` cited | That field is gone at HEAD. The doc now names `CommonUtil.TEST_OFFLINE` (reads the property), `PackageUtil.isOffline()` (delegates to it), and `RemoteCentral.OFFLINE` |
| Lock regenerated with a bare `bal build` | Now routed through `./gradlew resolveBallerinaDependencies`, which is what CI runs, plus copying the lock back. Also documents a trap found while checking this: the task short-circuits on a hash of the *committed* `Dependencies.toml`, which a `Ballerina.toml` bump does not change, so it logs `skipping dependency resolution` and does nothing unless `build/ballerina_dependencies` is cleared first |
| Distribution-over-cache precedence stated absolutely | Scoped to the default test path, with the workspace / custom-repository caveat |
| Editorial phrasing, unlabelled fences, ambiguous working directories | All fixed. Every command block now states the directory it runs from and was verified there — the repo-wide greps from the repository root, `./gradlew` and `build/` paths from `packages/ballerina-language-server` (there is no `gradlew` at the root) |

## UI Component Development
N/A — no UI changes.

## User stories
N/A

## Release note
N/A — developer tooling and in-repo documentation only; no user-facing change.

## Documentation
N/A — the change *is* the documentation. It is in-repo (two docs, a build comment and a contributor skill) and has no product documentation impact.

## Training
N/A

## Certification
N/A — no functional or user-facing behaviour to examine.

## Marketing
N/A

## Automation tests
 - Unit tests
   > No code change, so no coverage change. The `gradle.properties` diff is comment-only (every changed line is a comment; no pin value differs from the base branch).
 - Integration tests
   > N/A. Every command in both docs was run from the working directory its block names — the repository root for the repo-wide greps, `packages/ballerina-language-server` for `./gradlew` and anything under `build/` — and its output checked. Every symbol, path and behavioural claim was re-verified against the current branch head after merging staging, not against the checkout the first revision was written on — which is how the stale counts got in.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A — no Java or other source code changed.

## Samples
N/A

## Related PRs
None. Related commits for context: 23d85849ea (raised `stdlibAiVersion`), 008f4cc370 (reverted it), 01ff87ece8 (regrouped the stdlib pins, removing the comment line the first revision of this PR replaced).

## Migrations (if applicable)
N/A

## Test environment
N/A — no code executed by this change. The investigation behind it ran on macOS 15 (Darwin 24.1.0) with Ballerina 2201.13.4 and JDK 21.

## Learning

**Counts in a doc are a liability, and more so in a skill.** Review caught a subclass count that was off by one because the grep that produced it also matched `Class<? extends AbstractLSTest>` in the base class — and by the time the fix landed, two other counts had rotted on unrelated commits. A doc that is wrong in a checkable detail loses authority on the parts a reader cannot check, and an agent will not check. Every count is now the command that produces it.

**A doc in `.claude/` is not documentation.** The content was right and the placement made it unreadable to the audience the PR claimed to serve.

**Live `updateConfig` calls are a real defect.** A handful of call sites — not classes, and the number has moved three times during review, which is the point — call `updateConfig` uncommented. Because the call sits inside the mismatch branch, each rewrites its own expected output and goes green on the next run, absorbing any later regression. The guide warns not to treat them as precedent; re-committing them needs its own PR, and a check in `check` would beat a warning in a doc. Filing an issue for it.

## Summary

- Added contributor guides for writing, extending, regenerating, and debugging language-server JVM tests.
- Documented fixture dependency version management, lock regeneration, offline execution, and distribution behavior.
- Condensed the language-server test-writing skill into rules and links to the new guides.
- Linked the guidance from contributor and test documentation.
- Clarified that Gradle standard-library version pins inject distribution packages and do not define fixture dependency versions.
- Made documentation commands reproducible and corrected guidance on test registration, expected-output updates, dependency resolution, and concurrent builds.
- Updated the data-mapper issues URL to the Product Integrator repository.

No production code or dependency pin values changed.


.claude/skills/ballerina-ls-test-writer/SKILL.md
NipunaRanasinghe, line 1: **Blocker: the `gradle.properties` change this PR promises is not in the diff.**

The net diff is this one file (296 additions, 0 deletions). `43191425e` did edit `packages/ballerina-language-server/gradle.properties`, but the staging merge `2be6b2d4f` resolved that edit away. The current file on the branch contains no reference to `build-config/` or to this skill, so the second stated goal ("stop `gradle.properties` from directing contributors to edit `stdlib*Version` pins") is unmet.

Two things to note when restoring it:
- The "Versions must match …" line the description says it replaces was already removed upstream in `01ff87ece8` ("Group stdlib versions by dependency level"), so the diff will not look like the one in `43191425e`.
- That commit's replacement text pointed at `.claude/skills/ballerina-ls-test-deps/SKILL.md`, which this PR renames away — restoring it verbatim creates a dangling reference. It needs the `ballerina-ls-test-writer` path.

**Also: the description no longer matches the change.** It says the skill "replaces `ballerina-ls-test-deps`" — that skill exists only inside this PR's own history, never on the base branch, so a reviewer looking for the deletion will not find one. And the verification section cites numbers that disagree with the file it is verifying (158 commented `updateConfig` call sites vs. the skill's 160) and with the tree (169 subclasses — see the comment on lines 34-40).

.claude/skills/ballerina-ls-test-writer/SKILL.md
```diff
@@ -0,0 +1,296 @@
+---
+name: ballerina-ls-test-writer
+description: Use when adding, changing or debugging a language-server unit test in packages/ballerina-language-server - writing a new AbstractLSTest case, creating or regenerating a config/source fixture, working out why a golden mismatches, or changing the Ballerina package versions the fixtures compile against.
+---
+
+# Ballerina LS Test Writer
+
+These are JVM tests that start a real language server, call one LSP method over it,
+and compare the JSON response against a checked-in expected value.
```
NipunaRanasinghe, line 9: **Placement: this is the right content in a place humans will not read.**

The premise checks out — `docs/TEST_GUIDE.md` is the canonical test guide (linked from `CONTRIBUTING.md:165`), and its L3 tier (`docs/TEST_GUIDE.md:134`) is the *jest* headless-LS layer, so the JVM Gradle suite genuinely has no home there today.

But `.claude/skills/` is loaded by Claude Code only. The stated goal — "give a contributor one place to learn how to write, extend, regenerate and debug a language-server test" — is not achieved for a human contributor, who has no reason to open this path.

Suggestion: put the content in `docs/` (either a new `docs/LS_TEST_GUIDE.md` or a section in `TEST_GUIDE.md`) and keep the skill as a thin pointer to it. At minimum, link it from `docs/TEST_GUIDE.md` and `AGENTS.md` so it is discoverable without knowing the skill exists.

.claude/skills/ballerina-ls-test-writer/SKILL.md
```diff
@@ -0,0 +1,296 @@
...
+
+> Scope: work only inside `packages/ballerina-language-server`. Do **not** edit the
+> `submodules/` tree. All paths below are relative to that package.
+
+Changing the Ballerina package versions fixtures compile against is a
+self-contained sub-topic with its own rules — jump to
+[Fixture dependency versions](#fixture-dependency-versions).
```
NipunaRanasinghe, line 16: This paragraph makes the case for splitting the file: 296 lines in one `SKILL.md`, and the frontmatter description bundles two unrelated triggers (writing/debugging a test vs. bumping a fixture dependency version).

The sibling skill already establishes the convention — `.claude/skills/ballerina-e2e-writer/` keeps `SKILL.md` at 154 lines and puts the rest in `USER_GUIDE.md`.

Since the "Fixture dependency versions" section is self-contained (as this paragraph says), moving it to a reference file in the skill directory would keep `SKILL.md` focused on the task that actually triggers it.

.claude/skills/ballerina-ls-test-writer/SKILL.md
```diff
@@ -0,0 +1,296 @@
...
+
+```
+<module>/src/test/java/.../XyzTest.java          the test class - one per LSP method
+<module>/src/test/resources/<resource_dir>/
+    config/       one .json per case  <- the data provider walks this
+    source/       .bal fixtures the cases point at
+```
+
+`find_matching_type/config/` holds 11 `.json` files, and the class reports 11 tests.
+That ratio is the whole model: **a case is a file, not a method.**
```
NipunaRanasinghe, line 32: Already stale: `flow-model-generator/modules/flow-model-generator-ls-extension/src/test/resources/find_matching_type/config/` holds **12** `.json` files (`config1..11` plus `negative1.json`), not 11.

```
$ ls .../find_matching_type/config | wc -l
12
```

(The `build/resources/test/` copy still has 11, which is how a stale count survives a local run.)

The point being made — a case is a file, not a method — does not need the number. Suggest dropping the count, or giving the command that produces it.

.claude/skills/ballerina-ls-test-writer/SKILL.md
```diff
@@ -0,0 +1,296 @@
...
+
+Two `AbstractLSTest` classes exist, and they are not the same shape. 169 classes
+extend one of them directly:
+
+| base class | used by | what it gives you |
+|---|---|---|
+| `io.ballerina.modelgenerator.commons.AbstractLSTest` | the model generators (flow, service, architecture, sequence, …) — 164 classes | LS lifecycle **plus** the config walker, `getResponse`, `updateConfig`, `compareJsonElements` |
+| `org.ballerinalang.langserver.AbstractLSTest` | `langserver-core` — 5 classes | LS lifecycle and package mocking only; no config walker |
```
NipunaRanasinghe, line 40: The count is 168, not 169 — and the off-by-one comes from the verification method, which matters more than the number.

```
$ grep -rn -E "^(public |abstract |public abstract )?class [A-Za-z0-9_]+ extends AbstractLSTest" \
    --include='*.java' . | grep -v '/build/' | wc -l
168        # 163 model-generator + 5 langserver-core
```

A file-level `grep -l "extends AbstractLSTest"` returns 169 because it also matches the base class itself: `model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/AbstractLSTest.java:424` declares `protected abstract Class<? extends AbstractLSTest> clazz();`. So the table's "164 classes" should be 163, and line 34 (plus line 52) should say 168.

The PR description offers 169 as evidence that the skill was verified against the tree, so it is worth correcting there too.

Broader point: every count in this file (11 configs, 169/164/5 classes, 160/10 `updateConfig` sites, 29 pins) rots on the next commit that adds a test, and a doc that is wrong in a checkable detail loses authority on the parts a reader cannot check. Suggest replacing the counts with the commands that produce them, and keeping only the qualitative claim (two base classes, most tests use the model-generator one).

.claude/skills/ballerina-ls-test-writer/SKILL.md
```diff
@@ -0,0 +1,296 @@
...
+failure message prints. Many existing configs leave it out, and their failures read
+`Failed test: 'null' (…/config1.json)`, which tells the next reader nothing.
+
+**Register the class in the module's `testng.xml`.** The test task runs
+`useTestNG() { suites "src/test/resources/testng.xml" }`, so a class that is not
+listed there silently never runs:
+
+```xml
+<class name="io.ballerina.flowmodelgenerator.extension.typesmanager.FindMatchingTypeTest"/>
+```
```
NipunaRanasinghe, line 100: **Incomplete, and misleading for the module lines 42-45 send the reader to.** Suites also register by package, so "a class that is not listed there silently never runs" is not universally true.

- `langserver-core/src/test/resources/testng.xml:29-56` is package-based: a `<packages>` list of ~30 `org.ballerinalang.langserver.*.*` entries, plus only three explicit `<class>` entries. A new test there needs no class entry — which is exactly the module lines 42-45 route people to.
- `flow-model-generator/modules/flow-model-generator-ls-extension/src/test/resources/testng.xml:134-142` has a `<packages>` block covering seven subpackages (`modelprovidermanager`, `embeddingprovidermanager`, `vectorstoremanager`, `knowledgebasemanager`, `dataloadermanager`, `chunkermanager`, `toolkitmanager`).

Checked across all 168 subclasses: 134 are registered by explicit `<class>` entry, 34 only by a package wildcard, 0 by neither. So the hazard is real for most modules but the instruction is wrong for a third of the tree.

Suggested fix: state both mechanisms and tell the reader to check which one their target module's `testng.xml` uses before adding an entry — a redundant `<class>` entry in a package-covered suite runs the class twice.

.claude/skills/ballerina-ls-test-writer/SKILL.md
```diff
@@ -0,0 +1,296 @@
...
+   response as correct, not making a test green. `updateConfig` writes the repo's
+   committed JSON formatting, so the diff is only the semantic change.
+4. Re-run to confirm green.
+5. Re-comment before committing.
+
+**Never commit a live `updateConfig`.** It does not silence the failing run — it
+destroys the expectation, so the *next* run passes unconditionally against whatever
+the server just returned. The test goes red once and green forever after, absorbing
+any regression from then on. 160 call sites are correctly commented; 10 are
+currently live, so do not treat an uncommented one you find as precedent.
```
NipunaRanasinghe, line 134: 160 / 10 is correct as of this commit (I count 160 commented call sites, and 10 live calls across 9 classes), so the PR description's "158" is the stale figure, not this line.

The mechanism described here is right and worth documenting. Two asks:

1. Same rot problem as the other counts — "10 are currently live" will be wrong the moment one is re-commented, and a reader who counts 9 will distrust the paragraph. Consider "a few call sites are currently live" plus the grep that finds them.
2. Those 10 tests currently cannot fail. That is a real defect, and right now it exists only as prose in this PR's description — file an issue for the re-commit so it survives the merge, and link it here.

.claude/skills/ballerina-ls-test-writer/SKILL.md
```diff
@@ -0,0 +1,296 @@
...
+## 6. Debugging a failure
+
+The console shows only the assertion. **The diff goes to the test's stderr**, so
+either re-run with `-i`:
+
+```bash
+./gradlew :flow-model-generator:flow-model-generator-ls-extension:test \
+    --tests "io.ballerina.flowmodelgenerator.extension.typesmanager.FindMatchingTypeTest" -i
+```
+
```
NipunaRanasinghe, line 166: The stderr half is verified — an existing `build/test-results/test/TEST-…AvailableNodesTest.xml` carries the diff inside `<system-err>`:

```
INFO: - Value mismatch at '[0].items[0].items[0].codedata.version'
  actual: "1.2.3"
  expected: "1.0.1"
```

(`log` is slf4j at `AbstractLSTest.java:64,85`, bound to `slf4j-jdk14` per `build.gradle:279`, so JUL writes it to stderr.)

But "the console shows only the assertion" looks wrong: `gradle/javaProject.gradle:99-104` sets `showStandardStreams = true`, and this module applies that file (`flow-model-generator-ls-extension/build.gradle:19`). Gradle adds `STANDARD_ERROR` to the logged events and prints it at the default log level, so the diff should already be on the console without `-i`.

Worth confirming with one deliberately-failing run. If it does print by default, drop the `-i` step and keep the `grep` on the results XML as the recipe for a run that has already finished — that part is genuinely useful and I confirmed it works.

.claude/skills/ballerina-ls-test-writer/SKILL.md
```diff
@@ -0,0 +1,296 @@
...
+- **`Failed to initialize the compiler plugin in package: ... "ctxData" is null`**
+  is a known flaky compiler-plugin initialisation failure, not your change. It moves
+  between tests run to run. Re-run the class alone before investigating.
+- **`failed to connect to the docker API at unix:///var/run/docker.sock`** is
+  harmless noise; it prints on fully passing runs too.
+- **A golden that passes locally but fails on CI** usually means the local run was
+  green off a stale distribution. `copyStdlibs` copies into
+  `build/extracted-distribution` without deleting, so old package versions
+  accumulate there. List what you have, then wipe and re-run:
+
```
NipunaRanasinghe, line 196: The accumulation hazard is real, but the path is not universal — only the newer modules use `extracted-distribution`:

- `flow-model-generator/modules/flow-model-generator-ls-extension/build.gradle:93`, `flow-model-generator-core/build.gradle:104`, `xsd-service/…/build.gradle:95` → `build/extracted-distribution/jballerina-tools-<ver>`
- `misc/diagram-util/build.gradle:106` → `build/jballerina-tools-<ver>` (no `extracted-distribution` at all)
- `misc/ballerinalang-data-mapper/build.gradle:66` → `extracted-distributions` (plural)

The `find` command below handles all three because it matches on directory name, so only the prose and the table row on line 250 need to stop stating one path as universal.

.claude/skills/ballerina-ls-test-writer/SKILL.md
```diff
@@ -0,0 +1,296 @@
...
+  ```bash
+  find . -type d \( -name extracted-distribution -o -name 'jballerina-tools-*' \) -prune -print
+  ```
+
+  Delete those directories plus `build/ballerina_dependencies`, then re-run the
+  class — the next build re-extracts and re-provisions both.
+
+- **Never run two Gradle builds against this project at once**, and do not edit
+  sources while a suite is running. Both silently corrupt the result.
```
NipunaRanasinghe, line 205: This one is unsupported, and it is the only claim in the file with no traceable basis in the tree. Gradle takes a lock on the project and per-build-directory state; a second concurrent invocation normally blocks with "Waiting to acquire…" or fails outright rather than silently producing a wrong result.

Either cite the failure that was actually observed (which task, what the corruption looked like) or soften it to what is defensible — e.g. that these tasks write into a shared `build/extracted-distribution` and `build/ballerina_dependencies`, so a concurrent build can extract over a distribution another run is reading.

Same for "do not edit sources while a suite is running": true in a general sense, but as written it reads as a project-specific hazard the reader should have a reason to believe.

.claude/skills/ballerina-ls-test-writer/SKILL.md
```diff
@@ -0,0 +1,296 @@
...
+
+## 7. Tests must not reach Ballerina Central
+
+Every `Test` task in every module gets `-Dls.test.offline=true` and
+`BALLERINA_HOME_DIR` pointed at the build-owned Ballerina home, and depends on
+`resolveBallerinaDependencies` to populate it first (`build.gradle`, `allprojects`
+block). `PackageUtil.FORCE_OFFLINE` and `RemoteCentral.getInstance()` read that
+property; the latter returns an `OfflineCentral` that refuses every call.
+
```
NipunaRanasinghe, line 214: `PackageUtil.FORCE_OFFLINE` does not exist — a reader who greps for it finds nothing. The actual symbols:

- `langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java:107` — `public static final boolean TEST_OFFLINE = Boolean.getBoolean("ls.test.offline");` (this is the one that reads the property)
- `model-generator-commons/…/PackageUtil.java:79` — `public static boolean isOffline()`, which just returns `CommonUtil.TEST_OFFLINE`
- `flow-model-central-client/…/RemoteCentral.java:46` — `private static final boolean OFFLINE = Boolean.getBoolean("ls.test.offline");`, which is what makes `getInstance()` hand back `OfflineCentral`

The rest of the section checks out: `build.gradle:397-401` does set `ls.test.offline`, point `BALLERINA_HOME_DIR` at `ballerinaTestHomeDir`, and depend on `resolveBallerinaDependencies` in the `allprojects` block, and the verification snippet's path matches `homeBalaDir` in `build.gradle` (`repositories/central.ballerina.io/bala`).

.claude/skills/ballerina-ls-test-writer/SKILL.md
```diff
@@ -0,0 +1,296 @@
...
+1. Edit the `[[dependency]]` version in
+   `build-config/ballerina_dependencies/Ballerina.toml`.
+2. Regenerate the lock in place:
+
+   ```bash
+   cd build-config/ballerina_dependencies && bal build
+   ```
+
+   Jar-conflict and `provided`-scope warnings in that output are normal.
```
NipunaRanasinghe, line 266: This procedure diverges from what the build actually does, in a way that contradicts section 7 of this same file.

`resolveBallerinaDependencies` (`build.gradle:134-185`) copies `build-config/ballerina_dependencies` to `build/ballerina_dependencies`, sets `BALLERINA_HOME_DIR` to the build-owned home, and runs `bal build --sticky` there. Running `bal build` in the source directory instead:

- resolves against the developer's machine-wide `~/.ballerina` cache — the exact thing `ballerinaTestHomeDir` exists to avoid ("test runs never read from or write to the developer's machine-wide package cache", per the comment above `build.gradle:141`);
- drops a `target/` directory inside `build-config/ballerina_dependencies` (harmless for commits — `.gitignore:44` covers `target/` — but it is left behind);
- omits `--sticky` (`sticky = true` is in `[build-options]`, so probably equivalent, but the two commands are no longer obviously the same).

Suggest documenting the command with `BALLERINA_HOME_DIR` set the way the Gradle task sets it, or pointing at `./gradlew resolveBallerinaDependencies` and describing how to get the regenerated `Dependencies.toml` back into `build-config/`. As written, the lock a contributor commits was produced by a different resolution than the one CI performs.

.claude/skills/ballerina-ls-test-writer/SKILL.md
```diff
@@ -0,0 +1,296 @@
...
+version per package, so a second version can only have been downloaded.
+
+```bash
+H=build/ballerina_dependencies/home/repositories/central.ballerina.io/bala
+for pkg in "$H"/*/*; do
+  [ "$(ls "$pkg" | wc -l)" -gt 1 ] && echo "PULLED: $pkg -> $(ls "$pkg" | tr '\n' ' ')"
+done
+```
+
+Silence means clean.
```
NipunaRanasinghe, line 228: Style nit, applies in a few places: "Silence means clean", "That ratio is the whole model" (line 32), "so the fixtures stay honest" (line 296), "you are approving the server's response as correct, not making a test green" (line 121).

The mechanical statements in this file are its strength — this register is editorial and adds nothing a reader acts on. Prefer the literal behaviour: "No output means no package has two versions."

NipunaRanasinghe (changes requested):

NipunaRanasinghe:
One general point, since it covers three of the inline comments: I'd drop the exact counts. "11 config files", "169 classes", "29 pins", "160 commented / 10 live" all go stale on commits that have nothing to do with this file, and a reader can't tell when they have. Counts that are part of the design are fine to keep — two `AbstractLSTest` classes, four hooks — because those only change when the structure changes.

This matters more in a skill than in normal docs, since it gets read as fact by an agent that won't check it. The 169 came from a grep that also matched `Class<? extends AbstractLSTest>` in the base class, and from there it went into the PR description as evidence the skill had been verified.

Where a count is actually doing work, the command that produces it works better: `grep -rn 'updateConfig(' --include='*.java' | grep -v '//'` says more than "10 are live", and for that one a grep in `check` would beat a warning in a doc.

.claude/skills/ballerina-ls-test-writer/SKILL.md
pasindufernando1: Restored in b7bc3d9037. My local branch was 360 commits behind the pushed head, which is why I was reading the pre-merge file.

Rewritten for the 01ff87ece8 layout, pointing at `docs/LS_FIXTURE_DEPENDENCIES.md`. Description fixed too.

.claude/skills/ballerina-ls-test-writer/SKILL.md
pasindufernando1: Done in b7bc3d9037 — content is `docs/LS_TEST_GUIDE.md` + `docs/LS_FIXTURE_DEPENDENCIES.md`, linked from `docs/TEST_GUIDE.md`, `CONTRIBUTING.md` and `AGENTS.md`. `SKILL.md` is now a pointer.

.claude/skills/ballerina-ls-test-writer/SKILL.md
pasindufernando1: Split: dependency versions are their own file, `SKILL.md` is 296 → ~60 lines.

.claude/skills/ballerina-ls-test-writer/SKILL.md
pasindufernando1: Dropped. Both counts were right — `negative1.json` postdates the checkout I wrote against, which rather makes your point.

.claude/skills/ballerina-ls-test-writer/SKILL.md
pasindufernando1: Dropped, with the 164/5 split. Table says "the large majority" / "langserver-core", grep alongside. The `Class<? extends AbstractLSTest>` match is exactly how 169 reached the description.

.claude/skills/ballerina-ls-test-writer/SKILL.md
pasindufernando1: Fixed — both mechanisms documented, read your module's `testng.xml` first, and both failure modes stated (missing entry never runs, redundant entry runs twice).

.claude/skills/ballerina-ls-test-writer/SKILL.md
pasindufernando1: Both done.

1. Counts gone, replaced by `grep -rnE '^[[:space:]]*updateConfig\('` — anchored at line start, so neither the commented form nor the declaration matches. Returns 11 now, so the number moved during review.
2. #932.

.claude/skills/ballerina-ls-test-writer/SKILL.md
pasindufernando1: You were right — `-i` dropped. `javaProject.gradle:101` sets `showStandardStreams = true`, so the diff already prints. Results-XML grep kept for finished runs.

.claude/skills/ballerina-ls-test-writer/SKILL.md
pasindufernando1: Fixed — all three layouts named, `find` widened to `'extracted-distribution*'`, table row no longer claims one path.

.claude/skills/ballerina-ls-test-writer/SKILL.md
pasindufernando1: Softened to the shared build state that actually collides (module test distributions, root `build/ballerina_dependencies`). Dropped the "do not edit sources" line — no basis for it.

.claude/skills/ballerina-ls-test-writer/SKILL.md
pasindufernando1: Fixed. `FORCE_OFFLINE` is gone at HEAD entirely — upstream replaced it with `isOffline()`. Guide now tables all three: `CommonUtil.TEST_OFFLINE` reads the property, `PackageUtil.isOffline()` delegates, `RemoteCentral.OFFLINE` swaps in `OfflineCentral`.

.claude/skills/ballerina-ls-test-writer/SKILL.md
pasindufernando1: Rewritten to use `./gradlew resolveBallerinaDependencies` plus copying the lock back.

Found one more trap doing it: the task hashes the committed `Dependencies.toml`, which a `Ballerina.toml` bump does not change — so it runs, logs `Dependencies.toml unchanged — skipping dependency resolution`, and regenerates nothing unless `build/ballerina_dependencies` is cleared first. Now documented.

.claude/skills/ballerina-ls-test-writer/SKILL.md
pasindufernando1: All four replaced with the literal behaviour.

.claude/skills/ballerina-ls-test-writer/SKILL.md
pasindufernando1: #932 — includes the `check` guard suggestion.

pasindufernando1:
Pushed b7bc3d9037. All 13 threads answered inline.

- Content moved to `docs/LS_TEST_GUIDE.md` + `docs/LS_FIXTURE_DEPENDENCIES.md`, linked from `docs/TEST_GUIDE.md`, `CONTRIBUTING.md`, `AGENTS.md`. `SKILL.md` 296 → ~60 lines.
- `gradle.properties` restored — my local branch was 360 commits behind the pushed head.
- Counts replaced by the commands that produce them. Two of them had already rotted, which settles the argument.
- #932 filed for the live `updateConfig` calls.

Also found: `resolveBallerinaDependencies` hashes the committed `Dependencies.toml`, so a `Ballerina.toml` bump skips resolution and regenerates nothing. Documented.

@coderabbitai nitpicks covered — fences labelled, commands root-relative, try-with-resources in the example, `updateConfig` shown as failing the current run, precedence claim scoped.

docs/LS_FIXTURE_DEPENDENCIES.md
```diff
@@ -0,0 +1,99 @@
...
+the versions the shipped extension resolves.
+
+For writing and debugging the tests themselves, see
+[LS_TEST_GUIDE.md](LS_TEST_GUIDE.md). **All paths below are relative to
+`packages/ballerina-language-server`.**
+
+## The one rule
+
+**Fixture dependency versions live in `build-config/ballerina_dependencies/`. Never
+change them by editing a `stdlib*Version` pin in `gradle.properties`.**
```
NipunaRanasinghe: This rule doesn't hold for packages that are also `stdlib*Version` pins, and `ai` — the package from the incident that triggered this PR — is one of them.

`gradle.properties:91` has `stdlibAiVersion=1.11.2`; the lock has `ballerina/ai = 1.13.0`. Both versions are recorded in fixtures of the same module: `agents_manager/config/agent_template_ballerina.json` expects `ballerina:ai:1.11.2`, its siblings expect `1.13.0`. So neither layer owns the version uniformly, and someone who follows this rule (edit the lock, regenerate) can bump the lock and see nothing change on the paths that resolve through the distribution.

Worth adding a check before the bump procedure: grep `build.gradle` for `ballerinaStdLibs ".*:${stdlib*Version}"` to see whether the package is injected at all, and if it is, say that fixtures may record the injected version instead. The `ai` split is a good worked example, since it's the case the doc exists to explain.

docs/LS_TEST_GUIDE.md
```diff
@@ -0,0 +1,266 @@
...
+
+| base class | used by | what it gives you |
+|---|---|---|
+| `io.ballerina.modelgenerator.commons.AbstractLSTest` | the model generators (flow, service, architecture, sequence, …) — the large majority of tests | LS lifecycle **plus** the config walker, `getResponse`, `updateConfig`, `compareJsonElements` |
+| `org.ballerinalang.langserver.AbstractLSTest` | `langserver-core` | LS lifecycle and package mocking only; no config walker |
+
+```bash
+# every class extending a base directly (read its import to see which of the two)
+grep -rnE 'class [A-Za-z0-9_]+ extends AbstractLSTest' --include='*.java' \
+  packages/ballerina-language-server | grep -v '/build/'
```
NipunaRanasinghe, line 51: Line 14 says all paths are relative to `packages/ballerina-language-server`, but this command (and the `updateConfig` grep at line 163) is repo-root-relative and fails from that directory. The `./gradlew` commands are the other way round — there is no `gradlew` at the repo root, so they only work from inside the package.

Both work, they just need different working directories. Easiest fix is a `cd packages/ballerina-language-server` line in the gradle blocks and leaving the greps root-relative, or labelling each block.

docs/LS_TEST_GUIDE.md
```diff
@@ -0,0 +1,266 @@
...
+```
+
+Everything below describes the model-generator base. In `langserver-core`, extend
+the feature-specific abstract test instead (`CompletionTest`,
+`AbstractCodeActionTest`) — those declare their own `dataProvider()` and
+`getTestResourceDir()`, still over a `config` directory.
```
NipunaRanasinghe: `CompletionTest` declares both `dataProvider()` and `getTestResourceDir()`; `AbstractCodeActionTest` declares `dataProvider()` and `getResourceDir()` (line 386), not `getTestResourceDir()`. Small thing, but someone extending the code-action base will look for the wrong method name.

docs/LS_TEST_GUIDE.md
```diff
@@ -0,0 +1,266 @@
...
+
+For the Ballerina package versions the test fixtures compile against, see
+[LS_FIXTURE_DEPENDENCIES.md](LS_FIXTURE_DEPENDENCIES.md).
+
+## 1. How an LS test is shaped
+
+A test is **data-driven**: it does not declare its own cases. The base class walks a
+config directory and feeds every `*.json` in it to a single `test(Path config)`
+method, one TestNG invocation per file. A case is a file, not a method — the number
+of tests a class reports is the number of `.json` files in its config directory.
```
NipunaRanasinghe: Two details: the provider uses `Files.walk`, so nested directories under `config/` are picked up too, and anything in `skipList()` is filtered out. So it's every `.json` found recursively, minus the skip list.

docs/LS_TEST_GUIDE.md
```diff
@@ -0,0 +1,266 @@
...
+
+```java
+public class FindMatchingTypeTest extends AbstractLSTest {
+
+    @Override
+    @Test(dataProvider = "data-provider")
+    public void test(Path config) throws IOException {
+        Path configJsonPath = configDir.resolve(config);
+        TestConfig testConfig;
+        try (BufferedReader reader = Files.newBufferedReader(configJsonPath)) {
```
NipunaRanasinghe, line 86: The real `FindMatchingTypeTest` doesn't use try-with-resources — it's `gson.fromJson(Files.newBufferedReader(configJsonPath), TestConfig.class)` on one line. The version here is better, but the text says "modelled on `FindMatchingTypeTest`", so a reader who opens that file finds something else. Either note that the sample is cleaned up, or fix the class and keep them in sync.

docs/LS_TEST_GUIDE.md
```diff
@@ -0,0 +1,266 @@
...
+- a `<packages>` block that picks up every class in a package. `langserver-core` is
+  almost entirely package-based, and
+  `flow-model-generator-ls-extension` covers several subpackages this way.
+
+Read your target module's `testng.xml` before adding anything. In a suite that lists
+classes explicitly, a class you forget to add **silently never runs**. In a
+package-covered suite, a redundant `<class>` entry makes the class run twice.
+
+```bash
+grep -n '<packages>' -A20 <module>/src/test/resources/testng.xml
```
NipunaRanasinghe: `-A20` truncates the block you most want to read: `langserver-core` lists 30 `<package>` entries. `sed -n '/<packages>/,/<\/packages>/p'` prints whatever is there and won't rot.

docs/LS_TEST_GUIDE.md
```diff
@@ -0,0 +1,266 @@
...
+If you add a resolution path, go through `PackageUtil`. Do not read the property
+yourself and do not construct a `CentralAPIClient` directly.
+
+To check a run stayed offline: the home is provisioned from a lock with exactly one
+version per package, so a second version can only have been downloaded.
+
+```bash
+H=build/ballerina_dependencies/home/repositories/central.ballerina.io/bala
+for pkg in "$H"/*/*; do
+  [ "$(ls "$pkg" | wc -l)" -gt 1 ] && echo "PULLED: $pkg -> $(ls "$pkg" | tr '\n' ' ')"
```
NipunaRanasinghe: The path is right (`build.gradle:158` builds the same one) and the org/package glob depth is right. Two rough edges: if the home hasn't been provisioned the glob stays literal and `ls` errors, and `[ ... ] && echo` leaves a non-zero exit when the last package has one version — fine when pasted into a shell, breaks if anyone drops it into a `set -e` script. A `[ -d "$H" ] || exit` guard and an `if` instead of `&&` would make it safe to reuse.

docs/LS_FIXTURE_DEPENDENCIES.md
```diff
@@ -0,0 +1,99 @@
...
+- `"version": "<version>"`
+- `"packageInfo": "<org>:<pkg>:<version>"`
+- icon URLs — `ballerina_<pkg>_<version>.png`
+
+So grep for the **old** version before concluding nothing refers to it. From the
+repository root:
+
+```bash
+grep -rn 'ballerina_http_2\.16\.3\|"2\.16\.3"' --include='*.json' \
+  packages/ballerina-language-server
```
NipunaRanasinghe, line 126: `2.16.3` reads as a current value; committed `http` is `2.16.6`. A placeholder (`<old-version>`) avoids the same rot the counts had.

NipunaRanasinghe (commented):
Went through this against the code rather than reading it, so first the parts that check out — this revision is much better verified than the last one:

- The offline story is accurate: `CommonUtil.TEST_OFFLINE` (`CommonUtil.java:107`), `PackageUtil.isOffline()` (`PackageUtil.java:79`), `RemoteCentral.OFFLINE` (`RemoteCentral.java:46`), and `build.gradle:398-401` setting both `BALLERINA_HOME_DIR` and `ls.test.offline` on every `Test` task.
- `AbstractLSTest` matches what sections 1, 3, 4, 5 and 6 say, including the fallback when `config/` is missing (`:78`), the `.json` filter (`:113`), `updateConfig` writing after the comparison has already run, the `filePath` and `icon` normalisation (`:293-299`), and all three diff messages quoted word for word.
- `resolveBallerinaDependencies` behaves as described — hash over the committed `Dependencies.toml`, the `skipping dependency resolution` line, and `rm -rf build/ballerina_dependencies` as the way out.
- The `showStandardStreams` claim is correct and not obviously so: `javaProject.gradle` sets it before `events "failed"`, and Gradle's `events(...)` method adds to the set rather than replacing it, so it survives. Dropping the `-i` advice was right.
- All three distribution directory layouts, the `ballerinaStdLibs` → `unpackStdLibs` → `copyStdlibs` chain and the copy-without-delete accumulation, `GetServiceInitModelTest` as the skip-list example, `lockingMode = "LOCKED"` plus the `as _` imports, and gson's pretty printing matching the committed 2-space JSON all hold. Every embedded command runs.

One thing I'd like fixed before merge: the "one rule" in `LS_FIXTURE_DEPENDENCIES.md` doesn't hold for packages that are also `stdlib*Version` pins, and `ai` is one — `stdlibAiVersion=1.11.2` against a locked `1.13.0`, with fixtures in one module expecting each. Details inline. That's the exact failure this doc was written to prevent, so it's worth naming.

The rest is small: mixed working directories between the greps and the `./gradlew` commands, and a handful of precision fixes.

Two claims in the PR description that don't hold, for the record rather than for the diff:

- "every command verified to run from the repository root" — not possible for the `./gradlew` ones, there's no `gradlew` at the repo root, nor for the `build/...` paths.
- "follows the split the sibling `ballerina-e2e-writer` skill already established" — that skill is 154 lines and links to no docs at all, so there's no split to follow. Moving the content to `docs/` is still the right call, just not a precedent.
- Minor: the description says 11 classes call `updateConfig` uncommented; the doc's own grep returns 10 call sites on the base branch, and they're call sites rather than classes. Doesn't affect the docs, which carry no counts now.

Agree with your own conclusion on #932 — a grep in `check` would do more than the warning in the guide.

docs/LS_FIXTURE_DEPENDENCIES.md
pasindufernando1: Fixed in 1063b70f7c. Section is now "The rule, and where it stops", with the injected-package check (`grep -n 'ballerinaStdLibs .*<package>-ballerina' build.gradle`) before the bump procedure and `ai` as the worked example — `1.11.2` injected at `build.gradle:107` against `1.13.0` locked, with `agent_template_ballerina.json` and `agent_call_flow_node_8/9.json` in one directory expecting each. Table row softened from a flat **no** too.

docs/LS_TEST_GUIDE.md
pasindufernando1: Fixed. Header no longer claims a single working directory, and all 11 command blocks now name theirs — including the three indented in the bump procedure, which I missed first time.

docs/LS_TEST_GUIDE.md
pasindufernando1: Fixed — `AbstractCodeActionTest:386` is `getResourceDir()`, `CompletionTest:118` is `getTestResourceDir()`. Now a two-row table rather than one name.

docs/LS_TEST_GUIDE.md
pasindufernando1: Fixed in §1 and §4: every `.json` found recursively, minus `skipList()` and dot-files.

docs/LS_TEST_GUIDE.md
pasindufernando1: Fixed — "adapted from" now, and the doc says outright that the committed version reads the config on one line without closing the reader. Left the test class alone as out of scope for a docs PR.

docs/LS_TEST_GUIDE.md
pasindufernando1: Fixed — `sed -n '/<packages>/,/<\/packages>/p'`. Verified it prints all 30 entries.

docs/LS_TEST_GUIDE.md
pasindufernando1: Both fixed — `[ -d "$H" ]` guard and `if` instead of `&&`. Tested the unprovisioned path and under `set -e`.

docs/LS_FIXTURE_DEPENDENCIES.md
pasindufernando1: Fixed — `OLD='<old-version>'` placeholder, with the dot-escaping done by expansion.

pasindufernando1:
Pushed 1063b70f7c. All 8 threads answered inline.

The `ai` finding was the important one — it's the exact case the doc was written for and the rule as written would have sent someone to bump the lock and watch half the fixtures not move. That section now says where the rule stops, with the injected-package check before the bump procedure.

Description corrected on all three counts: the root-relative claim, the `ballerina-e2e-writer` precedent, and the `updateConfig` classes-vs-call-sites number.

PR #822, opened 2026-08-10
Add the integration from the project chooser, skipping the wizard

gigara:
The project overview's **Add Integration or Library** button still handed off to the three-step integration wizard (Name → Type → Configure). The welcome-view Create flow stopped doing that in b17d7915e8 (#806), so the two entry points to the same operation disagreed: one screen from Create, four from the project overview.

Give the project overview's Add button the same single-screen experience. Adding a library is unchanged.

Same collapse #806 applied to `CreateProjectChooser`, applied to `AddProjectForm`, which had the identical three-screen shape (`chooser` → `integration` → `library`):

- The `integration` screen that mounted `CreateIntegrationWizard` is gone. The wizard itself is untouched and still reachable from `BIProjectForm`, `EmbeddedBIProjectForm`, `StandaloneCreateChooser` and `AddIntegrationPanel`.
- The name is asked for inline on the chooser when the starting point is an integration, and submit goes through the **same `addProjectToWorkspace` call the library route already used** — with no artifact, which is the payload the wizard's "Create Empty Integration" skip sent. Convert-and-add still works, as "Convert & Add Integration".

**Both routes now share one component.** `AddLibraryFields` already did everything the integration needs — collision-checked naming against the target project, default-name indexing, the folder/package-name coupling, the debounced diagnostic — and differed only in its label and in owning the package details. It becomes `AddComponentFields`, reading the existing `formData.isLibrary` for both, and renders Advanced Configurations for the library alone. Two things the integration route needs that the library never did, because it renders on the chooser rather than on a screen the user reaches once the project is settled:

- **The target project can still move under it.** The convert flow edits the project name and location on that same screen, so the component keys its listing on the resolved path instead of seeding once. Equivalent for the library, whose path cannot change while it is mounted.
- **Its default name would race the parent's suggested defaults** over the same `integrationName`, so seeding waits for them. Released even when that call fails, so a failure cannot strand the field.

Submit-time name rejections have nowhere to go on this route — there is no package field to hang them off — so they are reported on the name field itself, and the live diagnostic wins once it is edited again.

Also drops the project-structure lookup that only named the project in the wizard's progress copy, now that nothing reads it.

## UI Component Development
- [x] Added reusable UI components to the ui-toolkit — N/A. No new component: this reuses `TextField` and merges two existing local form components into one.
- [x] Use ui-toolkit components wherever possible — yes, `TextField` from the toolkit; layout uses the existing local `FieldGroup`/`FormSection`.
- [x] Matches with the native VSCode look and feel — unchanged. The field is the same `TextField` the wizard's Name step used, in the same shell.

As a developer adding an integration to an existing project, I want to name it and have it created from one screen, the same way the Create flow already works, instead of stepping through a wizard whose remaining steps I skip.

Adding an integration from the project overview is now a single screen: the integration is named alongside the project and created directly, instead of continuing into the three-step integration wizard. Adding a library is unchanged.

N/A — no documented behaviour changes; this shortens an existing flow and collects the same field the wizard's first step did. Any screenshots of the Add flow's integration path would be stale, matching the existing staleness from #806 on the Create path.

## Training
N/A

## Certification
N/A — no concept or capability is added or removed; the same operation takes fewer steps.

## Marketing
N/A

## Automation tests
 - Unit tests
   > None added. The change is form wiring and component composition in the webview; this area has no existing unit-test harness (`ballerina-visualizer` has no tests over `ProjectForm`), and the logic that carries risk — collision-checked naming and default indexing — is the code path `AddLibraryFields` already shipped, now shared rather than rewritten. Host-side suite still green at 112 passing.
 - Integration tests
   > None added. Labelled `Checks/Run Ballerina UI Tests` so the existing UI suite covers it. Note the repo's own `project-creation.spec.ts` is currently `test.skip`ped pending a product-integrator release, so the Add flow is not covered end-to-end there either way.

## Security checks

## Samples
N/A

## Related PRs
- #806 — collapsed the same wizard hand-off in the welcome-view Create flow. This PR brings the project overview's Add button in line with it.

## Migrations (if applicable)
N/A — no persisted state or config.

macOS 15 (darwin 25.5.0), Node 20. Verified by typechecking `ballerina-visualizer` and `ballerina-extension` and a full production webpack build of the visualizer.

**Not yet exercised in a running extension.** Four paths worth clicking before merge: add an integration to an open project; add a library (should still go to its own screen via "Next"); convert-and-add an integration; and a plain convert with nothing added. Two specific things to watch — `autoFocus` on the name field is now conditional on `isLibrary`, because on the chooser the convert flow's Project Name field already claims focus; and the collision diagnostic should re-evaluate when the convert flow's project location is edited.

The first cut of this added a second component beside `AddLibraryFields` and re-implemented its naming logic. Diffing the two showed they differed only in an import block and one JSX section, so it became one component instead — 213 lines lighter, and the naming behaviour now has a single definition rather than two that can drift.

Worth noting for anyone touching this next: the collision check needs `getProjectComponentNames`, which is only exposed on the WS bridge, so the field has to sit under a `BiWsClientProvider`. That is why it is a component rather than inlined into the chooser — a hook call cannot live in the component that renders its own provider.

## Summary

- Simplifies integration creation by collecting the integration name on the project overview screen.
- Submits integrations through `addProjectToWorkspace` without requiring an artifact or editable organization name.
- Omits empty organization handles so project creation can use the fallback username.
- Preserves convert-and-add and library creation flows.
- Replaces `AddLibraryFields` with shared `AddComponentFields` logic for integrations and libraries.
- Retains name collision checks, default-name indexing, folder and package-name coupling, and debounced diagnostics.
- Keeps Advanced Configurations available only for libraries.
- Improves validation when component types or project paths change.
- Releases initialization state after setup failures.
- Updates add-to-project guidance to reflect the separate integration and library flows.


packages/ballerina-visualizer/src/views/BI/ProjectForm/AddProjectForm.tsx
```diff
@@ -472,11 +437,56 @@ export function AddProjectForm() {
...
+                    <>
+                        {pathValidationError && <SubmitError>{pathValidationError}</SubmitError>}
+                        <Button
+                            disabled={
+                                nextDisabled ||
+                                !!componentNameError ||
+                                !isFormValidAddProject(formData, isInProject, addNewAfterConvert) ||
+                                isLoading ||
+                                (isConvert && !!convertPathError)
+                            }
```
NipunaRanasinghe, line 521: This gates the integration route's only submit button on the organization name, but the route renders no organization field and suppresses the organization error, so every state that fails the check disables the button with nothing on screen to explain it and no field to correct it.

`isFormValidAddProject` requires `validateOrgName(formData.orgName) === null`. On this route `AddComponentFields` is rendered without `isOrgDataLoaded` and `organizations`, so `orgNameError` is computed as `null` (`(!isOrgLocked && isOrgDataLoaded) ? … : null`) and Advanced Configurations is not rendered at all. Three states reach it:

1. `formData.orgName` is `""` until `getDefaultOrgName` resolves. This happens on every open of the form. Transient, but the button is dead for that window with no indication why.
2. The resolved value fails `validateOrgName` — the reserved `wso2`/`ballerina`/`ballerinax`, or any value containing a dot or hyphen, or starting with a digit. Permanently disabled. Worth confirming what `getDefaultOrgName` returns on a machine where the project's org handle is `wso2`.
3. `getDefaultOrgName` rejects while `platformExtState.userInfo.organizations` is empty or undefined. `useDefaultOrgName`'s catch branch calls `onOrgResolved` only when the list is non-empty, so `orgName` stays `""`. Permanently disabled.

The library route survives all three: it passes `isOrgDataLoaded`, so `hasAdvancedConfigError` auto-expands Advanced Configurations and the user can type an organization name. This route has no equivalent.

Two ways out, either is fine:

- Drop the organization term from this button's `disabled` expression. The organization is not editable here, and `addProjectToWorkspace` is already called with `orgName: formData.orgName || undefined`, so an unset value is not fatal to the request.
- Render the organization error as a `SubmitError` in the footer beside `pathValidationError`, so at least the reason is visible.

The first is simpler and matches the fact that this route deliberately collects no package details.

packages/ballerina-visualizer/src/views/BI/ProjectForm/AddProjectForm.tsx
```diff
@@ -472,11 +437,56 @@ export function AddProjectForm() {
...
+                            formData={formData}
+                            onFormDataChange={handleFormDataChange}
+                            defaultsReady={defaultsReady}
+                            projectPath={componentTargetPath}
+                            reservedFolders={currentIntegrationDirName ? [currentIntegrationDirName] : undefined}
+                            nameTouched={nameTouched}
+                            onNameTouched={markNameTouched}
+                            nameValidationError={componentNameValidationError || undefined}
+                            onNameErrorChange={setComponentNameError}
+                        />
```
NipunaRanasinghe, line 494: This is the other half of the organization issue on the submit button below.

`isOrgDataLoaded` and `organizations` are not passed here, which is deliberate — this route renders no Advanced Configurations — but the effect is that `orgNameError` inside `AddComponentFields` evaluates to `null` unconditionally, so an invalid organization name cannot be reported anywhere on this screen while still blocking submit through `isFormValidAddProject`.

Whichever fix is chosen on the button, no change is needed in this component: keeping the organization out of the integration route is the right call, and the gate is what should move.

packages/ballerina-visualizer/src/views/BI/ProjectForm/AddProjectForm.tsx
gigara: Confirmed and fixed in 17cc5bad2b. All three states reproduce, and state 2 is the one that worries me: `useDefaultOrgName` passes the resolved handle through verbatim in the `isInProject && isLocked` branch, so a project whose org handle is `wso2` hits `RESERVED_ORG_NAMES` and disables the button permanently, with no field to correct it. That is the likely case for this repo's own users, not a corner.

Took your first option. `isFormValidAddProject` now takes `{ requireOrgName }`, and the integration route passes `false` — the gate moves, the component does not, as you said on the other thread.

**One correction, because it matters if anyone applies the suggestion as written:** an unset organization *is* fatal to the request. `createBIProjectPure` writes

```ts
org = "${orgHandle ?? finalOrgName}"
```

and `??` does not fall through on `""`. `orgHandle` is `sanitizeOrgHandle(formData.orgName)`, which returns `""` for an empty input, so dropping the gate on its own would have traded a dead button for `org = ""` in the generated `Ballerina.toml` — the `finalOrgName` username fallback never gets a chance to apply. It is now sent as `orgHandle: orgHandle || undefined`, which is only reachable at all now that submit is no longer gated on the organization.

So the fix is both halves: stop validating what the route cannot show, and stop sending empty what the backend cannot default.

packages/ballerina-visualizer/src/views/BI/ProjectForm/AddProjectForm.tsx
gigara: Agreed, and no change here. Keeping the organization out of the integration route is deliberate — it collects no package details at all — so `AddComponentFields` still receives no `isOrgDataLoaded`/`organizations` and still renders no Advanced Configurations.

Your diagnosis of the mechanism is what made the button fix obvious: `orgNameError` is `(!isOrgLocked && isOrgDataLoaded) ? … : null`, so on this route it is unconditionally `null` and there is nowhere for the error to appear. The gate was the thing in the wrong place. Fixed on the button in 17cc5bad2b.

gigara:
@coderabbitai — both nitpicks addressed, one fixed and one skipped.

**`defaultsReady` on a discovery failure — valid, fixed in 17cc5bad2b.** You are right that the `finally` covered only `getSuggestedProjectDefaults`. A rejection from `getWorkspaceRoot` or `getWorkspaceType` left the flag `false` forever, so the inline integration name field never seeded its default — the exact stranding the flag exists to prevent. The whole initialization sequence is now one `try`/`finally`, with the outer `catch` logging the discovery failure rather than swallowing it silently.

**Propagating `addProjectToWorkspace` failures — skipping, out of scope.** The `void … .catch((): undefined => undefined)` is not introduced here; it is unchanged from before this PR (`git show HEAD~1` on the same file has it verbatim) and is shared by the library and convert routes, so changing it would alter behaviour on two routes this PR does not touch.

It also looks deliberate rather than accidental: on success the extension reloads or re-navigates the window, so awaiting the call can mean awaiting something that never settles — which is presumably why it was written fire-and-forget rather than with a `finally` that clears `isLoading`.

That said, your underlying point stands: a failed create leaves the form spinning with no error, and the integration route now inherits that. It deserves a fix that covers all three routes and reasons about the reload, which is its own change rather than a rider on this one.

---

Also for the record, since it will look related: the `Run Ballerina e2e Test (group1)` failure on this PR was `flow-nodes.spec.ts`, unrelated to these files, and it passed on re-run. The panel dump showed `undefined symbol 'msg'` — a prior step's `form.fill` cmEditor no-op, which is the slow-CI failure mode the spec itself documents at line 88.

NipunaRanasinghe (dismissed):
LGTM

NipunaRanasinghe (approved):

PR #823, opened 2026-08-10
Add an artifact from the empty integration, not a second wizard

gigara:
The integration overview offered two different ways to add to an integration, depending on whether it had anything in it yet:

- **Add Artifact**, in the Design header, once artifacts existed — goes to the artifact list;
- **Add Integration**, in the empty state — opened the Create Integration wizard *in place of the overview* to collect an artifact type and configuration.

Same intent, two entry points, two destinations, and the label on the second one described creating an integration rather than adding to the one already open.

image

One way to add an artifact, reachable whichever state the integration is in.

The empty state now offers **Add Artifact** in the spot Add Integration occupied, calling the same handler the header button already used. The header button is untouched — still shown only when the integration has artifacts — so exactly one is on screen in either state.

That removes the last caller of the wizard-in-place path, so it goes too: the `showAddIntegration` state, the `closeAddIntegration` callback, the lazy import, and the render branch that swapped the overview out for the wizard. `AddIntegrationPanel` is deleted; nothing else referenced it.

`CreateIntegrationWizard` itself is untouched and still reachable from `BIProjectForm`, `EmbeddedBIProjectForm`, `StandaloneCreateChooser` and `AddProjectForm`.

**The mid-turn guard moves with the button.** The old one was disabled while Copilot was running, because reverting a generation restores a pre-turn checkpoint and drops whatever is not in it — including an artifact added while it runs. That reasoning is about adding an artifact, not about which wizard did it, so the guard stays. Worth noting the header button has never had it, so the two are still asymmetric; happy to align them either way if that asymmetry is not deliberate.

Net: 5 insertions, 102 deletions.

## UI Component Development
- [x] Added reusable UI components to the ui-toolkit — N/A. No new component: an existing `Button` changes its label and handler, and a view is deleted.
- [x] Use ui-toolkit components wherever possible — unchanged; the button and its container are the ones already there.
- [x] Matches with the native VSCode look and feel — unchanged; same component in the same position.

As a developer on an empty integration, I want the button in front of me to add an artifact to it — the same thing the button offers once it has artifacts — rather than reopening a creation wizard for the integration I am already in.

The integration overview now offers **Add Artifact** on an empty integration, in place of **Add Integration**. Both the empty and populated states lead to the same artifact list.

N/A — removes a path to an existing wizard and relabels a button. No new concept or workflow to document.

## Training
N/A

## Certification
N/A — nothing is added; one of two routes to the same outcome is removed.

## Marketing
N/A

## Automation tests
 - Unit tests
   > None added. The change is a button's label and handler plus the deletion of a view; there is no logic here to assert that would not simply restate the JSX.
 - Integration tests
   > None added. Labelled `Checks/Run Ballerina UI Tests` so the existing UI suite runs against it — worth watching for any test that reached the wizard by clicking **Add Integration** from an empty integration, since that route no longer exists.

## Security checks

## Samples
N/A

## Related PRs
None.

## Migrations (if applicable)
N/A — no persisted state.

macOS 15 (darwin 25.5.0), Node 20. Verified by typecheck and a production webpack build of the visualizer (0 errors). **Not yet exercised in a running extension** — the empty state's layout with the relabelled button is worth an eyeball before merge.

`AddIntegrationPanel` pointed the wizard at an already-created package (`existingPackagePath`) so it generated in place rather than reloading the window — a genuinely different embedding from the other four wizard call sites. Confirming it had no other caller was what made deleting it safe rather than just unlinking it.

## Summary

- Changed the empty integration action from **Add Integration** to **Add Artifact**.
- Reused artifact-list navigation for empty and populated integrations.
- Removed the unused `AddIntegrationPanel` wizard path and related logic.
- Preserved the Copilot-running guard for the empty-state artifact button.
- Added `PackageOverview` tests for button visibility and navigation.
- Updated end-to-end test comments and artifact helper logic without changing test behavior.


packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
```diff
@@ -1248,19 +1217,16 @@ export function PackageOverview(props: PackageOverviewProps) {
...
-                                                    artifact list, so the guided flow can be resumed. */}
                                                 {/* Disabled rather than hidden mid-turn: reverting the
                                                     generation restores a pre-turn checkpoint and drops
                                                     whatever is not in it, including an artifact added now. */}
                                                 <Button
                                                     appearance="primary"
-                                                    onClick={() => setShowAddIntegration(true)}
+                                                    onClick={handleAddConstruct}
                                                     disabled={agentWorking}
                                                     tooltip={agentWorking ? "Available once Copilot finishes" : undefined}
```
NipunaRanasinghe: On the open question in the Approach section about keeping this guard: I'd drop it, and the reason is that the change to `onClick` above moves the operation out from under what the guard was written to cover.

The comment's reasoning holds for a button that generates the artifact in place during the current session — that is what a revert restoring a pre-turn checkpoint can drop. `handleAddConstruct` does not do that; it only navigates:

```js
rpcClient.getVisualizerRpcClient().openView({
    type: EVENT_TYPE.OPEN_VIEW,
    location: { view: MACHINE_VIEW.BIComponentView },
});
```

So after this change the guard is mismatched in both directions:

- it blocks navigation, which is harmless while a turn is running;
- it does not block the creation itself, which now happens later from `ComponentListView`, where there is no equivalent guard.

The result is that a user cannot open the artifact list from the empty state while Copilot runs, can open the same list from the Design header button, and in neither case is anything stopping the artifact creation that follows. That reads as worse than the asymmetry noted in the description, because the remaining guard suggests a protection that is no longer in place.

Suggest removing `disabled={agentWorking}` and the `tooltip` here so both entry points behave identically. If the revert-drops-the-artifact risk is real — and it sounds like it is — it wants guarding at the creation action inside `ComponentListView`, which is out of scope here and probably worth its own issue.

packages/ballerina-visualizer/src/views/BI/PackageOverview/AddIntegrationPanel.tsx
```diff
@@ -1,63 +0,0 @@
...
-            fill
-        >
-            {/* The wizard's model/diagnostics calls go over the WS bridge, the same
-                seam the pre-project Create flow uses. */}
-            <BiWsClientProvider onBack={onClose}>
-                <CreateIntegrationWizard
-                    embedded
-                    showHeader={false}
-                    existingPackagePath={packageRoot}
-                    onArtifactAdded={onClose}
```
NipunaRanasinghe, line 56: These two props have no other consumer, so deleting this file leaves them dead on `CreateIntegrationWizard`. Searching the repo for `existingPackagePath` and `onArtifactAdded` returns exactly two files: this one and the wizard's own definition.

That means the wizard's generate-into-an-already-created-package mode becomes unreachable with this PR. The Learning section makes the point that this was "a genuinely different embedding from the other four wizard call sites" — which is the argument for removing the mode along with its only caller, rather than leaving props that nothing passes for the next person to work out.

Either drop `existingPackagePath` / `onArtifactAdded` and whatever branches on them inside the wizard, or add a line to the description saying the mode is deliberately retained and for what. Happy either way; it just should not be left implicit.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
```diff
@@ -1248,19 +1217,16 @@ export function PackageOverview(props: PackageOverviewProps) {
                                                 </StatusRow>
                                             )}
                                             <ButtonContainer>
-                                                {/* An empty integration means the creation wizard was
-                                                    skipped — offer it again here rather than the raw
-                                                    artifact list, so the guided flow can be resumed. */}
```
NipunaRanasinghe, line 1253: This comment recorded a deliberate decision that the PR reverses, and it goes away with the behaviour rather than being answered.

It says the empty state offers the wizard *rather than* the raw artifact list, specifically so a user who skipped the guided flow at creation time can resume it. The description's case for the change is consistency — one operation, one destination — which is a good reason, but it argues past this one instead of against it. Both are legitimate; they point opposite ways, and the person who loses out is the user on a brand-new empty integration, who now lands in the artifact list with no guidance on what to pick.

Not asking for the behaviour to change — this looks like a deliberate direction across the recent PRs, and the consistency argument is sound. But the decision to drop the guided-resume path deserves a design owner's sign-off and a line in the description, so it does not simply vanish with the comment that justified it.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
```diff
@@ -1204,6 +1171,8 @@ export function PackageOverview(props: PackageOverviewProps) {
                             {!isLibrary && (
                                 <DiagramHeaderContainer withPadding={true}>
                                     <Title variant="h2">Design</Title>
+                                    {/* An empty integration has its own copy of this below,
+                                        centred in the empty state, so only one is ever on screen. */}
```
dulajdilshan, line 1175: Could we encode this `exactly one Add Artifact entry point` rule in tests? The current Playwright helper (`e2e-playwright-tests/utils/helpers/artifacts.ts`) still waits for either **Add Artifact** or **Add Integration** and follows the old wizard branch, while `automation.spec.ts` and `expression-editor-advanced.spec.ts` also explicitly tolerate both flows. As a result, all E2E groups can remain green even if the old behavior remains or is accidentally restored—the behavior this PR removes is still considered valid by the tests. Please update the current-build helper/specs to require the **Add Artifact → BIComponentView → direct card** path, and add an L2 render invariant for `PackageOverview`: empty and populated integrations each expose exactly one Add Artifact action in the intended location, and clicking either dispatches `BIComponentView`. That verifies destination/exclusivity rather than merely restating the JSX.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
gigara: Removed in 262d3389fe.

You're right, and the framing is sharper than the asymmetry I put in the description. I had it as "one entry point guards, the other doesn't"; the real problem is that the guard no longer corresponds to anything. It was written for a button that generated the artifact in place during the session — which is exactly what a revert restoring a pre-turn checkpoint can drop — and `handleAddConstruct` only navigates. So it blocked navigation, which costs the user something and protects nothing, while the creation that follows in `ComponentListView` was unguarded either way. A guard implying a protection that is not there is worse than no guard.

Both entry points are identical again: same handler, same enablement.

Agreed the underlying risk wants guarding at the creation action in `ComponentListView`, and that it is out of scope here.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
gigara: Good catch, and the important half of it is that the suites would have stayed green either way — so they were not evidence of anything. Both parts done in 262d3389fe.

**Helper** (`utils/helpers/artifacts.ts`) — no longer races the two buttons. It waits for **Add Artifact**, then asserts exactly one of it (so a future change cannot reintroduce a second by rendering both states' buttons) and fails if **Add Integration** is present at all. `submitArtifactCreation` tightened from `/^Create( Integration)?$/` to `/^Create$/`, since the wizard's Configure step was the only source of the other.

**Specs** — `automation.spec.ts` and both expression-editor specs accepted either post-submit landing, which was the second place the old flow stayed valid. They now assert the direct-to-diagram landing rather than falling back to clicking the entry node. That also made three copies of a local `clickUntil` helper dead, so those went too.

**L2 invariant** — `views/BI/PackageOverview/index.test.tsx`, using the existing jsdom harness (`test/rpcHarness`, same pattern as `TypeDiagram/index.test.tsx`). Six cases: each state exposes exactly one Add Artifact action, the empty state's sits inside the empty state rather than the header, clicking either dispatches `BIComponentView`, and neither state offers Add Integration.

I mutation-checked it rather than trusting it green — reverted `index.tsx` to its pre-change state and ran the suite: **4 of the 6 fail**, the 2 that pass being the populated-integration cases, whose behaviour this PR does not change. So it fails for the reason it exists.

One thing I did not tighten: the automation spec's `/Create( New)? Automation/i` heading regex still tolerates both forms. The heading is composed from artifact metadata rather than a literal, so I could not confirm the exact string without running the suite, and pinning it to a guess seemed worse than leaving it. Flagged in a comment there.

packages/ballerina-visualizer/src/views/BI/PackageOverview/AddIntegrationPanel.tsx
gigara: Confirmed — `existingPackagePath` and `onArtifactAdded` now appear only in the wizard's own definition, so that mode is unreachable as of this PR.

I have not removed it here, and I want to be straight that this is a scope call rather than a disagreement. The mode is not the two props plus a handler; it is threaded through ~13 sites in `CreateIntegrationWizard/index.tsx`, all keyed off `isExistingPackage`:

- `firstStep = isExistingPackage ? TYPE_STEP : NAME_STEP` (~106)
- four effect guards (~261, ~301, ~311, ~335) and a condition at ~369
- `handleAddArtifactToExistingPackage` (~511) and the submit branch at ~562
- `steps={isExistingPackage ? WIZARD_STEPS.slice(1) : WIZARD_STEPS}` (~631) and the "Create Empty Integration" skip label (~697)

Unpicking that touches the wizard's step sequencing and submit path, which four live call sites depend on (`BIProjectForm`, `EmbeddedBIProjectForm`, `StandaloneCreateChooser`, `AddProjectForm`). Doing it inside a PR whose own diff is a relabelled button would put the risk in the wrong place, and it would not get the review attention it deserves here.

So: taking your second option for this PR, with a follow-up queued that carries the site list above. I have **not** added the description line yet — that is with the requester along with the design point on the sibling thread, and I will add both together. Happy to instead do the removal here if you would rather it not be split.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
gigara: Taking the point, and I think you have named the actual cost precisely: a user on a brand-new empty integration now lands in the flat artifact list with no guidance on what to pick, where the wizard's Type step was a curated subset. The consistency argument does not answer that — it just weighs differently.

You are also right about the mechanism, which is the part worth remembering: the comment justifying the old behaviour was deleted along with the behaviour, so nothing in the tree records that a guided-resume path ever existed or why. That is how a deliberate decision becomes invisible.

What I have not done: added the description line. The change was requested directly, so the direction is not in doubt, but "the requester asked for it" is not the same as a design owner signing off on losing the guided path, and I would rather that be recorded explicitly than assumed from a code review thread. I am checking with them and will add it here alongside the note on the sibling thread about the wizard's now-unreachable mode.

Nothing else in this PR depends on the answer, so it should not block the rest of the review.

NipunaRanasinghe (approved):
LGTM

PR #829, opened 2026-08-10
Fix issues in review mode diff and navigation

RNViththagan:
Related to https://github.com/wso2/product-integrator/issues/2069
Related to https://github.com/wso2/product-integrator/issues/2070

- Close review mode when a new generation starts, so a later turn's changes are not reviewed against
  the previous generation's diagrams
- Give `ReviewMode` the `key={remountKey}` every other view in `MainPanel` already has, so reopening
  it for a different generation actually reloads its data
- Carry a `generationId` on `ReviewModeData` and include it in `navTarget`, so the existing remount
  machinery can tell two generations apart
- Strip the URI scheme when matching a semantic diff to its package — diffs arrive as `ai://…` while
  package paths are plain, so in a workspace project every diff lost its package name and only the
  first type diff produced a view
- Correct the `useFileSchema` comment, which had `ai://` and `file://` the wrong way round

## Summary

- Close review mode before a new agent generation starts.
- Reload review data when the generation changes.
- Track `generationId` in review data and navigation targets.
- Improve semantic-diff attribution across URI schemes, path separators, and multiple packages.
- Add tests for package matching and type-change views.
- Correct the `useFileSchema` documentation comment.

### DIFF
--- a/packages/ballerina-core/src/state-machine-types.ts
+++ b/packages/ballerina-core/src/state-machine-types.ts
@@ -247,10 +247,9 @@ export interface ReviewViewItem {
 export interface ReviewModeData {
     views: ReviewViewItem[];
     currentIndex: number;
-    // Distinguishes one generation's review from another's. MainPanel keys its remount on this:
-    // without it, opening review for a later generation reuses the mounted ReviewMode, whose data
-    // is only read on mount, so the previous generation's diagrams stay on screen.
-    generationId?: string;
+    /** MainPanel keys ReviewMode's remount on this; without it a later generation reuses the
+     *  mounted view, whose data is only read on mount. */
+    generationId: string;
     onAccept?: string;
     onReject?: string;
     semanticDiffs?: object[];
--- a/packages/ballerina-core/src/state-machine-types.ts
+++ b/packages/ballerina-core/src/state-machine-types.ts
@@ -247,8 +247,7 @@ export interface ReviewViewItem {
 export interface ReviewModeData {
     views: ReviewViewItem[];
     currentIndex: number;
-    /** MainPanel keys ReviewMode's remount on this; without it a later generation reuses the
-     *  mounted view, whose data is only read on mount. */
+    /** MainPanel keys ReviewMode's remount on this; its data is only read on mount. */
     generationId: string;
     onAccept?: string;
     onReject?: string;
--- a/packages/ballerina-core/src/state-machine-types.ts
+++ b/packages/ballerina-core/src/state-machine-types.ts
@@ -247,6 +247,10 @@ export interface ReviewViewItem {
 export interface ReviewModeData {
     views: ReviewViewItem[];
     currentIndex: number;
+    // Distinguishes one generation's review from another's. MainPanel keys its remount on this:
+    // without it, opening review for a later generation reuses the mounted ReviewMode, whose data
+    // is only read on mount, so the previous generation's diagrams stay on screen.
+    generationId?: string;
     onAccept?: string;
     onReject?: string;
     semanticDiffs?: object[];
--- a/packages/ballerina-extension/src/features/ai/agent/index.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/index.ts
@@ -132,12 +132,6 @@ export async function generateAgent(params: GenerateAgentCodeRequest): Promise<b
         // no separate temp copy anymore (see existingTempPath below).
         finalizeLastGeneration(projectRootPath, threadId);
 
-        // The previous generation's review is no longer the one being worked on. Imported lazily:
-        // ApprovalViewManager reaches the webview layer and vscode-languageclient through RPCLayer,
-        // which cannot load under jest, and this module is imported by unit tests.
-        const { approvalViewManager } = await import("../state/ApprovalViewManager");
-        approvalViewManager.closeReviewModeIfOpen();
-
         // Create config using factory function. existingTempPath makes the agent operate
         // directly on the real project root instead of AICommandExecutor creating a
         // throwaway temp copy — file edits land live in the real workspace (M1+), so there's
--- a/packages/ballerina-extension/src/features/ai/agent/index.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/index.ts
@@ -32,6 +32,7 @@ import { getHashedProjectId } from "../../telemetry/common/project-id";
 import { runEventStore } from "../utils/run-event-store";
 import { sendSaveChatNotification } from "../utils/ai-utils";
 import { finalizeRevertibleGeneration } from "../utils/generation-response";
+import { approvalViewManager } from "../state/ApprovalViewManager";
 
 // ==================================
 // Agent Generation Functions
@@ -131,6 +132,9 @@ export async function generateAgent(params: GenerateAgentCodeRequest): Promise<b
         // Nothing to clean up: edits already land directly in the real workspace, and there's
         // no separate temp copy anymore (see existingTempPath below).
         finalizeLastGeneration(projectRootPath, threadId);
+
+        // The previous generation's review is no longer the one being worked on.
+        approvalViewManager.closeReviewModeIfOpen();
 
         // Create config using factory function. existingTempPath makes the agent operate
         // directly on the real project root instead of AICommandExecutor creating a
--- a/packages/ballerina-extension/src/features/ai/agent/index.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/index.ts
@@ -32,7 +32,6 @@ import { getHashedProjectId } from "../../telemetry/common/project-id";
 import { runEventStore } from "../utils/run-event-store";
 import { sendSaveChatNotification } from "../utils/ai-utils";
 import { finalizeRevertibleGeneration } from "../utils/generation-response";
-import { approvalViewManager } from "../state/ApprovalViewManager";
 
 // ==================================
 // Agent Generation Functions
@@ -133,7 +132,10 @@ export async function generateAgent(params: GenerateAgentCodeRequest): Promise<b
         // no separate temp copy anymore (see existingTempPath below).
         finalizeLastGeneration(projectRootPath, threadId);
 
-        // The previous generation's review is no longer the one being worked on.
+        // The previous generation's review is no longer the one being worked on. Imported lazily:
+        // ApprovalViewManager reaches the webview layer and vscode-languageclient through RPCLayer,
+        // which cannot load under jest, and this module is imported by unit tests.
+        const { approvalViewManager } = await import("../state/ApprovalViewManager");
         approvalViewManager.closeReviewModeIfOpen();
 
         // Create config using factory function. existingTempPath makes the agent operate
--- a/packages/ballerina-extension/src/features/ai/state/ApprovalViewManager.ts
+++ b/packages/ballerina-extension/src/features/ai/state/ApprovalViewManager.ts
@@ -595,20 +595,15 @@ export class ApprovalViewManager {
     }
 
     /**
-     * Close ReviewMode if it is the current view. A review belongs to one generation, so leaving an
-     * earlier one open while the next starts means the user's next click lands on the previous
-     * generation's diagrams — and the bar hides its Review button while review mode is open, so
-     * there is no way back either.
+     * A review belongs to one generation, so an earlier one must not stay open into the next. Same
+     * steps as the panel's own Close (`goBack`), which likewise clears the open generation through
+     * `notifyReviewModeClosed`.
      */
     closeReviewModeIfOpen(): void {
-        const ctx = StateMachine.context();
-        if (ctx.view !== MACHINE_VIEW.ReviewMode) {
+        if (StateMachine.context().view !== MACHINE_VIEW.ReviewMode) {
             return;
         }
         console.log('[ApprovalViewManager] Closing ReviewMode — a new generation is starting');
-        this.openReviewGenerationId = null;
-        // Same as the panel's own Close: return to whatever the user was looking at before, rather
-        // than dropping them on the package overview.
         history.pop();
         updateView(false);
         this.notifyReviewModeClosed();
--- a/packages/ballerina-extension/src/features/ai/state/ApprovalViewManager.ts
+++ b/packages/ballerina-extension/src/features/ai/state/ApprovalViewManager.ts
@@ -22,7 +22,7 @@ import { AiPanelWebview } from '../../../views/ai-panel/webview';
 import { chatStateStorage } from '../../../views/ai-panel/chatStateStorage';
 import { sendReviewRestoreDidOpenBatch } from '../utils/project/ls-schema-notifications';
 import { VisualizerWebview } from '../../../views/visualizer/webview';
-import { openView as openMainView, StateMachine } from '../../../stateMachine';
+import { history, openView as openMainView, StateMachine, updateView } from '../../../stateMachine';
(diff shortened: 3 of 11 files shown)


packages/ballerina-visualizer/src/MainPanel.tsx
```diff
@@ -921,7 +921,7 @@ const MainPanel = () => {
                         case MACHINE_VIEW.ReviewMode: {
                             const { ReviewMode } = await import("./views/ReviewMode");
                             if (isStaleNavigation()) return;
-                            setViewComponent(<ReviewMode />);
+                            setViewComponent(<ReviewMode key={remountKey} />);
```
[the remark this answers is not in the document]
RNViththagan: Checked this against the messenger implementation and I don't think it holds here.

`vscode-messenger-webview@0.5.1` registers notification handlers into a map keyed by method:

```js
onNotification(type, handler) {
    this.handlerRegistry.set(type.method, handler);
    return this;
}
```

Two consequences:

- It is single-slot, not additive. A remount replaces the previous handler rather than stacking another one, so callbacks cannot accumulate and an older closure cannot receive a later `navigateReviewIndex` notification — the newest mount is always the one that handles it, which is the behaviour the remount key is there to get.
- There is no unsubscribe to return. `onNotification` returns the `Messenger`, and `BallerinaRpcClient.onNavigateReviewIndex` has no Disposable to hand back, so the effect has nothing to clean up. Introducing one would mean changing the shared RPC client surface for every `onXxx` method, which is beyond this PR.

The residual worth noting is different from the one described: because nothing unregisters, the last handler stays in the registry after `ReviewMode` unmounts, so a `navigateReviewIndex` arriving while review mode is closed would call into an unmounted component's closure. That is a no-op today and is unchanged by this PR. If we want real disposal, the right fix is to return the `Disposable` from the client's `onXxx` methods across the board — happy to raise that separately.

SasinduDilshara (dismissed):

packages/ballerina-visualizer/src/views/ReviewMode/path-utils.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+export function toComparablePath(uri: string): string {
+    return uri.replace(/^[a-z][a-z0-9+.-]*:\/\//, "").replace(/\\/g, "/");
+}
+
+/** Whether a semantic diff's uri falls under the given package root. */
+export function diffBelongsToPackage(uri: string, packagePath: string): boolean {
+    const uriPath = toComparablePath(uri);
+    const pkgPath = packagePath.replace(/\\/g, "/");
+    return uriPath.startsWith(`${pkgPath}/`) || uriPath === pkgPath;
```
NipunaRanasinghe: Extracting this is the right fix — the first loop already stripped the scheme and the second did not, which is exactly why every diff fell back to the workspace root. Since this is now the single place that decides a diff's package, it is worth making the normalisation complete. Three gaps, none of them regressions (the old inline code had them too), but each one silently reproduces the symptom this PR fixes: diff attributed to the workspace root, package name lost, type views collapsed into one.

**Percent-encoding.** `convertAiToFileScheme` in `rpc-managers/bi-diagram/rpc-manager.ts` round-trips URIs through `Uri.parse(uri).with({ scheme: 'file' }).toString()`, and `Uri.toString()` percent-encodes. A project at `/ws/my project/school` then arrives as `.../my%20project/school` and never matches the plain package path. Spaces in project paths are common enough to hit this. Worth confirming how the semantic-diff URIs are actually constructed — if they come from `Uri.toString()`, this wants a `decodeURIComponent`; if they are plain concatenation, it does not apply and is worth a note saying so.

**Windows, two sub-cases.** `file:///c:/ws/…` strips to `/c:/ws/…`, keeping a leading slash that will not match `c:/ws/…`. The test below only covers `ai://C:\…`, where the scheme is followed directly by the drive letter, so the three-slash form is not exercised. Separately, drive-letter and path case can differ between what VS Code emits and the stored package path, and `diffBelongsToPackage` compares case-sensitively.

**Reuse.** `@wso2/ballerina-core` exports `isSamePath`. Worth checking whether it already handles separator and case normalisation before hand-rolling a second path comparison here — at minimum the `uriPath === pkgPath` branch could delegate to it.

Also free while you are here: the scheme regex `^[a-z][a-z0-9+.-]*:\/\//` is lowercase-only, and URI schemes are case-insensitive per RFC 3986. Add the `i` flag.

packages/ballerina-extension/src/features/ai/state/ApprovalViewManager.ts
```diff
@@ -593,6 +594,26 @@ export class ApprovalViewManager {
...
+            return;
+        }
+        console.log('[ApprovalViewManager] Closing ReviewMode — a new generation is starting');
+        this.openReviewGenerationId = null;
+        // Same as the panel's own Close: return to whatever the user was looking at before, rather
+        // than dropping them on the package overview.
+        history.pop();
+        updateView(false);
+        this.notifyReviewModeClosed();
+    }
```
NipunaRanasinghe, line 606: `goBack` in `rpc-managers/visualizer/rpc-manager.ts` already does this same sequence:

```js
const wasReviewMode = StateMachine.context().view === MACHINE_VIEW.ReviewMode;
history.pop();
updateView(false, params?.identifier);
if (wasReviewMode) { approvalViewManager.notifyReviewModeClosed(); }
```

So there are now two implementations of "close review mode", and they differ in exactly one line: this one clears `openReviewGenerationId`, that one does not.

That asymmetry is the part worth resolving, because it points at a question rather than just duplication. If clearing `openReviewGenerationId` is necessary when review mode closes — and this method treats it as necessary — then the panel's own Close leaves it stale, and closing manually then reopening the same generation behaves differently from closing via a new generation. Whatever that field gates, the two paths disagree about it.

Pre-existing rather than introduced here, but this is the first code to treat the reset as required, which is what surfaces it. Suggest having both paths call one shared close method so they cannot drift, or a line saying why only this trigger needs the reset.

Minor, same method: `history.pop()` is unconditional. If review mode is the only history entry, `updateView(false)` then has no last view to restore — the comment says the intent is to return to whatever preceded it, so it is worth checking what happens when nothing did.

packages/ballerina-extension/src/features/ai/agent/index.ts
```diff
@@ -132,6 +132,12 @@ export async function generateAgent(params: GenerateAgentCodeRequest): Promise<b
         // no separate temp copy anymore (see existingTempPath below).
         finalizeLastGeneration(projectRootPath, threadId);
 
+        // The previous generation's review is no longer the one being worked on. Imported lazily:
+        // ApprovalViewManager reaches the webview layer and vscode-languageclient through RPCLayer,
+        // which cannot load under jest, and this module is imported by unit tests.
+        const { approvalViewManager } = await import("../state/ApprovalViewManager");
+        approvalViewManager.closeReviewModeIfOpen();
```
NipunaRanasinghe: The lazy import does not achieve what the comment says it does. `await import()` still resolves the module at runtime, the moment this line executes — so this only avoids the jest problem as long as no test ever exercises `generateAgent`. The first test that does will fail on `RPCLayer` failing to load, and the error will surface a long way from its cause.

Cleaner would be to hoist the call to whichever layer already knows about webviews and have it invoke `closeReviewModeIfOpen` around `generateAgent`, which removes the need for the dynamic import entirely and puts the view concern outside the agent module. If it has to stay here, injecting it would at least make the seam explicit rather than depending on a code path never being reached under test.

Separate question on placement: is `generateAgent` the only entry point that starts a generation? If repair, continue, or test-generation flows have their own, review mode stays open on those and the bug this fixes persists there.

packages/ballerina-visualizer/src/views/ReviewMode/path-utils.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    it("does not treat a package as a prefix of a longer sibling name", () => {
+        expect(diffBelongsToPackage(`ai://${SCHOOL}-archive/types.bal`, SCHOOL)).toBe(false);
+    });
+
+    // The regression this guards: comparing the raw uri never matched, so in a workspace every diff
+    // fell back to the workspace root — losing its package name and collapsing all packages' type
+    // views into a single one.
+    it("would never match with the scheme left on the uri", () => {
+        expect(`ai://${SCHOOL}/types.bal`.startsWith(`${SCHOOL}/`)).toBe(false);
+    });
```
NipunaRanasinghe: This case asserts a property of `String.prototype.startsWith`, not of anything in the repo — it would still pass if `path-utils.ts` were deleted, so it cannot fail and gives no protection. The comment above it does the documenting job on its own; the assertion adds only the appearance of coverage.

Same concern about the last case, `gives each package its own type view instead of collapsing them` (lines 63-70). The name promises coverage of the type-view construction, but the body just re-runs `diffBelongsToPackage` over the same two URIs as the case above it with a `Set` around the result. The view-building loop in `index.tsx` — the thing that actually collapsed — stays untested.

The other five are good, and the `school-archive` vs `school` prefix case is a genuinely easy bug to ship. I would drop these two and spend the space on the gaps flagged on `path-utils.ts`: a percent-encoded URI and the `file:///c:/…` three-slash form, both of which would fail against the current implementation.

packages/ballerina-core/src/state-machine-types.ts
```diff
@@ -247,6 +247,10 @@ export interface ReviewViewItem {
 export interface ReviewModeData {
     views: ReviewViewItem[];
     currentIndex: number;
+    // Distinguishes one generation's review from another's. MainPanel keys its remount on this:
+    // without it, opening review for a later generation reuses the mounted ReviewMode, whose data
+    // is only read on mount, so the previous generation's diagrams stay on screen.
+    generationId?: string;
```
NipunaRanasinghe: The whole remount fix depends on this being present: if it is absent, `navTarget`'s new segment is empty and two different generations produce the same target, so `ReviewMode` is reused and the bug returns.

Both producers set it today — `AgentExecutor` from `context.messageId` and `buildReviewModeData` from its `generationId` — so this works. Making it required would turn that from something that happens to hold into something the type enforces, and would catch a future third construction site at compile time rather than as the original symptom reappearing.

packages/ballerina-visualizer/src/views/ReviewMode/ReadonlyFlowDiagram.tsx
```diff
@@ -210,7 +210,7 @@ export function ReadonlyFlowDiagram(props: ReadonlyFlowDiagramProps): JSX.Elemen
     }, [filePath, position, oldPosition, viewMode, expectedMetadata, changeType, rpcClient]);
 
     // Fetch one version of the enclosing function's flow model.
-    // useFileSchema=true reads the frozen original (file://); false reads the modified temp content (ai://).
+    // useFileSchema=true reads the frozen original (ai://); false reads the live edits (file://).
```
NipunaRanasinghe, line 213: Confirmed correct, and worth having traced — a confidently backwards comment is worse than none. Three independent sources agree with the new wording: `index.tsx` passes `useFileSchema={effectiveViewMode === "old"}`; `rpc-managers/bi-diagram/rpc-manager.ts` says "filePath defaults to `ai://` (the frozen baseline, used for 'old'); convert to `file://` (live) unless the caller explicitly wants the old view"; and `diagnostics-utils.ts` independently describes `ai://` as the frozen pre-generation baseline.

The underlying problem is the parameter name: `useFileSchema=true` means "do not use the file scheme". That is why the comment was wrong in the first place, and a comment is the weakest available fix — the next reader who skips it will make the same inference. The name appears in four files under `ReviewMode/` (`ReadonlyFlowDiagram`, `ReadonlyTypeDiagram`, `ReadonlyComponentDiagram`, `position-utils`), so a rename to `useBaseline` or `useOriginal` is a contained change that would make the comment unnecessary. Not for this PR, but worth queueing while the tracing is fresh.

packages/ballerina-visualizer/src/views/ReviewMode/path-utils.ts
RNViththagan: Checked against the LS rather than guessing. `SemanticDiffComputer.resolveUri` does `"ai" + filePath.toUri().toString().substring(4)`, so percent-encoding applies — now decoded, with a try/catch for a stray `%`.

The three-slash form is broader than described: `substring(4)` means every URI comes out `ai:///…`, not just Windows. On POSIX the leading slash is what makes it match an absolute package path, which is why it went unnoticed; on Windows it leaves `/c:/ws/…`. Stripped only when a drive letter follows.

On reuse — `isPathInside` is a closer fit than `isSamePath`: same predicate, sibling-prefix case included. Now delegating to it. It normalises the drive letter but not separators, so a `\` package path still missed a `/` child; that's the one `replace` left on the parent.

Regex has the `i` flag. `096cb536be`

packages/ballerina-extension/src/features/ai/state/ApprovalViewManager.ts
RNViththagan: The asymmetry isn't there — `notifyReviewModeClosed()` → `resetReviewNavigationState()` clears `openReviewGenerationId`, and both paths call it. My explicit assignment was redundant, so it's deleted; the indirection is what made it look one-sided.

Left separate otherwise: `goBack` pops for any view and forwards an `identifier`, this only fires in review mode.

On the unconditional `pop()` — empty-stack `pop()` is a no-op and `updateView` reads the last entry with `?.`, so nothing breaks. Review mode always opens on top of an existing view, so no guard added. `a2ea3467bc`

packages/ballerina-extension/src/features/ai/agent/index.ts
RNViththagan: `generateAgent` isn't the only entry point — the command executors start generations via `finalizePreviousGeneration()`, so review mode stayed open on those flows and the bug survived there.

Both paths already go through `finalizeRevertibleGeneration`, so the call lives there now: one site, both covered, dynamic import gone.

It has to run before that function's early return, since the open review belongs to the previous turn either way — there's a test for that case. A top-level import does break `finalizeLastGeneration.test.ts` on `vscode-languageclient`, so it's an explicit `jest.mock` in that suite, which also let me assert the close rather than just permit it. `b02df5cd1b`

packages/ballerina-visualizer/src/views/ReviewMode/path-utils.test.ts
RNViththagan: Both dropped. The `startsWith` one asserts a property of the standard library and would pass with the file deleted.

Replaced with the cases you listed: percent-encoded path, malformed `%`, `file:///c:/…`, and a differing drive-letter case. The last two caught the separator gap in `isPathInside`.

The view-building loop is still untested — that needs the view, not the helper. Not attempted here. `096cb536be`

packages/ballerina-core/src/state-machine-types.ts
RNViththagan: Required now. Both packages typecheck unchanged, which confirms both producers already set it — enforcement rather than a behaviour change. `5aa63f32cb`

packages/ballerina-visualizer/src/views/ReviewMode/ReadonlyFlowDiagram.tsx
RNViththagan: The name is the underlying problem — `useFileSchema=true` meaning "don't use the file scheme" is why the comment was backwards. Left out of this PR; queued a follow-up to rename it to `useBaseline` across the four `ReviewMode/` files.

PR #830, opened 2026-08-10
Allow stopping a generation while an approval prompt is showing

RNViththagan:
Related to https://github.com/wso2/product-integrator/issues/2050
Related to https://github.com/wso2/product-integrator/issues/2073

- Add a stop control to the approval footers, so a generation can be stopped while a plan, task,
  web-tool or skill-enable prompt is showing — those footers replace the composer, which is what
  normally owns Stop and Escape
- Extend the same control to the clarify (Ask) footer, at the end of its title row
- Stop the raw abort error from duplicating the interruption marker, so stopping leaves one clean
  `[Request interrupted by user]` line instead of a red error box beside it
- Clear the plan-approval and skill-enable cards when a generation stops or aborts, so a resolved
  prompt does not stay on screen with dead buttons
- Keep Escape from aborting a run it was meant to dismiss a menu for: the sessions dropdown now
  claims the key in the capture phase, and the footer hook ignores an Escape already handled
- Keep the composer mounted behind an approval footer instead of unmounting it, so a message typed
  (and files attached) while waiting for the prompt are still there once it is answered

## Summary

- Added stop controls and Escape-key handling to approval and clarification footers.
- Preserved composer input and attachments while approval footers are displayed.
- Prevented duplicate interruption messages after aborts.
- Cleared pending approval cards when generation stops or aborts.
- Added `SKIPPED` skill events when users decline activation.
- Improved Escape-key handling for session history menus.
- Added tests for hidden footer behavior and composer preservation.

### DIFF
--- a/packages/ballerina-extension/src/features/ai/agent/tools/skill-tool/index.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/tools/skill-tool/index.ts
@@ -21,6 +21,7 @@ import { Skill } from '../../skills/types';
 import { readProjectSkillContent, readUserSkillContent } from './skill-reader';
 import { buildAllDisabledSet } from '../../skills/context';
 import { approvalManager } from '../../../state/ApprovalManager';
+import { SkillEnableStage } from '@wso2/ballerina-core';
 
 export const SKILL_TOOL_NAME = "invoke_skill";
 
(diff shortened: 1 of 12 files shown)


RNViththagan:
The L2 test for `hidden` is in `2a628f6eca` — `Footer/index.test.tsx`. It asserts the container is hidden, the composer is still mounted, and (third case) that the composer node keeps its identity across a `hidden` toggle, so reintroducing conditional rendering fails the test rather than silently discarding the draft again.

@coderabbitai review

RNViththagan:
@coderabbitai review

RNViththagan:
Fixed — no `as any` left in the file.

- `baseProps` is now typed `React.ComponentProps<typeof Footer>`, so the fixture is checked against the real prop contract rather than asserted past it. That earns its keep immediately: a required prop added to `Footer` now fails this test at compile time.
- `IS_REACT_ACT_ENVIRONMENT` is a `declare global` instead of a cast on `globalThis`.
- `aiChatInputRef` uses `AIChatInputRef`; `tagOptions`/`attachmentOptions` are real values now.

Typing it properly caught something a cast was hiding: `placeholderTags: {}` does not satisfy `PlaceholderTagMap`, which is a mapped type over every command. It now uses the exported `placeholderTags` fixture, which in turn needed `Command` added to the core mock to initialise.

`node scripts/typecheck-tests.js` → 0 test-code errors across 9 packages; visualizer suite 223 passing.

SasinduDilshara (approved):

SasinduDilshara (approved):

PR #842, opened 2026-08-11
Add human-in-the-loop approval for agent tools

lochana-chathura:
## Purpose
$subject.

Fixes https://github.com/wso2/product-integrator/issues/2131

## Preview
**add tool → Use Function:**
- Conditional approval: create a new predicate function

https://github.com/user-attachments/assets/1cf13697-cc54-433e-880e-e6849273a4bc

- Conditional approval:  pick an existing function

https://github.com/user-attachments/assets/f7101b74-3a7d-4c21-b133-aba4c3de458d

- Always require approval (no predicate function)

https://github.com/user-attachments/assets/e74d0c7c-6662-4d9a-918f-80c3b56f056f

**add tool → Use Connection:**

https://github.com/user-attachments/assets/b0daed71-1d3e-48c3-a5eb-114417346995

**add tool → Create Custom Tool:**

https://github.com/user-attachments/assets/aeea5f6e-3b13-4e63-a752-17126c82a833

**add tool → Use Agent:**

https://github.com/user-attachments/assets/83e409c3-b109-4fd0-8a25-26a7043b1c19

## Summary

- Added human-in-the-loop approval for function, connection, custom, and agent tools.
- Added unconditional approval and existing or generated approval predicates.
- Added approval metadata handling, validation, and source generation.
- Added tests for approval configuration and source generation.
- Improved form diagnostics, optional labels, conditional layout, autocomplete configuration, and field value handling.

### DIFF
--- a/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
+++ b/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
@@ -655,10 +655,13 @@ public class AgentToolBuilder extends NodeBuilder {
         if (predicateName.isEmpty()) {
             return;
         }
-        // Mirror the tool's own signature exactly, using the same param list generate() built it
-        // from (stashed on ctx) — each ToolKind resolves params differently, so this must not
-        // recompute with a fixed kind.
+        // Mirror the tool's own signature, using the same param list generate() built it from
+        // (stashed on ctx) — each ToolKind resolves params differently, so this must not recompute
+        // with a fixed kind. The leading `ai:Context ctx` param is excluded: the `ai` compiler
+        // plugin strips it from the tool's own signature before comparing against the predicate's,
+        // so including it here would make the predicate's signature look mismatched.
         String paramDecls = ctx.resolvedParams.stream()
+                .filter(param -> !(ctx.includeContext && "ctx".equals(param.name())))
                 .map(ToolParam::decl)
                 .collect(Collectors.joining(", "));
         ctx.sb.token()
--- a/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
+++ b/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
@@ -628,13 +628,13 @@ public class AgentToolBuilder extends NodeBuilder {
 
     // Emit a sibling `isolated function` predicate stub for the @ai:AgentTool `requiresApproval` gate,
     // when the form requested one (codedata.data.generateApprovalFunction == "true"). The predicate
-    // mirrors the tool wrapper's own parameter list (the exact list its signature was built from,
-    // via the same FUNCTION.resolveParams) and returns boolean, satisfying the RequiresApproval
-    // contract (the ai module binds the proposed call's arguments to the predicate by name). Its
-    // name is the `requiresApproval` annotation value. Must be called AFTER the tool declaration has
-    // been flushed via textEdit(DECLARATION) (so the token buffer is empty); it appends a second
-    // DECLARATION text edit to the same file, which build() returns alongside the tool's. FUNCTION
-    // kind only — matches the original feature's scope (Create Tool from Function).
+    // mirrors the tool wrapper's own parameter list (the exact list its signature was built from) and
+    // returns boolean, satisfying the RequiresApproval contract (the ai module binds the proposed
+    // call's arguments to the predicate by name). Its name is the `requiresApproval` annotation value.
+    // Must be called AFTER the tool declaration has been flushed via textEdit(DECLARATION) (so the
+    // token buffer is empty); it appends a second DECLARATION text edit to the same file, which
+    // build() returns alongside the tool's. Called from buildFunctionBody, buildRemoteActionBody, and
+    // buildResourceActionBody — i.e. every ToolKind that reaches this via a real wrapped node.
     private static void appendApprovalPredicate(ToolGenContext ctx) {
         Map<String, Object> data = ctx.wrappedNode != null ? ctx.wrappedNode.codedata().data() : null;
         if (data == null || !"true".equals(String.valueOf(data.get("generateApprovalFunction")))) {
@@ -645,6 +645,8 @@ public class AgentToolBuilder extends NodeBuilder {
         if (predicateName.isEmpty()) {
             return;
         }
+        // resolveParams is the shared enum-level default (only AGENT_CALL overrides it), so any
+        // ToolKind constant yields the same result here — FUNCTION is used as a stand-in.
         String paramDecls = ToolKind.FUNCTION.resolveParams(ctx).stream()
                 .map(ToolParam::decl)
                 .collect(Collectors.joining(", "));
@@ -682,7 +684,7 @@ public class AgentToolBuilder extends NodeBuilder {
                 .stepOut()
                 .functionParameters(flowNode, ignoredKeys);
 
-        return finishActionBody(sourceBuilder, flowNode, returnType);
+        return finishActionBody(ctx, flowNode, returnType);
     }
 
     private static Map<Path, List<TextEdit>> buildResourceActionBody(ToolGenContext ctx, ReturnInfo returnInfo) {
@@ -748,7 +750,7 @@ public class AgentToolBuilder extends NodeBuilder {
                 .stepOut()
                 .functionParameters(flowNode, ignoredKeys);
 
-        return finishActionBody(sourceBuilder, flowNode, returnType);
+        return finishActionBody(ctx, flowNode, returnType);
     }
 
     private static void beginActionBody(SourceBuilder sourceBuilder, FlowNode flowNode, String returnType) {
@@ -762,8 +764,9 @@ public class AgentToolBuilder extends NodeBuilder {
         }
     }
 
-    private static Map<Path, List<TextEdit>> finishActionBody(SourceBuilder sourceBuilder, FlowNode flowNode,
+    private static Map<Path, List<TextEdit>> finishActionBody(ToolGenContext ctx, FlowNode flowNode,
                                                                 String returnType) {
+        SourceBuilder sourceBuilder = ctx.sb;
         if (!returnType.isEmpty()) {
             sourceBuilder.token().keyword(SyntaxKind.RETURN_KEYWORD)
                     .name(flowNode.getProperty(Property.VARIABLE_KEY).orElseThrow().value().toString())
@@ -772,6 +775,9 @@ public class AgentToolBuilder extends NodeBuilder {
         sourceBuilder.token().keyword(SyntaxKind.CLOSE_BRACE_TOKEN);
         sourceBuilder.textEdit(SourceBuilder.SourceKind.DECLARATION);
         sourceBuilder.acceptImport();
+        // If the form asked to scaffold a conditional-approval predicate, emit it as a sibling
+        // declaration reusing the tool wrapper's exact parameter list (mirrors buildFunctionBody).
+        appendApprovalPredicate(ctx);
         return sourceBuilder.build();
     }
 
--- a/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
+++ b/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
@@ -233,6 +233,7 @@ public class AgentToolBuilder extends NodeBuilder {
     private static Map<Path, List<TextEdit>> generate(ToolKind kind, ToolGenContext ctx) {
         ctx.sb.acceptImport(Constants.Ai.BALLERINA_ORG, Constants.Ai.AI_PACKAGE);
         List<ToolParam> params = kind.resolveParams(ctx);
+        ctx.resolvedParams = params;
         ReturnInfo returnInfo = kind.resolveReturn(ctx);
 
         emitDoc(ctx, params, returnInfo);
@@ -639,12 +640,12 @@ public class AgentToolBuilder extends NodeBuilder {
     // call's arguments to the predicate by name). Its name is the `requiresApproval` annotation value.
     // Must be called AFTER the tool declaration has been flushed via textEdit(DECLARATION) (so the
     // token buffer is empty); it appends a second DECLARATION text edit to the same file, which
-    // build() returns alongside the tool's. Called from buildFunctionBody, buildRemoteActionBody,
-    // buildResourceActionBody, and CUSTOM's buildBody — every ToolKind whose form exposes the
-    // "Requires Approval" gate.
+    // build() returns alongside the tool's. Called from every ToolKind's body builder
+    // (buildFunctionBody, buildRemoteActionBody, buildResourceActionBody, CUSTOM's buildBody, and
+    // buildAgentCallBody) — every path whose form exposes the "Requires Approval" gate.
     private static void appendApprovalPredicate(ToolGenContext ctx) {
-        // CUSTOM tools have no wrapped node — generateApprovalFunction/requiresApproval live
-        // directly on the tool's own data (ctx.data) instead, mirroring emitAnnotation's fallback.
+        // CUSTOM and AGENT_CALL tools have no wrapped node — generateApprovalFunction/requiresApproval
+        // live directly on the tool's own data (ctx.data) instead, mirroring emitAnnotation's fallback.
         Map<String, Object> data = ctx.wrappedNode != null ? ctx.wrappedNode.codedata().data() : ctx.data;
         if (data == null || !"true".equals(String.valueOf(data.get("generateApprovalFunction")))) {
             return;
@@ -654,9 +655,10 @@ public class AgentToolBuilder extends NodeBuilder {
         if (predicateName.isEmpty()) {
             return;
         }
-        // resolveParams is the shared enum-level default (only AGENT_CALL overrides it), so any
-        // ToolKind constant yields the same result here — FUNCTION is used as a stand-in.
-        String paramDecls = ToolKind.FUNCTION.resolveParams(ctx).stream()
+        // Mirror the tool's own signature exactly, using the same param list generate() built it
+        // from (stashed on ctx) — each ToolKind resolves params differently, so this must not
+        // recompute with a fixed kind.
+        String paramDecls = ctx.resolvedParams.stream()
                 .map(ToolParam::decl)
                 .collect(Collectors.joining(", "));
         ctx.sb.token()
@@ -814,6 +816,7 @@ public class AgentToolBuilder extends NodeBuilder {
                 .endOfStatement();
         sourceBuilder.token().keyword(SyntaxKind.CLOSE_BRACE_TOKEN);
         sourceBuilder.textEdit(SourceBuilder.SourceKind.DECLARATION).acceptImport();
+        appendApprovalPredicate(ctx);
         return sourceBuilder.build();
     }
 
@@ -1119,6 +1122,10 @@ public class AgentToolBuilder extends NodeBuilder {
         private final String agentReceiver;
         private final String hostClassName;
         private final boolean includeContext;
+        // The tool signature's parameter list, stashed by generate() so appendApprovalPredicate can
+        // mirror it exactly (each ToolKind resolves params differently — AGENT_CALL uses `query`,
+        // not the wrapped node's params — so recomputing with a fixed kind would be wrong).
+        private List<ToolParam> resolvedParams = List.of();
 
         private ToolGenContext(SourceBuilder sb, FlowNode wrappedNode, Map<String, Object> data,
                                String connection, String description,
--- a/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
+++ b/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
@@ -288,27 +288,35 @@ public class AgentToolBuilder extends NodeBuilder {
         }
     }
 
-    private static void emitAnnotation(ToolGenContext ctx) {
-        // CUSTOM tools (and AGENT_CALL) have no wrapped node — auth/requiresApproval live directly
-        // on the tool's own data (ctx.data) instead. Checking both keys (not just "auth") keeps a
-        // custom tool's requiresApproval from being dropped when auth was never set.
-        boolean useTopLevelData = ctx.data != null
-                && (ctx.data.containsKey("auth") || ctx.data.containsKey("requiresApproval"));
-        Map<String, Object> data = useTopLevelData
-                ? ctx.data
-                : ctx.wrappedNode != null ? ctx.wrappedNode.codedata().data() : null;
-        if (data == null) {
-            ctx.sb.token().name("@ai:AgentTool").name(System.lineSeparator());
-            return;
+    // Looks up a single annotation-data key across both places it can live: the tool's own data
+    // (ctx.data — authoritative for CUSTOM/AGENT_CALL, which have no wrapped node) and the wrapped
+    // node's data (where the form originally writes auth/requiresApproval for FUNCTION/REMOTE/
+    // RESOURCE tools). ctx.data wins when both have the key. Resolving key-by-key — instead of
+    // picking one map wholesale for every field — means a key present in only one of the two maps
+    // is never silently dropped, and emitAnnotation/appendApprovalPredicate can't disagree about
+    // where a given field lives.
+    private static Object resolveToolData(ToolGenContext ctx, String key) {
+        if (ctx.data != null && ctx.data.containsKey(key)) {
+            return ctx.data.get(key);
         }
+        if (ctx.wrappedNode != null) {
+            Map<String, Object> wrappedData = ctx.wrappedNode.codedata().data();
+            if (wrappedData != null && wrappedData.containsKey(key)) {
+                return wrappedData.get(key);
+            }
+        }
+        return null;
+    }
 
+    private static void emitAnnotation(ToolGenContext ctx) {
         // Each entry is a fully-rendered mapping field (already indented) to place inside the
         // `@ai:AgentTool { ... }` record. When empty, a bare `@ai:AgentTool` is emitted instead.
         List<String> annotationFields = new ArrayList<>();
 
         // auth: { ... } — OAuth client configuration block.
-        if (data.containsKey("auth")) {
-            String authStr = data.get("auth").toString();
+        Object authValue = resolveToolData(ctx, "auth");
+        if (authValue != null) {
+            String authStr = authValue.toString();
             JsonObject authConfig = gson.fromJson(authStr, JsonObject.class);
 
             List<String> fields = new ArrayList<>();
@@ -352,9 +360,9 @@ public class AgentToolBuilder extends NodeBuilder {
 
         // requiresApproval: <boolean|isolated function> — human-in-the-loop gate. The value is
         // emitted verbatim: "true" for the toggle, or an identifier for a predicate function pointer.
-        if (data.containsKey("requiresApproval")) {
-            Object approvalValue = data.get("requiresApproval");
-            String approval = approvalValue == null ? "" : approvalValue.toString().trim();
+        Object approvalValue = resolveToolData(ctx, "requiresApproval");
+        if (approvalValue != null) {
+            String approval = approvalValue.toString().trim();
             if (!approval.isEmpty()) {
                 annotationFields.add("    requiresApproval: " + approval);
             }
@@ -644,13 +652,12 @@ public class AgentToolBuilder extends NodeBuilder {
     // (buildFunctionBody, buildRemoteActionBody, buildResourceActionBody, CUSTOM's buildBody, and
     // buildAgentCallBody) — every path whose form exposes the "Requires Approval" gate.
     private static void appendApprovalPredicate(ToolGenContext ctx) {
-        // CUSTOM and AGENT_CALL tools have no wrapped node — generateApprovalFunction/requiresApproval
-        // live directly on the tool's own data (ctx.data) instead, mirroring emitAnnotation's fallback.
-        Map<String, Object> data = ctx.wrappedNode != null ? ctx.wrappedNode.codedata().data() : ctx.data;
-        if (data == null || !"true".equals(String.valueOf(data.get("generateApprovalFunction")))) {
+        // Same per-key resolution as emitAnnotation, so the two can't disagree about whether the
+        // gate is set or which name it uses.
+        if (!"true".equals(String.valueOf(resolveToolData(ctx, "generateApprovalFunction")))) {
             return;
         }
-        Object approvalValue = data.get("requiresApproval");
+        Object approvalValue = resolveToolData(ctx, "requiresApproval");
         String predicateName = approvalValue == null ? "" : approvalValue.toString().trim();
         if (predicateName.isEmpty()) {
             return;
--- a/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
+++ b/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
@@ -291,60 +291,79 @@ public class AgentToolBuilder extends NodeBuilder {
         Map<String, Object> data = ctx.data != null && ctx.data.containsKey("auth")
                 ? ctx.data
                 : ctx.wrappedNode != null ? ctx.wrappedNode.codedata().data() : null;
-        if (data == null || !data.containsKey("auth")) {
+        if (data == null) {
             ctx.sb.token().name("@ai:AgentTool").name(System.lineSeparator());
             return;
         }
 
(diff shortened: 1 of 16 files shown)


packages/ballerina-visualizer/src/views/BI/AIChatAgent/AgentToolForm.tsx
```diff
@@ -579,6 +698,12 @@ export function AgentToolForm(props: AgentToolFormProps): JSX.Element {
                                     `@ai:AgentTool {\n    ${configBlock}\n}`);
                             }
                         }
+                        // Edits reference the typed approval-function name as-is (no stub scaffolding —
+                        // this path rewrites already-generated source text directly, bypassing
+                        // AgentToolBuilder's codegen where the predicate stub is normally emitted).
+                        annotationStr = upsertScalarAnnotationField(
+                            annotationStr, "requiresApproval", resolveApprovalSource(data)
+                        );
```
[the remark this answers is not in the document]
lochana-chathura: With b552b3d, edits now do not allow creating new functions. Only existing functions can be picked there.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/formUtils.ts
```diff
@@ -294,6 +294,141 @@ export const stripCodeFences = (text: string): string => text.replace(CODE_FENCE
...
+    }
+    const approvalFn = typeof data["approvalFunction"] === "string" ? data["approvalFunction"].trim() : "";
+    const result: { requiresApproval?: string; generateApprovalFunction?: string } = {
+        requiresApproval: approvalFn || "true",
+    };
+    if (approvalFn && !compatibleFunctions.includes(approvalFn)) {
+        result.generateApprovalFunction = "true";
+    }
+    return result;
+}
```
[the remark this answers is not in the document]
lochana-chathura: Addressed in 454b342

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
```diff
@@ -288,63 +289,87 @@ private static void emitSignature(ToolGenContext ctx, List<ToolParam> params, Re
     }
 
     private static void emitAnnotation(ToolGenContext ctx) {
-        Map<String, Object> data = ctx.data != null && ctx.data.containsKey("auth")
+        // CUSTOM tools (and AGENT_CALL) have no wrapped node — auth/requiresApproval live directly
+        // on the tool's own data (ctx.data) instead. Checking both keys (not just "auth") keeps a
+        // custom tool's requiresApproval from being dropped when auth was never set.
+        boolean useTopLevelData = ctx.data != null
```
dan-niles: There is a small bug here: add a Function tool, fill in any one OAuth field (e.g. Client ID), tick `Requires Approval` with a new predicate name, save. You get:

```ballerina
@ai:AgentTool {
    auth: { clientId: "abc" }          // requiresApproval: isHighValue is missing
}
isolated function refundOrderTool(string orderId) returns string|error {
    ...
}

// generated anyway, but nothing references it
isolated function isHighValue(string orderId) returns boolean {
    return true;
}
```

The gate is gone, so the tool never pauses — but the predicate was still generated, so it sits there unused. No error anywhere. Leave the OAuth field empty and it works fine, which is what makes it easy to miss.

Cause: the form puts both `auth` and `requiresApproval` on the **wrapped node's** `codedata.data`. `buildAgentToolNode` (`utils.ts:606`) then copies `auth` onto the **tool node**, but not `requiresApproval`. Line 295 here switches to reading the tool node's data as soon as it finds `auth` — and `requiresApproval` isn't there. Meanwhile `appendApprovalPredicate` (line 649) always reads the wrapped node, so it still sees the flag and writes the stub.

OAuth settings are in `Advanced Configurations`:

Image

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
```diff
@@ -604,6 +633,54 @@ private static Map<Path, List<TextEdit>> buildFunctionBody(ToolGenContext ctx, R
...
+        String predicateName = approvalValue == null ? "" : approvalValue.toString().trim();
+        if (predicateName.isEmpty()) {
+            return;
+        }
+        // Mirror the tool's own signature exactly, using the same param list generate() built it
+        // from (stashed on ctx) — each ToolKind resolves params differently, so this must not
+        // recompute with a fixed kind.
+        String paramDecls = ctx.resolvedParams.stream()
+                .map(ToolParam::decl)
+                .collect(Collectors.joining(", "));
```
dan-niles, line 673: Another bug: Add Tool → Agent → pick an agent, tick **Pass agent context** under Advanced, tick `Requires Approval` and type a predicate name that doesn't exist yet (`needsReview`), save. You get:

```ballerina
@ai:AgentTool {
    requiresApproval: needsReview
}
isolated function helperAgentTool(ai:Context ctx, string query) returns string|error {
    ...
}

isolated function needsReview(ai:Context ctx, string query) returns boolean {
    return true;
}
```

`bal build` then fails on the `needsReview` reference inside the annotation:

```
ERROR: the `requiresApproval` predicate for the tool 'helperAgentTool' must have the
same parameter signature as the tool and return `boolean`
```

The predicate is supposed to match the tool's params **excluding** `ai:Context`. The ai compiler plugin strips a leading `ai:Context` from the *tool's* list before comparing, but not from the predicate's — so it compares `[string query]` against `(ai:Context ctx, string query)` and rejects it. The runtime agrees: `ai:Context` isn't `anydata`, so `bindApprovalPredicateArguments` in `approval.bal` skips it and the call fails safe to "pause" anyway.

One thing that make this easy to miss: the predicate name has to be **new** (pick an existing function from the dropdown and no stub is generated).

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AIAgentSidePanel.tsx
```diff
@@ -1255,6 +1318,22 @@ export function AIAgentSidePanel(props: BIFlowDiagramProps) {
...
+        // Mark the tool as requiring human-in-the-loop approval. This rides along in codedata.data
+        // (like `auth` above) and is rendered into the @ai:AgentTool annotation by the language
+        // server. The gate is the checkbox itself, not the function field — unchecking it yields an
+        // empty object, so a function picked earlier and left registered can't leak (buildApprovalToolData
+        // reads the checkbox first), which matters since CheckBoxConditionalEditor doesn't clear the
+        // "on"-state sub-field on uncheck.
+        if (targetNode) {
+            const approvalData = buildApprovalToolData(data, compatibleApprovalFunctionsRef.current);
+            if (Object.keys(approvalData).length > 0) {
+                targetNode.codedata.data = {
```
NipunaRanasinghe, line 1334: I think this drops the approval gate for function/connection tools whenever OAuth auth is also configured. Tracing it through:

`targetNode` here is `clonedFunctionNode || clonedFlowNode` (line 1304) — the **wrapped** node, not the AGENT_TOOL node. So `requiresApproval` and `generateApprovalFunction` land on the wrapped node's data.

The AGENT_TOOL node is then built by `buildAgentToolNode` in `utils.ts` (called from `NewTool.tsx` and `AgentDefinitionDesigner.tsx`), and that function lifts **only** `auth` up to the tool's own data:

```ts
const auth = wrappedNode.codedata.data?.auth;
const data: AgentToolData = { node: wrappedNode, ..., ...(typeof auth === "string" ? { auth } : {}) };
```

So on the Java side `ctx.data` ends up with `auth` but no approval keys, while `ctx.wrappedNode.codedata().data()` has both.

Now `emitAnnotation` sets `useTopLevelData = ctx.data.containsKey("auth") || ctx.data.containsKey("requiresApproval")`. With auth configured that's true, so it reads `ctx.data` only — and `requiresApproval` isn't there. The annotation comes out with `auth` and no gate. Meanwhile `appendApprovalPredicate` reads the wrapped node instead, finds `generateApprovalFunction`, and emits the predicate stub anyway. Net result: an orphan `isolated function` in the file and a tool that silently isn't gated.

Without auth it works, because `ctx.data` has neither key and the fallback to the wrapped node kicks in. That's why the CUSTOM-only fixtures pass.

Simplest fix is probably to lift the approval keys in `buildAgentToolNode` the same way `auth` is lifted, so the tool node's data is authoritative for both. Worth confirming against the OAuth + approval combination either way.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
```diff
@@ -288,63 +289,87 @@ private static void emitSignature(ToolGenContext ctx, List<ToolParam> params, Re
     }
 
     private static void emitAnnotation(ToolGenContext ctx) {
-        Map<String, Object> data = ctx.data != null && ctx.data.containsKey("auth")
+        // CUSTOM tools (and AGENT_CALL) have no wrapped node — auth/requiresApproval live directly
+        // on the tool's own data (ctx.data) instead. Checking both keys (not just "auth") keeps a
+        // custom tool's requiresApproval from being dropped when auth was never set.
+        boolean useTopLevelData = ctx.data != null
+                && (ctx.data.containsKey("auth") || ctx.data.containsKey("requiresApproval"));
```
NipunaRanasinghe: This picks one map for *all* annotation fields based on whether either key is present in it, which means both keys have to live in the same map or one of them gets dropped. Right now they don't for wrapped-node tools (see my comment on `AIAgentSidePanel.tsx:1330`).

Even after that's fixed, the either/or selection stays fragile — any future annotation field has to remember to land in whichever map won. Looking up each key independently would be sturdier:

```java
Map<String, Object> wrapped = ctx.wrappedNode != null ? ctx.wrappedNode.codedata().data() : null;
Object auth = value(ctx.data, wrapped, "auth");
Object approval = value(ctx.data, wrapped, "requiresApproval");
```

Then no single choice of map can silently swallow a field.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
```diff
@@ -604,6 +633,54 @@ private static Map<Path, List<TextEdit>> buildFunctionBody(ToolGenContext ctx, R
...
+    // call's arguments to the predicate by name). Its name is the `requiresApproval` annotation value.
+    // Must be called AFTER the tool declaration has been flushed via textEdit(DECLARATION) (so the
+    // token buffer is empty); it appends a second DECLARATION text edit to the same file, which
+    // build() returns alongside the tool's. Called from every ToolKind's body builder
+    // (buildFunctionBody, buildRemoteActionBody, buildResourceActionBody, CUSTOM's buildBody, and
+    // buildAgentCallBody) — every path whose form exposes the "Requires Approval" gate.
+    private static void appendApprovalPredicate(ToolGenContext ctx) {
+        // CUSTOM and AGENT_CALL tools have no wrapped node — generateApprovalFunction/requiresApproval
+        // live directly on the tool's own data (ctx.data) instead, mirroring emitAnnotation's fallback.
+        Map<String, Object> data = ctx.wrappedNode != null ? ctx.wrappedNode.codedata().data() : ctx.data;
```
NipunaRanasinghe: The comment says this mirrors `emitAnnotation`'s fallback, but the precedence is actually the other way round:

- `emitAnnotation`: prefer `ctx.data` when it has the keys, else the wrapped node.
- here: prefer the wrapped node whenever there is one, else `ctx.data`.

So the two functions can read different maps for the same tool, which is what produces the annotation-without-predicate / predicate-without-annotation mismatch described on `AIAgentSidePanel.tsx:1330`. Whatever resolution you settle on, it'd be good to have both read it through one shared helper so they can't drift apart again.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/java/io/ballerina/flowmodelgenerator/extension/agentsmanager/SourceGeneratorTest.java
```diff
@@ -61,6 +61,9 @@ protected Object[] getConfigsList() {
                 {Path.of("agent_model_source_default.json")},
                 {Path.of("memory_manager_source.json")},
                 {Path.of("custom_agent_tool_source.json")},
+                {Path.of("custom_agent_tool_approval_unconditional_source.json")},
+                {Path.of("custom_agent_tool_approval_existing_source.json")},
+                {Path.of("custom_agent_tool_approval_generate_source.json")},
```
NipunaRanasinghe, line 66: One gap: all three use `"toolKind": "CUSTOM"`, which has no wrapped node, so both `emitAnnotation` and `appendApprovalPredicate` read `ctx.data` and agree. The paths where they disagree — function and connection tools, which do have a wrapped node — aren't covered.

A fixture with a wrapped node plus `auth` and `requiresApproval` set together would have caught the issue I flagged on `AIAgentSidePanel.tsx:1330`, and would keep the two data-map lookups honest going forward.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AIAgentSidePanel.tsx
```diff
@@ -762,6 +772,36 @@ export function AIAgentSidePanel(props: BIFlowDiagramProps) {
...
+            const response = await rpcClient.getBIDiagramRpcClient().search(request);
+            const names = new Set<string>();
+            collectLocalFunctionNames((response?.categories ?? []) as (Category | AvailableNode)[], names);
+            if (toolFunctionSymbol) {
+                names.delete(toolFunctionSymbol);
+            }
+            return Array.from(names);
+        } catch (error) {
+            console.error(">>> Error fetching compatible approval functions", error);
+            return [];
```
NipunaRanasinghe: Returning `[]` on failure is silent, and it changes behaviour rather than just degrading it: `buildApprovalToolData` treats any name absent from the candidate list as new, so with an empty list every pick becomes `generateApprovalFunction: "true"`. The test at `formUtils.test.ts` ("treats every name as new when the candidate list is empty") codifies exactly that.

The consequence is that if this search fails or is slow, a user picking their existing `isHighValue` gets a second `isHighValue` generated, and the file no longer compiles.

Since the empty list can't be distinguished from "no functions in this project", could the failure be surfaced instead — hide or disable the picker, or fall back to the plain checkbox, when candidates couldn't be fetched? Same applies to the copies in `AgentToolForm.tsx` and `UseAgentToolForm.tsx`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
```diff
@@ -288,63 +289,87 @@ private static void emitSignature(ToolGenContext ctx, List<ToolParam> params, Re
...
         }
 
-        if (fields.isEmpty()) {
+        // requiresApproval: <boolean|isolated function> — human-in-the-loop gate. The value is
+        // emitted verbatim: "true" for the toggle, or an identifier for a predicate function pointer.
+        if (data.containsKey("requiresApproval")) {
+            Object approvalValue = data.get("requiresApproval");
+            String approval = approvalValue == null ? "" : approvalValue.toString().trim();
+            if (!approval.isEmpty()) {
+                annotationFields.add("    requiresApproval: " + approval);
```
NipunaRanasinghe, line 367: The value goes into source verbatim, and I can't find anything upstream that validates it as a Ballerina identifier. The picker is an AUTOCOMPLETE with `allowItemCreate: true` on create flows, so a user can type anything — `my function`, `2fast`, `check foo()` — and it lands here as-is. `appendApprovalPredicate` then also emits `isolated function my function(...)`, so you get two broken declarations from one typo.

An identifier check would be worth having, ideally on the form field so the user sees it before submitting rather than as a compile error afterwards.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/formUtils.ts
```diff
@@ -294,6 +294,150 @@ export const stripCodeFences = (text: string): string => text.replace(CODE_FENCE
 export const stripCodeFencesInline = (text: string): string =>
     text.replace(CODE_FENCE_REGEX, "").replace(/\n/g, " ").trim();
 
+// Categories in the FUNCTION search response that are NOT the project's own functions: standard
+// library and imported/third-party modules (plus agent tools). Approval-predicate pickers offer
+// only the user's own module-level functions as candidates, so these are skipped.
+const NON_LOCAL_FUNCTION_CATEGORIES = ["Standard Library", "Agent Tools"];
```
NipunaRanasinghe, line 303: Filtering on the category's display label means a wording change on the LS side ("Standard Library" → "Stdlib", or a localised label) silently turns the filter off, and stdlib functions start showing up as approval-predicate candidates. Nothing fails loudly — the list just gets wrong.

Is there a structural signal on the category or node to key off instead, e.g. `codedata.org` / `packageName` being the current project? That'd survive label changes. If the label really is all that's available, worth saying so in the comment so the next person doesn't go looking.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AgentToolForm.tsx
```diff
@@ -173,6 +184,62 @@ function toAuthSource(key: string, value: unknown, isExpression: boolean): strin
...
+    const name = existing?.functionName;
+    return name && !candidates.includes(name) ? [...candidates, name] : candidates;
+}
+
+// Insert, replace, or remove a scalar `key: value` field (no nested braces, unlike auth's record
+// value) inside an @ai:AgentTool annotation string. Mirrors the auth block's matchBraced-based
+// upsert in spirit, but auth's value is itself a `{ ... }` record while requiresApproval's value is
+// a bare boolean or identifier, so a brace-unaware match up to the next comma/close-brace suffices.
+function upsertScalarAnnotationField(annotationStr: string, key: string, source: string | undefined): string {
+    const match = new RegExp(`${key}\\s*:\\s*[^,}]+`).exec(annotationStr);
```
NipunaRanasinghe: Two small things about this regex, and the same shape in `parseRequiresApproval` above:

The pattern isn't anchored to a field boundary, so it can match inside a longer key. An auth field named `my_requiresApproval` would match starting at `requiresApproval`, and the replace would corrupt it into `my_requiresApproval: <newvalue>`. A `\b` (or requiring `{`/`,`/whitespace before the key) rules that out cheaply. Contrived today, but it's free to prevent.

Also, on the removal path: for `{\n    requiresApproval: true,\n    auth: {...}\n}` the match ends at the comma, `lead` doesn't hit, `trail` eats the comma, and what's left is a line of orphaned indentation before `auth`. Harmless but it'll show up in generated source — trimming the whole line would be tidier.

packages/ballerina-side-panel/src/components/editors/AutoCompleteEditor.tsx
```diff
@@ -46,9 +46,12 @@ export function AutoCompleteEditor(props: AutoCompleteEditorProps) {
                 required: buildRequiredRule({ isRequired: !field.optional, label: field.label }),
                 value: getValueForDropdown(field)
             })}
-            label={capitalize(field.label)}
+            // Append "(Optional)" AFTER capitalize() — capitalize is lodash startCase, which strips
+            // parentheses and title-cases words, so a paren'd label can't be passed through directly.
+            // Symmetric with required fields showing "*".
+            label={field.optional ? `${capitalize(field.label)} (Optional)` : capitalize(field.label)}
```
NipunaRanasinghe: This changes every optional AUTOCOMPLETE field, not just the approval picker, so it's a wider UI change than the PR title suggests.

I looked for an existing convention and didn't find one: the ordinary editors don't append anything, and the only precedent is `HeaderSetEditor`, which renders a separate styled `(optional)` element in lowercase. So after this, optional autocompletes will read `Approval Function (Optional)` while every other optional field in the same form shows a bare label.

Not saying it's wrong — just that it's a product-wide labelling decision riding along in a feature PR. Worth either confirming it's intended for all autocompletes, or scoping it to this field.

packages/ballerina-side-panel/src/components/editors/CheckBoxConditionalEditor.tsx
```diff
@@ -205,6 +212,7 @@ function mapPropertiesToFormFields(properties: { [key: string]: PropertyModel; }
             advanced: property.advanced,
             diagnostics: [],
             items,
+            allowItemCreate: (property as { allowItemCreate?: boolean }).allowItemCreate,
```
NipunaRanasinghe: The inline cast works, but it hides that `PropertyModel` has no `allowItemCreate` while `FormField` now does — so anyone setting it on a property has no type-level hint it's supported. Since you're already adding the field to `FormField` in `Form/types.ts`, adding it to `PropertyModel` too would let this be a plain `property.allowItemCreate`.

Same note for the `align-items: start` change above: it applies to every conditional editor in the product, not only the approval field. The reasoning in the comment is clear, I just want it to be a deliberate global change rather than an incidental one.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/UseAgentToolForm.tsx
```diff
@@ -74,26 +74,54 @@ export function UseAgentToolForm(props: UseAgentToolFormProps): JSX.Element {
...
+            setFields([
+                ...buildAgentToolFields(
+                    `${agentVarName}Tool`,
+                    `Delegates a query to ${agentLabel === "Agent" ? "the generic agent" : agentLabel}.`
+                ),
+                buildRequiresApprovalField(createRequiresApprovalField(), approvalCandidates),
+            ]);
             setReady(true);
         })();
     }, [agentNode]);
```
NipunaRanasinghe: This effect is now async on every path (previously the `hostClass` case returned synchronously), which introduces two things it didn't have before:

No `cancelled` guard — if `agentNode` changes while the search is in flight, the stale response still calls `setFields` and overwrites the newer one. The equivalent effect in `AgentToolForm.tsx` does use a `cancelled` flag; worth matching here.

The dep array is still `[agentNode]`, but the body now also reads `hostClass`, `agentVarName` and `agentLabel`. If any of those change without `agentNode` changing, the fields keep the old name and description.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/formUtils.ts
```diff
@@ -294,6 +294,150 @@ export const stripCodeFences = (text: string): string => text.replace(CODE_FENCE
...
+                        optional: true,
+                        editable: true,
+                    },
+                },
+            },
+            { metadata: { label: "Off" }, properties: {} },
+        ],
+    } as unknown as FormField;
+}
+
```
dan-niles, line 397: There is no identifier validation here, so a free-typed `my predicate` or `check` goes straight into both `requiresApproval: <name>` and `isolated function <name>(…)`. The `Tool Name` field right above is `IDENTIFIER` and does get validated.

Repro: Add Tool → Use Function → pick any function, tick Requires Approval, then type `my predicate` (with a space) into Approval Function rather than picking from the dropdown. Save. You get:

```ballerina
@ai:AgentTool {
    requiresApproval: my predicate
}
isolated function fileReadBytesTool(string path) returns byte[]|error {
    ...
}

isolated function my predicate(string path) returns boolean {
    return true;
}
```

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
lochana-chathura: Fixed in d7ea8f6

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
lochana-chathura: Fixed in d7ea8f6 and now it generates
```bal
isolated function needsReview(string query) returns boolean {
    return true;
}
```
as expected.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/formUtils.ts
lochana-chathura: Fixed in d7ea8f6. The following error is now shown.

Screenshot 2026-08-14 at 21 02 08

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AIAgentSidePanel.tsx
lochana-chathura: Same as https://github.com/ballerina-platform/ballerina-vscode/pull/842#discussion_r3782267539. Fixed in d7ea8f6.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
lochana-chathura: Same as https://github.com/ballerina-platform/ballerina-vscode/pull/842#discussion_r3782518213. Fixed in d7ea8f6.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
lochana-chathura: Fixed in 927f8a9. Added a `resolveToolData` helper that checks each key on its own instead of picking one map for everything.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
lochana-chathura: You're right, this was the real cause of the mismatch. Now uses the same `resolveToolData` helper as `emitAnnotation`, so they can't disagree anymore.

Fixed in 927f8a9.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/java/io/ballerina/flowmodelgenerator/extension/agentsmanager/SourceGeneratorTest.java
lochana-chathura: Fixed in 927f8a9. Added a test case for a FUNCTION tool with `auth` and `requiresApproval` together.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AIAgentSidePanel.tsx
lochana-chathura: Agreed, changed the fetch to return `null` on failure instead of `[]`. Create flows now fall back to the plain checkbox if the fetch fails.

Fixed in 927f8a9

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AgentToolForm.tsx
lochana-chathura: Both fixed. Added word boundaries so it won't match inside a longer key, and the removal now cleans up the leftover blank line too.

Fixed in d7ea8f6

packages/ballerina-visualizer/src/views/BI/AIChatAgent/UseAgentToolForm.tsx
lochana-chathura: Fixed in 927f8a9. Added the same cancelled guard the sibling form uses, and added the missing dependencies so it won't go stale.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/formUtils.ts
lochana-chathura: It only becomes a problem if the language server ever renames the `"Standard Library"` or `"Agent Tools"` category labels, which would silently let stdlib functions leak into the picker

Added a comment in dfadcd8, saying this only matches on labels because that's all the category gives us, and it'll break quietly if the labels change.

packages/ballerina-side-panel/src/components/editors/AutoCompleteEditor.tsx
lochana-chathura: Good point. Added a `showOptionalSuffix flag` on the field, off by default, only turned on for the approval field. Every other optional autocomplete goes back to its plain label. Addressed via ab63963

packages/ballerina-side-panel/src/components/editors/CheckBoxConditionalEditor.tsx
lochana-chathura: Added `allowItemCreate` to `PropertyModel` so no cast is needed anymore.

On the align-items change, that one's a real layout fix, not a preference. Without it, a revealed sub-field with a height:100% control stretches over the next section. It's not approval-specific, any conditional field with that kind of sub-field would hit it, so I'd rather keep it applied everywhere.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/formUtils.ts
```diff
@@ -294,6 +294,175 @@ export const stripCodeFences = (text: string): string => text.replace(CODE_FENCE
...
+                        // pick-list of existing functions.
+                        // The identifier rule guards the free-typed path: without it, a value like
+                        // "my predicate" goes straight into both `requiresApproval: <name>` and the
+                        // generated `isolated function <name>(...)`, producing invalid Ballerina source.
+                        types: [{
+                            fieldType: "AUTOCOMPLETE", selected: true,
+                            validations: [{ rule: "common.validate.identifier", message: "Invalid identifier" }],
+                        }],
+                        allowItemCreate: allowCreate,
+                        showOptionalSuffix: true,
```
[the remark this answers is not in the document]
lochana-chathura: Good catch. Added `showOptionalSuffix` to `PropertyModel` and copied it in `CheckBoxConditionalEditor`'s mapper alongside `allowItemCreate`. The suffix actually shows now.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
```diff
@@ -604,6 +641,56 @@ private static Map<Path, List<TextEdit>> buildFunctionBody(ToolGenContext ctx, R
...
+        if (predicateName.isEmpty()) {
+            return;
+        }
+        // Mirror the tool's own signature, using the same param list generate() built it from
+        // (stashed on ctx) — each ToolKind resolves params differently, so this must not recompute
+        // with a fixed kind. The leading `ai:Context ctx` param is excluded: the `ai` compiler
+        // plugin strips it from the tool's own signature before comparing against the predicate's,
+        // so including it here would make the predicate's signature look mismatched.
+        String paramDecls = ctx.resolvedParams.stream()
+                .filter(param -> !(ctx.includeContext && "ctx".equals(param.name())))
```
NipunaRanasinghe, line 671: `resolveToolData` came out well — reading each key on its own is exactly the property that stops the two emitters disagreeing, and the new `function_agent_tool_approval_with_auth_source.json` fixture pins the case that was broken.

One small thing on this line: the context parameter is identified by the name `"ctx"`. The invariant it's actually relying on is positional — when `includeContext` is set, the `ai:Context` param is the one the generator prepends. Keying on the name means that if the generated parameter is ever renamed, this silently stops stripping and you're back to the signature mismatch dan-niles hit, with no test failing in between (the fixtures wouldn't notice). It also drops any user parameter that happens to be called `ctx`, though a duplicate name would be invalid Ballerina anyway, so that half is harmless.

`ctx.resolvedParams.stream().skip(ctx.includeContext ? 1 : 0)` would tie it to the invariant that's actually true. Minor, but it's the kind of thing that only resurfaces as a compile error in a user's project.

NipunaRanasinghe (dismissed):
LGTM except for the below non-blocking comment

dan-niles (dismissed):
LGTM

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AgentToolForm.tsx
```diff
@@ -173,6 +184,64 @@ function toAuthSource(key: string, value: unknown, isExpression: boolean): strin
...
 }
 
+// Parse an existing `requiresApproval: <boolean|identifier>` field out of an @ai:AgentTool
+// annotation, to prefill the "Requires Approval" control when editing a custom tool that already
+// has the gate set.
+function parseRequiresApproval(annotationValue: string): ExistingApprovalConfig | undefined {
+    const match = /\brequiresApproval\b\s*:\s*([^,}]+)/.exec(annotationValue);
+    if (!match) return undefined;
+    const raw = match[1].trim();
+    return raw === "true" ? {} : { functionName: raw };
```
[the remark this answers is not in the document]
lochana-chathura: I will test this scenario, and fix it as a bug fix, separately.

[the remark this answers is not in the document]
lochana-chathura: Well, it was reproducible locally. Fixed in 9de46de and tested.

NipunaRanasinghe (approved):
LGTM

dan-niles (approved):
LGTM

PR #843, opened 2026-08-11
Show approval-required tools with a shield badge on the agent

lochana-chathura:
## Purpose
$subject.

Fixes https://github.com/wso2/product-integrator/issues/2130

## Preivew
Screenshot 2026-08-12 at 08 35 41

Tooltip:
Screenshot 2026-08-13 at 10 45 47

## Summary

Adds `requiresApproval` metadata to AI tools and propagates it through tool discovery and agent metadata. Displays an approval badge for tools that require approval in the agent diagram.


packages/bi-diagram/src/components/nodes/AgentNode/AgentNodeWidget.tsx
```diff
@@ -1159,6 +1159,36 @@ export function AgentNodeWidget(props: AgentNodeWidgetProps) {
...
+                                                color: ThemeColors.SECONDARY,
+                                            }}
+                                        >
+                                            {/* bi-shield glyph (wso2-vscode font, U+F18A) */}
+                                            {String.fromCharCode(0xf18a)}
+                                        </div>
+                                    </Tooltip>
+                                </foreignObject>
+                            )}
+
```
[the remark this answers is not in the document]
lochana-chathura: Fixed in 5d7c905

packages/bi-diagram/src/components/nodes/AgentNode/AgentNodeWidget.tsx
```diff
@@ -1159,6 +1159,36 @@ export function AgentNodeWidget(props: AgentNodeWidgetProps) {
...
+                                                fontFamily: "wso2-vscode",
+                                                fontSize: "10px",
+                                                color: ThemeColors.SECONDARY,
+                                            }}
+                                        >
+                                            {/* bi-shield glyph (wso2-vscode font, U+F18A) */}
+                                            {String.fromCharCode(0xf18a)}
+                                        </div>
+                                    </Tooltip>
+                                </foreignObject>
```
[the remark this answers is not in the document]
lochana-chathura: This is just a UI badge on existing data, not new command/RPC/LS logic. No test infra for these widgets yet, so deferring this one.

packages/bi-diagram/src/components/nodes/AgentCallNode/AgentCallNodeWidget.tsx
```diff
@@ -876,6 +876,31 @@ export function AgentCallNodeWidget(props: AgentCallNodeWidgetProps) {
                                 </div>
                             </foreignObject>
 
+                            {/* Approval-required badge: a shield on the tool's top-right corner when the
```
dan-niles: This badge code is duplicated in `AgentCallNodeWidget` and `AgentCallWidget`. Can we extract to a common component?

lochana-chathura: Addressed in [de146b9](https://github.com/ballerina-platform/ballerina-vscode/pull/843/commits/de146b925f75a77c0b29b02334d17af2c269f78f)

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
```diff
@@ -1723,6 +1728,58 @@ && unwrapConstant(annot.attachmentValue().get()) instanceof Map<?, ?> map) {
...
+     * function-typed field ({@code RequiresApproval = boolean | isolated function}), which makes the
+     * compiler's constant-value construction throw for every AgentTool annotation.
+     */
+    public static boolean readRequiresApproval(Symbol toolSymbol, Project project) {
+        try {
+            Optional<Location> location = toolSymbol.getLocation();
+            if (location.isEmpty()) {
+                return false;
+            }
+            Document document = CommonUtils.getDocument(project, location.get());
```
NipunaRanasinghe: This looks up the document in the entry `project`, but the agent class doesn't have to live there. `isWorkspaceClass` (line 1652) accepts a class from any package in the workspace, since `getProjectsForModule` walks all the workspace projects.

So for an agent defined in a sibling package, `getDocument` returns null here, `document.syntaxTree()` throws an NPE, the catch at the bottom swallows it, and every tool comes back as `requiresApproval = false`. The badge just never shows up and nothing is logged.

`workspaceSystemPrompt` (lines 1578-1596) already handles the same situation: it loops over `getProjectsForModule(org, packageName, project)` and null-checks the document for each candidate. Could this do the same? At the very least a `if (document == null) { return false; }` guard, so an NPE isn't doing the control flow.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
```diff
@@ -1723,6 +1728,58 @@ && unwrapConstant(annot.attachmentValue().get()) instanceof Map<?, ?> map) {
...
+            NonTerminalNode node = CommonUtil.findNode(toolSymbol, document.syntaxTree()).orElse(null);
+            while (node != null && !(node instanceof FunctionDefinitionNode)) {
```
NipunaRanasinghe: This walks up to the nearest enclosing function definition, but nothing checks that it's the function that actually declares `toolSymbol`.

It matters because of the call site in `CodeAnalyzer` (line 807): that branch passes whatever `semanticModel.symbol(element)` returns, which isn't always a `FunctionSymbol` — only `McpToolKit` variables get filtered out above it. If some other symbol's location happens to sit inside a function that itself carries `@ai:AgentTool`, this walk-up finds that outer function's annotation and reports approval for the wrong thing.

An early `if (!(toolSymbol instanceof FunctionSymbol)) { return false; }` would rule it out, or you could compare the found function's name against the symbol name.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
```diff
@@ -1723,6 +1728,58 @@ && unwrapConstant(annot.attachmentValue().get()) instanceof Map<?, ?> map) {
...
+            }
+            if (node == null) {
+                return false;
+            }
+            FunctionDefinitionNode functionDefinition = (FunctionDefinitionNode) node;
+            if (functionDefinition.metadata().isEmpty()) {
+                return false;
+            }
+            for (AnnotationNode annotation : functionDefinition.metadata().get().annotations()) {
+                if (!annotation.annotReference().toSourceCode().trim().endsWith(AGENT_TOOL_ANNOT)) {
```
NipunaRanasinghe: `endsWith("AgentTool")` will also match an annotation from a completely different module, e.g. `@myorg:CustomAgentTool`.

Just above, `toolMethodsOf` uses `hasAiAnnotation(method.annotAttachments(), AGENT_TOOL_ANNOT)`, which checks the module is `ai` before accepting it. Worth matching that here — either compare the whole `ai:AgentTool` reference including the prefix, or resolve the annotation through the semantic model and check its module.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
```diff
@@ -1723,6 +1728,58 @@ && unwrapConstant(annot.attachmentValue().get()) instanceof Map<?, ?> map) {
...
+                }
+                return annotation.annotValue()
+                        .flatMap(mapping -> mapping.fields().stream()
+                                .filter(field -> field instanceof SpecificFieldNode specificField
+                                        && REQUIRES_APPROVAL.equals(specificField.fieldName().toSourceCode().trim()))
+                                .map(field -> (SpecificFieldNode) field)
+                                .findFirst())
+                        // Gated unless the value is the literal `false`; absent field -> not gated.
+                        .map(specificField -> !"false".equals(
+                                specificField.valueExpr().map(expr -> expr.toSourceCode().trim()).orElse("")))
```
NipunaRanasinghe: The "anything but the literal `false` counts as gated" rule makes sense for the predicate-function case, but it also catches a plain boolean reference: `requiresApproval: approvalDisabled` where that constant is `false` will show the badge.

Probably fine in practice, and I'd leave it as is. Could you just extend the javadoc to say it outright — that any non-literal expression, including a `boolean`-typed reference that evaluates to false, is treated as gated? Right now the doc only mentions `true` or a predicate function, so the next reader won't expect that case.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
```diff
@@ -1723,6 +1728,58 @@ && unwrapConstant(annot.attachmentValue().get()) instanceof Map<?, ?> map) {
...
+                        .orElse(false);
+            }
+            return false;
+        } catch (RuntimeException e) {
+            // Never let annotation reading break flow-model generation.
```
NipunaRanasinghe, line 1803: Not letting this break flow-model generation is the right call, but catching silently means a real failure is invisible in the LS logs — you'd see no badge and have nothing to go on.

`getProjectsForModule` in this same file logs at `Level.FINE` when it gives up. Could this do the same?

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
```diff
@@ -1723,6 +1728,58 @@ && unwrapConstant(annot.attachmentValue().get()) instanceof Map<?, ?> map) {
...
+     * {@code requiresApproval} field is present with any value other than the literal {@code false}
+     * (i.e. {@code true} or a predicate-function reference). A bare {@code @ai:AgentTool}, an explicit
+     * {@code requiresApproval: false}, or a missing field are all not gated.
+     * <p>
+     * The annotation value is read syntactically rather than via
+     * {@code AnnotationAttachmentSymbol.attachmentValue()} because the AgentTool config record has a
+     * function-typed field ({@code RequiresApproval = boolean | isolated function}), which makes the
+     * compiler's constant-value construction throw for every AgentTool annotation.
+     */
+    public static boolean readRequiresApproval(Symbol toolSymbol, Project project) {
```
NipunaRanasinghe, line 1748: Thanks for the javadoc here — explaining why the syntax tree is read instead of `attachmentValue()` will save whoever touches this next a lot of time.

One ask: could we get a test for it? This reads the tree directly and falls back to `false` on anything unexpected, so if it breaks it breaks quietly. The four cases the javadoc describes are cheap to lock in with flow-model config cases: `requiresApproval: true`, `requiresApproval: false`, a predicate-function reference, and a bare `@ai:AgentTool`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
```diff
@@ -1806,7 +1863,12 @@ private static AgentToolData parseToolMetadata(Map<?, ?> toolMap) {
             return null;
         }
         boolean isMcp = "MCP_TOOLKIT".equals(constantString(toolMap.get("kind")));
-        return new AgentToolData(name, constantString(toolMap.get("icon")), null, isMcp ? "MCP Server" : null);
+        // Prebuilt/published agents carry tool metadata in the compiled @display{agentMetadata}. The
+        // requiresApproval flag is read here defensively so it works once the generation side emits it;
+        // absent → false.
+        boolean requiresApproval = "true".equals(constantString(toolMap.get(REQUIRES_APPROVAL)));
```
NipunaRanasinghe, line 1897: The read itself works either way — `constantString` goes through `toString()`, so a `Boolean` and a `String` both land on `"true"`.

Since the generation side doesn't emit the field yet, this path always yields `false` today, which means prebuilt and published agents show no badge at all. Worth a line in the PR description so a reviewer testing that path doesn't file it as a bug.

packages/bi-diagram/src/components/nodes/AgentNode/AgentNodeWidget.tsx
```diff
@@ -34,7 +34,7 @@ import {
     NODE_WIDTH,
     NodeTypes,
 } from "../../../resources/constants";
-import { Button, Icon, Item, Menu, MenuItem, Popover, ThemeColors, getAIModuleIcon, DefaultLlmIcon } from "@wso2/ui-toolkit";
+import { Button, Icon, Item, Menu, MenuItem, Popover, ThemeColors, Tooltip, getAIModuleIcon, DefaultLlmIcon } from "@wso2/ui-toolkit";
```
NipunaRanasinghe: `Tooltip` isn't used anywhere in this file — the only other match is the unrelated `AgentIdTooltip` styled div, and the actual tooltip now lives inside `ApprovalBadge`. Looks like a leftover from before the component was pulled out. (`AgentCallNodeWidget` was already importing it for "View Agent", so it's only stale here.)

packages/bi-diagram/src/components/nodes/AgentWidget/ApprovalBadge.tsx
```diff
@@ -0,0 +1,54 @@
...
+    background: string;
+}
+
+/**
+ * A shield badge on the tool's top-right corner shown when the tool's
+ * @ai:AgentTool annotation gates it for human-in-the-loop approval.
+ */
+export function ApprovalBadge({ background }: ApprovalBadgeProps) {
+    return (
+        <foreignObject x="88.5" y="-0.5" width="17" height="17" style={{ overflow: "visible" }}>
```
NipunaRanasinghe, line 36: These coordinates come from the tool circle's geometry (`cx=80 cy=24 r=22`), which is the same in both widgets, so the badge lands on the 45° point of the arc in both. That checks out.

It is an invisible coupling though — if either widget's circle moves, the badge silently drifts off it. A short comment saying the numbers are derived from that circle would be enough.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
lochana-chathura: Fixed in 847623b. Fixed by walking `getProjectsForModule` like `workspaceSystemPrompt` does, with a null-check instead of relying on the NPE fallthrough.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
lochana-chathura: Fixed in 847623b. Added the early guard.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
lochana-chathura: Fixed in 847623b. Now short-circuits on `hasAiAnnotation(...)` (same module check `toolMethodsOf` uses) and matches the syntax node via `QualifiedNameReferenceNode.identifier()` instead of a suffix compare.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
lochana-chathura: Added `LOGGER.log(Level.FINE, ...)`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
lochana-chathura: Added `AiUtilsReadRequiresApprovalTest` covering all four cases.

packages/bi-diagram/src/components/nodes/AgentNode/AgentNodeWidget.tsx
lochana-chathura: Removed

packages/bi-diagram/src/components/nodes/AgentWidget/ApprovalBadge.tsx
lochana-chathura: Added a comment

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
lochana-chathura: Called this out explicitly in the comment.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
lochana-chathura: Updated the javadoc

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
lochana-chathura: Reverted this test case temporary with 8703898 due to LS test failing. Will add it in a separate PR.

NipunaRanasinghe: Thanks for flagging the revert rather than quietly dropping it.

Two things before this merges, though. Was the failure in the test itself (fixture wiring, distribution not available in that job) or did it actually catch something in `readRequiresApproval`? Those have very different implications — the second would mean the parser still has a hole, and the rest of this PR rests on it.

And could you link the follow-up issue here? `readRequiresApproval` is now the one piece of this PR with no coverage at all, it's the piece most likely to break quietly (every failure path returns `false`, which looks identical to "not gated"), and it's just been rewritten — cross-package project walking, the `FunctionSymbol` guard, and the qualified-name matching all landed after the original review. A "separate PR" without a tracking issue tends not to happen.

The rest of the changes in 847623b look right to me: the workspace walk mirrors `workspaceSystemPrompt` with the null check, the `instanceof FunctionSymbol` + `hasAiAnnotation` short-circuit removes the wrong-function risk, and matching on `QualifiedNameReferenceNode.identifier()` is tighter than the suffix compare I complained about.

lochana-chathura: `readRequiresApproval` functionality correctness was tested with a built vsix with the all changes. My suspicion is the ai library version bump caused the test failure.  So in your terms, failure is the test itself.

Created issue https://github.com/wso2/product-integrator/issues/2158 to track it.

NipunaRanasinghe (approved):
LGTM

dan-niles (approved):
LGTM

PR #844, opened 2026-08-12
Add approval UI for human-in-the-loop AI chat pauses

lochana-chathura:
## Purpose
$subject.

Fixes https://github.com/wso2/product-integrator/issues/2132

## Preview
https://github.com/user-attachments/assets/08c4793c-d171-4224-9101-3a6c9c4956a4

## Summary

- Add human-in-the-loop approval support to AI chat.
- Extend agent chat RPC contracts and clients with `submitDecision`.
- Handle pending approvals, decisions, repeated approval pauses, and final responses.
- Add `ApprovalCard` for individual or batch approval and rejection.
- Preserve approval state in chat history and disable chat input while approval is pending.

### DIFF
--- a/packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-handler.ts
+++ b/packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-handler.ts
@@ -32,7 +32,9 @@ import {
     getSessionInfo,
     getAvailableChatAgents,
     switchChatAgent,
-    SwitchAgentRequest
+    SwitchAgentRequest,
+    submitDecision,
+    SubmitDecisionRequest
 } from "@wso2/ballerina-core";
 import { Messenger } from "vscode-messenger";
 import { AgentChatRpcManager } from "./rpc-manager";
@@ -41,6 +43,7 @@ export function registerAgentChatRpcHandlers(messenger: Messenger) {
     const rpcManger = new AgentChatRpcManager();
     messenger.onRequest(getChatMessage, (args: ChatReqMessage) => rpcManger.getChatMessage(args));
     messenger.onNotification(abortChatRequest, () => rpcManger.abortChatRequest());
+    messenger.onRequest(submitDecision, (args: SubmitDecisionRequest) => rpcManger.submitDecision(args));
     messenger.onRequest(getTracingStatus, (params) => rpcManger.getTracingStatus(params));
     messenger.onNotification(showTraceView, (args: TraceInput) => rpcManger.showTraceView(args));
     messenger.onNotification(showSessionOverview, (args: SessionInput) => rpcManger.showSessionOverview(args));
--- a/packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-manager.ts
+++ b/packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-manager.ts
@@ -32,7 +32,11 @@ import {
     SessionInfoResponse,
     AvailableAgentsResponse,
     SwitchAgentRequest,
-    SwitchAgentResponse
+    SwitchAgentResponse,
+    SubmitDecisionRequest,
+    DecisionMessage,
+    HumanResponse,
+    PendingApprovalInfo
 } from "@wso2/ballerina-core";
 import * as vscode from 'vscode';
 import * as fs from 'fs';
@@ -81,7 +85,18 @@ export class AgentChatRpcManager implements AgentChatAPI {
                     payload,
                     this.currentAbortController.signal
                 );
-                if (response && response.message) {
+                if (response && response.pendingApproval) {
+                    const pendingApproval: PendingApprovalInfo = response.pendingApproval;
+
+                    this.addMessageToHistory(sessionId, {
+                        type: 'approval',
+                        text: '',
+                        isUser: false,
+                        pendingApproval
+                    });
+
+                    resolve({ message: '', pendingApproval } as ChatRespMessage);
+                } else if (response && response.message) {
                     // Find trace and extract tool calls and execution steps
                     const trace = this.findTraceForMessage(extension.agentChatContext.chatSessionId, params.message, response.message);
                     const executionSteps = trace ? this.extractExecutionSteps(trace) : undefined;
@@ -149,6 +164,100 @@ export class AgentChatRpcManager implements AgentChatAPI {
         AgentChatRpcManager.chatHistoryMap.get(sessionId)!.push(message);
     }
 
+    // Merges the decisions just submitted into whichever 'approval' history entry they belong
+    // to (a batch may be resolved across several submitDecision calls, one request at a time),
+    // so a later getChatHistory()/switchChatAgent() replay renders resolved requests collapsed
+    // rather than as if still pending.
+    private resolvePendingApprovalInHistory(sessionId: string, decisions: Record<string, HumanResponse>): void {
+        const history = AgentChatRpcManager.chatHistoryMap.get(sessionId);
+        if (!history) {
+            return;
+        }
+        const decidedIds = Object.keys(decisions);
+        for (let i = history.length - 1; i >= 0; i--) {
+            const entry = history[i];
+            if (entry.type !== 'approval' || !entry.pendingApproval) {
+                continue;
+            }
+            const belongsToThisBatch = entry.pendingApproval.requests.some(r => decidedIds.includes(r.id));
+            if (belongsToThisBatch) {
+                entry.decisions = { ...(entry.decisions || {}), ...decisions };
+                break;
+            }
+        }
+    }
+
+    // The dispatcher attaches `chat` and `decision` as sibling resources under the same
+    // service base path (see ai:Listener's ChatDispatcherService), and `chatEp` is always
+    // built ending in `/chat` (see buildChatEndpoint in features/tryit/activator.ts).
+    private toDecisionEndpoint(chatEp: string): string {
+        return chatEp.replace(/\/chat$/, '/decision');
+    }
+
+    async submitDecision(params: SubmitDecisionRequest): Promise<ChatRespMessage> {
+        return new Promise(async (resolve, reject) => {
+            try {
+                if (
+                    !extension.agentChatContext.chatEp || typeof extension.agentChatContext.chatEp !== 'string' ||
+                    !extension.agentChatContext.chatSessionId || typeof extension.agentChatContext.chatSessionId !== 'string'
+                ) {
+                    throw new Error('Invalid Agent Chat Context: Missing or incorrect ChatEP or ChatSessionID!');
+                }
+
+                const sessionId = extension.agentChatContext.chatSessionId;
+                const decisionEp = this.toDecisionEndpoint(extension.agentChatContext.chatEp);
+                const payload: DecisionMessage = { sessionId, decisions: params.decisions };
+
+                this.currentAbortController = new AbortController();
+                const response = await this.fetchTestData(decisionEp, payload, this.currentAbortController.signal);
+
+                this.resolvePendingApprovalInHistory(sessionId, params.decisions);
+
+                if (response && response.pendingApproval) {
+                    // The run resumed only to pause again (e.g. a further gated call surfaced
+                    // in the same turn), so this is a fresh batch, not a resolution of the last one.
+                    const pendingApproval: PendingApprovalInfo = response.pendingApproval;
+
+                    this.addMessageToHistory(sessionId, {
+                        type: 'approval',
+                        text: '',
+                        isUser: false,
+                        pendingApproval
+                    });
+
+                    resolve({ message: '', pendingApproval } as ChatRespMessage);
+                } else if (response && response.message) {
+                    const trace = this.findTraceForMessage(sessionId, '', response.message);
+                    const executionSteps = trace ? this.extractExecutionSteps(trace) : undefined;
+
+                    this.addMessageToHistory(sessionId, {
+                        type: 'message',
+                        text: response.message,
+                        isUser: false,
+                        traceId: trace?.traceId,
+                        executionSteps
+                    });
+
+                    resolve({
+                        message: response.message,
+                        traceId: trace?.traceId,
+                        executionSteps
+                    } as ChatRespMessage);
+                } else {
+                    reject(new Error("Invalid response format from the decision endpoint."));
+                }
+            } catch (error) {
+                const errorMessage =
+                    error && typeof error === "object" && "message" in error
+                        ? String(error.message)
+                        : "An unknown error occurred";
+                reject(new Error(errorMessage));
+            } finally {
+                this.currentAbortController = null;
+            }
+        });
+    }
+
     abortChatRequest(): void {
         if (this.currentAbortController) {
             this.currentAbortController.abort();
@@ -178,7 +287,21 @@ export class AgentChatRpcManager implements AgentChatAPI {
             });
 
             if (!response.ok) {
-                const errorData = await response.json();
+                const errorData: any = await response.json().catch(() => undefined);
+
+                // The dispatcher maps a paused run (ai:ApprovalRequiredError) to HTTP 403 with a
+                // structured `{ requests: ApprovalRequest[] }` body - not a real error, so surface
+                // it to the caller instead of throwing.
+                if (response.status === 403 && errorData && Array.isArray(errorData.requests)) {
+                    return { pendingApproval: { requests: errorData.requests } };
+                }
+
+                // A `decision` resume naming a session/id with nothing pending maps to 404/400
+                // with a stable `errorType` discriminator (ApprovalNotFoundError/UnknownApprovalIdError).
+                if (errorData?.errorType === 'ApprovalNotFoundError' || errorData?.errorType === 'UnknownApprovalIdError') {
+                    throw new Error(errorData.message || 'The pending approval could not be resumed.');
+                }
+
                 switch (response.status) {
                     case 400:
                         throw new Error("Bad Request: The server could not understand the request.");
--- a/packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-manager.ts
+++ b/packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-manager.ts
@@ -49,6 +49,9 @@ import { v4 as uuidv4 } from "uuid";
 import { updateChatSessionId } from "../../features/tryit/activator";
 
 export class AgentChatRpcManager implements AgentChatAPI {
+    // Shared by getChatMessage and submitDecision. Safe only because the UI serializes them -
+    // input stays disabled while an approval is pending, so the two never overlap. If that
+    // ever changes, give each operation its own controller.
     private currentAbortController: AbortController | null = null;
     // Store chat history per session ID
     private static chatHistoryMap: Map<string, ChatHistoryMessage[]> = new Map();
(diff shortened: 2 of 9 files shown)


packages/ballerina-visualizer/src/views/AgentChatPanel/Components/ChatInterface.tsx
```diff
@@ -529,6 +557,10 @@ const ChatInterface: React.FC = () => {
     // Check if we have any traces (to enable/disable Session Traces button)
     const hasTraces = messages.some(msg => !msg.isUser && msg.traceId);
 
+    // Block free-text input while the latest turn is an unresolved approval request
+    const lastMessage = messages[messages.length - 1];
+    const isApprovalPending = !!lastMessage && isApprovalUnresolved(lastMessage);
+
```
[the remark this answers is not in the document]
lochana-chathura: Fixed in [1ad63f6](https://github.com/ballerina-platform/ballerina-vscode/pull/844/commits/1ad63f662b0ae3ac16f13b123a0bb65f15a4df7b)

packages/ballerina-visualizer/src/views/AgentChatPanel/Components/ChatInterface.tsx
```diff
@@ -750,6 +771,68 @@ const ChatInterface: React.FC = () => {
...
+                    ...prev,
+                    {
+                        type: ChatMessageType.MESSAGE,
+                        text: response.message,
+                        isUser: false,
+                        traceId: response.traceId,
+                        executionSteps: response.executionSteps
+                    },
+                ]);
+            }
```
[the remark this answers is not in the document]
lochana-chathura: Fixed in [1ad63f6](https://github.com/ballerina-platform/ballerina-vscode/pull/844/commits/1ad63f662b0ae3ac16f13b123a0bb65f15a4df7b)

packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-handler.ts
```diff
@@ -41,6 +43,7 @@ export function registerAgentChatRpcHandlers(messenger: Messenger) {
     const rpcManger = new AgentChatRpcManager();
     messenger.onRequest(getChatMessage, (args: ChatReqMessage) => rpcManger.getChatMessage(args));
     messenger.onNotification(abortChatRequest, () => rpcManger.abortChatRequest());
+    messenger.onRequest(submitDecision, (args: SubmitDecisionRequest) => rpcManger.submitDecision(args));
```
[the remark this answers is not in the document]
lochana-chathura: this is quite complicated. Not going to address in this PR

packages/ballerina-visualizer/src/views/AgentChatPanel/Components/ChatInterface.tsx
```diff
@@ -750,6 +779,85 @@ const ChatInterface: React.FC = () => {
...
+
+            console.error("Submit decision error:", error);
+
+            setMessages((prev) => [...prev, {
+                type: ChatMessageType.ERROR,
+                text: errorMessage,
+                isUser: false,
+                traceId,
+                executionSteps
+            }]);
```
dan-niles, line 865: This part is duplicated from `handleSendMessage` function in this same file. Can we refactor to avoid duplication?

dan-niles (commented):
LGTM!

Just as a side note, we could add a dedicated view for `resolve_human_approval` spans in the dev tracing UI in the future as well. No need to include it in this PR. We can track in a seperate issue and do it later.

image

packages/ballerina-visualizer/src/views/AgentChatPanel/Components/ChatInterface.tsx
lochana-chathura: Addressed in 7482d4d

dan-niles (dismissed):

packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-manager.ts
```diff
@@ -149,6 +164,100 @@ export class AgentChatRpcManager implements AgentChatAPI {
...
+                    this.addMessageToHistory(sessionId, {
+                        type: 'approval',
+                        text: '',
+                        isUser: false,
+                        pendingApproval
+                    });
+
+                    resolve({ message: '', pendingApproval } as ChatRespMessage);
+                } else if (response && response.message) {
+                    const trace = this.findTraceForMessage(sessionId, '', response.message);
```
NipunaRanasinghe, line 238: Passing `''` as the user message defeats the matching inside `findTraceForMessage`. It does:

```ts
const inputMatches = inputMessages && inputMessages.includes(userMessage);
```

and `String.prototype.includes('')` is always true, so the first `invoke_agent` span with the right session id matches and the function returns immediately — before `outputMatches` is even considered. In effect this returns "the most recent trace for this session" rather than the trace for this response.

Usually that's the right trace, but not always: if the resumed run's trace hasn't been exported yet, this attaches the *previous* turn's `traceId` to the resumed reply, and "View Trace" then opens the wrong trace.

Since you do have `response.message` here, matching on the output alone would be correct. Guarding the input branch would do it:

```ts
const inputMatches = userMessage && inputMessages && inputMessages.includes(userMessage);
```

packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-manager.ts
```diff
@@ -149,6 +164,100 @@ export class AgentChatRpcManager implements AgentChatAPI {
...
+                }
+
+                const sessionId = extension.agentChatContext.chatSessionId;
+                const decisionEp = this.toDecisionEndpoint(extension.agentChatContext.chatEp);
+                const payload: DecisionMessage = { sessionId, decisions: params.decisions };
+
+                this.currentAbortController = new AbortController();
+                const response = await this.fetchTestData(decisionEp, payload, this.currentAbortController.signal);
+
+                this.resolvePendingApprovalInHistory(sessionId, params.decisions);
```
NipunaRanasinghe: This marks the decisions resolved in history as soon as the fetch returns, before the response shape has been checked. If the response is neither a `pendingApproval` nor a `message`, the `else` at the bottom rejects — but history already says resolved.

The webview only merges decisions into its own message on success, so at that point the two disagree: the UI shows an error with the card still pending (correctly), while the extension's history says the batch was decided. Switch agents and back, and the replayed card renders collapsed as if the decision went through — when it never did.

Moving this call into the two success branches would keep the extension's history and the webview in step.

packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-manager.ts
```diff
@@ -149,6 +164,100 @@ export class AgentChatRpcManager implements AgentChatAPI {
...
+                break;
+            }
+        }
+    }
+
+    // The dispatcher attaches `chat` and `decision` as sibling resources under the same
+    // service base path (see ai:Listener's ChatDispatcherService), and `chatEp` is always
+    // built ending in `/chat` (see buildChatEndpoint in features/tryit/activator.ts).
+    private toDecisionEndpoint(chatEp: string): string {
+        return chatEp.replace(/\/chat$/, '/decision');
```
NipunaRanasinghe, line 200: If `chatEp` ever doesn't end in `/chat` — a trailing slash, a query string, or `buildChatEndpoint` changing shape — `replace` is a no-op and returns the chat endpoint unchanged. The `DecisionMessage` then gets POSTed to `/chat`, where it's a malformed chat request rather than an obvious wiring bug, so the failure surfaces as a confusing chat error.

The comment documents the invariant, which is good, but it'd be better enforced than assumed:

```ts
if (!chatEp.endsWith('/chat')) {
    throw new Error(`Unexpected chat endpoint shape: ${chatEp}`);
}
```

packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-manager.ts
```diff
@@ -149,6 +164,100 @@ export class AgentChatRpcManager implements AgentChatAPI {
...
+                    } as ChatRespMessage);
+                } else {
+                    reject(new Error("Invalid response format from the decision endpoint."));
+                }
+            } catch (error) {
+                const errorMessage =
+                    error && typeof error === "object" && "message" in error
+                        ? String(error.message)
+                        : "An unknown error occurred";
+                reject(new Error(errorMessage));
```
NipunaRanasinghe: Two differences from `getChatMessage`'s catch that look unintentional:

It doesn't call `addMessageToHistory` with an `'error'` entry. The webview does render the error, but it's never recorded, so switching agents and back makes it disappear while the still-pending card remains — the user loses the only explanation for why the card didn't resolve.

It also rejects with a bare message rather than the `JSON.stringify({ message, traceInfo })` envelope. `parseAgentError` falls back gracefully, so nothing breaks, but a failed decision can't carry a traceId the way a failed chat can.

packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-manager.ts
```diff
@@ -149,6 +164,100 @@ export class AgentChatRpcManager implements AgentChatAPI {
...
+                    !extension.agentChatContext.chatSessionId || typeof extension.agentChatContext.chatSessionId !== 'string'
+                ) {
+                    throw new Error('Invalid Agent Chat Context: Missing or incorrect ChatEP or ChatSessionID!');
+                }
+
+                const sessionId = extension.agentChatContext.chatSessionId;
+                const decisionEp = this.toDecisionEndpoint(extension.agentChatContext.chatEp);
+                const payload: DecisionMessage = { sessionId, decisions: params.decisions };
+
+                this.currentAbortController = new AbortController();
```
NipunaRanasinghe, line 217: `currentAbortController` is now shared between `getChatMessage` and `submitDecision`. Assigning here overwrites whatever a chat request stored, and the `finally` nulls it unconditionally — so if the two ever overlap, `abortChatRequest()` aborts the wrong one or nothing at all.

The UI does keep them apart today (the input is disabled while an approval is pending), so I don't think it's reachable. Worth a line saying that's why one controller is enough, or one field per operation if you'd rather not depend on the UI for it.

packages/ballerina-visualizer/src/views/AgentChatPanel/Components/ChatInterface.tsx
```diff
@@ -525,10 +553,20 @@ const ChatInterface: React.FC = () => {
...
+    const sessionTokenRef = useRef(0);
 
     // Check if we have any traces (to enable/disable Session Traces button)
     const hasTraces = messages.some(msg => !msg.isUser && msg.traceId);
 
+    // Block free-text input while any turn has an unresolved approval request. Scanning the
+    // whole list (not just the last message) matters because a failed decision submission
+    // appends an error message after the approval card, which would otherwise become
+    // `lastMessage` and incorrectly report nothing pending.
+    const isApprovalPending = messages.some(isApprovalUnresolved);
```
NipunaRanasinghe, line 573: This can leave the chat permanently unusable. If a pending approval can no longer be resumed — the service restarted, the run timed out — history replays the card as unresolved, `isApprovalPending` stays true, and the input is disabled. Submitting a decision returns `ApprovalNotFoundError`, which you handle, but handling it just appends an error; the card stays pending, so the input stays disabled. There's no way out except Clear Chat, which isn't obviously the remedy from the user's point of view.

Since you already recognise `ApprovalNotFoundError`/`UnknownApprovalIdError` specifically, you have what you need to treat it as terminal — mark that batch resolved (or otherwise dismissable) so the conversation unblocks, and say why in the error text.

packages/ballerina-visualizer/src/views/AgentChatPanel/Components/ChatInterface.tsx
```diff
@@ -694,49 +716,114 @@ const ChatInterface: React.FC = () => {
...
+
+            if (sessionTokenRef.current !== requestToken) {
+                return;
+            }
+
+            setMessages((prev) => {
+                const updated = [...prev];
+                const target = updated[msgIndex];
+                if (target) {
+                    updated[msgIndex] = { ...target, decisions: { ...(target.decisions || {}), ...decisions } };
```
NipunaRanasinghe: Using the array index as the identity of the approval message works right now, because everything that mutates `messages` between the click and the response either appends (indices stay valid) or replaces the whole array behind a session-token bump. But it's held together by those two facts rather than by anything local, and the `if (target)` guard means a mismatch writes the decisions nowhere and fails silently.

Matching on the request ids you already have would be self-evidently correct and wouldn't care about ordering:

```ts
const ids = Object.keys(decisions);
updated.findIndex(m => m.pendingApproval?.requests.some(r => ids.includes(r.id)))
```

packages/ballerina-visualizer/src/views/AgentChatPanel/Components/ChatInterface.tsx
```diff
@@ -694,49 +716,114 @@ const ChatInterface: React.FC = () => {
...
+
+    const handleSendMessage = async (text: string) => {
+        if (!text.trim()) return;
+
+        setMessages((prev) => [...prev, { type: ChatMessageType.MESSAGE, text, isUser: true }]);
+        setIsLoading(true);
+
+        try {
+            const chatResponse = await rpcClient.getAgentChatRpcClient().getChatMessage({ message: text });
+            appendAgentResponse(chatResponse);
```
NipunaRanasinghe, line 792: `handleApprovalDecision` captures `sessionTokenRef` and drops a response that arrives after a session change, but `handleSendMessage` doesn't — so a chat reply that lands after the user switched agents still gets appended to the new session's transcript.

That's a pre-existing gap rather than something this PR introduces, but the PR is what adds the mechanism to fix it, and the two handlers now sitting side by side with different guarantees invites the reader to assume both are covered. Applying the same token check here would be a couple of lines.

packages/ballerina-visualizer/src/views/AgentChatPanel/Components/ApprovalCard.tsx
```diff
@@ -0,0 +1,389 @@
...
+}
+
+export const ApprovalCard: React.FC<ApprovalCardProps> = ({ requests, decisions, onSubmit }) => {
+    const [expandedArgs, setExpandedArgs] = useState<Record<string, boolean>>({});
+    const [rejectingId, setRejectingId] = useState<string | null>(null);
+    const [reasonText, setReasonText] = useState("");
+    const [submitting, setSubmitting] = useState(false);
+
+    const pendingRequests = requests.filter(r => !decisions?.[r.id]);
+    const isFullyResolved = pendingRequests.length === 0;
```
NipunaRanasinghe: `pendingRequests.length === 0` is also true when `requests` is empty, which takes the collapsed-summary branch and renders an empty bordered card with nothing in it.

Shouldn't happen if the service never sends an empty batch, but it costs one condition to be sure:

```ts
const isFullyResolved = requests.length > 0 && pendingRequests.length === 0;
```

and an early `if (!requests.length) return null;` would cover it even more directly.

lochana-chathura: Added extra guard as suggested

packages/ballerina-visualizer/src/views/AgentChatPanel/Components/ChatInterface.tsx
lochana-chathura: Added the same session token check here that handleApprovalDecision already had, so a late reply after switching agents gets dropped instead of appended to the wrong session.

packages/ballerina-visualizer/src/views/AgentChatPanel/Components/ChatInterface.tsx
lochana-chathura: Switched to matching the approval message by request id instead of array index

packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-manager.ts
lochana-chathura: Added a comment explaining why one controller is fine today

packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-manager.ts
lochana-chathura: Updated the catch block to match `getChatMessage`: it now records an error entry in history and rejects with the same `{message, traceInfo}` envelope so `parseAgentError` behaves consistently.

packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-manager.ts
lochana-chathura: Added a check that throws if `chatEp` doesn't end in /chat, so a shape change fails loudly instead of quietly posting to the wrong endpoint.

packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-manager.ts
lochana-chathura: Moved `resolvePendingApprovalInHistory` into the two success branches instead of calling it right after the fetch. Now history only gets updated once we know the response shape is valid.

packages/ballerina-extension/src/rpc-managers/agent-chat/rpc-manager.ts
lochana-chathura: Fixed by guarding the input match on a non-empty userMessage in findTraceForMessage, so an empty string can no longer match everything.

packages/ballerina-visualizer/src/views/AgentChatPanel/Components/ChatInterface.tsx
lochana-chathura: Added a "Dismiss" link on the approval card that appears once a decision submission has failed at least once. Clicking it marks that card resolved locally so the chat input unblocks, without the user having to Clear Chat. The card otherwise stays fully interactive after a failure so retries (e.g. approving one item, then resolving the rest in a second round) keep working normally.

packages/ballerina-visualizer/src/views/AgentChatPanel/Components/ChatInterface.tsx
```diff
@@ -750,6 +811,79 @@ const ChatInterface: React.FC = () => {
...
+    const handleDismissApproval = (requestIds: string[]) => {
+        setMessages((prev) => {
+            const updated = [...prev];
+            const targetIndex = findApprovalMessageIndex(updated, requestIds);
+            if (targetIndex !== -1) {
+                updated[targetIndex] = { ...updated[targetIndex], unresolvable: true };
+            }
+            return updated;
+        });
+    };
```
NipunaRanasinghe, line 885: The Dismiss affordance is a good answer to the dead-end, but it only holds until the next history replay, which is the case that motivated it.

`unresolvable` is set here in webview state only — nothing tells the extension, so it's never written into `AgentChatRpcManager.chatHistoryMap`. Switch agents and back (or anything else that re-runs `getChatHistory`/`switchChatAgent`) and `toChatMessage` reads `unresolvable: undefined`, the card renders unresolved again, `isApprovalPending` goes true, and the input is disabled again.

Getting the Dismiss link back then needs another failed submission, because `hasFailed` is also component-local and resets when `ApprovalCard` remounts. So the escape hatch is there, but it doesn't stick — and a stale approval that can never be resumed is exactly the scenario where the user is most likely to switch away and come back.

You've already plumbed `unresolvable` through `ChatHistoryMessage` and `toChatMessage`, so the read side is done; it's the write side that's missing. A small RPC to mark the batch dismissed in the extension's history would close it. If keeping it session-only is deliberate, then the field on `ChatHistoryMessage` is misleading as written — worth saying in the comment that it never round-trips.

NipunaRanasinghe (approved):

dan-niles (approved):
LGTM

PR #847, opened 2026-08-12
Land on the newly created integration

### LINKED ISSUE #2088: New integration overview does not stay open when added to an existing project
### Description

When adding a new integration to a project that already has other integrations, the editor briefly opens the new integration's own overview, then jumps back to the project overview a moment later. Creating a new integration in a brand-new project doesn't have this problem — it opens on the integration's overview and stays there.

### Steps to Reproduce

1. Open an existing Ballerina/Integrator project that already contains at least one integration.
2. Add a new integration (e.g. "Add Integration" from the project overview, or through the integration creation wizard).
3. Watch the view right after the integration is created.

https://github.com/user-attachments/assets/584b8d58-2f67-4ef4-b425-fbd51877b73c

### Version

5.1.0-Alpha

gigara:
Creating an integration does not reliably end up on the integration that was just created.

Fixes: https://github.com/wso2/product-integrator/issues/2088

- **From the project creation form**, the window lands on the project overview. The new integration is one click away, on a page the user did not ask for.
- **From the Add button on the project overview**, it lands on the new integration correctly — so the same action behaves differently depending on where it was started.

There is also an intermittent version of the first symptom: the new integration appears and is then replaced by the project overview a second or two later.

Creating an integration ends on that integration, from either entry point, every time.

Three separate defects sat behind one symptom. They are separate commits.

### 1. The landing was split by design

`resolveCreateLandingContext` decided where to go:

```ts
landing: isNewProject ? "project" : "package"
```

A submit that also created the project landed on the project, on the reasoning that the project was the thing the user had just made. That is the primary bug, and it explains the difference between the two entry points: Add from the project overview goes through the in-place path, where the project already exists, so `isNewProject` is false and it lands on the package.

Every create now lands on the new integration. That leaves the choice with one outcome, so the machinery carrying it is gone: the `landing` field on the pending pointer, `PendingIntegrationLanding`, `resolveLandsOnPackage`, and the `landOnPackageOverview` parameter threaded through generation. `CreateLandingContext` keeps only the naming it also carried and is renamed `CreateNamingContext` to match. Pointers written by an older build simply lose a field that no longer has a reader, so a create spanning the upgrade still lands correctly.

### 2. The landing stood down whenever anything else had navigated

`ensureLandedOnNewIntegration` acted only while the machine was still in `extensionReady`:

```ts
if (StateMachine.service().getSnapshot().value !== "extensionReady") return;
```

`checkAndRunPendingArtifact` fires the moment the machine reaches `extensionReady` and then awaits reading the payload and generating the artifact. Other startup navigation lands during those awaits, so by the time the create tried to land it saw `viewReady` and gave up.

The guard existed to avoid overriding generation that had navigated somewhere deliberate. That is now stated directly rather than inferred from machine state: `generatePendingArtifact` returns whether it claimed the view, which only the AI Chat Agent does, since it hands off to its own wizard. Everything else lands unconditionally.

### 3. Startup navigations arrive behind the landing

Even landing last is not enough, because the create is not the only thing navigating. Instrumenting `openView` across two runs of the same action showed three navigations in different orders:

```
worked   (bare) -> openPackageOverview -> Workspace Overview   [redirected]
failed   openPackageOverview -> (bare) -> Workspace Overview   [not redirected]
```

The bare navigation is startup asking for the visualizer with nothing specific; the workspace overview follows it. The create's position among them shifts because it waits on an artifact-generation RPC — which is why this was intermittent rather than consistently broken.

So the landing is claimed as well as performed. `claimCreateLanding` records the package, and a workspace-overview navigation arriving afterwards is redirected to it. The rules that make it safe:

- **Only the post-reload path claims.** The in-place path — Add from the project overview — runs in a settled window with no startup navigation racing it, and the one navigation that can follow it, the untracked-package fallback in `updateProjectArtifacts`, is already covered there by that function's own `alreadyViewingAddedPackage` check. Claiming there would plant a claim nothing ever spends, and the project explorer's Show Overview would then walk into it: that handler passes no `exactView`, so it would be redirected back to the package just left and the click would appear dead. Caught in review on the backport.
- **Only a workspace-overview navigation spends the claim.** An earlier version spent it on the first navigation of any kind, which let the bare one consume a claim meant for the workspace overview — that is the failing run above.
- **A bare navigation passes through untouched.** It resolves against the context, which the landing has already pointed at the new package, so nothing is lost by ignoring it.
- **A navigation naming any other view releases the claim.** That is the user moving on, and the claim must not survive to redirect a workspace overview they ask for later.
- **Home passes `exactView` and releases the claim**, so the deliberate route to the workspace overview always works.

The claim is bounded by the navigation it answers rather than by a duration. An earlier revision used a timeout; that was wrong about who owns the competing navigation — the project explorer only *invokes* `BI.project-explorer.overview`, a command registered in this repo, so there was nothing to out-wait.

## UI Component Development
- [x] Added reusable UI components to the ui-toolkit — N/A, no new UI. This changes which existing view a create ends on.

As a developer creating an integration, I want to land on the integration I just created and stay there, whichever entry point I started from.

Creating an integration now opens the new integration instead of the project overview, including when the same action creates the project. Fixes a related issue where the new integration would briefly appear and then be replaced by the project overview.

N/A — no documented behaviour changes. The screens and navigation controls are unchanged; only which view a create ends on differs.

## Training
N/A

## Certification
N/A — no user-facing concept or workflow is added or removed.

## Marketing
N/A

## Automation tests
 - Unit tests
   > `singleIntegrationLanding.test.ts` covers `resolveCreateLandingOverride`: the redirect itself, that a bare navigation survives it (the intermittency above, pinned as a regression case), that it is spent by the navigation it answers and not held for the next, and that a navigation elsewhere releases it. Full host suite: 120 passing.
   > Not covered: the landing itself. Reproducing it needs an awaited rebuild in a notification handler racing a navigation in the create flow, against a real state machine — the host suite mocks `vscode` and has no extension host, so this would need scaffolding well beyond the fix. Verified by hand instead, over repeated runs, since a single passing run proves nothing about an intermittent fault.
 - Integration tests
   > None added. Labelled `Checks/Run Ballerina UI Tests` so the existing UI suite runs against it. Worth e2e coverage — create from the form, assert the integration overview is still showing once the structure settles — which would need a wait on the post-create rebuild rather than a fixed delay.

## Security checks

## Samples
N/A

## Related PRs
#810 (merged) landed the single-integration rule in the same `openView` seam. The two overrides are independent and ordered there: the create's claim is consulted first, then the single-integration rule.

## Migrations (if applicable)
N/A — no persisted state beyond the pending-create pointer, which loses a field rather than gaining one. A create started on an older build and finished on this one still lands correctly.

macOS 15 (darwin 25.5.0), VS Code, Node 20. Verified by hand across both entry points and repeated runs of the create-form path, which is the one that was intermittent.

The fault was intermittent, and three successive fixes were reasoned from what could navigate rather than from what did. Each addressed a real defect and none resolved the symptom, because the remaining cause was an ordering that only shows up about half the time — a passing run looked like a fix.

Instrumenting `openView` to log every navigation with its caller settled it in one round trip, and should have been the first step rather than the fourth. The specific thing it revealed, which no amount of reading would have: the competing navigations are two separate startup calls, and the create's position between them moves because it waits on an RPC.

Worth recording for whoever touches this next: `goBack` does not route through `openView` at all — it replays history via `updateView` — so navigation rules added there cannot affect Back. That fact is what identified a stale `projectPath` write in the `webViewLoaded` assign, rather than the redirect, as the cause of a regression that broke Home and Back together.

## Summary

- Ensure integration creation consistently opens the new integration overview.
- Add navigation-based landing claims to prevent workspace navigation from replacing the new integration view.
- Remove obsolete landing state and simplify naming-context resolution.
- Update pending integration generation and navigation flows to use the new claim model.
- Add unit tests for landing overrides, claim consumption, preservation, and release.

### DIFF
--- a/packages/ballerina-extension/src/features/bi/pending-artifact.ts
+++ b/packages/ballerina-extension/src/features/bi/pending-artifact.ts
@@ -43,34 +43,6 @@ import {
 /** Payload file location inside the scaffolded project (target/ is gitignored by the scaffold). */
 const PENDING_ARTIFACT_RELATIVE_PATH = path.join("target", ".wizard-pending-artifact.json");
 
-let finishingIntegrationCreate = false;
-
-/**
- * Whether a Create Integration submit is being finished right now.
- *
- * Scaffolding a package makes the language server publish artifacts for a package the project
- * structure does not know yet, and `updateProjectArtifacts` answers that by navigating to the
- * workspace overview. During a create that is wrong: the create decides where to land, and the
- * fallback would replace the new integration's overview a moment after it opened.
- */
-export function isFinishingIntegrationCreate(): boolean {
-    return finishingIntegrationCreate;
-}
-
-/**
- * Runs `task` with {@link isFinishingIntegrationCreate} reporting true. Not re-entrant, and
- * does not need to be: the post-reload path runs once per activation, and the in-place path is
- * driven by the wizard webview, which is not open then.
- */
-async function whileFinishingIntegrationCreate<T>(task: () => PromiseLike<T>): Promise<T> {
-    finishingIntegrationCreate = true;
-    try {
-        return await task();
-    } finally {
-        finishingIntegrationCreate = false;
-    }
-}
-
 /** Human-readable labels for progress and error messages, per artifact kind. */
 const ARTIFACT_KIND_LABELS = INTEGRATION_ARTIFACT_LABELS;
 
@@ -128,10 +100,6 @@ export async function schedulePendingIntegration(schedule: PendingIntegrationSch
  * activation; never throws. No progress toast: the startup screen already narrates the wait.
  */
 export async function checkAndRunPendingArtifact(): Promise<void> {
-    return whileFinishingIntegrationCreate(runPendingArtifact);
-}
-
-async function runPendingArtifact(): Promise<void> {
     try {
         const stored = readPendingIntegrationPointer();
         if (!stored) {
@@ -247,10 +215,10 @@ export async function generateArtifactInPlace(
     }
 
     try {
-        const claimedView = await whileFinishingIntegrationCreate(() => window.withProgress(
+        const claimedView = await window.withProgress(
             { location: ProgressLocation.Notification, title: `Generating your ${label}...` },
             () => generatePendingArtifact(payload, packageRoot)
-        ));
+        );
         if (!claimedView) {
             openPackageOverview(packageRoot);
         }
--- a/packages/ballerina-extension/src/features/bi/pending-artifact.ts
+++ b/packages/ballerina-extension/src/features/bi/pending-artifact.ts
@@ -43,6 +43,34 @@ import {
 /** Payload file location inside the scaffolded project (target/ is gitignored by the scaffold). */
 const PENDING_ARTIFACT_RELATIVE_PATH = path.join("target", ".wizard-pending-artifact.json");
 
+/**
+ * Depth rather than a boolean: the in-place and post-reload paths can overlap, and the first
+ * one to finish must not clear the flag out from under the other.
+ */
+let integrationCreateDepth = 0;
+
+/**
+ * Whether a Create Integration submit is being finished right now.
+ *
+ * Scaffolding a package makes the language server publish artifacts for a package the project
+ * structure does not know yet, and `updateProjectArtifacts` answers that by navigating to the
+ * workspace overview. During a create that is wrong: the create flow decides where to land, and
+ * the fallback would replace the new integration's overview a moment after it opened.
+ */
+export function isFinishingIntegrationCreate(): boolean {
+    return integrationCreateDepth > 0;
+}
+
+/** Runs `task` with {@link isFinishingIntegrationCreate} reporting true. */
+async function whileFinishingIntegrationCreate<T>(task: () => PromiseLike<T>): Promise<T> {
+    integrationCreateDepth++;
+    try {
+        return await task();
+    } finally {
+        integrationCreateDepth--;
+    }
+}
+
 /** Human-readable labels for progress and error messages, per artifact kind. */
 const ARTIFACT_KIND_LABELS = INTEGRATION_ARTIFACT_LABELS;
 
@@ -104,6 +132,10 @@ export async function schedulePendingIntegration(schedule: PendingIntegrationSch
  * activation; never throws. No progress toast: the startup screen already narrates the wait.
  */
 export async function checkAndRunPendingArtifact(): Promise<void> {
+    return whileFinishingIntegrationCreate(runPendingArtifact);
+}
+
+async function runPendingArtifact(): Promise<void> {
     try {
         const stored = readPendingIntegrationPointer();
         if (!stored) {
@@ -252,10 +284,10 @@ export async function generateArtifactInPlace(
     }
 
     try {
-        await window.withProgress(
+        await whileFinishingIntegrationCreate(() => window.withProgress(
             { location: ProgressLocation.Notification, title: `Generating your ${label}...` },
             () => generatePendingArtifact(payload, packageRoot, landOnPackageOverview)
-        );
+        ));
         // A non-silent refresh lands on the workspace overview, which would clobber the
         // package overview navigated to above.
         StateMachine.refreshProjectInfo({ silent: landOnPackageOverview });
--- a/packages/ballerina-extension/src/features/bi/pending-artifact.ts
+++ b/packages/ballerina-extension/src/features/bi/pending-artifact.ts
@@ -165,14 +165,14 @@ async function runPendingArtifact(): Promise<void> {
         // An empty integration has no payload: there is nothing to generate, only
         // the landing view below to open.
         if (!payload) {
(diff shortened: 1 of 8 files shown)


packages/ballerina-extension/src/utils/state-machine-utils.ts
```diff
@@ -28,6 +28,54 @@ import { getConstructBodyString } from "./history/util";
...
+    }
+    if (viewLocation.view === MACHINE_VIEW.WorkspaceOverview) {
+        const claimed = createLanding;
+        createLanding = undefined;
+        return { view: MACHINE_VIEW.PackageOverview, projectPath: claimed };
+    }
+    if (viewLocation.view && viewLocation.view !== MACHINE_VIEW.PackageOverview) {
+        createLanding = undefined;
+    }
+    return undefined;
```
NipunaRanasinghe: Blocking, and it covers three things at once: the claim is consumed by `openView()` being *called* rather than by the navigation actually happening, and it has no lifetime bound.

**Consumed on a dropped event.** `OPEN_VIEW` is handled in exactly two states — `extensionReady` (stateMachine.ts:341) and `viewActive.viewReady` (stateMachine.ts:451). There is no root-level `on`, so an `OPEN_VIEW` sent during `viewInit`, `resolveMissingDependencies`, `webViewLoading` or `webViewLoaded` is dropped. This function clears `createLanding` and returns the override *before* `stateService.send` runs in `openView`, so a workspace-overview navigation arriving while a webview is mid-load burns the claim and produces no navigation. The project explorer's second overview call — the one the comment on `resolveSingleIntegrationOverride` describes as arriving "seconds after startup, once its tree finishes loading" — then finds no claim and lands on the workspace overview. Original symptom, with the safety net already spent.

**Never consumed.** Nothing guarantees a workspace-overview navigation arrives at all. A standalone/single-package create resolves Show Overview to `{ view: PackageOverview }` (features/bi/activator.ts:156), which the exemption on line 73 lets through, so the claim just sits. Same for any ordering where the workspace overview lands ahead of the create — the PR's evidence is that this ordering varies run to run, and two traces aren't a guarantee it is always behind. A surviving claim then hijacks the next workspace-overview navigation: the untracked-package fallback in `updateProjectArtifacts` (utils/project-artifacts.ts:309), or Show Overview in a workspace. That is the dead-click bug the second commit fixed for the in-place path — narrowing *where* a claim is planted did not bound how long one lives.

**The `PackageOverview` exemption is wider than its test.** The test case is `{ view: PackageOverview, projectPath: CREATED }`, but line 73 exempts any package overview, including a different package — reachable from `configGenerator.ts:77`/`:100`, `tryit/activator.ts:627`/`:676`, `project/cmds/doc.ts:92`, `bi-diagram/rpc-manager.ts:858`. Create → land → open another package → Show Overview redirects back to the created one.

One change addresses all three: don't consume on first match. Hold the claim and redirect every workspace-overview navigation, releasing it on explicit signals only — `exactView`/Home as now, a navigation naming another view as now, and one positive signal that the landing actually happened (the machine reaching `viewReady` on the created package) instead of the first `openView` call that happens to match. That is what makes `resolveSingleIntegrationOverride` robust: it is a pure function re-applied to every navigation, so a dropped event costs it nothing. Worth having the same property here, given the whole PR is about an ordering nobody controls.

`it("is spent by the navigation it answers, not held for the next one")` would invert under that change.

packages/ballerina-extension/src/features/bi/pending-artifact.ts
```diff
@@ -89,14 +86,30 @@ export async function schedulePendingIntegration(schedule: PendingIntegrationSch
...
+ *
+ * The in-place path deliberately uses {@link openPackageOverview} instead. It runs in a settled
+ * window with no startup navigation to race, and the one navigation that can follow it — the
+ * untracked-package fallback in `updateProjectArtifacts` — is already covered there by that
+ * function's own `alreadyViewingAddedPackage` check. A claim planted there would be one nothing
+ * ever spends, and the project explorer's Show Overview would walk into it.
+ */
+function landOnNewIntegrationAfterReload(projectRoot: string): void {
+    openPackageOverview(projectRoot);
+    claimCreateLanding(projectRoot);
```
NipunaRanasinghe: Follows from the `resolveCreateLandingOverride` thread. `openPackageOverview` is fire-and-forget into a machine that handles `OPEN_VIEW` only in `extensionReady` and `viewActive.viewReady`, and the claim is planted whether or not the navigation was delivered. So "claimed but never landed" is a reachable state, and if no workspace-overview navigation follows to spend the claim, the window stays wherever startup left it — silently.

Worth noting that the deleted `getSnapshot().value !== "extensionReady"` check was doing two jobs and only one has been replaced. `claimedView` covers "generation navigated somewhere deliberate". The other job was guaranteeing the landing event was deliverable at all: sending it while the machine was in `extensionReady` meant it would be handled. Nothing covers that now.

The claim compensates by accident today — if the landing is dropped but a workspace overview arrives behind it, the redirect performs the landing the dropped event didn't. Making the claim persistent rather than single-shot (other thread) turns that accident into the mechanism.

PR #851, opened 2026-08-12
Land on the newly created integration (5.14.0)

gigara:
**Backport of #847 to `staging/5.14.0`.** Cherry-picked and squashed; the resulting diff is byte-identical to the one reviewed on `main`. `staging/5.14.0` already carries #810, so the `openView` seam this builds on is present and no adaptation was needed.

Creating an integration does not reliably end up on the integration that was just created.

- **From the project creation form**, the window lands on the project overview. The new integration is one click away, on a page the user did not ask for.
- **From the Add button on the project overview**, it lands on the new integration correctly — so the same action behaves differently depending on where it was started.

There is also an intermittent version of the first symptom: the new integration appears and is then replaced by the project overview a second or two later.

Creating an integration ends on that integration, from either entry point, every time.

Three separate defects sat behind one symptom. They are separate commits.

### 1. The landing was split by design

`resolveCreateLandingContext` decided where to go:

```ts
landing: isNewProject ? "project" : "package"
```

A submit that also created the project landed on the project, on the reasoning that the project was the thing the user had just made. That is the primary bug, and it explains the difference between the two entry points: Add from the project overview goes through the in-place path, where the project already exists, so `isNewProject` is false and it lands on the package.

Every create now lands on the new integration. That leaves the choice with one outcome, so the machinery carrying it is gone: the `landing` field on the pending pointer, `PendingIntegrationLanding`, `resolveLandsOnPackage`, and the `landOnPackageOverview` parameter threaded through generation. `CreateLandingContext` keeps only the naming it also carried and is renamed `CreateNamingContext` to match. Pointers written by an older build simply lose a field that no longer has a reader, so a create spanning the upgrade still lands correctly.

### 2. The landing stood down whenever anything else had navigated

`ensureLandedOnNewIntegration` acted only while the machine was still in `extensionReady`:

```ts
if (StateMachine.service().getSnapshot().value !== "extensionReady") return;
```

`checkAndRunPendingArtifact` fires the moment the machine reaches `extensionReady` and then awaits reading the payload and generating the artifact. Other startup navigation lands during those awaits, so by the time the create tried to land it saw `viewReady` and gave up.

The guard existed to avoid overriding generation that had navigated somewhere deliberate. That is now stated directly rather than inferred from machine state: `generatePendingArtifact` returns whether it claimed the view, which only the AI Chat Agent does, since it hands off to its own wizard. Everything else lands unconditionally.

### 3. Startup navigations arrive behind the landing

Even landing last is not enough, because the create is not the only thing navigating. Instrumenting `openView` across two runs of the same action showed three navigations in different orders:

```
worked   (bare) -> openPackageOverview -> Workspace Overview   [redirected]
failed   openPackageOverview -> (bare) -> Workspace Overview   [not redirected]
```

The bare navigation is startup asking for the visualizer with nothing specific; the workspace overview follows it. The create's position among them shifts because it waits on an artifact-generation RPC — which is why this was intermittent rather than consistently broken.

So the landing is claimed as well as performed. `claimCreateLanding` records the package, and a workspace-overview navigation arriving afterwards is redirected to it. The rules that make it safe:

- **Only the post-reload path claims.** The in-place path — Add from the project overview — runs in a settled window with no startup navigation racing it, and the one navigation that can follow it, the untracked-package fallback in `updateProjectArtifacts`, is already covered there by that function's own `alreadyViewingAddedPackage` check. Claiming there would plant a claim nothing ever spends, and the project explorer's Show Overview would then walk into it: that handler passes no `exactView`, so it would be redirected back to the package just left and the click would appear dead. Caught in review on the backport.
- **Only a workspace-overview navigation spends the claim.** An earlier version spent it on the first navigation of any kind, which let the bare one consume a claim meant for the workspace overview — that is the failing run above.
- **A bare navigation passes through untouched.** It resolves against the context, which the landing has already pointed at the new package, so nothing is lost by ignoring it.
- **A navigation naming any other view releases the claim.** That is the user moving on, and the claim must not survive to redirect a workspace overview they ask for later.
- **Home passes `exactView` and releases the claim**, so the deliberate route to the workspace overview always works.

The claim is bounded by the navigation it answers rather than by a duration. An earlier revision used a timeout; that was wrong about who owns the competing navigation — the project explorer only *invokes* `BI.project-explorer.overview`, a command registered in this repo, so there was nothing to out-wait.

**Known edge, accepted rather than solved.** "Bounded by the navigation it answers" assumes such a navigation arrives, and nothing guarantees one does — a standalone create resolves Show Overview to `PackageOverview` instead, and the re-assert after generation can plant a claim once startup's navigations have already been answered. A claim with nothing to spend it survives until the user navigates to another view, so if their very next action is Show Overview in a workspace, it is redirected to the package they just came from and the click looks dead; a second click works.

Two alternatives were tried and are worse. Releasing the claim once the landing renders fires *before* the competing navigation — the traces above put it 1.4s behind — so the original symptom returns. Holding the claim instead of spending it leaves Show Overview redirected on *every* click until the user goes elsewhere, which is worse than one. The competing navigation and a deliberate click are the same command with nothing to tell them apart, so any bound is a trade; this is the one chosen, with the requester.

## UI Component Development
- [x] Added reusable UI components to the ui-toolkit — N/A, no new UI. This changes which existing view a create ends on.

As a developer creating an integration, I want to land on the integration I just created and stay there, whichever entry point I started from.

Creating an integration now opens the new integration instead of the project overview, including when the same action creates the project. Fixes a related issue where the new integration would briefly appear and then be replaced by the project overview.

N/A — no documented behaviour changes. The screens and navigation controls are unchanged; only which view a create ends on differs.

## Training
N/A

## Certification
N/A — no user-facing concept or workflow is added or removed.

## Marketing
N/A

## Automation tests
 - Unit tests
   > `singleIntegrationLanding.test.ts` covers `resolveCreateLandingOverride`: the redirect itself, that a bare navigation survives it (the intermittency above, pinned as a regression case), that it is spent by the navigation it answers and not held for the next, and that a navigation elsewhere releases it. Full host suite: 120 passing.
   > Not covered: the landing itself. Reproducing it needs an awaited rebuild in a notification handler racing a navigation in the create flow, against a real state machine — the host suite mocks `vscode` and has no extension host, so this would need scaffolding well beyond the fix. Verified by hand instead, over repeated runs, since a single passing run proves nothing about an intermittent fault.
 - Integration tests
   > None added. Labelled `Checks/Run Ballerina UI Tests` so the existing UI suite runs against it. Worth e2e coverage — create from the form, assert the integration overview is still showing once the structure settles — which would need a wait on the post-create rebuild rather than a fixed delay.

## Security checks

## Samples
N/A

## Related PRs
#847 — the same change against `main`. This is its backport; review there.

#810 (merged, and already on this branch) landed the single-integration rule in the same `openView` seam. The two overrides are independent and ordered there: the create's claim is consulted first, then the single-integration rule.

## Migrations (if applicable)
N/A — no persisted state beyond the pending-create pointer, which loses a field rather than gaining one. A create started on an older build and finished on this one still lands correctly.

macOS 15 (darwin 25.5.0), VS Code, Node 20. Verified by hand on the `main` version across both entry points and repeated runs of the create-form path, which is the one that was intermittent. On this branch: cherry-pick applied without conflict, diff verified identical to the reviewed change, typecheck and the 120 host unit tests pass.

The fault was intermittent, and three successive fixes were reasoned from what could navigate rather than from what did. Each addressed a real defect and none resolved the symptom, because the remaining cause was an ordering that only shows up about half the time — a passing run looked like a fix.

Instrumenting `openView` to log every navigation with its caller settled it in one round trip, and should have been the first step rather than the fourth. The specific thing it revealed, which no amount of reading would have: the competing navigations are two separate startup calls, and the create's position between them moves because it waits on an RPC.

Worth recording for whoever touches this next: `goBack` does not route through `openView` at all — it replays history via `updateView` — so navigation rules added there cannot affect Back. That fact is what identified a stale `projectPath` write in the `webViewLoaded` assign, rather than the redirect, as the cause of a regression that broke Home and Back together.

## Summary

- Open newly created integrations in the created package overview.
- Apply bounded create-landing claims during navigation.
- Preserve claims until deliverable workspace-overview navigation occurs.
- Exclude AI Chat Agent and self-navigating integrations from landing redirects.
- Remove obsolete landing metadata from scheduling and startup state.
- Update artifact generation and project naming metadata handling.
- Add unit tests for redirect, claim, release, and consumption behavior.
- Host tests pass: 120 tests. UI tests are configured to run.

### DIFF
--- a/packages/ballerina-extension/src/features/bi/pending-artifact.ts
+++ b/packages/ballerina-extension/src/features/bi/pending-artifact.ts
@@ -28,7 +28,7 @@ import {
     MACHINE_VIEW,
     PendingIntegrationArtifactPayload,
 } from "@wso2/ballerina-core";
-import { openView, StateMachine } from "../../stateMachine";
+import { openView, StateMachine, whenNavigationDeliverable } from "../../stateMachine";
 import { claimCreateLanding } from "../../utils/state-machine-utils";
 import { ServiceDesignerRpcManager } from "../../rpc-managers/service-designer/rpc-manager";
 import { BiDiagramRpcManager } from "../../rpc-managers/bi-diagram/rpc-manager";
@@ -108,7 +108,14 @@ export async function schedulePendingIntegration(schedule: PendingIntegrationSch
  * function's own `alreadyViewingAddedPackage` check. A claim planted there would be one nothing
  * ever spends, and the project explorer's Show Overview would walk into it.
  */
-function landOnNewIntegrationAfterReload(projectRoot: string): void {
+async function landOnNewIntegrationAfterReload(projectRoot: string): Promise<void> {
+    // Awaited because the machine handles `OPEN_VIEW` only in `extensionReady` and
+    // `viewActive.viewReady`. Generation is asynchronous, so by the time it returns the machine may
+    // be mid-load, where the landing would be dropped without trace and the window would silently
+    // stay wherever startup left it. The deleted `extensionReady` guard used to make delivery
+    // certain as a side effect of when it allowed the landing at all; this replaces that half of
+    // its job, `claimedView` having replaced the other.
+    await whenNavigationDeliverable();
     openPackageOverview(projectRoot);
     claimCreateLanding(projectRoot);
 }
@@ -154,14 +161,14 @@ export async function checkAndRunPendingArtifact(): Promise<void> {
         // An empty integration has no payload: there is nothing to generate, only
         // the landing view below to open.
         if (!payload) {
-            landOnNewIntegrationAfterReload(stored.projectRoot);
+            await landOnNewIntegrationAfterReload(stored.projectRoot);
             return;
         }
 
         const label = ARTIFACT_KIND_LABELS[payload.kind];
         if (!label || payload.version !== 1) {
             console.error(`[IntegrationWizard] Unsupported pending artifact payload:`, payload);
-            landOnNewIntegrationAfterReload(stored.projectRoot);
+            await landOnNewIntegrationAfterReload(stored.projectRoot);
             return;
         }
 
@@ -187,7 +194,7 @@ export async function checkAndRunPendingArtifact(): Promise<void> {
         // because something had already navigated is what left the window on the workspace
         // overview — a create is the last word on where a create ends up.
         if (!claimedView) {
-            landOnNewIntegrationAfterReload(stored.projectRoot);
+            await landOnNewIntegrationAfterReload(stored.projectRoot);
         }
     } catch (error) {
         console.error("[IntegrationWizard] Unexpected error while checking pending artifact:", error);
--- a/packages/ballerina-extension/src/features/bi/pending-artifact.ts
+++ b/packages/ballerina-extension/src/features/bi/pending-artifact.ts
@@ -29,6 +29,7 @@ import {
     PendingIntegrationArtifactPayload,
 } from "@wso2/ballerina-core";
 import { openView, StateMachine } from "../../stateMachine";
+import { claimCreateLanding } from "../../utils/state-machine-utils";
 import { ServiceDesignerRpcManager } from "../../rpc-managers/service-designer/rpc-manager";
 import { BiDiagramRpcManager } from "../../rpc-managers/bi-diagram/rpc-manager";
 import { extension } from "../../BalExtensionContext";
@@ -38,7 +39,6 @@ import {
     clearPendingIntegrationPointer,
     isPendingPointerFresh,
     PendingIntegrationArtifactPointer,
-    PendingIntegrationLanding,
     readPendingIntegrationPointer,
     writePendingIntegrationPointer,
 } from "./startup-progress";
@@ -67,14 +67,12 @@ export interface PendingIntegrationSchedule {
     isNewProject?: boolean;
     /** Defaults to "integration" on read. */
     componentLabel?: IntegrationComponentLabel;
-    /** Where the reloaded window should land. */
-    landing: PendingIntegrationLanding;
 }
 
 /**
  * Records the create so the reloaded window can finish it. Written even for an empty
- * integration or a library — it is also what lets the new window narrate the create and
- * know where to land. Call right before `openInVSCode(openRoot)`.
+ * integration or a library — it is also what lets the new window narrate the create.
+ * Call right before `openInVSCode(openRoot)`.
  */
 export async function schedulePendingIntegration(schedule: PendingIntegrationSchedule): Promise<void> {
     const { packageRoot, payload } = schedule;
@@ -92,11 +90,9 @@ export async function schedulePendingIntegration(schedule: PendingIntegrationSch
         projectName: schedule.projectName,
         isNewProject: schedule.isNewProject,
         componentLabel: schedule.componentLabel,
-        landing: schedule.landing,
     });
     console.log(
-        `[IntegrationWizard] Scheduled pending ${payload?.kind ?? "empty"} integration for project: ${packageRoot} ` +
-        `(landing=${schedule.landing})`
+        `[IntegrationWizard] Scheduled pending ${payload?.kind ?? "empty"} integration for project: ${packageRoot}`
     );
 }
 
@@ -138,19 +134,17 @@ export async function checkAndRunPendingArtifact(): Promise<void> {
             return;
         }
 
-        const landOnPackageOverview = resolveLandsOnPackage(stored, opensStoredPackage);
-
         // An empty integration has no payload: there is nothing to generate, only
         // the landing view below to open.
         if (!payload) {
-            ensureLandedOnNewIntegration(stored, landOnPackageOverview);
+            openPackageOverview(stored.projectRoot);
             return;
         }
 
         const label = ARTIFACT_KIND_LABELS[payload.kind];
         if (!label || payload.version !== 1) {
             console.error(`[IntegrationWizard] Unsupported pending artifact payload:`, payload);
-            ensureLandedOnNewIntegration(stored, landOnPackageOverview);
+            openPackageOverview(stored.projectRoot);
             return;
         }
 
@@ -158,10 +152,11 @@ export async function checkAndRunPendingArtifact(): Promise<void> {
         console.log(
             `[IntegrationWizard] Pending artifact: kind=${payload.kind}, projectRoot=${stored.projectRoot}, ` +
             `opensStoredPackage=${opensStoredPackage}, insideOpenWorkspace=${insideOpenWorkspace}, ` +
-            `addedIntoWorkspace=${addedIntoWorkspace}, landOnPackageOverview=${landOnPackageOverview}`
+            `addedIntoWorkspace=${addedIntoWorkspace}`
         );
+        let claimedView = false;
         try {
-            await generatePendingArtifact(payload, stored.projectRoot, landOnPackageOverview);
+            claimedView = await generatePendingArtifact(payload, stored.projectRoot);
         } catch (error) {
             const message = error instanceof Error ? error.message : String(error);
             console.error(`[IntegrationWizard] Failed to generate pending ${payload.kind} artifact:`, error);
@@ -170,47 +165,16 @@ export async function checkAndRunPendingArtifact(): Promise<void> {
                 `Your integration was created; you can add the artifact from the Artifacts panel.`
             );
         }
-        // Whatever generation did (navigated, didn't, or failed), never leave the window
-        // on the startup screen.
-        ensureLandedOnNewIntegration(stored, landOnPackageOverview);
+        // Unconditional, and it has to be: this runs after awaiting generation, and the
+        // project explorer's own startup navigation lands during that await. Standing down
+        // because something had already navigated is what left the window on the workspace
+        // overview — a create is the last word on where a create ends up.
+        if (!claimedView) {
+            openPackageOverview(stored.projectRoot);
+        }
     } catch (error) {
         console.error("[IntegrationWizard] Unexpected error while checking pending artifact:", error);
     }
-}
-
-/**
- * Whether this create lands on the new package's own overview rather than the project
- * overview. The submit already decided (`landing`); a pointer written by an older build
- * has no such field, so it falls back to the previous rule — package overview only when
- * the package itself is the opened folder.
- */
-function resolveLandsOnPackage(
-    pointer: PendingIntegrationArtifactPointer,
-    opensStoredPackage: boolean
-): boolean {
-    return pointer.landing ? pointer.landing === "package" : opensStoredPackage;
-}
-
-/**
- * Guarantees the window lands on a real view after a wizard create. Acts only when
- * nothing has navigated yet (machine still in `extensionReady`), so it stays a no-op on
- * paths that navigate themselves.
- */
-function ensureLandedOnNewIntegration(
-    pointer: PendingIntegrationArtifactPointer,
-    landOnPackageOverview: boolean
-): void {
-    // Read the raw machine value rather than `StateMachine.state()`: the shared
-    // `MachineStateValue` type predates the startup states and does not include
-    // `extensionReady`, which is exactly the one being tested here.
-    if (StateMachine.service().getSnapshot().value !== "extensionReady") {
-        return;
-    }
-    if (landOnPackageOverview) {
-        openPackageOverview(pointer.projectRoot);
-        return;
-    }
-    openView(EVENT_TYPE.OPEN_VIEW, { view: MACHINE_VIEW.WorkspaceOverview });
 }
 
 /** Reads and immediately deletes the payload file; undefined when missing (empty integration) or unreadable. */
@@ -245,8 +209,7 @@ function consumePendingArtifactPayload(projectRoot: string): PendingIntegrationA
  */
 export async function generateArtifactInPlace(
     packageRoot: string,
-    payload: PendingIntegrationArtifactPayload,
-    landOnPackageOverview = false
+    payload: PendingIntegrationArtifactPayload
 ): Promise<void> {
     const label = ARTIFACT_KIND_LABELS[payload.kind];
     if (!label || payload.version !== 1) {
@@ -255,13 +218,16 @@ export async function generateArtifactInPlace(
     }
 
     try {
-        await window.withProgress(
+        const claimedView = await window.withProgress(
             { location: ProgressLocation.Notification, title: `Generating your ${label}...` },
-            () => generatePendingArtifact(payload, packageRoot, landOnPackageOverview)
+            () => generatePendingArtifact(payload, packageRoot)
         );
-        // A non-silent refresh lands on the workspace overview, which would clobber the
-        // package overview navigated to above.
-        StateMachine.refreshProjectInfo({ silent: landOnPackageOverview });
+        if (!claimedView) {
+            openPackageOverview(packageRoot);
+        }
+        // Silent: a non-silent refresh lands on the workspace overview, which would clobber
+        // the package overview navigated to above.
+        StateMachine.refreshProjectInfo({ silent: true });
     } catch (error) {
         const message = error instanceof Error ? error.message : String(error);
         console.error(`[IntegrationWizard] Failed to generate ${payload.kind} artifact in place:`, error);
@@ -273,16 +239,18 @@ export async function generateArtifactInPlace(
 }
 
 /**
- * Runs the kind-specific generation and navigates to the result. All files target
- * `projectRoot` (the new package). `landOnPackageOverview`: true when the new package is
- * the thing to show (standalone, or added into an existing project); false when the
- * project itself was just created, so the window stays on the project overview.
+ * Runs the kind-specific generation. All files target `projectRoot` (the new package).
+ *
+ * Returns whether generation claimed the view for a destination of its own, in which case the
+ * caller leaves it alone. Only the agent does that — it hands off to a wizard rather than
+ * finishing here. Everything else leaves the caller to land on the new integration, which is
+ * the thing the user came to see whether it went into a project that already existed or one
+ * this same submit created.
  */
 async function generatePendingArtifact(
     payload: PendingIntegrationArtifactPayload,
-    projectRoot: string,
-    landOnPackageOverview: boolean
-): Promise<void> {
+    projectRoot: string
+): Promise<boolean> {
     switch (payload.kind) {
         case "SERVICE": {
             if (!payload.serviceInitModel) {
@@ -295,10 +263,7 @@ async function generatePendingArtifact(
                 projectPath: projectRoot,
                 serviceInitModel: payload.serviceInitModel,
             });
-            if (landOnPackageOverview) {
-                openPackageOverview(projectRoot);
-            }
-            return;
+            return false;
         }
         case "AUTOMATION":
         case "WORKFLOW": {
@@ -315,10 +280,7 @@ async function generatePendingArtifact(
             if (payload.flowNode.codedata?.node === "DURABLE_AGENT") {
                 await configureDurableAgentModelProvider(projectRoot);
             }
-            if (landOnPackageOverview) {
-                openPackageOverview(projectRoot);
-            }
-            return;
+            return false;
         }
         case "AI_CHAT_AGENT": {
             // Pragmatic v1: the agent's multi-RPC orchestration stays webview-side —
@@ -328,7 +290,7 @@ async function generatePendingArtifact(
                 view: MACHINE_VIEW.AIChatAgentWizard,
                 identifier: payload.aiAgent?.name,
             });
-            return;
+            return true;
         }
         default:
             throw new Error(`Unsupported artifact kind: ${(payload as PendingIntegrationArtifactPayload).kind}`);
@@ -367,5 +329,10 @@ async function configureDurableAgentModelProvider(projectRoot: string): Promise<
  */
 export function openPackageOverview(projectRoot: string): void {
     openView(EVENT_TYPE.OPEN_VIEW, { view: MACHINE_VIEW.PackageOverview, projectPath: projectRoot });
+    // Claimed as well as navigated: the project explorer issues its own Open Overview once its
+    // tree finishes loading, which can be after this lands, and would otherwise replace the
+    // integration the user just made. Claimed AFTER the navigation above, so that navigation
+    // does not consume its own claim.
+    claimCreateLanding(projectRoot);
 }
 
--- a/packages/ballerina-extension/src/features/bi/pending-artifact.ts
+++ b/packages/ballerina-extension/src/features/bi/pending-artifact.ts
@@ -38,7 +38,6 @@ import { isAIAuthenticated } from "../ai/migration/orchestrator";
 import {
     clearPendingIntegrationPointer,
     isPendingPointerFresh,
-    PendingIntegrationArtifactPointer,
     readPendingIntegrationPointer,
     writePendingIntegrationPointer,
 } from "./startup-progress";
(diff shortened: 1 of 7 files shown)


packages/ballerina-extension/src/features/bi/pending-artifact.ts
```diff
@@ -335,5 +297,10 @@ async function generatePendingArtifact(
  */
 export function openPackageOverview(projectRoot: string): void {
     openView(EVENT_TYPE.OPEN_VIEW, { view: MACHINE_VIEW.PackageOverview, projectPath: projectRoot });
+    // Claimed as well as navigated: the project explorer issues its own Open Overview once its
+    // tree finishes loading, which can be after this lands, and would otherwise replace the
+    // integration the user just made. Claimed AFTER the navigation above, so that navigation
+    // does not consume its own claim.
+    claimCreateLanding(projectRoot);
```
NipunaRanasinghe: Claiming here is unconditional, but only the reload path has a navigation to spend the claim. On the in-place path nothing consumes it, and it survives until the user's next navigation — which it then swallows.

Tracing the in-place path:

- `createIntegration` → `generateArtifactInPlace` → `openPackageOverview` plants the claim, then `StateMachine.refreshProjectInfo({ silent: true })`. Silent, so the `if (!event.silent)` branch in `stateMachine.ts` never fires `openView(WorkspaceOverview)`.
- The no-artifact branch in `integration-wizard.ts` is the same shape: `refreshProjectInfoAndWait()` runs *before* the landing, so it cannot spend the claim either.
- `addIntegrationArtifact` — same.

The file-watcher refresh in `views/visualizer/webview.ts` is also silent. So after an in-place create the claim just sits there.

The failure: create via the Add button → land on the package overview → click **Show Overview** in the project explorer. That command handler in `features/bi/activator.ts` issues

```ts
openView(EVENT_TYPE.OPEN_VIEW, { view: MACHINE_VIEW.WorkspaceOverview });
```

with no `exactView`, so `resolveCreateLandingOverride` redirects it back to the package the user just left. The click does nothing. A second click works, because the first spent the claim.

The description's safety argument covers Home, and Home does pass `exactView` — but Home and this command handler are the only two deliberate routes to the workspace overview, and I could not find `exactView` anywhere else in the repo. The project explorer button is the uncovered one.

Worth flagging that this lands on the entry point the description calls out as already working correctly.

Two ways out:

1. Claim only where the race actually exists — move `claimCreateLanding` out of `openPackageOverview` and into `checkAndRunPendingArtifact`, or gate it behind a parameter. The in-place path has no competing navigation to defend against, so it does not need a claim at all.
2. Pass `{ exactView: true }` from the `SHOW_OVERVIEW` handler in `activator.ts`, on the same reasoning as Home.

The first is narrower and matches what the claim is documented to be for.

packages/ballerina-extension/src/features/bi/pending-artifact.ts
```diff
@@ -29,13 +29,13 @@ import {
     PendingIntegrationArtifactPayload,
 } from "@wso2/ballerina-core";
 import { openView, StateMachine } from "../../stateMachine";
+import { claimCreateLanding } from "../../utils/state-machine-utils";
 import { ServiceDesignerRpcManager } from "../../rpc-managers/service-designer/rpc-manager";
 import { BiDiagramRpcManager } from "../../rpc-managers/bi-diagram/rpc-manager";
 import {
     clearPendingIntegrationPointer,
     isPendingPointerFresh,
     PendingIntegrationArtifactPointer,
```
NipunaRanasinghe: `PendingIntegrationArtifactPointer` is now unused — this import is the only reference left in the file. Both functions that took it, `resolveLandsOnPackage` and `ensureLandedOnNewIntegration`, are deleted in this change. Typecheck passes, so `noUnusedLocals` is presumably off, but the import should go with them.

packages/ballerina-extension/src/features/bi/pending-artifact.ts
gigara: Confirmed, and taking option 1. Fixed in 12a7c7f202 here, and in ee004077ba on #847.

Your trace is right, and the consequence is worse than the edge I had noted in the description. I had it as "immediately after a create, one click is swallowed". In fact nothing on the in-place path ever spends the claim, so it is not time-limited at all — it survives until the next navigation naming another view, and Show Overview is the click it eats on the way.

`claimCreateLanding` has moved out of `openPackageOverview` into a `landOnNewIntegrationAfterReload` helper used only by the three post-reload call sites. The in-place path keeps `openPackageOverview` unchanged.

The reason that is safe is the same one used earlier in this PR to delete `isFinishingIntegrationCreate`, so it is consistent rather than a fresh assumption: the in-place path runs in a settled window with no startup navigation to race, and the one navigation that *can* follow it — the untracked-package fallback in `updateProjectArtifacts` — is already covered there by `alreadyViewingAddedPackage`. Of its two orderings, resolving before the landing is harmless because the landing follows, and resolving after hits that check.

On `exactView`: your reading of the repo is correct, `goHome` is the only caller. I preferred narrowing the claim over adding it to the `SHOW_OVERVIEW` handler, because that handler serves both the explorer's automatic startup navigation and a deliberate button click, with nothing to tell them apart — `exactView` there would have suppressed the redirect in the case the claim exists for.

Also fair on the last point: this landed on the entry point the description calls already-correct. The description has been rewritten since; I will make sure that framing is not left in it.

packages/ballerina-extension/src/features/bi/pending-artifact.ts
gigara: Right, removed in the same commit — 12a7c7f202 here, ee004077ba on #847.

`noUnusedLocals` is indeed off for this package, so nothing flagged it. Both functions that took the type are gone in this change, and the import was the only reference left in the file; the type itself is still used in `startup-progress.ts`, so only the import goes.

NipunaRanasinghe:
Verified this is a faithful backport: `gh pr diff 847` and `gh pr diff 851` are identical ignoring `index` lines. The two rationales the change leans on are both present on `staging/5.14.0` — `alreadyViewingAddedPackage` in `updateProjectArtifacts` and #810's `resolveSingleIntegrationOverride` — so no adaptation was needed, as the description says.

Reviewing on #847 to keep the threads in one place. Holding this one until that lands, since the fix will need cherry-picking again.

packages/ballerina-extension/src/utils/state-machine-utils.ts
```diff
@@ -28,6 +28,54 @@ import { getConstructBodyString } from "./history/util";
...
+    }
+    if (viewLocation.view === MACHINE_VIEW.WorkspaceOverview) {
+        const claimed = createLanding;
+        createLanding = undefined;
+        return { view: MACHINE_VIEW.PackageOverview, projectPath: claimed };
+    }
+    if (viewLocation.view && viewLocation.view !== MACHINE_VIEW.PackageOverview) {
+        createLanding = undefined;
+    }
+    return undefined;
```
NipunaRanasinghe: Blocking, and it covers three things at once: the claim is consumed by `openView()` being *called* rather than by the navigation actually happening, and it has no lifetime bound.

**Consumed on a dropped event.** `OPEN_VIEW` is handled in exactly two states — `extensionReady` (stateMachine.ts:341) and `viewActive.viewReady` (stateMachine.ts:451). There is no root-level `on`, so an `OPEN_VIEW` sent during `viewInit`, `resolveMissingDependencies`, `webViewLoading` or `webViewLoaded` is dropped. This function clears `createLanding` and returns the override *before* `stateService.send` runs in `openView`, so a workspace-overview navigation arriving while a webview is mid-load burns the claim and produces no navigation. The project explorer's second overview call — the one the comment on `resolveSingleIntegrationOverride` describes as arriving "seconds after startup, once its tree finishes loading" — then finds no claim and lands on the workspace overview. Original symptom, with the safety net already spent.

**Never consumed.** Nothing guarantees a workspace-overview navigation arrives at all. A standalone/single-package create resolves Show Overview to `{ view: PackageOverview }` (features/bi/activator.ts:156), which the exemption on line 73 lets through, so the claim just sits. Same for any ordering where the workspace overview lands ahead of the create — the PR's evidence is that this ordering varies run to run, and two traces aren't a guarantee it is always behind. A surviving claim then hijacks the next workspace-overview navigation: the untracked-package fallback in `updateProjectArtifacts` (utils/project-artifacts.ts:309), or Show Overview in a workspace. That is the dead-click bug the second commit fixed for the in-place path — narrowing *where* a claim is planted did not bound how long one lives.

**The `PackageOverview` exemption is wider than its test.** The test case is `{ view: PackageOverview, projectPath: CREATED }`, but line 73 exempts any package overview, including a different package — reachable from `configGenerator.ts:77`/`:100`, `tryit/activator.ts:627`/`:676`, `project/cmds/doc.ts:92`, `bi-diagram/rpc-manager.ts:858`. Create → land → open another package → Show Overview redirects back to the created one.

One change addresses all three: don't consume on first match. Hold the claim and redirect every workspace-overview navigation, releasing it on explicit signals only — `exactView`/Home as now, a navigation naming another view as now, and one positive signal that the landing actually happened (the machine reaching `viewReady` on the created package) instead of the first `openView` call that happens to match. That is what makes `resolveSingleIntegrationOverride` robust: it is a pure function re-applied to every navigation, so a dropped event costs it nothing. Worth having the same property here, given the whole PR is about an ordering nobody controls.

`it("is spent by the navigation it answers, not held for the next one")` would invert under that change.

packages/ballerina-extension/src/features/bi/pending-artifact.ts
```diff
@@ -89,14 +86,30 @@ export async function schedulePendingIntegration(schedule: PendingIntegrationSch
...
+ *
+ * The in-place path deliberately uses {@link openPackageOverview} instead. It runs in a settled
+ * window with no startup navigation to race, and the one navigation that can follow it — the
+ * untracked-package fallback in `updateProjectArtifacts` — is already covered there by that
+ * function's own `alreadyViewingAddedPackage` check. A claim planted there would be one nothing
+ * ever spends, and the project explorer's Show Overview would walk into it.
+ */
+function landOnNewIntegrationAfterReload(projectRoot: string): void {
+    openPackageOverview(projectRoot);
+    claimCreateLanding(projectRoot);
```
NipunaRanasinghe, line 113: Follows from the `resolveCreateLandingOverride` thread. `openPackageOverview` is fire-and-forget into a machine that handles `OPEN_VIEW` only in `extensionReady` and `viewActive.viewReady`, and the claim is planted whether or not the navigation was delivered. So "claimed but never landed" is a reachable state, and if no workspace-overview navigation follows to spend the claim, the window stays wherever startup left it — silently.

Worth noting that the deleted `getSnapshot().value !== "extensionReady"` check was doing two jobs and only one has been replaced. `claimedView` covers "generation navigated somewhere deliberate". The other job was guaranteeing the landing event was deliverable at all: sending it while the machine was in `extensionReady` meant it would be handled. Nothing covers that now.

The claim compensates by accident today — if the landing is dropped but a workspace overview arrives behind it, the redirect performs the landing the dropped event didn't. Making the claim persistent rather than single-shot (other thread) turns that accident into the mechanism.

packages/ballerina-extension/src/utils/state-machine-utils.ts
gigara: All three verified and fixed. The dropped-event mechanism is the root of it, and I had not spotted that `OPEN_VIEW` has no root-level handler — confirmed: the root `on` block carries seven events and none of them is `OPEN_VIEW`, so only `extensionReady` and `viewActive.viewReady` handle it.

**Consumed on a dropped event** — fixed. `resolveCreateLandingOverride` now takes a `deliverable` flag, passed as `handlesOpenView()` from `openView`, and returns without touching the claim when the navigation would be dropped. A navigation that will not happen is not an event the claim has any business reacting to.

**The `PackageOverview` exemption** — fixed. Only the claimed package's own overview leaves the claim standing; a different package's releases it, via `isSamePath` against the claim. Your list of reachable call sites is in the code comment, and the test now covers a different package explicitly.

**Never consumed** — this is where I did not take your suggestion, and the reasoning is worth putting on the record.

Your third release signal was the machine reaching `viewReady` on the created package. That fires *before* the competing navigation, not after: the traces in the description put the workspace overview 1.4s behind the landing, so the claim would be released and the original symptom would return. A persistent claim without that signal has the opposite problem — Show Overview stays redirected on every click until the user navigates elsewhere, which is worse than the single dead click the previous commit fixed.

So the claim stays single-shot, and the specific hole you found is closed at its cause instead: with the deliverability check, a claim can no longer be spent by a navigation that never happened, which is the property you were after ("a dropped event costs it nothing"). What remains is a claim that outlives a create where no workspace overview ever arrives, and one later Show Overview absorbing it. That is a deliberate trade rather than an oversight, and it was the one taken with the requester.

`it("is spent by the navigation it answers, not held for the next one")` therefore still holds. Three cases were added around it: a claim surviving an undeliverable navigation, surviving the claimed package's own overview, and being released by a different package's.

Separately, and following from your other thread — the landing no longer relies on the claim to compensate for a dropped event. See there.

packages/ballerina-extension/src/features/bi/pending-artifact.ts
gigara: Correct on every point, and this was the most valuable comment on the PR — the deleted `extensionReady` check was doing two jobs and I had only noticed one. "Claimed but never landed" was reachable, and the claim covering for it by accident is exactly the wrong shape to rely on.

Fixed, though not on the first attempt, and the detour is worth recording.

I first made the landing wait for a deliverable state (`await whenNavigationDeliverable()`). That closed the hole but blocked for up to the 15s bound on every create, and e2e group1 went green → red on that commit alone. Setup creates an integration for nearly every spec, so the added latency cascaded into unrelated 30s view timeouts. Reverted.

What is on the branch now solves it by ordering rather than waiting: the landing is sent **before** generation, from the `extensionReady` subscription this code already runs on, where `OPEN_VIEW` is certain to be handled. It is then re-asserted after generation, best-effort — if that one is dropped the first already landed, and the claim it refreshes still answers a workspace overview arriving later.

Three things make that safe:

- Landing early is invisible. The webview stays on the create's progress screen until the artifact appears in the project structure.
- It is closer to the original timing than either previous version — before this PR the landing was issued from *inside* `generatePendingArtifact`, mid-generation, so concurrent view-load and generation is not new behaviour.
- Nothing blocks. `handlesOpenView` remains, but only as the synchronous guard on spending the claim.

So the claim no longer compensates for a dropped landing — the landing is delivered on its own, and the claim is back to doing only the job it was introduced for.

packages/ballerina-extension/src/features/bi/pending-artifact.ts
```diff
@@ -138,30 +151,37 @@ export async function checkAndRunPendingArtifact(): Promise<void> {
...
-            `addedIntoWorkspace=${addedIntoWorkspace}, landOnPackageOverview=${landOnPackageOverview}`
+            `addedIntoWorkspace=${addedIntoWorkspace}`
         );
+        // Land BEFORE generating, not only after. `OPEN_VIEW` is handled in `extensionReady` and
+        // `viewActive.viewReady` only, and this runs from the `extensionReady` subscription, so a
+        // navigation sent here is certain to be acted on. Generation is asynchronous and startup
+        // navigates while it runs, so one sent only afterwards can arrive in a state that drops it.
+        // Invisible to the user: the webview stays on the create's progress screen until the
+        // artifact appears in the project structure.
+        landOnNewIntegrationAfterReload(stored.projectRoot);
```
NipunaRanasinghe: Blocking. "Invisible to the user" is true for three of the four kinds. The hold that makes it invisible only covers `KINDS_WRITING_AN_ARTIFACT = ["SERVICE", "AUTOMATION", "WORKFLOW"]` (ballerina-visualizer/src/Visualizer.tsx:207), so an AI Chat Agent create has no hold. This pre-landing navigates to the new package's overview with nothing covering it, then `generatePendingArtifact` replaces it with the wizard — a visible flash where it used to go straight there, plus an extra history entry.

Cheap fix: skip the pre-landing for the kind that navigates itself. That's the same distinction `claimedView` already makes, just needed before generation rather than after.

Separately, on the comment itself: "this runs from the `extensionReady` subscription, so a navigation sent here is certain to be acted on" isn't quite certain. `await clearPendingIntegrationPointer()` is a `globalState.update` sitting between the subscription firing and this line, so a startup navigation can land in that gap. It doesn't hurt anything now — the `deliverable` flag makes a dropped pre-landing free, and the re-assert follows — but it's the same "fires the moment the machine reaches extensionReady and then awaits" shape as the original defect 2, and the comment reads as a guarantee. Worth softening to "very likely to be acted on, and free if it isn't".

packages/ballerina-extension/src/features/bi/pending-artifact.ts
```diff
@@ -170,49 +190,19 @@ export async function checkAndRunPendingArtifact(): Promise<void> {
...
-        // on the startup screen.
-        ensureLandedOnNewIntegration(stored, landOnPackageOverview);
+        // Re-assert, in case startup navigated away while generation ran. Best-effort by design:
+        // if this one is dropped the landing above already happened, and the claim it refreshes
+        // still answers a workspace overview arriving later. Skipped when generation navigated
+        // somewhere deliberate of its own — only the agent does, handing off to its wizard, whose
+        // navigation also releases the claim.
+        if (!claimedView) {
+            landOnNewIntegrationAfterReload(stored.projectRoot);
+        }
```
NipunaRanasinghe, line 216: Note, not blocking. The re-assert re-plants the claim after generation, which can be after startup's navigations have already been answered — so a claim planted here has nothing left to spend it. It survives until the user navigates, and if their first move is Show Overview in a workspace (features/bi/activator.ts:154, no `exactView`), it gets redirected to the package they just came from and the click looks dead.

Much narrower than the earlier revision, since any other-view navigation now releases. But the claim still has no bound, and this commit makes a late one more likely rather than less. Fine to leave if it's a known edge — worth saying so in the description, since the current text presents the claim as bounded by the navigation it answers.

packages/ballerina-extension/src/utils/state-machine-utils.ts
```diff
@@ -28,6 +28,68 @@ import { getConstructBodyString } from "./history/util";
...
+        createLanding = undefined;
+        return { view: MACHINE_VIEW.PackageOverview, projectPath: claimed };
+    }
+    // Only the claimed package's own overview leaves the claim standing. Any other view — including
+    // a different package's overview, reachable from the config generator, Try It, the doc command
+    // and the BI diagram manager — is the user somewhere else, and must not leave a claim behind to
+    // redirect them back here.
+    const staysOnClaimedPackage =
+        viewLocation.view === MACHINE_VIEW.PackageOverview &&
+        (!viewLocation.projectPath || isSamePath(viewLocation.projectPath, createLanding));
```
NipunaRanasinghe: Nit. `!viewLocation.projectPath` counts as staying on the claimed package, but a package overview with no path resolves against `context.projectPath`, which may well be another package — see rpc-managers/visualizer/rpc-manager.ts:152/166/211 and features/bi/activator.ts:156/165. Comparing against the path `openView` actually resolves (`viewLocation.projectPath || StateMachine.context().projectPath`) would make it exact. Narrow, and it fails in the safe direction (claim kept rather than dropped), so only worth doing if it's cheap.

packages/ballerina-extension/src/stateMachine.ts
```diff
@@ -1004,6 +1006,17 @@ export const StateMachine = {
...
 
+/**
+ * Whether an `OPEN_VIEW` sent right now would be acted on.
+ *
+ * The machine handles it in `extensionReady` and `viewActive.viewReady` only — there is no
+ * root-level handler — so one sent while a view is mid-load is dropped without trace.
+ */
+function handlesOpenView(): boolean {
+    const value = stateService.getSnapshot().value;
+    return value === "extensionReady" || StateMachine.isReady();
```
NipunaRanasinghe, line 1019: Checked this and it's right — `isReady()` is exactly `viewActive.viewReady` (line 973), so the predicate covers precisely the two states that handle `OPEN_VIEW`, and there's no root-level handler to widen it.

Only note: this is now the load-bearing condition for the whole claim mechanism and it has no test, since `stateMachine` is mocked in the unit suite. If the state names or `isReady()` ever change, the claim silently stops working rather than failing. A one-line assertion somewhere, or a comment on `isReady` pointing back here, would keep the two from drifting apart.

packages/ballerina-extension/src/features/bi/pending-artifact.ts
gigara: Both correct. Fixed in 05fd2e8ffb.

Verified `KINDS_WRITING_AN_ARTIFACT` is `["SERVICE", "AUTOMATION", "WORKFLOW"]` at Visualizer.tsx:207, so "invisible to the user" held for three kinds of the four and the agent had nothing covering the pre-landing. Visible flash plus a spare history entry, on the one path that previously went straight to the wizard.

`KINDS_NAVIGATING_THEMSELVES` now carries that distinction and the caller consults it both before generation and after. Your framing — the same distinction `claimedView` already makes, just needed earlier — pointed at something further: a value returned from `generatePendingArtifact` can only be read after the fact, so it could never serve the pre-landing. Keeping it as well would have left the same fact in two places. The boolean is gone and the in-place path reads the same list.

On the comment: right, and worth being exact about. `await clearPendingIntegrationPointer()` sits between the subscription firing and the pre-landing, so a startup navigation can land in that gap — it is very likely to be handled, not certain. Reworded to say so, and to say why it is free when it is not (the `deliverable` guard keeps a dropped pre-landing from spending the claim, and the re-assert follows). Also noted that a landing sent *only* after generation is the one likely to be dropped, which is the actual reason for the ordering.

packages/ballerina-extension/src/utils/state-machine-utils.ts
gigara: Taken — it was cheap. Fixed in 05fd2e8ffb.

`resolveCreateLandingOverride` now receives the context's `projectPath` and compares the claim against `viewLocation.projectPath || contextProjectPath`, the same expression `openView` resolves a line later, so the two cannot disagree. Two tests cover it: a pathless package overview resolving to another package releases the claim, and one resolving to the claimed package keeps it.

You are right that it failed in the safe direction, which is why it was a nit rather than a bug — but the asymmetry was still wrong, and the fix is three lines.

packages/ballerina-extension/src/stateMachine.ts
gigara: Thanks for checking it rather than taking it on trust — and the drift risk is real, since a change to either would make the claim stop working silently rather than fail.

Added the cross-reference in 05fd2e8ffb: `isReady` now says it is also the `viewActive` half of `handlesOpenView` and that the two must stay in step. Went that way rather than the assertion because the unit suite mocks `stateMachine` wholesale, so any test would be asserting against the mock rather than the real machine — worse than nothing, since it would look like coverage.

If it is worth pinning properly, the place is the L4 LS-integration config, which runs against a real machine. Happy to add it there as its own change if you would rather not have this resting on a comment.

packages/ballerina-extension/src/features/bi/pending-artifact.ts
gigara: Agreed, and documented — the description now states it rather than implying the opposite.

Your reading of the interaction is right: the re-assert makes a late unspent claim more likely, not less, because it can plant one after startup's navigations have already been answered. Added under the claim rules as **Known edge, accepted rather than solved**, with the concrete failure (Show Overview as the next action gets redirected, second click works), the standalone case where no workspace-overview navigation ever arrives, and why the two alternatives are worse:

- Releasing once the landing renders fires *before* the competing navigation — the traces put it 1.4s behind — so the original symptom returns.
- Holding the claim rather than spending it leaves Show Overview redirected on every click until the user goes elsewhere, which is worse than one.

The competing navigation and a deliberate click are the same command with nothing to distinguish them, so any bound here is a trade rather than a fix. This is the one chosen, and the description says so, including that it was agreed with the requester rather than decided quietly.

NipunaRanasinghe (approved):
LGTM

PR #853, opened 2026-08-12
Fix the durable agent Attach Activity form and its code generation

hasithaa:
Fixes the durable agent **Attach Activity** form and its code generation, covering all six problems reported in wso2/product-integrator#2094.

## Fixes

1. **Selecting an activity now always opens the registration form** (retry policy, approval gating, connection bindings) instead of appending the entry with defaults. Previously the form only appeared when the activity needed a binding; configuring anything else required re-opening the attached activity. The form is titled with the chosen activity, like the edit path.
2. **Form description corrected** to `Register a workflow activity as a durable agent activity`.
3. **The activity selector is hidden when the activity is already chosen** (both the add-from-list and edit paths) — the entry's activity is its identity, not a choice. The edit path now passes the activity symbol in the template request.
4. **Selecting Auto Retry does not make retry configurations mandetory
5. **Tool Name / Description are no longer presented.** They mirrored the module `ActivityDecl`'s optional `name`/`description` fields from when activities registered as agent tools. They remain as hidden carriers so entries declared with them in source round-trip through an edit unchanged. (The Register Tool form never had them — `ToolDecl` has no such fields.)
6. **Saving the form no longer generates invalid code.** Two defects:
   - `FlowNodeForm` re-stamps the submitted node's `lineRange` from `targetLineRange`, which for capability edits is the expression-editor probe range at the *declaration start* — the LS capability replace then spliced the entry text into the declaration line (`validateClaimfinal workflow:DurableAgent ...`). The entry's own range is now restored on submit for every durable capability kind.
   - Declared `retryPolicy` and `bindings` fields were not hydrated into the edit form, so an edit-save silently dropped both from the regenerated entry. `CodeAnalyzer.collectDeclaredCapabilities` now decomposes `retryPolicy` into the retry form's values and explodes `bindings` into the per-parameter selectors.

## Tests

New goldens covering the agent capability flows:
- `node_template`: Register Activity from the palette and pre-selected (hidden selector, hidden identity fields, retry defaults, binding selector for a client parameter); Register Tool pre-selected (tool reference + ToolDecl gating only).
- `to_source`: activities-list append (retry record + bindings emitted) and rewrite at the entry's own range (the regression guard for the invalid code-gen); tools-list bare entry, gated `ToolDecl` mapping, and an ai toolkit variable registration (the MCP server flow's final write).
- `search`: the Add Activity list for a project with activities and for an empty project (empty category behind the create-new CTA), both with the Use MCP Server entry.

Existing goldens (builtin activities, Call Activity, diagram) regenerated for the Auto Retry defaults — each changes exactly the three hidden sub-field values.

`NodeTemplateTest`, `SourceGeneratorTest`, `SearchTest`, `ModelGeneratorTest`, `WorkflowNodeTemplateTest` and checkstyle are green locally; `rush build --to ballerina-visualizer` passes.

## Summary

- Fixed durable agent activity selection and registration flows.
- Corrected the activity form description and hid selectors for preselected activities.
- Made `AutoRetry` settings optional and added grouped form fields.
- Preserved retry values, bindings, and source ranges during edits and code generation.
- Hid tool name and description fields while preserving source round-tripping.
- Added support for grouped properties and expanded test coverage for retry configuration and registration flows.

### DIFF
--- a/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
+++ b/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
@@ -80,13 +80,16 @@ import java.util.Set;
  * @param dynamicFormFields maps dropdown option values to their conditional sub-fields keyed by property name
  *                          (for DropdownChoiceForm)
  * @param itemOptions       dropdown item options for DROPDOWN_CHOICE fields (id, content, value)
+ * @param advanceProperties the child properties of a GROUP_SECTION property, keyed by property name; the UI
+ *                          renders them behind the group's Expand/Collapse toggle
  * @since 1.0.0
  */
 public record Property(Metadata metadata, List<PropertyType> types, Object value, Object oldValue,
                        String placeholder, boolean optional, boolean editable, boolean advanced, boolean hidden,
                        Boolean modified, Diagnostics diagnostics, PropertyCodedata codedata, Object advancedValue,
                        Map<String, String> imports, String defaultValue, CommentProperty comment,
-                       Map<String, Map<String, Property>> dynamicFormFields, List<ItemOption> itemOptions) {
+                       Map<String, Map<String, Property>> dynamicFormFields, List<ItemOption> itemOptions,
+                       Map<String, Property> advanceProperties) {
 
     private static final Gson gson = new GsonBuilder().setPrettyPrinting().disableHtmlEscaping().create();
 
@@ -374,6 +377,11 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
         ADVANCE_PARAM_LIST,
         DROPDOWN_CHOICE,
         /**
+         * A property type that renders a labelled group with an Expand/Collapse toggle. The group's child
+         * properties are carried in {@link Property#advanceProperties()} and render behind that toggle.
+         */
+        GROUP_SECTION,
+        /**
          * A property type that renders a connection picker in the UI. The compatible
          * connection kinds are advertised through {@link PropertyCodedata#searchNodesKind()}.
          * The property value carries the picked connection variable name, or
@@ -411,6 +419,7 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
         private CommentProperty commentProperty;
         private Map<String, Map<String, Property>> dynamicFormFields;
         private List<ItemOption> itemOptions;
+        private Map<String, Property> advanceProperties;
 
         /**
          * Creates a builder pre-populated with all fields from an existing property.
@@ -453,6 +462,8 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
             }
             builder.itemOptions = original.itemOptions() != null
                     ? new ArrayList<>(original.itemOptions()) : null;
+            builder.advanceProperties = original.advanceProperties() != null
+                    ? new LinkedHashMap<>(original.advanceProperties()) : null;
             if (original.metadata() != null) {
                 builder.metadataBuilder = new Metadata.Builder<>(builder);
                 builder.metadataBuilder.label(original.metadata().label())
@@ -610,6 +621,11 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
 
         public Builder<T> itemOptions(List<ItemOption> itemOptions) {
             this.itemOptions = itemOptions;
+            return this;
+        }
+
+        public Builder<T> advanceProperties(Map<String, Property> advanceProperties) {
+            this.advanceProperties = advanceProperties;
             return this;
         }
 
@@ -1412,7 +1428,7 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
                     diagnosticsBuilder == null ? null : diagnosticsBuilder.build(),
                     codedataBuilder == null ? null : codedataBuilder.build(), advancedValue,
                     imports == null ? null : imports, defaultValue, commentProperty, dynamicFormFields,
-                    itemOptions);
+                    itemOptions, advanceProperties);
             this.metadataBuilder = null;
             this.types = new ArrayList<>();
             this.value = null;
@@ -1430,6 +1446,7 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
             this.imports = null;
             this.dynamicFormFields = null;
             this.itemOptions = null;
+            this.advanceProperties = null;
             return property;
         }
     }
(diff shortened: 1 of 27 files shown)


packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
```diff
@@ -80,13 +80,16 @@
...
+ *                          renders them behind the group's Expand/Collapse toggle
  * @since 1.0.0
  */
 public record Property(Metadata metadata, List<PropertyType> types, Object value, Object oldValue,
                        String placeholder, boolean optional, boolean editable, boolean advanced, boolean hidden,
                        Boolean modified, Diagnostics diagnostics, PropertyCodedata codedata, Object advancedValue,
                        Map<String, String> imports, String defaultValue, CommentProperty comment,
-                       Map<String, Map<String, Property>> dynamicFormFields, List<ItemOption> itemOptions) {
+                       Map<String, Map<String, Property>> dynamicFormFields, List<ItemOption> itemOptions,
+                       Map<String, Property> advanceProperties) {
```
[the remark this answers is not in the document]
rdulmina: Checked this against the contract and the packaging, and the premise does not hold: neither addition is new RPC surface. Both have been in the shared client contract for a while, and the Java record was the half that could not express them.

| Contract element | Where | Added |
| --- | --- | --- |
| `advanceProperties?: NodeProperties` on `Property` | `packages/ballerina-core/src/interfaces/bi.ts:279` | `26519cad6e`, 2025-06-30 |
| `"GROUP_SECTION"` in `FormFieldInputType` | `packages/ballerina-core/src/interfaces/bi.ts` | `737d685864`, 2026-03-20 |
| Rendering a `GROUP_SECTION` inside `dynamicFormFields` from `advanceProps` | `packages/ballerina-side-panel/src/components/editors/DropdownChoiceForm.tsx:108-141` | `43d614440e`, 2026-04-06 (on `main`) |

The visualizer has been reading `property.advanceProperties` since 2025 (`packages/ballerina-visualizer/src/utils/node-property-utils.ts:121`) and already renders exactly this shape. `cf47117b2e` makes the language server emit a payload the client already declares and handles.

On the "older visualizer" scenario: the language server is not versioned independently of the visualizer. `copyLS` (`packages/ballerina-extension/package.json:1535`) copies `ballerina-language-server-<package.json version>.jar` into `ls/`, which is packaged into the same VSIX as the webview bundle, so a released build always pairs the two at the same version. There is no cross-version pairing to gate.

On deserialization: the component is nullable and Gson omits nulls, so every payload that does not use a group is byte-identical to before. The regenerated goldens show `advanceProperties` on exactly one property - the AutoRetry group - and nowhere else. On the client side the absence is a no-op: `convertNodePropertiesToFormFields(undefined)` iterates nothing.

Keeping the change as it stands. No contract version bump or compatibility shim is warranted for filling in a field the contract already specifies.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
```diff
@@ -409,11 +416,15 @@ public static void addRetryPolicyFormProperties(NodeBuilder nodeBuilder, String
 
         Map<String, Map<String, Property>> dynamicFields = new LinkedHashMap<>();
         dynamicFields.put(NO_RETRY_VALUE, Map.of());
-        dynamicFields.put(AUTO_RETRY_VALUE, autoRetryFields);
+        // The tuning fields sit behind the AutoRetry branch's collapsible group instead of being
```
dulajdilshan, line 419: **Wire values from the new group back to the root retry fields before generating source.** `retryPolicyEntryValue()` only reads `maxRetries`, `retryDelay`, `retryBackoff`, and `maxRetryDelay` from the root property map. The new form places the editable controls under `retryPolicy.AutoRetry.autoRetryOptions.advanceProperties`; `updateNodeProperties()` preserves that nested group but does not copy its edited children into those root properties. Consequently, choosing Auto Retry and filling the expanded fields emits `{}` (or the pre-edit root values), not the entered configuration. Please flatten/synchronise the group values during form-to-node conversion (or teach source generation to read the group), and add an end-to-end form submission/source test rather than only template and direct-to-source fixtures.

rdulmina: Checked this end to end, and the values do reach the root fields — the group's children are not a separate storage location, they share the key.

The chain has three seams, and each is now covered by a test on this branch (`9b3cc76aec`):

1. **Property → form field.** `convertNodePropertyToFormField` maps a GROUP_SECTION's `advanceProperties` into `advanceProps` (`node-property-utils.ts:121`), and `convertNodePropertiesToFormFields` keys each child by its own property name — so the child is the field `maxRetries`, not `retryPolicy.AutoRetry.autoRetryOptions.maxRetries`. Test: *"becomes a form field under its own key, which is the key its value is stored under"*.

2. **Form field → form value.** `DropdownChoiceForm` renders the group's `advanceProps` through `FieldFactory` (`DropdownChoiceForm.tsx:108-141`), which registers each editor under `field.key`. So editing Max Retries writes to the form value `maxRetries`. Test: *"renders the group's child under its own key once expanded, so its edit reaches the root property"*.

3. **Form value → node property.** `updateNodeProperties` iterates the submitted values and assigns each to the node property of the same key (`node-property-utils.ts:376-420`). The root `maxRetries` property is exactly that key, so it receives the edit — and it is what `retryPolicyEntryValue()` / `autoRetryRecordLiteral()` read. Test: *"carries an edited group field to the root property source generation reads"*.

This is the same mechanism the retry sub-fields already relied on before the group existed, and the one `RestActivityStrategy` uses for `method`/`message`: the field inside the branch is a definition carrying an empty value, and the root property of the same key is the storage. The comment in `addRetryPolicyFormProperties` says this, and it is now enforced rather than just described.

I also covered the case behind "or the pre-edit root values": opening a node that already has `maxRetries: 5` and expanding the group must not overwrite it with the branch field's own empty value. Test: *"keeps the loaded value when the group is expanded"* — the loaded value survives.

On the testing ask more broadly, that part was fair and I have taken it: the three tests above are the form-submission coverage that the template and to-source fixtures could not give. What still is not covered anywhere is a real VS Code round trip (webview → RPC → LS); that needs the Playwright suite, and I have not added one here.

Suites after the change: side panel 145/145, visualizer 223/223.

dulajdilshan (dismissed):

PR #856, opened 2026-08-12
Fix silent data loss in workflow form hydration

hasithaa:
Fixes silent data-loss defects in the workflow designer forms (found by the survey in `docs/dx/workflow-form-semantics.md`): values present in source that the matching form never received, so an edit-save regenerated the construct without them.

## Fixes

- **Peer capabilities** — the keyword-escaped `'wait` field is now unescaped before the hydration lookup, so editing an async peer no longer turns it blocking or drops its `callbackChannel`. `PeerDecl.requiresApproval`/`userRoles` hydrate, appear as advanced fields, and are emitted — reviewer roles through `roleSource`, so an expression-mode reference is written as-is rather than quoted into a role literal of the same spelling.
- **Named-argument call sites** — one shared `argumentExpression(index, name)` resolves every anchor parameter from either the positional or the named form: `callActivity`'s `activityFunction`/`args` (user-defined *and* builtin REST/SOAP/email), `workflow:run`'s `processFunction`/`input`, `childWorkflow`, and the durable agent calls, which previously opened blank. `workflow:run` also types its `input` from the target workflow instead of the library's `anydata`.
- **Positional `awaitHumanTask`** — positional 3/4/5 map to `title`/`description`/`timeout` per the signature. `description` now also reaches the generated call: it is a reserved key, so the signature-derived form stores it as `$description` and the old plain-key read found nothing.
- **`retryPolicy` parsing** — the record split is nesting- and string-aware and the `NoRetry`/`ManualRetry`/`HumanReview` sentinels match exact identifiers, so `int:max(2, 3)` parses whole and a user identifier that merely contains a sentinel word round-trips verbatim.
- **Generated file name** — workflow and durable-agent source is written to `workflows.bal`, matching `functions.bal` / `connections.bal` / `agents.bal`.

Closes https://github.com/wso2/product-integrator/issues/2107
Closes https://github.com/wso2/product-integrator/issues/2108
Closes https://github.com/wso2/product-integrator/issues/2109
Closes https://github.com/wso2/product-integrator/issues/2110
Closes https://github.com/wso2/product-integrator/issues/2382

## Tests

Every fix is locked by a golden or a unit test — including the first non-empty capability list in any golden (`durable_agent_box`, which needed a fixture fix: `inputType: map<json>` is a type descriptor, not an expression, and the parser swallowed every field after it). `ModelGeneratorTest`, `NodeTemplateTest`, `SourceGeneratorTest`, the `flow-model-generator-core` unit tests and checkstyle are green locally.

Remaining survey findings are tracked separately: wso2/product-integrator#2111, #2112, #2113, #2114.


packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1992,20 +2034,25 @@ private void populateFallbackHumanTaskProperties(RemoteMethodCallActionNode call
...
-        if (descValue == null && args.size() > 3 && args.get(3) instanceof PositionalArgumentNode posArg3) {
-            descValue = posArg3.expression().toSourceCode().strip();
+        if (descValue == null && args.size() > 4 && args.get(4) instanceof PositionalArgumentNode posArg4) {
+            descValue = posArg4.expression().toSourceCode().strip();
         }
-        // timeout: named arg only
         String timeoutValue = namedArgs.get(HumanTaskBuilder.TIMEOUT_KEY);
+        if (timeoutValue == null && args.size() > 5 && args.get(5) instanceof PositionalArgumentNode posArg5) {
+            timeoutValue = posArg5.expression().toSourceCode().strip();
+        }
```
dulajdilshan: **[Standards]** Possible **Duplicated Code**. Four hand-rolled `args.size() > N && args.get(N) instanceof PositionalArgumentNode posArgN` blocks — exactly what `agentArgumentSource(args, index, name)` (:1161, over the `argumentExpression` helper added at :1703 in this same diff) already does, including the named-argument scan. `namedArgs` stays in use above for `taskName`/`userRoles`, so this replacement compiles as-is.

```suggestion
        String payloadValue = agentArgumentSource(args, 2, HumanTaskBuilder.PAYLOAD_KEY);
        String titleValue = agentArgumentSource(args, 3, HumanTaskBuilder.TITLE_KEY);
        String descValue = agentArgumentSource(args, 4, HumanTaskBuilder.DESCRIPTION_KEY);
        String timeoutValue = agentArgumentSource(args, 5, HumanTaskBuilder.TIMEOUT_KEY);
```

dulajdilshan: **[Spec]** The #2109 fix duplicates the helper added for #2108. The same PR introduces `argumentExpression(args, index, namedArgName)` for exactly this "positional-or-named" read, but this fallback block hand-rolls `args.size() > n && args.get(n) instanceof PositionalArgumentNode` four more times. Two spellings of one rule in one file will drift.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1614,14 +1621,20 @@ private void collectDeclaredCapabilities(ExpressionNode listExpr, String capabil
...
             values.put(key, value);
         }
     }
 
+    // Unescapes a keyword-escaped identifier ('wait -> wait), the way the builtin activity
+    // hydration already does for record keys ('from -> from).
+    private static String stripKeywordEscape(String identifier) {
+        return identifier.startsWith("'") ? identifier.substring(1) : identifier;
+    }
+
```
dulajdilshan: **[Standards]** Possible **Duplicated Code**. `stripKeywordEscape` re-implements the existing shared `ParamUtils.removeLeadingSingleQuote(String)` (`core/utils/ParamUtils.java:107`), which is identical apart from also being null-safe. Suggest deleting this helper and calling `ParamUtils.removeLeadingSingleQuote(...)` at the single use site (:1561).

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1614,14 +1621,20 @@ private void collectDeclaredCapabilities(ExpressionNode listExpr, String capabil
...
     private static void putIfNotBlank(Map<String, String> values, String key, String value) {
         if (value != null && !value.isBlank()) {
             values.put(key, value);
         }
     }
 
+    // Unescapes a keyword-escaped identifier ('wait -> wait), the way the builtin activity
+    // hydration already does for record keys ('from -> from).
+    private static String stripKeywordEscape(String identifier) {
+        return identifier.startsWith("'") ? identifier.substring(1) : identifier;
```
dulajdilshan: **[Spec]** The new helper is left half-applied. Its own comment cites *"the way the builtin activity hydration already does for record keys (`'from` -> `from`)"*, but that site was not switched to it — `CodeAnalyzer.java:2175` still reads `rawKey.startsWith("'") ? rawKey.substring(1) : rawKey` inline.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -2280,22 +2327,116 @@ private static boolean isRoleLiteral(String expression) {
...
+                    if (depth == 0) {
+                        return i;
+                    }
+                }
+                default -> {
+                }
+            }
+        }
+        return -1;
+    }
```
dulajdilshan: **[Standards]** Possible **Duplicated Code**. `splitTopLevel` and `topLevelColonIndex` are the same bracket/string scanner written twice — identical `inString`/`depth`/escape handling — differing only in the terminating character and what they return. One scanner parameterised by the delimiter, or one that returns key/value pairs directly, removes ~40 lines.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -553,7 +553,7 @@ public void visit(RemoteMethodCallActionNode remoteMethodCallActionNode) {
                 nodeBuilder.codedata().module(ACTIVITY_MODULE);
                 populateBuiltinActivityProperties(remoteMethodCallActionNode, builtinSymbol);
             } else {
-                overrideSymbolFromFirstArg(remoteMethodCallActionNode.arguments());
+                overrideSymbolFromFirstArg(remoteMethodCallActionNode.arguments(), "activityFunction");
```
dulajdilshan: **[Standards]** Possible **Primitive Obsession** / duplicated literal. Anchor parameter names arrive as bare inline strings throughout the change, though constants already hold the same values. `"activityFunction"` appears unconstanted three times (:556, :1753, :1805); `"eventName"`/`"data"`/`"token"`/`"query"`/`"input"` (:1047-1049, :1109-1122) duplicate the `*_KEY` constants passed on the very same call line; `"childWorkflow"` at :578; and `"processFunction"` at :4494 duplicates `Constants.Workflow.RUN_PROCESS_FUNCTION_PARAM`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -4350,7 +4491,7 @@ private void addModelProviderMetadata(SeparatedNodeList<FunctionArgumentNode> ar
     private void handleWorkflowFunctionSymbol(FunctionCallExpressionNode functionCallExpressionNode,
                                               FunctionSymbol functionSymbol) {
         if (isWorkflowOperation(functionSymbol, RUN_METHOD_NAME)) {
-            overrideSymbolFromFirstArg(functionCallExpressionNode.arguments());
+            overrideSymbolFromFirstArg(functionCallExpressionNode.arguments(), "processFunction");
```
dulajdilshan: **[Standards]** Same finding as the comment at :556 — this one has a constant sitting ready for it: `Constants.Workflow.RUN_PROCESS_FUNCTION_PARAM` is literally `"processFunction"`.

```suggestion
            overrideSymbolFromFirstArg(functionCallExpressionNode.arguments(),
                    Constants.Workflow.RUN_PROCESS_FUNCTION_PARAM);
```

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentPeerBuilder.java
```diff
@@ -144,6 +146,37 @@ public void setConcreteTemplateData(TemplateContext context) {
...
+                            + "e.g. \"support-lead\" or [\"finance\", \"manager\"].")
+                    .stepOut()
+                .type().fieldType(Property.ValueType.EXPRESSION)
+                    .ballerinaType("string|string[]").selected(true).stepOut()
+                .placeholder("")
+                .editable(true)
+                .optional(true)
+                .advanced(true)
+                .stepOut()
+                .addProperty(USER_ROLES_KEY);
```
dulajdilshan: **[Standards]** Possible **Duplicated Code**. Third verbatim copy of the approval-gate property pair (cf. `DurableAgentAddActivityBuilder.java:275-300` and `DurableAgentRegisterToolBuilder`); only the description wording differs. Suggest extracting a shared `addApprovalGateProperties(String doc)`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/HumanTaskBuilder.java
```diff
@@ -324,11 +325,15 @@ public static void relabelHumanTaskFormProperties(Map<String, Property> properti
     }
 
     private static void relabel(Map<String, Property> properties, String key, String label, String description) {
-        Property existing = properties.get(key);
+        // Same reserved-key escaping as resolveParameterProperty: `description` lands under
+        // `$description` on the signature-derived path, so relabelling only the plain key left
+        // that field showing the compiler-derived doc instead of the form's own.
+        String actualKey = properties.containsKey(key) ? key : FlowNodeUtil.getPropertyKey(key);
```
dulajdilshan: **[Standards]** Possible **Duplicated Code**. The plain-or-escaped key fallback is now written twice, differently, in this one file — here, and as `resolveParameterProperty` at :438. Suggest one `resolveKey`-style helper used by both, so the reserved-key rule has a single spelling.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/HumanTaskBuilder.java
```diff
@@ -324,11 +325,15 @@ public static void relabelHumanTaskFormProperties(Map<String, Property> properti
...
+        Property existing = properties.get(actualKey);
         if (existing == null) {
             return;
         }
-        properties.put(key, Property.Builder.copyFrom(existing)
+        properties.put(actualKey, Property.Builder.copyFrom(existing)
```
dulajdilshan: **[Spec]** Scope creep. This fixes reserved-key escaping (`description` -> `$description`) on the *signature-derived* path, but #2109 is scoped strictly to the unresolved-symbol fallback: *"The fallback hydration of `ctx->awaitHumanTask(...)` (used when the symbol is unresolved) mis-maps the positional arguments"*. No issue mentions the escaped-key/emitter defect. It is a real bug — but note it is the one human-task change carrying goldens, while the requirement the spec actually states has none (the PR acknowledges that gap).

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/resources/diagram_generator/config/durable_agent_box.json
```diff
@@ -0,0 +1,2229 @@
...
+            "model": {
+              "name": "deskModel",
+              "path": "https://bcentral-packageicons.azureedge.net/images/ballerina_ai_1.13.0.png",
+              "type": "Wso2ModelProvider"
+            },
+            "activities": [],
+            "humanTasks": [],
+            "tools": [],
+            "events": [],
+            "peers": [],
```
dulajdilshan: **[Spec]** The #2107 hydration fix is not actually locked by this golden. Spec: *"`PeerDecl.requiresApproval` and `userRoles` are neither hydrated into the form, nor presented, nor written"* and *"`'wait` never hydrates."* The fixture does declare a gated async peer (`durable_agent_box.bal:34`), but the recorded golden has every capability array empty — `"activities": []`, `"humanTasks": []`, `"tools": []`, `"events": []`, `"peers": []` — and the node records `hasDiagnostics: true`. So neither `stripKeywordEscape` on `'wait` nor the new `requiresApproval`/`userRoles` map entries are exercised; this golden would be byte-identical before the fix. The *presented* and *written* halves of #2107 are covered (`node_template/config/durable_agent_peer.json`, `to_source/config/durable_agent_peer_gated_to_source.json`); the *hydrated* half is not.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
dulajdilshan: **[Standards] Duplicated Code.** Four hand-rolled positional reads, right beside the helper that already does this: `agentArgumentSource` (:1161) covers both the named and positional form. `namedArgs` stays in use above, so this compiles as-is.

```suggestion
        String payloadValue = agentArgumentSource(args, 2, HumanTaskBuilder.PAYLOAD_KEY);
        String titleValue = agentArgumentSource(args, 3, HumanTaskBuilder.TITLE_KEY);
        String descValue = agentArgumentSource(args, 4, HumanTaskBuilder.DESCRIPTION_KEY);
        String timeoutValue = agentArgumentSource(args, 5, HumanTaskBuilder.TIMEOUT_KEY);
```

dulajdilshan: **[Spec]** The #2109 fix hand-rolls the positional-or-named read four more times instead of using `argumentExpression`, added in this same PR for #2108. Two spellings of one rule in one file will drift.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
dulajdilshan: **[Standards] Duplicated Code.** `stripKeywordEscape` duplicates the shared `ParamUtils.removeLeadingSingleQuote` (`core/utils/ParamUtils.java:107`), which is also null-safe. Drop it and call that at the one use site (:1561).

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
dulajdilshan: **[Spec]** Helper left half-applied. The comment cites the builtin-activity path as precedent, but that site (:2175) still inlines `rawKey.startsWith("'") ? rawKey.substring(1) : rawKey`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
dulajdilshan, line 2397: **[Standards] Duplicated Code.** `splitTopLevel` and `topLevelColonIndex` are the same bracket/string scanner twice — same `depth`/`inString`/escape handling, only the terminator differs. One parameterised scanner drops ~40 lines.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
dulajdilshan: **[Standards] Primitive Obsession / duplicated literal.** Anchor parameter names are inline strings where constants exist: `"activityFunction"` x3 (:556, :1753, :1805), `"childWorkflow"` (:578), `"query"`/`"input"`/`"instanceId"`/`"eventName"`/`"data"`/`"token"` (:1047-1122, each duplicating the `*_KEY` constant on the same call line), `"processFunction"` (:4494).

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
dulajdilshan: **[Standards]** Same as :556 — and this one has its constant ready: `Constants.Workflow.RUN_PROCESS_FUNCTION_PARAM` is literally `"processFunction"`.

```suggestion
            overrideSymbolFromFirstArg(functionCallExpressionNode.arguments(),
                    Constants.Workflow.RUN_PROCESS_FUNCTION_PARAM);
```

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentPeerBuilder.java
dulajdilshan: **[Standards] Duplicated Code.** Third copy of the approval-gate property pair (cf. `DurableAgentAddActivityBuilder.java:275-300`, `DurableAgentRegisterToolBuilder`); only the wording differs. Extract one `addApprovalGateProperties(String doc)`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/HumanTaskBuilder.java
dulajdilshan: **[Standards] Duplicated Code.** The plain-or-escaped key fallback is now written twice in this file, differently — here and `resolveParameterProperty` (:438). One helper, one spelling.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/HumanTaskBuilder.java
dulajdilshan: **[Spec] Scope creep.** #2109 covers only the unresolved-symbol fallback (*"used when the symbol is unresolved"*); this fixes reserved-key escaping on the signature-derived path. A real bug — but it's the human-task change that got goldens, while the one the spec asks for has none.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/resources/diagram_generator/config/durable_agent_box.json
dulajdilshan: **[Spec]** This golden doesn't lock the #2107 hydration fix. The fixture declares a gated async peer, yet every capability array here is empty (`"peers": []`, ...) and the node carries `hasDiagnostics: true` — `'wait`, `requiresApproval` and `userRoles` hydration is never exercised, so the golden would be byte-identical before the fix. #2107's *presented* and *written* halves are covered elsewhere; *hydrated* isn't.

dulajdilshan (commented):
**[Spec] No fixture for `workflow:run`.** #2108 asks that `workflow:run(processFunction = ...)` carry the target as the node symbol. Implemented at `CodeAnalyzer.java:4494`, but untested — `runChildWorkflow` and `callActivity` both got goldens.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Done — the fallback reads all six parameters through one rule, `fallbackHumanTaskArgumentValues(args)`, each by signature position or by name via `argumentExpression`. Extracting it also made #2109 testable: `HumanTaskFallbackArgumentsTest` covers full positional, partial positional (position 3 is the payload, the rest empty), named in any order, and all-named.

hasithaa: Done — see above: one spelling of positional-or-named now, `argumentExpression`, used by the fallback through `fallbackHumanTaskArgumentValues`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Done — removed; the use site calls `ParamUtils.removeLeadingSingleQuote`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Done — the builtin-activity site uses `ParamUtils.removeLeadingSingleQuote` as well.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Done — one scanner, `topLevelIndexOf(source, terminator, from)`; the comma split and the colon lookup are two calls to it (the colon lookup gains the template awareness it lacked).

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Done — `Constants.Workflow` now names every anchor the analyzer reads back: `CALL_ACTIVITY_FUNCTION_PARAM`, `CALL_ACTIVITY_ARGS_PARAM`, `CHILD_WORKFLOW_PARAM`, `AGENT_RUN_QUERY_PARAM`, `AGENT_RUN_INPUT_PARAM`, `AGENT_INSTANCE_ID_PARAM`, `AGENT_EVENT_NAME_PARAM`, `AGENT_DATA_PARAM`, `AGENT_TOKEN_PARAM`; `EXCLUDED_CALL_ACTIVITY_PARAMS` uses the same. One note on the agent calls: the `*_KEY` constants on those lines are form property keys and differ from the module's parameter names (`AGENT_ID_KEY` is "agentId", the parameter is `instanceId`), so these are new constants rather than the property keys.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Done — `RUN_PROCESS_FUNCTION_PARAM`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentPeerBuilder.java
hasithaa: Done — `WorkflowUtil.addApprovalGateProperties(builder, approvalKey, approvalDoc, userRolesKey, reviewerRolesDoc)`; the activity, tool and peer forms pass only their wording.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/HumanTaskBuilder.java
hasithaa: Done — one `presentKey(properties, key)`, used by `relabel` and by `resolveParameterProperty`. `HumanTaskKeyFallbackTest` covers both spellings for the relabel and the lookup.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/HumanTaskBuilder.java
hasithaa: Fair. The fallback path cannot be reached by a fixture project (it needs the module not to resolve), so #2109 now has an L1 test instead: `HumanTaskFallbackArgumentsTest` drives the extracted reader with parsed calls. The signature-path `$description` fix keeps its golden and gets `HumanTaskKeyFallbackTest`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/resources/diagram_generator/config/durable_agent_box.json
hasithaa: You were right, and the cause was the fixture, not the reader: `inputType: map<json>` is a parse error — a type descriptor is not an expression, and the parser takes `map<json>, activities: [` as the start of a query expression ("missing from keyword… missing select clause"), swallowing every field after it. The declaration reached the analyzer with no activities/events/humanTasks/peers at all, which is also where the "declares no data-event channel named hotelResults" diagnostic came from. The fixture now names the type (`type TripInput map<json>;`, declared last so no golden range moves) and the golden is regenerated with the capabilities hydrated — `'wait`/`callbackChannel`/`requiresApproval`/`userRoles` on the peer, the decomposed retryPolicy and bindings on the activity, the event and the human task — and without the spurious diagnostic. Worth noting: no golden in the repo had a non-empty capability list before this.

dulajdilshan (dismissed):

NipunaRanasinghe (dismissed):

NipunaRanasinghe (dismissed):
LGTM

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentPeerBuilder.java
```diff
@@ -231,6 +245,12 @@ public Map<Path, List<TextEdit>> toSource(SourceBuilder sourceBuilder) {
             entry.append(", 'wait: false");
             entry.append(", callbackChannel: ").append(WorkflowUtil.quoteIfPlain(callbackChannel));
         }
+        if (requiresApproval) {
+            entry.append(", requiresApproval: true");
+        }
+        if (!userRoles.isBlank()) {
+            entry.append(", userRoles: ").append(WorkflowUtil.quoteIfBareRole(userRoles));
```
NipunaRanasinghe: **Medium — the peer's reviewer roles bypass `roleSource`, so an expression-mode value gets quoted into a literal.**

`addApprovalGateProperties` gives this field the same multi-mode TEXT/EXPRESSION types the tool and activity forms have, and both of those read it through `WorkflowUtil.roleSource` (`DurableAgentRegisterToolBuilder:130`, `DurableAgentAddActivityBuilder:299-300`), which checks the selected mode first. Here `propertyValue` returns the raw `value.toString()` and hands it straight to `quoteIfBareRole`: a user who selects Expression and types `financeRoles` — a module-level variable of role type — gets `userRoles: "financeRoles"` written, exactly the rewrite `roleSource`'s comment says the mode check exists to prevent. A text-mode value has the opposite problem: `value()` is still the `string \`…\`` template that `toSourceCode()` would have unwrapped.

Route it through `sourceBuilder.getProperty(USER_ROLES_KEY).map(WorkflowUtil::roleSource).orElse("")` like the other two.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1106,20 +1115,22 @@ private void populateDurableAgentCallArguments(MethodCallExpressionNode callNode
...
+                "The running agent's instance ID", agentArgumentSource(arguments, 0, AGENT_INSTANCE_ID_PARAM));
         switch (nodeKind) {
             case DURABLE_AGENT_UPDATE -> {
                 // The argument's own source, not the unquoted name: a literal keeps its escapes
                 // verbatim, and a channel the form cannot represent as a literal is still shown
                 // instead of opening the field blank and writing that blank back on save.
                 addAgentCallProperty(DurableAgentUpdateBuilder.EVENT_NAME_KEY, "Data Event",
-                        "The channel the payload is sent on", positionalArgumentSource(arguments, 1));
+                        "The channel the payload is sent on",
+                                agentArgumentSource(arguments, 1, AGENT_EVENT_NAME_PARAM));
```
NipunaRanasinghe, line 1127: **Medium — `resolveDurableAgentDataEventName` still reads argument 1 positionally, so the new fixture loses `dataName`.**

`CodeAnalyzer:1266` and `:1271` take `arguments.get(1)` and require a `PositionalArgumentNode`. The fixture this PR adds writes `travelAgent.sendData(instanceId = instanceId, eventName = "hotelResults", data = {ok: true})` and `waitForDataResult(instanceId = instanceId, token = token)`, so both return null and `metadata().addData("dataName", ...)` at `:1240` never fires — `durable_agent_box.json` carries `agentName` and `waits` on all three agent-call nodes and no `dataName` anywhere, and the golden now locks that in.

Same two anchors as the properties here (`AGENT_EVENT_NAME_PARAM` / `AGENT_TOKEN_PARAM`), so `argumentExpression` covers it.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1758,10 +1805,12 @@ private void populateActivityCallProperties(RemoteMethodCallActionNode remoteMet
                 ActivityCallBuilder.ADVANCE_CONFIGURATIONS,
                 ActivityCallBuilder.ADVANCE_CONFIGURATIONS);
 
-        // Step 2: Get activity function params directly from the symbol (avoids expensive FunctionDataBuilder).
+        // Step 2: Get activity function params directly from the symbol (avoids expensive
+        // FunctionDataBuilder). The function reference may be positional or named.
         List<ParameterSymbol> activityParamSymbols = List.of();
-        if (!args.isEmpty() && args.get(0) instanceof PositionalArgumentNode firstArg) {
-            Optional<Symbol> resolvedSymbol = semanticModel.symbol(firstArg.expression());
+        Optional<ExpressionNode> activityRefExpr = argumentExpression(args, 0, CALL_ACTIVITY_FUNCTION_PARAM);
```
NipunaRanasinghe, line 1817: **Medium — the builtin-activity branch of the same call site was not converted.**

`resolveBuiltinActivitySymbol` (`:1670`) reads the same `activityFunction` anchor positionally, so `ctx->callActivity(activityFunction = activity:callRestAPI, args = {...})` returns null and the call falls into this generic branch instead of `CONNECTION_ACTIVITY_CALL` + `populateBuiltinActivityProperties` — the REST/SOAP/email form fields are gone and an edit-save regenerates the call without them. `populateBuiltinActivityProperties` (`:2104`) requires a positional argument 1 for the same reason, so it would not recover even if the symbol resolved.

Both take the same two parameters this line already resolves.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/Constants.java
```diff
@@ -107,6 +107,19 @@ private Workflow() {
         public static final String CONTEXT_CLASS_NAME = "Context";
         public static final String RUN_METHOD_NAME = "run";
         public static final String RUN_PROCESS_FUNCTION_PARAM = "processFunction";
+        public static final String RUN_INPUT_PARAM = "input";
```
NipunaRanasinghe, line 110: **Low — `RUN_INPUT_PARAM` is never read, because the one site that would use it still hand-rolls the lookup.**

`populateWorkflowRunProperties` resolves `processFunction` through `argumentExpression` at `CodeAnalyzer:4485`, then at `:4503-4511` reads the input with its own positional check plus a named-argument loop keyed on `WorkflowRunBuilder.INPUT_KEY`. Behaviour is the same, but it is the fourth copy of the rule this PR set out to unify, and it leaves this constant dead.

`argumentExpression(args, 1, RUN_INPUT_PARAM)` replaces the block; note it would also make `RUN_INPUT_PARAM` and `AGENT_RUN_INPUT_PARAM` two constants for the same `"input"` string.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/test/java/io/ballerina/flowmodelgenerator/core/model/node/HumanTaskKeyFallbackTest.java
```diff
@@ -0,0 +1,83 @@
...
+                HumanTaskBuilder.DESCRIPTION_KEY);
+
+        Map<String, Property> escaped = new LinkedHashMap<>();
+        escaped.put(ESCAPED_DESCRIPTION, property("\"escaped\""));
+        Assert.assertEquals(HumanTaskBuilder.presentKey(escaped, HumanTaskBuilder.DESCRIPTION_KEY),
+                ESCAPED_DESCRIPTION);
+
+        Assert.assertNotEquals(ESCAPED_DESCRIPTION, HumanTaskBuilder.DESCRIPTION_KEY,
+                "the test only means something if the two spellings differ");
+        Assert.assertNull(new LinkedHashMap<String, Property>()
```
NipunaRanasinghe: **Nit. This assertion cannot fail.**

```java
Assert.assertNull(new LinkedHashMap<String, Property>()
        .get(HumanTaskBuilder.presentKey(new LinkedHashMap<>(), HumanTaskBuilder.DESCRIPTION_KEY)));
```

`get` on a freshly constructed empty map returns null for every key, so `presentKey` could return anything at all and this still passes. The behaviour the comment describes — an absent key resolves to its escaped form — is what `assertEquals(presentKey(emptyMap, DESCRIPTION_KEY), ESCAPED_DESCRIPTION)` would actually check.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentPeerBuilder.java
rdulmina: Confirmed and fixed in f6faac1513.

`EXPRESSION_LIKE` is `[(){}:.]`, so a bare `financeRoles` matched none of it and fell through `quoteIfBareRole` → `quoteIfPlain`, giving `userRoles: "financeRoles"` exactly as you describe. It now reads through the mode check like the other two:

```java
String userRoles = sourceBuilder.getProperty(USER_ROLES_KEY)
        .map(WorkflowUtil::roleSource).orElse("");
```

and the `quoteIfBareRole` at the call site is gone, since `roleSource` already quotes the non-expression case.

Worth noting this was a regression from the #970 merge on this branch: folding staging's `addRoleFieldTypes` into `addApprovalGateProperties` gave the peer field the multi-mode types without updating its emitter to match.

The existing `durable_agent_peer_gated_to_source` fixture could not catch it — it sets `valueType: EXPRESSION` but carries no `types` array, so `isExpressionModeSelected` is false and the output was unchanged. Added `durable_agent_peer_gated_expression_roles_to_source.json` with `EXPRESSION` actually selected; verified it fails on the old code (`userRoles: "financeRoles"`) and passes on the new (`userRoles: financeRoles`).

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
rdulmina: Confirmed and fixed in f6faac1513. The golden had zero occurrences of `dataName`, so it was locking the loss in.

Both anchors now go through `argumentExpression`, and `stringLiteralArgument` became `stringLiteralValue` over a resolved expression so the `captureSendDataEvent` walk shares the same rule rather than keeping the positional one:

```java
return argumentExpression(arguments, 1, AGENT_EVENT_NAME_PARAM)
        .map(CodeAnalyzer::stringLiteralValue)
        .orElse(null);
```

`durable_agent_box.json` now carries `"dataName": "hotelResults"` on both agent-call nodes — the `sendData` node from the named `eventName`, and the `waitForDataResult` node by tracing the named `token` back through `findSendDataEventForToken`, so both branches are exercised. The `arguments.size() < 2` guard went too, since `argumentExpression` covers it.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
rdulmina: Confirmed and fixed in f6faac1513. Both sites converted:

- `resolveBuiltinActivitySymbol` → `argumentExpression(args, 0, CALL_ACTIVITY_FUNCTION_PARAM)`
- `populateBuiltinActivityProperties` → `argumentExpression(args, 1, CALL_ACTIVITY_ARGS_PARAM)`, which also subsumes its `args.size() <= 1` guard

You are right that the second one had to move as well — the symbol resolving alone would have left it returning early on the mapping-constructor check.

No golden changed, because no fixture writes a builtin activity with named arguments. Flagging that rather than leaving it implied: the fix is uncovered. Happy to add a `callActivity(activityFunction = activity:callRestAPI, args = {...})` fixture if you would like the branch locked.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/Constants.java
rdulmina: Confirmed and fixed in f6faac1513 — `RUN_INPUT_PARAM` was declared and referenced nowhere. The block is now one line:

```java
Node valueNode = argumentExpression(args, 1, RUN_INPUT_PARAM).orElse(null);
```

Behaviour is unchanged, as you say: the old code let the named loop overwrite the positional read, and `argumentExpression` prefers positional — but a valid call cannot supply one parameter both ways, so the orders cannot diverge. `workflow_run_named_arg.json` still passes.

On the two constants for `"input"`: I kept them separate. They name parameters of two unrelated signatures — `workflow:run(processFunction, input)` and `DurableAgent.run(query, input)` — that merely share a spelling, and `Constants.Workflow` already does this deliberately for `"run"` itself (`RUN_METHOD_NAME`, `AGENT_OBJECT_RUN_METHOD_NAME`, `AGENT_START_METHOD_NAME`). Collapsing them would couple the two APIs so that renaming a parameter in one silently changes the other. Say the word if you would rather have one shared constant.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/test/java/io/ballerina/flowmodelgenerator/core/model/node/HumanTaskKeyFallbackTest.java
rdulmina: Confirmed and fixed in f6faac1513. You are right that it was vacuous — `get()` on a fresh map returns null for every key, so `presentKey` could have returned anything.

Replaced with the assertion on `presentKey`'s own return, as you suggested:

```java
Assert.assertEquals(HumanTaskBuilder.presentKey(new LinkedHashMap<>(), HumanTaskBuilder.DESCRIPTION_KEY),
        ESCAPED_DESCRIPTION);
```

Kept a comment recording why the old shape proved nothing, so it does not come back.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentPeerBuilder.java
```diff
@@ -231,6 +249,12 @@ public Map<Path, List<TextEdit>> toSource(SourceBuilder sourceBuilder) {
             entry.append(", 'wait: false");
             entry.append(", callbackChannel: ").append(WorkflowUtil.quoteIfPlain(callbackChannel));
```
pasindufernando1, line 255: `callbackChannel` is only emitted when `!waits`, but `PeerDecl` declares it independently and this PR makes it hydrate into the form. A peer with `callbackChannel` and the default `'wait = true` now loads the channel into the form and drops it on the next save.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1962,60 +2014,35 @@ private void populateResolvedHumanTaskProperties() {
      * matching {@link HumanTaskBuilder}'s fallback shape. The result type uses the inferred {@code T} key so
```
pasindufernando1: This Javadoc documents `populateFallbackHumanTaskProperties`, but the new method was inserted between them — it now sits above `fallbackHumanTaskArgumentValues` (which has its own) and the method at line 2038 is left undocumented. Move it down.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/SourceBuilder.java
```diff
@@ -164,7 +164,7 @@ private Path resolvePath(Path inputPath, NodeKind node, LineRange lineRange, Boo
                 case NEW_CONNECTION, MODEL_PROVIDER, EMBEDDING_PROVIDER, VECTOR_STORE, KNOWLEDGE_BASE,
                      DATA_LOADER, CHUNKER, CLASS_INIT -> CONNECTIONS_BAL;
                 case DATA_MAPPER_DEFINITION -> DATA_MAPPINGS_BAL;
-                case WORKFLOW, DURABLE_AGENT -> WORKFLOW_BAL;
+                case WORKFLOW, DURABLE_AGENT -> WORKFLOWS_BAL;
```
pasindufernando1, line 167: A project that already has a `workflow.bal` from an earlier build ends up with both files: existing constructs keep editing in place via `codedata.lineRange`, new ones go to `workflows.bal`. Fine if nobody's expected to carry one over — just flagging the silent split.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1044,9 +1054,9 @@ private void populateDurableAgentObjectRun(MethodCallExpressionNode callNode, Ex
...
         SeparatedNodeList<FunctionArgumentNode> runArguments = callNode.arguments();
         addAgentCallProperty(DurableAgentStartBuilder.QUERY_KEY, "Query",
-                "The initial user query for the agent", positionalArgumentSource(runArguments, 0));
+                "The initial user query for the agent", agentArgumentSource(runArguments, 0, AGENT_RUN_QUERY_PARAM));
         addAgentCallProperty(DurableAgentStartBuilder.INPUT_KEY, "Input",
-                "The structured input the run was given", positionalArgumentSource(runArguments, 1));
+                "The structured input the run was given", agentArgumentSource(runArguments, 1, AGENT_RUN_INPUT_PARAM));
 
         SyntaxKind parentKind = callNode.parent().kind();
         boolean hasCheck = parentKind == SyntaxKind.CHECK_ACTION || parentKind == SyntaxKind.CHECK_EXPRESSION;
```
pasindufernando1, line 1062: **Re line 1089**, `applyChildWorkflowMetadata` just below (GitHub won't let me anchor there — not in the diff):

`target` is still read positional-only, while `overrideSymbolFromFirstArg` was fixed to accept `childWorkflow = ...`. In the named form `target` is null, so the method returns before `metadata().description(target)` and the node loses the target-workflow subtitle. `workflow_retry_policy_parsing.json:902` in this PR shows the generic library description instead of `childFlow`. `argumentExpression(..., 0, CHILD_WORKFLOW_PARAM)` fixes it.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1346,11 +1363,8 @@ private void captureWorkflowName(ExpressionNode initializer) {
         }
```
pasindufernando1, line 1365: **Re line 1360**, `ChildWorkflowHandleFinder.captureWorkflowName` (same reason for the offset):

Same positional-only check. A handle from `runChildWorkflow(childWorkflow = ...)` can't be traced back, so later `waitForChildWorkflow` / `sendDataToChildWorkflow` nodes are labelled with the handle variable instead of the workflow.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1614,7 +1636,7 @@ private void collectDeclaredCapabilities(ExpressionNode listExpr, String capabil
 
     // Capability declaration fields whose values render in text-mode form fields.
     private static final Set<String> TEXT_MODE_CAPABILITY_FIELDS =
-            Set.of("name", "title", "description", "roles");
+            Set.of("name", "title", "description", "roles", "callbackChannel");
```
pasindufernando1: `userRoles` is missing here, so the peer's Reviewer Roles field hydrates unstripped — the new golden has `"userRoles": "\"travel-lead\""` where the human-task field, feeding an identical field, gets `travel-lead`. It round-trips untouched, but adding a second role in that box yields `userRoles: "a", "b"`, which isn't valid Ballerina. (Pre-existing for tool/activity, newly inherited by peer.)

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentPeerBuilder.java
rdulmina: Confirmed and fixed in daa94f3040. `PeerDecl` declares the two independently, and the module documents the channel as *required when* `wait = false` rather than exclusive to it:

```ballerina
boolean 'wait = true;
string callbackChannel?;
```

So `{agent: a, name: "x", callbackChannel: "ch"}` hydrated the channel into the form and the `!waits` branch dropped it on save — the same data-loss class this PR is about, newly reachable because the field now hydrates. `'wait: false` stays gated on the flag; the channel is emitted whenever it is set, so the existing gated-peer golden is byte-identical.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
rdulmina: Fixed in daa94f3040 — moved back down onto `populateFallbackHumanTaskProperties`. Thanks, that was mine: I extracted `fallbackHumanTaskArgumentValues` between the Javadoc and the method it documented and left the two stacked.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/SourceBuilder.java
rdulmina: Correct, and it is deliberate — flagging it was the right call, so recording the decision here.

The split is accepted: durable workflows are unreleased, so no project in the wild carries a `workflow.bal`, and we would rather not carry migration or fallback code for a file nobody has. Confirmed with @rdulmina before the rename landed.

If one ever does turn up, the fix would be to prefer an existing `workflow.bal` in `resolvePath` when the project already has one, rather than to write both. Not adding that now. Tracked in https://github.com/wso2/product-integrator/issues/2382.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
rdulmina: Confirmed and fixed in daa94f3040 — but the anchor was only half of it, so flagging the divergence.

`argumentExpression(..., 0, CHILD_WORKFLOW_PARAM)` is in. That alone did not restore the subtitle: the golden still showed the generic description. A probe confirmed the anchor was fine —

```
PROBE applyChildWorkflowMetadata method=runChildWorkflow label=Run Child Workflow
PROBE resolved target=childFlow
```

— so `metadata().description("childFlow")` ran and was then overwritten. `NodeBuilder.build()` re-invokes `setConstData()` after the analysis (`ActivityCallBuilder:171` documents this for the node kind), and all four child workflow builders assigned `metadata().label(LABEL).description(DESCRIPTION)` unconditionally. So the subtitle was clobbered on **every** path — the positional form never showed one either, which is why no golden in the repo has a child workflow subtitle at all.

`Metadata.Builder` gains `labelIfAbsent` / `descriptionIfAbsent`, and the four builders use them, so the constants only fill in what the analysis left unset. The palette listing is unaffected — `buildAvailableNode()` starts from a fresh builder, so the constants still apply; `AvailableNodesTest` is green.

`workflow_retry_policy_parsing.json:902` now reads `"description": "childFlow"`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
rdulmina: Confirmed and fixed in daa94f3040 — `captureWorkflowName` now resolves the same anchor through `argumentExpression`, so a handle from `runChildWorkflow(childWorkflow = ...)` traces back and the later wait/send nodes name the workflow rather than the handle variable.

Note the same `build()` metadata reset described on the sibling thread applied here too, so this needed the `descriptionIfAbsent` change to be observable at all — the wait and send builders are two of the four that were fixed.

No golden covers the wait/send pair yet (no fixture writes them against a named-argument run), so this half is verified by inspection. Happy to add a fixture if you want it locked.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
rdulmina: Confirmed and fixed in daa94f3040 — `userRoles` added to `TEXT_MODE_CAPABILITY_FIELDS`. The golden showed exactly the asymmetry you describe:

```
peers      -> "\"travel-lead\""
humanTasks -> "travel-lead"
```

the human task reaching the same field through the `roles` key, which was already in the set.

Your failure mode checks out: unstripped, the box holds `"travel-lead"`, and a second role makes it `"travel-lead", "manager"` — which `quoteIfPlain` passes through untouched (already quote-delimited), emitting `userRoles: "travel-lead", "manager"` inside the record literal. Not a valid field.

Stripping is safe for the other shapes: `stripQuotes` only removes *surrounding* quotes, so `financeRoles` and `["a", "b"]` pass through unchanged, and on the way out `quoteIfPlain` leaves a `[`-prefixed value alone and re-quotes a bare role. Fixed for tool and activity at the same time, since all three read the same source field.

NipunaRanasinghe (approved):
Approving. All five findings from my earlier review are fixed and verified.


Two open questions, neither blocking:

**`userRoles` hydration when the source value is a reference.** Adding `userRoles` to `TEXT_MODE_CAPABILITY_FIELDS` handles literals and lists correctly. A peer declared `userRoles: financeRoles` strips nothing, and `addRoleFieldTypes` marks neither mode `selected` — so unless something downstream selects EXPRESSION, `roleSource` falls through to `quoteIfBareRole` and writes `"financeRoles"` back on save. `durable_agent_peer_gated_expression_roles_to_source.json` covers `toSource` with the mode already selected, not the hydration side. Is there a path that sets it?

**The `awaitHumanTask` positional order.** `fallbackHumanTaskArgumentValues` maps positional 0–5 to `taskName, userRoles, payload, title, description, timeout`, and `HumanTaskFallbackArgumentsTest` asserts that order against itself. The library signature isn't in this repo, so nothing here proves positional 3 is `title` rather than `description` — if it isn't, the fallback rewrites the call on save. Worth checking against the workflow module.

PR #858, opened 2026-08-13
Schedule sync branch with main

samithkavishke:
## Purpose

Release work lands on `staging/*` and has to flow back to `main`. Today that is done by hand, so `main` drifts behind the active staging branch and the eventual catch-up merge is large and conflict-prone.

## Goals

Keep `main` continuously up to date with the active staging branch, via one sync PR per staging branch that is updated in place each night rather than a new PR piling up alongside yesterday's unmerged one.

## Approach

Adds `.github/workflows/sync-staging-to-main.yml`, running on an 18:00 cron plus `workflow_dispatch`.

For each `staging/*` branch:

- A **deterministic** temp branch, `sync/main-<staging-branch>` — no timestamp. Re-pushing it updates the existing open PR's diff instead of opening a second PR.
- If an open sync PR already exists, `origin/$stagingBranch` is **merged into** the temp branch rather than the branch being reset to the staging head, so conflict resolution pushed to the sync branch survives the next run. On conflict the merge is aborted (with `reset --hard` / `clean -fd` as a self-heal) and the branch is skipped with a warning for a human to resolve.
- Branches with nothing ahead of `main` are skipped, and an open PR that has become empty is closed. This is an ancestry check (`git rev-list --count origin/main..$tempBranch`), not a content diff — once a sync PR is merged and `main` gains any further commit, the staging head is an ancestor of `main` while the trees still differ, and `gh pr create` rejects that with a 422.
- Pushes use `--force-with-lease`.
- Each branch is processed in its own subshell, so one failure does not abort the rest of the loop; the job fails at the end if any branch failed.

A `concurrency` group serializes a manual run against the scheduled one, since both would otherwise force-push the same temp branches.

## UI Component Development

N/A — CI-only change, no UI surface.

## User stories

N/A — internal release-engineering automation, no user-facing behaviour.

## Release note

N/A — no product change.

## Documentation

N/A — no product doc impact. The workflow's behaviour is documented in comments in the workflow file itself.

## Training

N/A — no product change.

## Certification

N/A — no product change, nothing certifiable.

## Marketing

N/A — internal automation.

## Automation tests

- Unit tests
  > N/A — a GitHub Actions workflow, not application code.
- Integration tests
  > Verified by `workflow_dispatch`. The nothing-to-sync guard was additionally verified against scratch repositories covering: staging fully merged with `main` ahead (skips), a genuinely new staging commit (proceeds), fully in sync (skips), and commits ahead with no net change (skips). The YAML parses and the extracted `run` block passes `bash -n`.

## Security checks

- Ran FindSecurityBugs plugin and verified report? **N/A** — no Java/application code in this change.
- Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? **yes**

Additionally: the job uses the built-in `github.token` with `contents: write` and `pull-requests: write` only, and all actions are pinned by commit SHA.

## Samples

N/A.

## Related PRs

None. Note for reviewers: GitHub does not fire `pull_request` workflow triggers for PRs opened with the built-in `GITHUB_TOKEN`, so `pull-request.yml` (the build and tests) **does not run on the sync PRs this workflow creates** — an absent check list on those PRs is not a passing one. This matches the existing behaviour of `sync-main-with-releases.yml`; changing it would require a PAT or GitHub App token and is tracked separately.

## Migrations (if applicable)

N/A — no migration.

## Test environment

GitHub Actions `ubuntu-latest`. No JDK or OS matrix applies; the workflow only runs `git` and `gh`.

## Learning

Two behaviours drove the design: `gh pr create` rejects a head branch with no commits ahead of the base (HTTP 422), which is an ancestry question and not the same as a content diff; and `git for-each-ref` matches with `WM_PATHNAME`, so `staging/*` does not cross a `/` — fine under the current `staging/<version>` convention.

Adds a scheduled and manual GitHub Actions workflow to synchronize `staging/*` branches with `main`.

The workflow creates or reuses deterministic temporary branches and pull requests. It skips branches with no changes, closes stale pull requests, and processes each staging branch independently.

### DIFF
--- a/.github/workflows/sync-staging-to-main.yml
+++ b/.github/workflows/sync-staging-to-main.yml
@@ -71,7 +71,10 @@ jobs:
                 git checkout -B "$tempBranch" "origin/$stagingBranch"
               fi
 
-              if git diff --quiet "origin/main" "$tempBranch" --; then
+              # Ancestry, not just contents: main moving ahead on its own makes the trees
+              # differ while staging has nothing to sync, and gh pr create 422s on that.
+              if [ "$(git rev-list --count "origin/main..$tempBranch")" -eq 0 ] \
+                || git diff --quiet "origin/main" "$tempBranch" --; then
                 if [ -n "$existingPr" ]; then
                   gh pr close "$existingPr" --comment "Closing: \`$stagingBranch\` has no changes left to sync into \`main\`." || true
                   echo "::notice::$stagingBranch has nothing new for main; closed stale PR $existingPr."
--- a/.github/workflows/sync-staging-to-main.yml
+++ b/.github/workflows/sync-staging-to-main.yml
@@ -1,0 +1,108 @@
+name: Sync Staging Branches with Main
+on:
+  schedule:
+    - cron: '0 18 * * *'
+  workflow_dispatch: {}
+
+concurrency:
+  group: sync-staging-to-main
+  cancel-in-progress: false
+
+jobs:
+  sync-staging:
+    if: github.repository == 'ballerina-platform/ballerina-vscode'
+    permissions:
+      contents: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          fetch-depth: 0
+
+      - name: Configure Git
+        run: |
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config --global user.name "github-actions[bot]"
+
+      # A deterministic temp branch per staging branch (no timestamp) is reused night
+      # over night: force-pushing it just updates the existing open PR's diff instead
+      # of a new PR piling up alongside yesterday's unmerged one.
+      - name: Sync each staging branch into main
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+
+          stagingBranches=$(git for-each-ref --format='%(refname:short)' 'refs/remotes/origin/staging/*' | sed 's|^origin/||')
+
+          if [ -z "$stagingBranches" ]; then
+            echo "::notice::No staging/* branches found; nothing to sync."
+            exit 0
+          fi
+
+          failed=0
+          for stagingBranch in $stagingBranches; do
+            set +e
+            (
+              set -euo pipefail
+
+              tempBranch="sync/main-${stagingBranch//\//-}"
+
+              echo "::group::Syncing $stagingBranch via $tempBranch"
+
+              existingPr=$(gh pr list -H "$tempBranch" -B main --state open --json url --jq '.[0].url // empty')
+
+              if [ -n "$existingPr" ] && git show-ref --verify --quiet "refs/remotes/origin/$tempBranch"; then
+                # Merge instead of reset so conflict-resolution work already pushed to the
+                # open sync PR survives, rather than being force-pushed away every run.
+                git checkout -B "$tempBranch" "origin/$tempBranch"
+                if ! git merge --no-edit "origin/$stagingBranch"; then
+                  git merge --abort || true
+                  # Self-heals even if the abort above failed, so a bad state can't survive into the next iteration.
+                  git reset --hard
+                  git clean -fd
+                  echo "::warning::Merge conflict syncing $stagingBranch into $tempBranch; skipping (resolve manually on the sync branch)."
+                  echo "::endgroup::"
+                  exit 0
+                fi
+              else
+                git checkout -B "$tempBranch" "origin/$stagingBranch"
+              fi
+
+              if git diff --quiet "origin/main" "$tempBranch" --; then
+                if [ -n "$existingPr" ]; then
+                  gh pr close "$existingPr" --comment "Closing: \`$stagingBranch\` has no changes left to sync into \`main\`." || true
+                  echo "::notice::$stagingBranch has nothing new for main; closed stale PR $existingPr."
+                else
+                  echo "::notice::$stagingBranch has nothing new for main; skipping."
+                fi
+                echo "::endgroup::"
+                exit 0
+              fi
+
+              git push --force-with-lease origin "$tempBranch"
+
+              if [ -n "$existingPr" ]; then
+                echo "::notice::Reusing existing sync PR: $existingPr"
+              else
+                pr=$(gh pr create -B main -H "$tempBranch" \
+                  --title "Sync ${stagingBranch} into main" \
+                  --body "Nightly sync of \`${stagingBranch}\` into \`main\`. Automated PR — updated in place each night this runs.")
+                echo "::notice::Opened sync PR: $pr"
+              fi
+
+              echo "::endgroup::"
+            )
+            status=$?
+            set -e
+            if [ "$status" -ne 0 ]; then
+              echo "::warning::Sync failed for $stagingBranch"
+              failed=1
+            fi
+          done
+
+          if [ "$failed" -ne 0 ]; then
+            exit 1
+          fi


.github/workflows/sync-staging-to-main.yml
```diff
@@ -0,0 +1,60 @@
...
+
+          stagingBranches=$(git for-each-ref --format='%(refname:short)' 'refs/remotes/origin/staging/*' | sed 's|^origin/||')
+
+          if [ -z "$stagingBranches" ]; then
+            echo "::notice::No staging/* branches found; nothing to sync."
+            exit 0
+          fi
+
+          for stagingBranch in $stagingBranches; do
+            tempBranch="sync/main-${stagingBranch//\//-}"
```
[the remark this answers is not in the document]
samithkavishke: currently our strategy for the staging branch's are staging/x.y.z ( ex: staging/5.1.0, ...) and ideally we should not have two staging/* branch's at once so no need to consider this

.github/workflows/sync-staging-to-main.yml
```diff
@@ -0,0 +1,60 @@
...
+
+            echo "::group::Syncing $stagingBranch via $tempBranch"
+
+            git checkout -B "$tempBranch" "origin/$stagingBranch"
+            git push --force origin "$tempBranch"
+
+            pr=$(gh pr list -H "$tempBranch" -B main --state open --json url --jq '.[0].url // empty')
+            if [ -z "$pr" ]; then
+              pr=$(gh pr create -B main -H "$tempBranch" \
```
NipunaRanasinghe: **Blocking:** the first run after a sync PR merges fails the whole job.

Once the sync PR is merged, `sync/main-staging-5.14.0` still exists and now points at a commit that is an ancestor of `main`. On the next run the force-push is a no-op, `gh pr list --state open` returns empty, and `gh pr create` gets a 422 `No commits between main and sync/main-staging-5.14.0`. `gh` exits non-zero, `set -e` kills the job. The same thing happens on any night `staging/*` has nothing new relative to `main`.

Guard before creating:

```bash
if [ "$(git rev-list --count "origin/main..$tempBranch")" -eq 0 ]; then
  echo "::notice::$stagingBranch has nothing new for main; skipping."
  echo "::endgroup::"
  continue
fi
```

.github/workflows/sync-staging-to-main.yml
```diff
@@ -0,0 +1,60 @@
...
+          set -euo pipefail
+
+          stagingBranches=$(git for-each-ref --format='%(refname:short)' 'refs/remotes/origin/staging/*' | sed 's|^origin/||')
+
+          if [ -z "$stagingBranches" ]; then
+            echo "::notice::No staging/* branches found; nothing to sync."
+            exit 0
+          fi
+
+          for stagingBranch in $stagingBranches; do
```
NipunaRanasinghe, line 46: **Blocking:** branches are not processed independently.

With `set -e` in effect, the first branch that fails any command aborts the loop, so every remaining `staging/*` branch is skipped silently for that run. Wrap the per-branch body in a subshell so one failure doesn't take out the rest:

```bash
for stagingBranch in $stagingBranches; do
  (
    ...
  ) || echo "::warning::sync failed for $stagingBranch"
done
```

.github/workflows/sync-staging-to-main.yml
```diff
@@ -0,0 +1,60 @@
...
+            tempBranch="sync/main-${stagingBranch//\//-}"
+
+            echo "::group::Syncing $stagingBranch via $tempBranch"
+
+            git checkout -B "$tempBranch" "origin/$stagingBranch"
+            git push --force origin "$tempBranch"
```
NipunaRanasinghe: The force-push discards conflict resolution on the sync branch.

`checkout -B` resets `$tempBranch` to the staging head, so if the sync PR conflicts with `main` and someone resolves it on the sync branch (or merges `main` into it), the 18:00 run overwrites that work and the PR goes back to conflicted. As written a conflicting sync PR can never be brought to a mergeable state that survives the night.

The comment above explains the force-push as the way to avoid stacking up PRs, which it does. The gap is only the conflict case. One option: when an open PR already exists and `$tempBranch` has commits not in `origin/$stagingBranch`, merge `origin/$stagingBranch` in instead of resetting.

.github/workflows/sync-staging-to-main.yml
```diff
@@ -0,0 +1,60 @@
...
+        run: |
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config --global user.name "github-actions[bot]"
+
+      # A deterministic temp branch per staging branch (no timestamp) is reused night
+      # over night: force-pushing it just updates the existing open PR's diff instead
+      # of a new PR piling up alongside yesterday's unmerged one.
+      - name: Sync each staging branch into main
+        env:
+          GH_TOKEN: ${{ github.token }}
```
NipunaRanasinghe, line 34: PRs opened with `github.token` do not trigger `pull_request` workflows, so `pull-request.yml` will not run on these sync PRs. Whatever comes from `staging/*` reaches `main` without the PR build or tests having run on the merged result.

This is the same in the existing `sync-main-with-releases.yml`, so it is not new behaviour introduced here, but it does mean the sync PR is only a manual review step. Fixing it needs a PAT or a GitHub App token in place of `github.token`.

.github/workflows/sync-staging-to-main.yml
```diff
@@ -0,0 +1,60 @@
+name: Sync Staging Branches with Main
+on:
+  schedule:
+    - cron: '0 18 * * *'
+  workflow_dispatch: {}
```
NipunaRanasinghe, line 5: Add a `concurrency` group. A `workflow_dispatch` run overlapping the scheduled run force-pushes the same temp branches at the same time.

```yaml
concurrency:
  group: sync-staging-to-main
  cancel-in-progress: false
```

.github/workflows/sync-staging-to-main.yml
```diff
@@ -0,0 +1,60 @@
...
+      # A deterministic temp branch per staging branch (no timestamp) is reused night
+      # over night: force-pushing it just updates the existing open PR's diff instead
+      # of a new PR piling up alongside yesterday's unmerged one.
+      - name: Sync each staging branch into main
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+
+          stagingBranches=$(git for-each-ref --format='%(refname:short)' 'refs/remotes/origin/staging/*' | sed 's|^origin/||')
```
NipunaRanasinghe, line 38: Minor: `for-each-ref` matches patterns with `WM_PATHNAME`, so `*` does not cross a `/`. A nested branch name is skipped with no output. Verified locally:

```
$ git for-each-ref --format='%(refname:short)' 'refs/heads/staging/*'
staging/5.14.0

$ git for-each-ref --format='%(refname:short)' 'refs/heads/staging/**'
staging/5.14.0
staging/5.14/hotfix
```

Fine for the current `staging/<version>` convention. Use `staging/**` if that ever changes. Note that with `**` the `${stagingBranch//\//-}` substitution maps `staging/a/b` and `staging/a-b` to the same temp branch.

.github/workflows/sync-staging-to-main.yml
```diff
@@ -0,0 +1,60 @@
...
+              pr=$(gh pr create -B main -H "$tempBranch" \
+                --title "Sync ${stagingBranch} into main" \
+                --body "Nightly sync of \`${stagingBranch}\` into \`main\`. Automated PR — updated in place each night this runs.")
+              echo "::notice::Opened sync PR: $pr"
+            else
+              echo "::notice::Reusing existing sync PR: $pr"
+            fi
+
+            echo "::endgroup::"
+          done
```
NipunaRanasinghe, line 107: Two smaller things at the end of the loop:

- Deleting a `staging/*` branch leaves its `sync/main-staging-*` branch in the repo permanently. Worth a cleanup pass over `sync/main-staging-*` branches whose source branch no longer exists.
- `sync-main-with-releases.yml` posts the PR URL to `EDITOR_TEAM_CHAT_API` after creating its sync PR. This workflow does not, so a nightly sync PR only surfaces if someone is watching the PR list. Consider reusing the same notification step for consistency.

NipunaRanasinghe (changes requested):

.github/workflows/sync-staging-to-main.yml
NipunaRanasinghe: The temp branch name reads in the wrong direction. `sync/main-staging-5.14.0` puts `main` first, which suggests `main` is the source, but the source is `staging/5.14.0` and `main` is the target. It also leaves `main-staging` next to each other in the middle of the name with nothing marking which is which.

Suggest source first, target last:

```bash
tempBranch="sync/${stagingBranch//\//-}-to-main"
```

which gives `sync/staging-5.14.0-to-main`.

One note: `sync-main-with-releases.yml` already produces `sync/main-<release>-<timestamp>`, so this makes the two workflows inconsistent. Either rename both or keep the current form here — worth deciding rather than ending up with one of each.

.github/workflows/sync-staging-to-main.yml
samithkavishke: Confirmed and fixed in 3d993364ec.

The root cause was that the existing guard compared trees (`git diff --quiet`) rather than commit ancestry, and those diverge in exactly this scenario: once the sync PR is merged and `main` gains any further commit, the staging head is an ancestor of `main` (`rev-list --count main..staging` = 0) while the trees still differ, so the guard fell through to `gh pr create` and hit the 422.

Rather than adding a separate pre-create check, the guard itself now asks the same question `gh` does:

```bash
if [ "$(git rev-list --count "origin/main..$tempBranch")" -eq 0 ] \
  || git diff --quiet "origin/main" "$tempBranch" --; then
```

The `diff` check is kept as a second condition for the narrow case where the branch is ahead but contributes no net change (a commit and its revert both synced), which `gh` would otherwise accept and open an empty-diff PR for.

Verified against scratch repos: staging fully merged with `main` ahead now skips; a genuinely new staging commit still proceeds; fully-in-sync skips. This also fixes the stale-PR-close path, which previously would not have closed a PR that became empty because it was merged.

.github/workflows/sync-staging-to-main.yml
samithkavishke: Addressed — the per-branch body is now isolated in a subshell (lines 46-107):

```bash
for stagingBranch in $stagingBranches; do
  set +e
  (
    set -euo pipefail
    ...
  )
  status=$?
  set -e
  if [ "$status" -ne 0 ]; then
    echo "::warning::Sync failed for $stagingBranch"
    failed=1
  fi
done

if [ "$failed" -ne 0 ]; then
  exit 1
fi
```

Each iteration gets its own `set -euo pipefail`, so a failure aborts only that branch; the remaining `staging/*` branches still run, and the job fails at the end rather than silently skipping them. This also covers CodeRabbit's "make per-branch failures fail closed" nitpick, which was raised against an earlier `(...) || echo ...` form that swallowed the exit status entirely — the current form both continues past a failure *and* propagates it.

One deliberate deviation worth flagging: the merge-conflict path `exit 0`s with a `::warning::` instead of failing. A conflict needs a human to resolve on the sync branch, and failing the nightly job every night until they do would just train people to ignore it. If you would rather a conflict go red, happy to change it.

.github/workflows/sync-staging-to-main.yml
samithkavishke: Addressed — when an open sync PR already exists, the temp branch is merged into rather than reset (lines 57-69):

```bash
if [ -n "$existingPr" ] && git show-ref --verify --quiet "refs/remotes/origin/$tempBranch"; then
  git checkout -B "$tempBranch" "origin/$tempBranch"
  if ! git merge --no-edit "origin/$stagingBranch"; then
    git merge --abort || true
    git reset --hard
    git clean -fd
    echo "::warning::Merge conflict syncing $stagingBranch into $tempBranch; skipping (resolve manually on the sync branch)."
    echo "::endgroup::"
    exit 0
  fi
else
  git checkout -B "$tempBranch" "origin/$stagingBranch"
fi
```

So conflict resolution pushed to the sync branch survives the next run: the branch is only reset from the staging head when there is no open PR to preserve. The `reset --hard` / `clean -fd` after the abort is a self-heal in case `merge --abort` itself fails, so a dirty worktree cannot leak into the next loop iteration.

Also closing out CodeRabbit's related nitpick on the same push, since it is the same line: line 88 is now `git push --force-with-lease origin "$tempBranch"`. I kept the bare form rather than the explicit `--force-with-lease=refs/heads/$b:$sha` it suggested — nothing re-fetches inside the loop, so the remote-tracking ref set at checkout is a valid lease, and the `concurrency` group (lines 7-9) already serializes the scheduled and manual runs that were the concern.

.github/workflows/sync-staging-to-main.yml
samithkavishke: Added at lines 7-9, as suggested — `group: sync-staging-to-main`, `cancel-in-progress: false` so a queued run waits rather than being dropped.

.github/workflows/sync-staging-to-main.yml
samithkavishke: Confirmed with the exact expression used here (`refs/remotes/origin/staging/*`), not just the `refs/heads/` illustration — a nested `staging/5.14/hotfix` is indeed skipped.

Keeping `*` though: our staging branches are `staging/x.y.z` and only one exists at a time. `.` is not `/`, so `staging/5.14.0` matches fine, and the nested shape never occurs.

PR #861, opened 2026-08-13
Add scheduled E2E workflow with flakiness reporting

samithkavishke:
## Purpose
E2E tests were disabled from the nightly build on 2026-08-02 due to flakiness. Add a standalone workflow that runs the Playwright suite every 6 hours against the same staging/main branch the nightly build targets, independent of the release pipeline.

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary

- Add a standalone Playwright E2E workflow that runs every six hours against the nightly VSIX.
- Run four E2E groups in parallel with retries and failed-test reruns.
- Aggregate JSON results and publish flakiness and skip-cause metrics.
- Persist E2E history to the canonical repository with duplicate detection.
- Add reusable actions for source-branch resolution and E2E execution.
- Update nightly workflows to use shared branch resolution and reusable E2E execution.
- Generate JSON Playwright reports for automated processing.
- Update workflow documentation and ignore generated E2E reports.

### DIFF
--- a/.github/actions/resolve-source-branch/action.yml
+++ b/.github/actions/resolve-source-branch/action.yml
@@ -2,6 +2,8 @@ name: 'Resolve source branch'
 description: >-
   Resolve the branch a nightly-style pipeline should run against: an explicit
   override if given, otherwise the latest staging/* branch if one exists, else main.
+  Runs 'git ls-remote --heads origin', so the calling job must check out the
+  repository (any ref) before this step, to have the 'origin' remote configured.
 inputs:
   sourceBranch:
     description: 'Explicit branch to use. Leave blank to auto-select.'
--- a/.github/actions/resolve-source-branch/action.yml
+++ b/.github/actions/resolve-source-branch/action.yml
@@ -1,0 +1,38 @@
+name: 'Resolve source branch'
+description: >-
+  Resolve the branch a nightly-style pipeline should run against: an explicit
+  override if given, otherwise the latest staging/* branch if one exists, else main.
+inputs:
+  sourceBranch:
+    description: 'Explicit branch to use. Leave blank to auto-select.'
+    required: false
+    default: ''
+outputs:
+  branch:
+    description: 'The resolved branch name'
+    value: ${{ steps.resolve.outputs.branch }}
+runs:
+  using: "composite"
+  steps:
+    - id: resolve
+      shell: bash
+      env:
+        INPUT_SOURCE_BRANCH: ${{ inputs.sourceBranch }}
+      run: |
+        set -euo pipefail
+
+        if [ -n "${INPUT_SOURCE_BRANCH}" ]; then
+          SOURCE_BRANCH="${INPUT_SOURCE_BRANCH}"
+        else
+          # No explicit source given: prefer the latest staging/* branch (e.g.
+          # staging/5.14.0) if one exists, so callers follow the active release
+          # staging branch without a workflow edit each time one is cut. Falls
+          # back to main once no staging/* branch remains.
+          latest_staging=$(git ls-remote --heads origin \
+            | awk -F'refs/heads/' '/refs\/heads\/staging\// { print $2 }' \
+            | sort -t/ -k2 -V | tail -n 1)
+          SOURCE_BRANCH="${latest_staging:-main}"
+        fi
+
+        SOURCE_BRANCH="$(git check-ref-format --branch "${SOURCE_BRANCH}")"
+        echo "branch=${SOURCE_BRANCH}" >> "$GITHUB_OUTPUT"
--- a/.github/actions/run-e2e-group/action.yml
+++ b/.github/actions/run-e2e-group/action.yml
@@ -1,0 +1,97 @@
+name: 'Run E2E test group'
+description: >-
+  Runs one matrix group of the Ballerina e2e Playwright suite (first attempt, then a
+  --last-failed re-run if needed) and uploads its results/recordings/project-data
+  artifacts. Shared by reusable-build.yml's BalE2ETest job and e2e-scheduled.yml's E2E
+  job, which differ only in how they prepare the workspace beforehand (a full
+  ExtBuild restore vs installing the nightly VSIX) — this action starts from the
+  assumption that workspace is already prepared (deps installed, Ballerina
+  distribution installed) in the calling job.
+inputs:
+  group:
+    description: 'Matrix group name, e.g. group1'
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Check for previous test results
+      shell: bash
+      if: github.run_attempt > 1
+      run: |
+        PREVIOUS_ATTEMPT=$((${{ github.run_attempt }} - 1))
+        echo "PREVIOUS_ATTEMPT=$PREVIOUS_ATTEMPT" >> $GITHUB_ENV
+
+    - name: Download previous test results
+      id: check-results
+      if: github.run_attempt > 1
+      uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4
+      continue-on-error: true
+      with:
+        name: Ballerina-e2e-test-results-linux-${{ inputs.group }}-${{ env.PREVIOUS_ATTEMPT }}
+        path: packages/ballerina-extension/test-results
+
+    - name: Install packages
+      shell: bash
+      run: |
+        sudo apt-get update -y
+        sudo apt-get install -y xvfb ffmpeg gnome-keyring libsecret-1.0 dbus-x11
+        cd packages/ballerina-extension && npx playwright install
+
+    - name: Run Ballerina e2e Test (First Attempt)
+      shell: bash
+      id: run-tests-first
+      if: github.run_attempt == 1 || steps.check-results.outcome == 'failure'
+      continue-on-error: true
+      env:
+        # A distinct filename from the re-run's report below: both are full Playwright
+        # JSON reports written by separate CLI invocations, so a shared name would let
+        # the re-run silently overwrite the first attempt's results instead of adding
+        # to them — aggregate-e2e-results.js merges both by matching this prefix.
+        PLAYWRIGHT_JSON_OUTPUT_FILE: test-results/e2e-results-first.json
+      run: |
+        export $(dbus-launch)
+        export DISPLAY=:98.0
+        cd packages/ballerina-extension
+        xvfb-run --server-num 98.0 -s "-ac -screen 0 1920x1080x24" bash -c "ffmpeg -y -nostdin -f x11grab -r 30 -s 1920x1080 -i :98.0 -c:v libx264 -preset ultrafast -pix_fmt yuv420p -movflags +faststart -hide_banner -loglevel error record.mp4 & pnpm exec playwright test -c e2e-test/playwright.config.js --grep ${{ inputs.group }}"
+
+    - name: Run Failed Ballerina e2e Test (Re-run)
+      shell: bash
+      if: steps.run-tests-first.outcome == 'failure' || steps.check-results.outcome == 'success'
+      env:
+        PLAYWRIGHT_JSON_OUTPUT_FILE: test-results/e2e-results-rerun.json
+      run: |
+        rm -rf packages/ballerina-extension/e2e-test/data/new-project
+        export $(dbus-launch)
+        export DISPLAY=:98.0
+        cd packages/ballerina-extension
+        xvfb-run --server-num 98.0 -s "-ac -screen 0 1920x1080x24" bash -c "ffmpeg -y -nostdin -f x11grab -r 30 -s 1920x1080 -i :98.0 -c:v libx264 -preset ultrafast -pix_fmt yuv420p -movflags +faststart -hide_banner -loglevel error record.mp4 & pnpm exec playwright test -c e2e-test/playwright.config.js --grep ${{ inputs.group }} --last-failed"
+
+    - name: Upload Ballerina e2e Test results
+      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+      if: always()
+      with:
+        name: Ballerina-e2e-test-results-linux-${{ inputs.group }}-${{ github.run_attempt }}
+        path: packages/ballerina-extension/test-results/**
+        retention-days: 5
+        include-hidden-files: true
+
+    - name: Upload Ballerina e2e Test recordings
+      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+      if: always()
+      with:
+        name: Ballerina-e2e-test-recording-linux-${{ inputs.group }}-${{ github.run_attempt }}
+        path: |
+          packages/ballerina-extension/e2e-test/test-resources/videos/**
+          packages/ballerina-extension/e2e-test/test-resources/screenshots/**
+          packages/ballerina-extension/e2e-test/test-resources/snapshots/**
+          packages/ballerina-extension/record.mp4
+        retention-days: 5
+
+    - name: Upload Ballerina e2e Test project data
+      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+      if: always()
+      with:
+        name: Ballerina-e2e-test-project-linux-${{ inputs.group }}-${{ github.run_attempt }}
+        path: |
+          packages/ballerina-extension/e2e-test/data/new-project/**
+        retention-days: 5
--- a/.github/actions/run-e2e-group/action.yml
+++ b/.github/actions/run-e2e-group/action.yml
@@ -3,10 +3,9 @@ description: >-
   Runs one matrix group of the Ballerina e2e Playwright suite (first attempt, then a
   --last-failed re-run if needed) and uploads its results/recordings/project-data
   artifacts. Shared by reusable-build.yml's BalE2ETest job and e2e-scheduled.yml's E2E
-  job, which differ only in how they prepare the workspace beforehand (a full
-  ExtBuild restore vs installing the nightly VSIX) — this action starts from the
-  assumption that workspace is already prepared (deps installed, Ballerina
-  distribution installed) in the calling job.
+  job — see "Scheduled E2E testing" in .github/workflows/README.md for the design
+  rationale. Assumes the workspace is already prepared (deps installed, Ballerina
+  distribution installed) by the calling job.
 inputs:
   group:
     description: 'Matrix group name, e.g. group1'
@@ -28,20 +27,7 @@ runs:
       continue-on-error: true
       with:
         name: Ballerina-e2e-test-results-linux-${{ inputs.group }}-${{ env.PREVIOUS_ATTEMPT }}
-        # Extracts to packages/ballerina-extension, not straight into test-results/: the
-        # upload step below now spans two path roots (test-results/ and e2e-reports/), so
-        # upload-artifact roots the artifact at their common ancestor instead of flattening
-        # test-results/'s own contents to the artifact root.
-        # e2e-reports/ from a restored previous attempt is kept, deliberately, not
-        # discarded: a GitHub-level "re-run failed jobs" is a continuation of the same
-        # logical test run, so a test's full attempt history across that continuation
-        # (attempt N-1's failures plus attempt N's retry) is the correct thing to
-        # record — the same principle the first-attempt/re-run merge within one
-        # attempt already relies on. Discarding it instead (tried once, reverted) means
-        # the Report job — which only downloads the highest-numbered artifact per
-        # group — sees *only* the small --last-failed subset from the latest attempt,
-        # silently losing every test that already passed earlier and erasing the
-        # failed-then-passed history for exactly the flaky tests this exists to catch.
+        # Restored e2e-reports/ is deliberately kept, not discarded — see README.
         path: packages/ballerina-extension
 
     - name: Install packages
@@ -57,11 +43,7 @@ runs:
       if: github.run_attempt == 1 || steps.check-results.outcome == 'failure'
       continue-on-error: true
       env:
-        # Not under test-results/: that's Playwright's own outputDir, which the re-run
-        # below wipes at its own startup — a report placed there would be deleted before
-        # ever being read, regardless of filename (verified locally). e2e-reports/ is
-        # never touched by that cleanup. The distinct filename from the re-run's report
-        # lets aggregate-e2e-results.js merge both instead of one replacing the other.
+        # Not under test-results/ (Playwright's outputDir, wiped every invocation).
         PLAYWRIGHT_JSON_OUTPUT_FILE: e2e-reports/e2e-results-first.json
       run: |
         export $(dbus-launch)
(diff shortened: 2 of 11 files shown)


packages/ballerina-extension/e2e-test/playwright.config.js
```diff
@@ -9,7 +9,7 @@ exports.default = (0, test_1.defineConfig)({
     retries: process.env.BI_E2E_RETRIES ? Number(process.env.BI_E2E_RETRIES) : 2,
     maxFailures: 10,
     workers: 1,
-    reporter: 'html',
+    reporter: [['html'], ['json', { outputFile: 'test-results/e2e-results.json' }]],
```
NipunaRanasinghe: **Blocking:** this file is written outside the path the workflow uploads, so the report is empty on every run.

Playwright resolves an explicit reporter `outputFile` against the **config** directory, not the cwd — `resolveOutputFile` in `playwright/lib/reporters/base.js`:

```js
if (!outputFile && options.outputFile)
  outputFile = path.resolve(options.configDir, options.outputFile);
```

configDir here is `packages/ballerina-extension/e2e-test`, so the JSON lands at `packages/ballerina-extension/e2e-test/test-results/e2e-results.json`. But `reusable-build.yml` uploads `packages/ballerina-extension/test-results/**`. Two different `test-results` directories.

The default `outputDir` resolves differently — it walks up to the nearest `package.json` — which is why it ends up at `packages/ballerina-extension/test-results`. Confirmed against a real artifact: `Ballerina-e2e-test-results-linux-group1-1` from run 31676542446 contains exactly `.last-run.json` and one trace directory, nothing from `e2e-test/`.

Result: `findResultFiles` returns `[]`, the aggregator prints `Groups reported: 0 · Total: 0` and "No retries or failures — every test passed on the first attempt", exits 0, writes no NDJSON, and the persist step logs "No test results to persist". A green report on every run, including total failures.

```suggestion
    reporter: [['html'], ['json', { outputFile: '../test-results/e2e-results.json' }]],
```

.github/scripts/aggregate-e2e-results.js
```diff
@@ -0,0 +1,169 @@
...
+const path = require('path');
+
+function findResultFiles(rootDir) {
+  const found = [];
+  const walk = (dir) => {
+    for (const entry of fs.readdirSync(dir, { withFileTypes: true })) {
+      const full = path.join(dir, entry.name);
+      if (entry.isDirectory()) {
+        walk(full);
+      } else if (entry.isFile() && entry.name === 'e2e-results.json') {
```
NipunaRanasinghe: **Blocking:** only one JSON survives per group, and on failing runs it is the wrong one.

`reusable-build.yml:337` runs Playwright a second time with `--last-failed` against the same config, so the JSON reporter rewrites the same `e2e-results.json` with a report covering only the re-run subset. On exactly the runs this workflow exists to analyse — the ones where the first attempt failed — `total`, `passed` and `flaky` collapse to that subset and every test that passed on the first attempt disappears from the counts.

`PLAYWRIGHT_JSON_OUTPUT_FILE` is checked before the config value, so the two steps can be given distinct filenames. The aggregator then needs to match on a pattern rather than the exact name here, and merge `results` per test across both files so the attempt count spans both invocations.

.github/workflows/e2e-scheduled.yml
```diff
@@ -0,0 +1,119 @@
...
+        continue-on-error: true
+        with:
+          pattern: Ballerina-e2e-test-results-linux-*-${{ github.run_attempt }}
+          path: e2e-results
+
+      - name: Aggregate results
+        id: aggregate
+        env:
+          E2E_SOURCE_BRANCH: ${{ needs.resolve-branch.outputs.branch }}
+        run: node .github/scripts/aggregate-e2e-results.js e2e-results e2e-run-stats.jsonl | tee -a "$GITHUB_STEP_SUMMARY"
```
NipunaRanasinghe: **Blocking:** the aggregator's exit code is discarded here.

This step has no `shell:` key, so it runs under GitHub's default `bash -e {0}`. `pipefail` is only added when you write `shell: bash` explicitly. Without it the pipeline exits with `tee`'s status, which is always 0, so `steps.aggregate.outcome` is never `failure` and that half of the notify condition on line 112 is dead. The script's stated contract — "Exits non-zero if any test's final status is not 'passed'/'skipped', so callers can gate a failure notification on it" — never takes effect.

```yaml
        run: |
          set -o pipefail
          node .github/scripts/aggregate-e2e-results.js e2e-results e2e-run-stats.jsonl | tee -a "$GITHUB_STEP_SUMMARY"
```

.github/workflows/e2e-scheduled.yml
```diff
@@ -0,0 +1,119 @@
...
+            git rm -rf --quiet . || true
+          fi
+          mkdir -p history
+          cat e2e-run-stats.jsonl >> history/e2e-history.jsonl
+          git add history/e2e-history.jsonl
+          git commit -m "E2E stats: run ${{ github.run_id }} attempt ${{ github.run_attempt }}"
+          git push origin e2e-metrics
+
+      - name: Notify on failure
+        if: ${{ always() && (steps.aggregate.outcome == 'failure' || contains(needs.*.result, 'failure')) && github.repository == 'ballerina-platform/ballerina-vscode' }}
```
NipunaRanasinghe: `contains(needs.*.result, 'failure')` misses `cancelled`. The E2E job has a 60-minute timeout per matrix group and this workflow has a `concurrency` group, so a timed-out or evicted run is a realistic outcome and it would go unreported.

`needs.E2E.result != 'success'` covers failed, cancelled and skipped in one condition.

packages/ballerina-extension/e2e-test/playwright.config.js
```diff
@@ -9,7 +9,7 @@ exports.default = (0, test_1.defineConfig)({
     retries: process.env.BI_E2E_RETRIES ? Number(process.env.BI_E2E_RETRIES) : 2,
     maxFailures: 10,
```
NipunaRanasinghe, line 10: Not changed by this PR, but it caps the data the new report is built on: `maxFailures: 10` aborts the group once 10 tests have failed, and the remaining tests never run and never appear in the JSON. The report will show them as neither passed nor failed, with nothing indicating the run was truncated.

Worth having the aggregator read `report.config` / the run status and print an explicit "run stopped early at maxFailures" line, otherwise a badly broken run looks smaller than a mildly broken one.

.github/workflows/e2e-scheduled.yml
```diff
@@ -0,0 +1,119 @@
...
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git fetch origin e2e-metrics || true
+          if git show-ref --verify --quiet refs/remotes/origin/e2e-metrics; then
+            git checkout -B e2e-metrics origin/e2e-metrics
+          else
+            git checkout --orphan e2e-metrics
+            git rm -rf --quiet . || true
+          fi
+          mkdir -p history
+          cat e2e-run-stats.jsonl >> history/e2e-history.jsonl
```
NipunaRanasinghe: `history/e2e-history.jsonl` grows without bound: one line per test, per group, four runs a day, appended forever, in a git branch that is also fetched on every run. Worth deciding the retention rule now — rotate per month (`history/2026-08.jsonl`) or truncate past N days — rather than after the file is large enough to be annoying to fix.

.github/workflows/e2e-scheduled.yml
```diff
@@ -0,0 +1,119 @@
...
+          git add history/e2e-history.jsonl
+          git commit -m "E2E stats: run ${{ github.run_id }} attempt ${{ github.run_attempt }}"
+          git push origin e2e-metrics
```
NipunaRanasinghe: No rebase or retry on a non-fast-forward push. The `concurrency` group prevents this workflow racing itself, so this only bites if anything else ever writes `e2e-metrics`, but the failure mode is a red job caused purely by the metrics bookkeeping. A `git pull --rebase origin e2e-metrics || true` before the push, or a short retry loop, keeps a push conflict from failing the run.

.github/workflows/e2e-scheduled.yml
```diff
@@ -0,0 +1,119 @@
+name: Scheduled E2E Tests
+
+on:
+  schedule:
+    - cron: '0 */6 * * *'
```
NipunaRanasinghe, line 8: Every 6 hours is four full runs a day: build stage plus a four-group E2E matrix at 60 minutes per group, on top of the 01:30 nightly. If the goal is flakiness data rather than fast feedback, every 12 hours gives most of the same signal at half the runner cost. Worth a deliberate choice rather than defaulting to the shorter interval.

.github/scripts/aggregate-e2e-results.js
```diff
@@ -0,0 +1,169 @@
...
+    lines.push('No retries or failures — every test passed on the first attempt.');
+    return lines.join('\n');
+  }
+
+  lines.push('| Group | Test | Attempts | Final status | Errors |');
+  lines.push('|---|---|---|---|---|');
+  for (const row of rows.sort((a, b) => b.attempts - a.attempts)) {
+    const errors = row.errorLines.length ? row.errorLines.join('<br>') : '';
+    lines.push(
+      `| ${row.group} | ${row.title} (${row.file}) | ${row.attempts} | ${row.finalStatus} | ${errors} |`
```
NipunaRanasinghe: A `|` in a test title or an error message breaks the table for that row. Playwright assertion messages routinely contain them (union types, regex alternation, `expect(...).toMatch(/a|b/)`).

Escape the cell contents before interpolating, e.g. a small `cell = (s) => String(s ?? '').replace(/\|/g, '\\|')` applied to title, file, finalStatus and errors.

.github/actions/resolve-source-branch/action.yml
```diff
@@ -0,0 +1,38 @@
+name: 'Resolve source branch'
+description: >-
+  Resolve the branch a nightly-style pipeline should run against: an explicit
+  override if given, otherwise the latest staging/* branch if one exists, else main.
```
NipunaRanasinghe, line 4: The action runs `git ls-remote --heads origin`, so it only works after a checkout has set up the remote. Both callers do check out first, but that precondition isn't inferable from the inputs — worth stating it in the description so the next caller doesn't put this before its checkout step.

Good extraction otherwise; this and `schedule.yml` were already drifting apart.

.github/workflows/e2e-scheduled.yml
samithkavishke: We have minimized the cost by not building the full one rather use the nightly build. And this is just for the starters after we manage to minimize the flakyness we will anyway remove this job

.github/workflows/e2e-scheduled.yml
```diff
@@ -0,0 +1,194 @@
...
+          fi
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+          pushed=false
+          for attempt in 1 2 3; do
+            git fetch origin e2e-metrics || true
+            if git show-ref --verify --quiet refs/remotes/origin/e2e-metrics; then
+              git checkout -B e2e-metrics origin/e2e-metrics
+            else
```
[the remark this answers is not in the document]
samithkavishke: Addressed

.github/actions/resolve-source-branch/action.yml
samithkavishke: Addressed

packages/ballerina-extension/e2e-test/playwright.config.js
samithkavishke: Agree which need to be address separately

.github/actions/run-e2e-group/action.yml
```diff
@@ -0,0 +1,97 @@
...
+      shell: bash
+      id: run-tests-first
+      if: github.run_attempt == 1 || steps.check-results.outcome == 'failure'
+      continue-on-error: true
+      env:
+        # A distinct filename from the re-run's report below: both are full Playwright
+        # JSON reports written by separate CLI invocations, so a shared name would let
+        # the re-run silently overwrite the first attempt's results instead of adding
+        # to them — aggregate-e2e-results.js merges both by matching this prefix.
+        PLAYWRIGHT_JSON_OUTPUT_FILE: test-results/e2e-results-first.json
```
NipunaRanasinghe: **Still broken:** distinct filenames don't survive the re-run, because both files live inside `outputDir` and Playwright deletes that directory at the start of every run.

`createRemoveOutputDirsTask` calls `removeFolders([outputDir])` with no carve-out. `--last-failed` still works only because `.last-run.json` is read at `testRunner.js:368`, before the task pipeline starts — the JSON reporter writes at the *end*, so `e2e-results-first.json` is already gone by then.

Verified on 1.55.1, the version pinned in `packages/ballerina-extension/package.json` — two invocations, one config, distinct `PLAYWRIGHT_JSON_OUTPUT_FILE` values:

```
=== first attempt (one test fails) ===
.last-run.json
e2e-results-first.json
=== re-run --last-failed ===
.last-run.json
e2e-results-rerun.json      <- e2e-results-first.json is gone
```

So `mergeGroupReports` never has two files to merge on exactly the runs that need merging, and the outcome is the same as before the fix.

Both reports need to be written outside `outputDir`, e.g. `e2e-reports/e2e-results-first.json` and `e2e-reports/e2e-results-rerun.json` (both cwd-relative, so `packages/ballerina-extension/e2e-reports/`), with `packages/ballerina-extension/e2e-reports/**` added to the upload on line 74.

One knock-on: with two paths, upload-artifact roots the artifact at the common ancestor `packages/ballerina-extension`, so the download on line 31 has to become `path: packages/ballerina-extension` or `.last-run.json` lands a level too deep and the run_attempt > 1 resume path stops working. `findResultFiles` and `groupNameFromPath` both still work under that layout.

.github/workflows/e2e-scheduled.yml
```diff
@@ -0,0 +1,207 @@
...
+
+      - name: Aggregate results
+        id: aggregate
+        shell: bash
+        env:
+          # Every run tests the VSIX from the 'nightly' GitHub Release/tag now, not a
+          # freshly resolved branch — this labels the history data accordingly. (Not
+          # 'builds/nightly' — that's schedule.yml's own internal build branch, a
+          # different thing from the 'nightly' release tag this workflow reads.)
+          E2E_SOURCE_TAG: nightly
```
NipunaRanasinghe, line 166: This is a constant, so no row in the history says which nightly build produced it.

The job comment above argues that testing the nightly VSIX "holds product code constant so variance is attributable to test flakiness rather than code churn" — but the `nightly` tag moves every night, so across the history the product code isn't constant, only within a day. Without the built commit in the row, a genuine regression introduced by last night's build and a flaky test look identical in the data.

`gh release view nightly --json targetCommitish --jq .targetCommitish` at download time, passed through as a second field, makes the history able to answer "did this start failing when the build changed?".

packages/ballerina-extension/e2e-test/playwright.config.js
NipunaRanasinghe, line 10: Not changed by this PR, but it caps the data the report is built on: `maxFailures: 10` aborts the group once 10 tests have failed, and the rest never run and never appear in the JSON. They'll be absent from the counts entirely, with nothing marking the run as truncated — so a badly broken run reads as smaller than a mildly broken one.

The aggregator now flags unparseable reports explicitly; a truncated run deserves the same treatment. Fine to defer, but worth not losing.

.github/workflows/e2e-scheduled.yml
```diff
@@ -0,0 +1,207 @@
...
+              # A prior failed iteration this run may have already left an orphan
+              # branch checked out here — detach and drop it first so re-orphaning
+              # can't collide with itself on retry.
+              git checkout --detach 2>/dev/null || true
+              git branch -D e2e-metrics-tmp 2>/dev/null || true
+              git checkout --orphan e2e-metrics-tmp
+              git rm -rf --quiet . || true
+            fi
+            mkdir -p history
+            cat e2e-run-stats.jsonl >> history/e2e-history.jsonl
```
NipunaRanasinghe: `history/e2e-history.jsonl` still grows without bound — one line per test, per group, four runs a day, appended forever, in a branch that's also fetched on every run. Rotating per month (`history/2026-08.jsonl`) costs one `date` call here and keeps any future reader from having to stream a single large file. Fine to defer, but cheaper to decide now than to split the file later.

.github/workflows/e2e-scheduled.yml
```diff
@@ -0,0 +1,207 @@
...
+        with:
+          persist-credentials: false
+
+      - name: Download nightly VSIX
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          mkdir -p packages/ballerina-extension/vsix
+          gh release download nightly --repo "${{ github.repository }}" \
```
NipunaRanasinghe: The VSIX path checks out — `setup.ts:32` resolves `extensionsFolder` to `packages/ballerina-extension/vsix`, which is where this downloads, and `resolveBallerinaVsixPath` filters out `ballerina-integrator-*.vsix` so the broader `ballerina-*.vsix` pattern here is safe.

One thing to know: if the download ever produces nothing, `setup.ts:167` falls back to running `rush build -t ballerina` instead of failing. `set -euo pipefail` here catches a `gh` error, but a release that exists with no matching asset would pass this step and turn into a silent full build — which is the one thing the job comment says this design avoids. Worth asserting the file exists after the download.

.github/workflows/e2e-scheduled.yml
```diff
@@ -0,0 +1,253 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          # This job only needs .github/scripts/ and .github/actions/failure-notification
+          # from the current tree — no history. The e2e-metrics push below does its git
+          # work in an isolated throwaway clone (see "Persist run stats"), so this
+          # checkout never pushes and doesn't need full history or persisted credentials.
+          persist-credentials: false
+          persist-credentials: false
```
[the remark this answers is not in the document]
samithkavishke: Addressed

.github/actions/run-e2e-group/action.yml
samithkavishke: Addressed

.github/workflows/e2e-scheduled.yml
samithkavishke: Addressed

.github/workflows/e2e-scheduled.yml
samithkavishke: Addressed

.github/workflows/e2e-scheduled.yml
samithkavishke: Addressed

.github/actions/run-e2e-group/action.yml
```diff
@@ -0,0 +1,114 @@
...
+        export $(dbus-launch)
+        export DISPLAY=:98.0
+        cd packages/ballerina-extension
+        xvfb-run --server-num 98.0 -s "-ac -screen 0 1920x1080x24" bash -c "ffmpeg -y -nostdin -f x11grab -r 30 -s 1920x1080 -i :98.0 -c:v libx264 -preset ultrafast -pix_fmt yuv420p -movflags +faststart -hide_banner -loglevel error record.mp4 & pnpm exec playwright test -c e2e-test/playwright.config.js --grep ${{ inputs.group }}"
+
+    - name: Run Failed Ballerina e2e Test (Re-run)
+      shell: bash
+      if: steps.run-tests-first.outcome == 'failure' || steps.check-results.outcome == 'success'
+      env:
+        PLAYWRIGHT_JSON_OUTPUT_FILE: e2e-reports/e2e-results-rerun.json
```
NipunaRanasinghe: On a GitHub "re-run failed jobs", this filename is reused, so attempt N's re-run overwrites the attempt N-1 re-run report that the download step just restored.

Attempt N-1's `e2e-results-first.json` survives, so the merge is still broadly right and the final verdict is unaffected. But the intermediate re-run's attempts are dropped, which undercounts exactly the failed-then-passed tests the comment on line 35 says it wants to keep across a continuation.

`e2e-results-rerun-${{ github.run_attempt }}.json` closes it — `mergeGroupReports` sorts filenames and that still sorts after `-first`, and `findResultFiles` already matches on the `e2e-results*.json` prefix.

.github/workflows/e2e-scheduled.yml
```diff
@@ -0,0 +1,252 @@
...
+
+      - name: Aggregate results
+        id: aggregate
+        shell: bash
+        env:
+          # Every run tests the VSIX from the 'nightly' GitHub Release/tag now, not a
+          # freshly resolved branch — this labels the history data accordingly. (Not
+          # 'builds/nightly' — that's schedule.yml's own internal build branch, a
+          # different thing from the 'nightly' release tag this workflow reads.)
+          E2E_SOURCE_TAG: nightly
```
NipunaRanasinghe, line 166: Still a constant, so no history row records which build it came from.

The `nightly` tag moves every night, so across the history the product code isn't held constant — only within a day. Without the built commit in the row, a regression introduced by last night's build and a genuinely flaky test are indistinguishable in the data, which is the main question this history exists to answer.

Cheaper to fix now than last round: `setup-ballerina` already fetches `refs/tags/nightly` for the `gradlePropertiesRef` pin, so the SHA is available in the E2E job — `gh release view nightly --json targetCommitish --jq .targetCommitish` in the Report job gets the same thing, as a second env var alongside this one.

packages/ballerina-extension/e2e-test/playwright.config.js
NipunaRanasinghe, line 10: The new missing-group check doesn't cover this case. `maxFailures: 10` aborts the group once 10 tests have failed, so it still produces a report — just a short one. `missingGroups` only catches a group that uploaded nothing at all, so a truncated group is counted as fully reported and the tests that never ran are silently absent from the totals. A badly broken run reads as smaller than a mildly broken one.

With `EXPECTED_GROUPS_JSON` already threaded through, an expected test count per group would close the gap, or read the run's own status from the report and print the same kind of warning line the unreadable/missing cases now get.

.github/workflows/e2e-scheduled.yml
```diff
@@ -0,0 +1,252 @@
...
+              # A prior failed iteration this run may have already left an orphan
+              # branch checked out here — detach and drop it first so re-orphaning
+              # can't collide with itself on retry.
+              git checkout --detach 2>/dev/null || true
+              git branch -D e2e-metrics-tmp 2>/dev/null || true
+              git checkout --orphan e2e-metrics-tmp
+              git rm -rf --quiet . || true
+            fi
+            mkdir -p history
+            cat "$STATS_FILE" >> history/e2e-history.jsonl
```
NipunaRanasinghe: `history/e2e-history.jsonl` still grows without bound — one line per test, per group, four runs a day, appended forever. The shallow single-branch clone keeps the fetch cheap, but the file itself is the thing that grows, and every run rewrites it whole.

Rotating per month (`history/$(date -u +%Y-%m).jsonl`) is a one-line change here and keeps any future reader from having to stream one large file. Fine to defer, cheaper to decide now than to split it later.

.github/workflows/e2e-scheduled.yml
```diff
@@ -0,0 +1,252 @@
...
+          # its project folders don't exist and Setup Rush/the build step fail outright.
+          submodules: recursive
+
+      - name: Download nightly VSIX
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          mkdir -p packages/ballerina-extension/vsix
+          gh release download nightly --repo "${{ github.repository }}" \
```
NipunaRanasinghe: Worth asserting a VSIX actually landed. `set -euo pipefail` catches a `gh` failure, but a release that exists with no matching asset exits 0 having downloaded nothing — and `setup.ts:167` then falls back to `rush build -t ballerina` rather than failing, which is the exact scenario the job comment above says this design avoids.

```bash
ls packages/ballerina-extension/vsix/ballerina-*.vsix >/dev/null
```

after the download turns a silent hour-long rebuild into an immediate, obvious failure.

NipunaRanasinghe (approved):
LGTM

PR #862, opened 2026-08-13
Add code cov support in the language server

pasindufernando1:
The language server has no coverage measurement. Each module can emit JaCoCo execution data, but nothing merges it, so there is no single number for the language server and no way to see which classes a change left untested. Reviewers have no signal beyond "the suite passed", and a PR that adds a code path with no test looks identical to one that does not.

- Produce one coverage figure for the whole language server, not per-module fragments.
- Surface it on every pull request that touches the language server, and keep a per-branch baseline to compare against.
- Make it visible locally too, so a developer can find uncovered classes without waiting for CI.
- Report coverage without gating merges. The number is unknown until the first full run, so it must not fail builds on day one.

**`packages/ballerina-language-server/build.gradle`** — adds a `createCodeCoverageReport` task (`JacocoReport`) that merges every module's execution data into one report, mirroring the task of the same name in `ballerina-lang`.

It reads two locations, because modules produce execution data differently: `build/coverage-reports/*.exec` for modules configured by `gradle/javaProject.gradle`, and JaCoCo's default `build/jacoco/*.exec` for modules that only get the base plugin from `allprojects`. Four projects contribute execution data but are excluded from the classes and sources being measured, because they are not language-server production code: `:checkstyle` (no sources, it only downloads the config), `:langserver-stdlib` (a Ballerina mock stdlib used as a test fixture), and the two index generators (build-time tools).

Three details are deliberate:

- **`mustRunAfter`, not `dependsOn`.** `./gradlew test createCodeCoverageReport` runs the report last, while `./gradlew createCodeCoverageReport` on its own re-reads existing build outputs without rebuilding or re-running anything. That is what lets CI publish coverage from a run whose suite failed — the execution data for everything that did run is still on disk, and that number is more useful than none. Declaring no dependency also avoids Gradle 8 rejecting the task for reading `compileJava` and `test` outputs it never declared.
- **`outputs.upToDateWhen { false }`.** Asking for the task always produces a report and always logs the percentage; an `UP-TO-DATE` skip would silently print nothing. Regeneration takes seconds against a suite that takes tens of minutes.
- **A `doLast` that parses the generated XML** and logs INSTRUCTION, BRANCH and LINE percentages, so the figure appears in the build log rather than only on Codecov.

**`codecov.yml`** (new, repo root — the only place Codecov reads it) — reporting policy.

Every status is `informational: true`: coverage appears on the PR and the dashboard but never fails a build. The comment marks where to flip that once the numbers have settled. `range: "30...80"` and `threshold: 1%` are first-pass values; the threshold absorbs the small run-to-run variation in which language-server tests execute.

Two settings exist to handle conditional CI. Language-server tests only run when language-server files change, so a PR touching nothing there produces no fresh upload. `max_report_age: off` stops Codecov discarding the carried-forward report as stale, and `carryforward: true` on the `language-server` flag makes it reuse the base commit's report instead of reporting 0%.

**`.github/actions/ls-test/action.yml`** — three steps appended to the existing language-server test action, all guarded by `always()` so they run even when the suite failed: generate the aggregated report, resolve upload metadata, then upload to Codecov and attach the HTML report as a build artifact.

Coverage is generated and published from **Linux only**. The Windows leg runs the same suite over the same sources, so a second upload would duplicate the report, and its Windows paths do not map onto the repository layout.

The metadata step handles one subtlety: it overrides the commit **only** when an explicit branch is supplied. `schedule.yml` checks out `matrix.branch` rather than the ref the schedule fired on, so its uploads must be attributed explicitly; on a `pull_request` event Codecov's own detection is correct, and `HEAD` is a merge commit that would be attributed to the wrong place. It also replaces `/` with `-` in the artifact label, because artifact names cannot contain slashes and both branch names (`builds/nightly`) and PR refs (`123/merge`) do.

`fail_ci_if_error: false` and an optional token mean a Codecov outage or a missing secret cannot fail the build; public repositories fall back to Codecov's tokenless flow. The Codecov action is pinned by commit SHA.

**Workflow plumbing** — `CODECOV_TOKEN` threaded through `pull-request.yml`, `devBuild.yml`, `reusable-build.yml`, `schedule.yml` and `release-pre-release.yml`, declared `required: false` in the reusable workflow. `schedule.yml` additionally passes `codecov-branch: ${{ matrix.branch }}`; those nightly runs are what give each tracked branch, `main` included, the baseline PRs compare against.

**`.gitignore`** — ignores `.jacoco/`, the aggregated report's output directory. Without this the report is untracked and one careless `git add` away from being committed; it is a 3,600-file directory.

**Documentation** — `.github/workflows/README.md` (+30) and the root `README.md` (+32) describe how to produce the report locally and how CI publishes it.

No production code changes.

## UI Component Development
N/A — no UI changes.

## User stories
N/A

## Release note
N/A — build and CI tooling only; no user-facing change.

## Documentation
Covered in this PR: `.github/workflows/README.md` explains the CI upload path, and the root `README.md` documents running `./gradlew test createCodeCoverageReport` locally and where the HTML report lands. No product documentation impact.

## Training
N/A

## Certification
N/A — no functional or user-facing behaviour to examine.

## Marketing
N/A

## Automation tests
 - Unit tests
   > No production code changed, so no coverage change of its own. The task itself is exercised by every run that produces a report.
 - Integration tests
   > The first CI run on this PR is the real verification: it establishes whether the aggregation finds execution data in both locations, whether the upload is attributed to the right commit and branch, and what the language server's coverage actually is. The `range` and `threshold` values in `codecov.yml` are placeholders to be revisited once that number is known.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? N/A — no Java or other source code changed.
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes. `CODECOV_TOKEN` is referenced only as a GitHub secret and is optional; the third-party Codecov action is pinned by commit SHA rather than a mutable tag.

## Samples
N/A

## Related PRs
None functionally. Note for merge order: #852 also adds a root-level task to `packages/ballerina-language-server/build.gradle`, so whichever lands second may need a trivial textual resolution in that file.

## Migrations (if applicable)
N/A

## Test environment
Task behaviour was exercised locally on macOS 15 (Darwin 24.1.0) with Ballerina 2201.13.4 and JDK 21. The CI paths — Linux-only generation, branch and commit attribution, artifact upload — are verified by this PR's own run.

## Learning
The non-obvious constraint was that coverage has to survive a failed suite. Wiring the report with `dependsOn` would have been the natural choice, but then a build whose tests failed would either skip the report or try to re-run the suite. Declaring only ordering, and letting the task re-read build outputs, is what makes `always()` in CI meaningful — a partial coverage number from a run that failed is still worth publishing.

The second was that conditional test execution and Codecov's staleness rules interact badly: because language-server tests only run on relevant path changes, the default configuration would have reported 0% for unrelated PRs. That is what `carryforward` and `max_report_age: off` are for.

## Summary

- Added aggregated JaCoCo coverage reporting for language-server modules.
- Added instruction, branch, and line coverage metrics.
- Configured informational Codecov statuses with carry-forward support.
- Updated Linux CI workflows to generate and upload coverage after test runs, including failed runs.
- Added optional `CODECOV_TOKEN` support across relevant workflows.
- Added local and CI coverage documentation.
- Excluded generated JaCoCo output from version control.
- Preserved coverage baselines when language-server tests are skipped.


.github/actions/ls-test/action.yml
```diff
@@ -64,3 +78,63 @@ runs:
...
+        # token must never fail the build.
+        fail_ci_if_error: false
+
+    # The XML is what Codecov consumes; the HTML report next to it is how a developer
+    # finds which classes are uncovered without re-running the suite locally.
+    - name: Upload coverage report
+      if: ${{ always() && runner.os == 'Linux' }}
+      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+      with:
+        name: ls-coverage-${{ steps.coverage.outputs.label }}
```
NipunaRanasinghe, line 144: This will break the nightly runs. `label` falls back to `github.ref_name` when `codecov-branch` is empty, and `schedule.yml`'s `ls-ubuntu-test` is a matrix over two branches that doesn't pass `codecov-branch`. `github.ref_name` is the same for both legs on a schedule event, so both try to upload an artifact named `ls-coverage-main`. `upload-artifact@v4` returns a 409 on a duplicate name and fails the step, and since the step is `if: always()` with `fail-fast: false`, whichever leg gets there second goes red every night.

Easiest fix is to base the label on something job-unique, e.g. `matrix.branch`, or pass `codecov-branch` from schedule.yml.

.github/workflows/schedule.yml
NipunaRanasinghe, line 1: The `ls-ubuntu-test` job (lines 240-269) never passes `codecov-branch` or `codecov-token` to `./.github/actions/ls-test`, even though it's the one caller that needs them.

That job checks out `refs/heads/${{ matrix.branch }}`, but with `override_branch`/`override_commit` empty, Codecov attributes the upload to the schedule's own ref and `GITHUB_SHA` (default branch head) while the report actually covers a different tree. So coverage measured on `builds/nightly` and `staging/*` gets recorded against a `main` commit with different sources, and both matrix legs land on the same commit + flag. It also means no branch ever gets a real baseline, so `carryforward: true` and `max_report_age: off` in codecov.yml have nothing to carry forward.

No `codecov-token` either, so this is a tokenless upload on a non-`pull_request` event. Codecov's tokenless path is a PR-only fallback, so it will most likely be rejected, and silently, given `fail_ci_if_error: false`.

Worth noting `.github/workflows/README.md` says "schedule.yml passes `codecov-branch`" and the input description in action.yml names schedule.yml explicitly, so the docs describe something that isn't wired up yet.

.github/actions/ls-test/action.yml
```diff
@@ -64,3 +78,63 @@ runs:
...
+    # did run, and the coverage number for it is more useful than none. createCodeCoverageReport
+    # declares no task dependencies, so this re-reads build outputs instead of rebuilding.
+    - name: Generate aggregated coverage report
+      if: ${{ always() && runner.os == 'Linux' }}
+      shell: bash
+      working-directory: packages/ballerina-language-server
+      env:
+        packageUser: ${{ inputs.package-user }}
+        packagePAT: ${{ inputs.package-pat }}
+      run: ./gradlew createCodeCoverageReport
```
NipunaRanasinghe, line 103: Every other step in the coverage path is defensive (`fail_ci_if_error: false`, `if-no-files-found: warn`) but this one isn't. If report generation fails the whole job fails, even when the test suite passed. Realistic causes: JaCoCo's `Can't add different class with same name` when two modules ship the same FQCN, an `XmlSlurper` failure in the `doLast`, or a configuration failure.

Suggest `continue-on-error: true` here so coverage stays informational.

packages/ballerina-language-server/build.gradle
```diff
@@ -429,6 +429,89 @@ task buildCommons {
...
+    // on its own — what CI runs after a suite that failed — must not rebuild or re-run
+    // anything. Without this, Gradle 8 rejects the task for reading compileJava and
+    // test outputs it never declared a dependency on.
+    mustRunAfter subprojects.collect { it.tasks.withType(JavaCompile) }
+    mustRunAfter subprojects.collect { it.tasks.withType(Test) }
+
+    // Asking for this task always produces a report and always logs the percentage;
+    // an UP-TO-DATE skip would silently print nothing. Regeneration is seconds against
+    // a suite that takes tens of minutes.
+    outputs.upToDateWhen { false }
```
NipunaRanasinghe, line 471: This doesn't quite do what the comment says. `JacocoReport` is `@CacheableTask` and `gradle.properties` has `org.gradle.caching=true`. `upToDateWhen { false }` only defeats the up-to-date check; the task is still eligible to be loaded FROM-CACHE, and a FROM-CACHE task's actions don't run, so the `doLast` that logs the percentages is skipped.

The flow in the README hits this: `./gradlew test createCodeCoverageReport`, then `./gradlew createCodeCoverageReport` with unchanged inputs gives a cache hit, restores the report and logs nothing. Same in CI, since `setup-gradle` restores the build cache across runs.

Adding `outputs.cacheIf { false }` next to this covers it.

.github/actions/ls-test/action.yml
```diff
@@ -64,3 +78,63 @@ runs:
...
+
+    # Coverage is aggregated and published from Linux only. The Windows leg runs the
+    # same suite over the same sources, so a second upload would only duplicate the
+    # report, and its Windows paths do not map onto the repository layout.
+    #
+    # always(): a failing suite still wrote JaCoCo execution data for everything that
+    # did run, and the coverage number for it is more useful than none. createCodeCoverageReport
+    # declares no task dependencies, so this re-reads build outputs instead of rebuilding.
+    - name: Generate aggregated coverage report
+      if: ${{ always() && runner.os == 'Linux' }}
```
NipunaRanasinghe: Minor: `always()` also covers the cancelled state, which isn't what the comment above describes. `pull-request.yml` sets `cancel-in-progress: true` and schedule.yml has per-job concurrency groups, so two pushes in quick succession make the cancelled run re-invoke Gradle and publish a partial report during teardown.

`!cancelled()` matches the intent here (and on lines 99, 117, 134).

packages/ballerina-language-server/build.gradle
```diff
@@ -429,6 +429,89 @@ task buildCommons {
...
+        def report = parser.parse(xmlReport)
+        ['INSTRUCTION', 'BRANCH', 'LINE'].each { type ->
+            def counter = report.counter.find { it.@type == type }
+            if (counter.isEmpty()) {
+                return
+            }
+            def missed = counter.@missed.toInteger()
+            def covered = counter.@covered.toInteger()
+            def total = missed + covered
+            def percentage = total == 0 ? 0 : (covered * 100.0d / total)
```
NipunaRanasinghe: Nit: the zero branch yields an `Integer`, which `%6.2f` on the next line rejects with `IllegalFormatConversionException: f != java.lang.Integer`. JaCoCo never emits a counter with `missed=0 covered=0` so it's unreachable in practice, but as written the guard turns a harmless case into a build failure. `0.0d` instead of `0`.

NipunaRanasinghe (changes requested):

packages/ballerina-language-server/README.md
```diff
@@ -35,6 +37,36 @@ You can find the Language Server User Guide [here](https://github.com/ballerina-
...
+
+```bash
+./gradlew test createCodeCoverageReport   # full suite, then the report
+./gradlew createCodeCoverageReport        # report from the last run's data, no rebuild
+```
+
+The output looks like this (illustrative figures — a report only reflects the tests that
+actually ran, so only a full `test` run gives the project's real coverage):
+
+```
```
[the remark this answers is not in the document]
pasindufernando1: Done — fenced as ```text in 2a669b4.

.github/actions/ls-test/action.yml
pasindufernando1: Fixed by passing `codecov-branch: ${{ matrix.branch }}` from schedule.yml, so the label comes from the matrix leg instead of `github.ref_name`. The two legs now produce `ls-coverage-main` and `ls-coverage-builds-nightly`.

.github/workflows/schedule.yml
pasindufernando1: Right, the action had the input and the caller never used it. `ls-ubuntu-test` now passes both:

```yaml
codecov-token: ${{ secrets.CODECOV_TOKEN }}
codecov-branch: ${{ matrix.branch }}
```

`codecov-branch` also switches on the commit override in the action, so each leg is attributed to the branch and commit it actually checked out, and each gets its own baseline. The docs now match what's wired.

One caveat: if `CODECOV_TOKEN` isn't defined at repo or org level it resolves to an empty string and we're back on the tokenless path, silently. Worth confirming the secret exists.

.github/actions/ls-test/action.yml
pasindufernando1: Agreed — added `continue-on-error: true` on the generate step. The downstream steps already degrade on their own, so a failed report just means no upload rather than a red job.

packages/ballerina-language-server/build.gradle
pasindufernando1: Good catch — up-to-date and caching are separate mechanisms and I only disabled the first. Added `outputs.cacheIf { false }` and corrected the comment to say why both are needed.

.github/actions/ls-test/action.yml
pasindufernando1: Switched all four to `!cancelled()`, and updated the comment plus the matching line in workflows/README.md. A superseded run publishing a partial report was not the intent.

packages/ballerina-language-server/build.gradle
pasindufernando1: Fixed — `0.0d`.

NipunaRanasinghe (approved):
LGTM

NipunaRanasinghe (approved):
LGTM

PR #868, opened 2026-08-14
Fix the workflow activity flows: creation, result binding, and run navigation

rdulmina:
This PR is based on https://github.com/ballerina-platform/ballerina-vscode/pull/853

Fixes the workflow activity flows reported for 5.1-alpha: creating an activity, what an activity call binds, and navigating from a run to the workflow it starts.

Fixes wso2/product-integrator#2126
Fixes wso2/product-integrator#2128
Fixes wso2/product-integrator#2143
Fixes https://github.com/wso2/product-integrator/issues/2141
Fixes https://github.com/wso2/product-integrator/issues/2094
Fixes https://github.com/wso2/product-integrator/issues/2090
Fixes https://github.com/wso2/product-integrator/issues/2097
Fixes https://github.com/wso2/product-integrator/issues/2099
Fixes https://github.com/wso2/product-integrator/issues/2041

### Creating an activity did nothing

Clicking **+** on Workflow Activities dispatches through `openView` with `resetHistory`, so the view lands in the main panel — but `MainPanel`'s switch had a case for every sibling form (function, workflow, durable agent) and none for the activity form, which was routed only in `PopupPanel`. The switch has no `default`, so the state machine transitioned, the webview received the location, and no component was ever set. The edit path opens the same view in the main panel, so it was broken too.

The gap dates to the commit that introduced both forms: it gave each a `PopupPanel` case but only the workflow a `MainPanel` one, which is why Workflows **+** worked and Activities **+** did not.

### An activity call bound a result the call did not produce

An activity that returns nothing generated `anydata result = check ctx->callActivity(...)`. The name was invented rather than asked for, so a second call to such an activity redeclared it. The builtin with no return value (email) had the opposite problem — it emitted a bare call statement. Neither compiled:

```
ERROR redeclared symbol 'result'
ERROR cannot infer the 'typedesc' argument for parameter 'T': expected an argument
      for the parameter or a contextually-expected type to infer the argument
```

What a call binds now follows from what it produces and whether its errors are checked:

| Activity | Check Error | Generated |
| --- | --- | --- |
| returns a value | ticked | `int total = check ctx->callActivity(...)` |
| returns a value | unticked | `int\|error total = ctx->callActivity(...)` |
| returns nothing | ticked | `() _ = check ctx->callActivity(...)` |
| returns nothing | unticked | `error? myResult = ctx->callActivity(...)` |

The wildcard is not cosmetic: `callActivity` is dependently typed (`typedesc<anydata> T = <>`), so a bare call statement cannot infer its `T`. All four shapes were compiled against `ballerina/workflow` to confirm they are accepted.

### Check Error on user-defined activities

`callActivity` is error-returning whatever the activity itself returns, so the choice always applies; user-defined activities carried only a hidden "Trigger error flow" flag, while the prebuilt activities had the checkbox. They now read the same. It sits in the form's advanced configurations — it refines how the call behaves rather than saying what it does — and stays a root property because entries in the activity call configurations list are emitted as named arguments to `callActivity`, which `checkError` is not.

An activity with no result has nothing to name while the box is ticked, so the **Result** field is carried in the flag's unchecked branch and appears only then. Its value lives in a root property seeded with a name unique among the visible symbols, so two unchecked calls to the same activity do not collide. Making that possible needed a shared-editor change: the checkbox editor rendered the box and nothing else, so any fields attached to a flag were dropped. It now reads them per state (`"true"` / `"false"`), the way the dropdown choice already reads them per option. A flag carrying no such fields renders exactly as before.

Reading a statement back mirrors all of this: the flag comes from whether the source wrote `check`, and a wildcard binding is left out of the form since it names nothing the user chose.

### Clicking a workflow icon opens the workflow

The Run Workflow node draws the workflow it runs beside it, and clicking that icon used to open it. It stopped when the node moved from the plain box — where `BaseNodeWidget` navigated through `getFunctionLocation` — to the action shape with the icon, because `ApiCallNodeWidget` only knows how to open connections, and a run has no connection property. The side icon of a workflow-target node (a run, and the child workflow statements sharing the shape) now resolves the workflow it names and opens it. A workflow that cannot be located falls back to the previous behaviour rather than swallowing the click.

### Testing

- New: `WorkflowRunNavigation.test.tsx` (icon click + fallback) and `MainPanel.routing.test.ts` (every view the activator dispatches has a `MainPanel` case). Both confirmed failing against the previous code.
- New language server fixtures: activity call templates and source generation for a nil-returning activity, checked and unchecked, plus the source read-back of an unchecked call.
- `NodeTemplateTest`, `ModelGeneratorTest`, `SourceGeneratorTest` and the workflow manager suites pass (756 tests); the visualizer suite passes (222 tests).
- The four generated shapes above were compiled with `bal build` against `ballerina/workflow` 0.8.3. Worth noting the goldens assert generated text, not that it compiles — both bugs fixed here were compiler errors that a text comparison would have accepted.

### Note for reviewers

This branch is stacked on unmerged work: `e9ba6b771d` and `d336dac9be` are #853, and `428e49ccec`, `e88a1922bf`, `04f239998f` come from `fix/alpha-round-2`. Only the four commits on top plus `cf47117b2e` are new here. The diff will shrink as those land.

## Summary

- Fixed workflow activity creation and editing through `MainPanel`.
- Improved activity-call bindings for typed, nil, error, and wildcard results.
- Restored activity error-check configuration and source round-tripping.
- Updated retry defaults and conditional branch-field rendering.
- Added workflow navigation from run and workflow-target nodes with fallback behavior.
- Added human-task rendering, role handling, sizing, and node support.
- Improved durable-agent activity configuration and model-provider setup.
- Improved Ballerina package indexing and module resolution.
- Added routing, navigation, language-server, source-generation, and visualizer test coverage.


packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -3849,6 +3934,22 @@ private void handleVariableNode(NonTerminalNode variableDeclarationNode) {
...
+        } else if (nodeBuilder instanceof ActivityCallBuilder && bindsNilResult()) {
+            // `error? r = ctx->callActivity(...)`: the activity produces no value, so what the statement
+            // names is the error. The creation template presents that through the Check Error branch and
+            // keeps the name in a root property the branch's Result field reads from, so the analysed
+            // node is given the same shape - a plain result variable here would be shadowed by the
+            // branch's own (empty) field, leaving the form blank.
+            ActivityCallBuilder.addNilResultVariableProperty(nodeBuilder,
+                    this.typedBindingPatternNode.bindingPattern().toSourceCode().strip());
+            ActivityCallBuilder.addCheckErrorProperty(nodeBuilder, false,
+                    ActivityCallBuilder.nilResultCheckErrorFields());
```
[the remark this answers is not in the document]
rdulmina: Stale — the feature this describes no longer exists on the branch.

Check Error has been removed entirely: an activity call is always generated with `check`, so there is no flag to preserve and no `addCheckErrorProperty` call left in `CodeAnalyzer` (`grep -rn "addCheckErrorProperty" --include="*.java"` returns nothing). What replaced it: a statement whose binding names nothing the user chose — a wildcard, or a nil-typed binding such as `error? r = ...` — is kept out of the form so it round-trips, and everything else binds under its own result type.

The concern behind the comment was real while the flag existed, and it was fixed at the time: the analysed node was given the same shape as the creation template so the name landed where the form reads it. That code has since been deleted along with the flag.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1541,6 +1542,35 @@ private void collectDeclaredCapabilities(ExpressionNode listExpr, String capabil
...
+                                : ((MappingConstructorExpressionNode) specificField.valueExpr().get()).fields()) {
+                            if (bindingField instanceof SpecificFieldNode bindingSpecific
+                                    && bindingSpecific.valueExpr().isPresent()) {
+                                values.put(DurableAgentAddActivityBuilder.BINDING_KEY_PREFIX
+                                                + bindingSpecific.fieldName().toSourceCode().trim(),
+                                        bindingSpecific.valueExpr().get().toSourceCode().trim());
+                            }
+                        }
+                        continue;
+                    }
```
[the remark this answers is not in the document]
rdulmina: Checked, and the premise holds: the field map for `activities` (`CodeAnalyzer.java:1481-1483`) carries `activity`, `name`, `description`, `requiresApproval` and `userRoles` — no `bindings` — so a `bindings` value that is not a mapping constructor misses this branch, finds no `fieldToPropertyKey` entry, and is dropped from the form. An edit-save then regenerates the entry without it.

Leaving it open rather than fixing it here. This block comes from `e9ba6b771d`, which belongs to #853, still open and under its own review; this branch only stacks on it. Fixing it in a stacked PR would collide with however #853 resolves the same comment there. It is worth carrying over to that PR.

Scope note for whoever picks it up: the same shape applies to `tools` and `peers`, whose maps also have no entry for their nested fields, so a fallback that keeps the raw source would want to be general rather than specific to `bindings`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
```diff
@@ -619,21 +690,17 @@ private Map<Path, List<TextEdit>> toSourceBuiltin(SourceBuilder sourceBuilder,
...
+        // `callActivity` is dependently typed and a bare call statement cannot infer its `T`.
+        ResultBinding binding = resolveResultBinding(sourceBuilder, hasReturnValue ? lhsType : null, "result");
 
         sourceBuilder.token()
+                .name(binding.resultType())
+                .whiteSpace()
+                .name(binding.variableName())
+                .whiteSpace()
+                .keyword(SyntaxKind.EQUAL_TOKEN)
+                .keyword(SyntaxKind.CHECK_KEYWORD)
```
NipunaRanasinghe: **`check` is now unconditional, so an existing unchecked activity call is silently rewritten on edit-save.**

`addCheckErrorProperty()` is gone, the `checkError` restore in `CodeAnalyzer.populateBuiltinActivityProperties` is gone, and both `toSource` paths (here and line 632) hard-code `.keyword(SyntaxKind.CHECK_KEYWORD)`. There is no longer any input that can produce an unchecked call.

Concrete repro, from the fixture added in this PR — `diagram_generator/source/workflow_activity_result_bindings.bal:15`:

```ballerina
int|error count = ctx->callActivity(countEvents, {});
io:println(count);
```

The matching golden `diagram_generator/config/workflow_activity_result_bindings.json` records `type.value = "int"` for that node and carries no `checkError` property. So opening this node in the form and saving regenerates it as:

```ballerina
int count = check ctx->callActivity(countEvents, {});
```

The error is no longer captured in `count`, and in any workflow whose signature is not `returns error?` (or inside a `do`/`on fail` that relied on the union) the file stops compiling. The user changed, say, a retry field, and got their error handling rewritten.

This also contradicts the PR description, which still documents `int|error total = ctx->callActivity(...)` (Check Error unticked) as a supported generated shape — the description and the "Always check an activity call" commit disagree. Worth deciding which is intended and, if unconditional `check` is the intent, at minimum preserving the union type/`checkError` on read-back so an existing statement round-trips instead of changing meaning.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -3852,6 +3902,11 @@ private void handleVariableNode(NonTerminalNode variableDeclarationNode) {
                             Property.VARIABLE_DOC, true, new HashSet<>(), true);
         } else if (nodeBuilder instanceof WaitDataBuilder) {
             // Variable/type info is embedded in the dataWaits property — skip generic handling
+        } else if (nodeBuilder instanceof ActivityCallBuilder && bindsNothingNamed()) {
```
NipunaRanasinghe: **`bindsNothingNamed()` also swallows `error` / `error?` bindings, so it deletes a variable the user did name and still references.**

`ActivityCallBuilder.isNilResultType(Node)` returns `true` for `ERROR_TYPE_DESC` and, via the `OPTIONAL_TYPE_DESC` recursion, for `error?`. A binding of that type is therefore treated as "nothing the user named" and this branch skips setting `VARIABLE_KEY` / `TYPE_KEY` entirely — even though the source declared a real, named variable.

Concrete repro, from the fixture added in this PR — `diagram_generator/source/workflow_activity_result_bindings.bal:19`:

```ballerina
error? r = ctx->callActivity(logEvent, {message: message});
io:println(r);
```

The golden `diagram_generator/config/workflow_activity_result_bindings.json` contains exactly **one** `variable` property in the whole model (`count`), confirming `r` is dropped from the flow node. Since the form no longer carries the name, `resolveResultBinding` falls into the `declaredType == null` case and regeneration emits:

```ballerina
() _ = check ctx->callActivity(logEvent, {message: message});
```

`r`'s declaration is deleted while `io:println(r)` on the next line still references it → `undefined symbol 'r'`, plus the error is now propagated rather than handled.

The wildcard case (`WILDCARD_BINDING_PATTERN`) and a literal `()` type are the ones that genuinely name nothing. Suggest narrowing `isNilResultType` to `NIL_TYPE_DESC` for this check, and keeping `error` / `error?` bindings as named results.

packages/ballerina-extension/src/features/bi/pending-artifact.ts
```diff
@@ -329,6 +334,25 @@ async function generatePendingArtifact(
...
+ * without the `[ballerina.ai.wso2ProviderConfig]` values the agent fails at startup. The
+ * package root is already known here, so the config file is targeted directly (no project
+ * quick-pick). Failures are non-fatal: the agent exists and the provider can be configured
+ * from the agent's model circle.
+ */
+async function configureDurableAgentModelProvider(projectRoot: string): Promise<void> {
+    try {
+        const configPath = await getConfigFilePath(extension.ballerinaExtInstance, projectRoot);
+        if (configPath) {
+            await addConfigFile(configPath, "model");
```
NipunaRanasinghe: **Unconditional `addConfigFile` signs the user out of Copilot when no token is available, and the `catch` cannot prevent it.**

`addConfigFile` → `getTokenForDefaultModel()`; when that returns `null` it dispatches

```ts
AIStateMachine.service().send(AIMachineEventType.LOGOUT);
```

*before* throwing (`features/ai/utils.ts`, inside the `withProgress` callback — and again in its own `catch`). The `try/catch` here swallows the rejected promise but the LOGOUT event has already been sent. Repro: a user who is not signed in to WSO2 Copilot creates a durable agent through the integration wizard → the artifact is generated, a progress notification flashes, and the AI state machine is silently logged out as a side effect. `FunctionForm` deliberately avoids this ("skip the config write rather than prompting for sign-in on an unknown project state"), but this path has no such guard.

Second issue on the same call: there is no probe at all here, so the Config.toml `[ballerina.ai.wso2ProviderConfig]` entry is written for every `DURABLE_AGENT` artifact. `DurableAgentBuilder.setConcreteTemplateData` only declares `wso2ModelProvider` when `resolveExistingModelProvider` returns `null`; when the generated agent instead reused an existing (possibly non-WSO2) provider, this writes provider config that nothing references.

Suggest mirroring `FunctionForm`: check whether the generation actually declared the default provider (or that the package had none beforehand) before calling `addConfigFile`.

packages/ballerina-visualizer/src/views/BI/FunctionForm/index.tsx
```diff
@@ -518,6 +502,9 @@ export function FunctionForm(props: FunctionFormProps) {
         }
 
         console.log("Updated function node: ", functionNodeCopy);
+        // Probed before generation on purpose — the durable agent's source generation declares
+        // the default provider itself, so an after-the-fact probe always finds one.
+        const hadModelProviderBeforeSave = isDurableAgent ? await projectHasModelProvider() : true;
```
NipunaRanasinghe, line 507: **The "did the generation declare a provider?" decision is now made twice, by two different rules that can disagree.**

Client side (here): `searchNodes({ filePath: projectPath, query: { kind: "MODEL_PROVIDER" } })` — any hit means "already had one", skip the config write.

Server side: `DurableAgentBuilder.resolveExistingModelProvider` loads and compiles the package and asks `DurableAgentRunBuilder.modelProviderOptions(semanticModel)` for **module-level `ai:ModelProvider` variables**; only when that is empty does it emit `final ai:Wso2ModelProvider wso2ModelProvider = check ai:getDefaultModelProvider();`.

When the client's search finds a provider that `modelProviderOptions` does not surface (different module, non-module-level declaration, or a stale/pre-compile workspace), `hadModelProviderBeforeSave` is `true`, the config write is skipped — but the LS has just declared `wso2ModelProvider` anyway. That is exactly the `ballerina.ai.wso2ProviderConfig is not configured correctly` failure this change is meant to fix, and it now happens for a different reason.

The `catch` returning `true` (line 385) widens the same hole: any probe failure suppresses the config write, while the LS still decides independently to declare the provider.

Rather than predicting the LS's decision, consider having the source-generation response report whether it declared the default provider (or scanning `sourceCode.artifacts` for the `wso2ModelProvider` declaration) and driving the Config.toml write off that.

packages/ballerina-visualizer/src/MainPanel.tsx
```diff
@@ -824,6 +824,24 @@ const MainPanel = () => {
...
+                            // Editing resolves the activity from the file that declares it
+                            // (`value.documentUri`); creating one falls back to the default
+                            // functions file, same as the workflow and function forms.
+                            const defaultFunctionsFile = await getDefaultFunctionsFile();
+                            if (isStaleNavigation()) return;
+                            setViewComponent(
+                                <FunctionForm
+                                    key={remountKey}
+                                    projectPath={value.projectPath}
+                                    filePath={defaultFunctionsFile}
```
NipunaRanasinghe: **The comment says editing resolves the activity from `value.documentUri`, but nothing here reads `documentUri`.**

Lines 829-831 promise "Editing resolves the activity from the file that declares it (`value.documentUri`)"; the case only ever computes `getDefaultFunctionsFile()` and passes it as `filePath`.

That matters because `FunctionForm` turns `filePath` into a lookup key:

```ts
const fileName = filePath.split(/[\\/]/).pop();   // index.tsx:76
...
rpcClient.getBIDiagramRpcClient().getFunctionNode({ functionName, fileName, projectPath });  // :318-326
```

So editing an activity that is declared anywhere other than the default functions file (a second `.bal`, a non-default module) asks the LS for it in `functions.bal`. `BIDurableAgentForm`, directly above at line 810, does honour `value.documentUri ?? <default>` — the same pattern applies here.

Either use `value.documentUri ?? defaultFunctionsFile` (and keep the comment), or drop the comment so it stops describing behaviour the code does not have.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/LSPackageLoader.java
```diff
@@ -160,6 +160,24 @@ public void loadModules(LanguageServerContext context) {
...
+
+                // The user-home repositories hold the centrally-pulled and locally-published
+                // packages that do NOT ship inside the distribution (ballerina/workflow, for
+                // one). Without loading them here, the "import module" code action and
+                // unimported-module completions only ever offer distribution packages — a
+                // module the project itself pulled would still get no import suggestion.
+                // The empty listener map skips the listener-metadata compilation pass.
+                lsClientLogger.logTrace("Loading packages from the Ballerina user home repositories");
+                BallerinaUserHome ballerinaUserHome = BallerinaUserHome.from(environment);
+                this.remoteRepoPackages.addAll(checkAndResolvePackagesFromRepository(
```
NipunaRanasinghe, line 177: **Two problems with resolving the whole user-home repository at startup: cost, and duplicate versions.**

**1. Startup cost.** `checkAndResolvePackagesFromRepository` iterates every `org -> name:version` entry the repository reports and calls `repository.getPackage(request, offline)` for each one that survives the skip list. For the distribution repo that is a bounded, known set; `ballerinaUserHome.remotePackageRepository()` is the central cache under `~/.ballerina/repositories/central.ballerina.io`, which on a working developer machine routinely holds hundreds of packages across many versions. All of that is now loaded inside the "Indexing" progress phase, before the LS reports ready. The stated goal is one module (`ballerina/workflow`) getting an import suggestion — worth checking whether this can be scoped (e.g. to the packages the current project's `Dependencies.toml` names) rather than resolving the whole cache.

**2. Duplicate `ModuleInfo`s.** The dedupe set passed in is `packagesList.keySet()`, a snapshot taken *before* the call, and inside the loop the check is `loadedPackages.contains(packageIdentifier)` where `packageIdentifier` is `org/name` with **no version**. Entries added during the same call are never added back to `loadedPackages`, so a cache holding `ballerina/workflow:0.8.2` and `ballerina/workflow:0.8.3` appends two `ModuleInfo`s with the same identifier to `this.remoteRepoPackages`. `getAllVisiblePackages` then collapses them with `!packagesList.containsKey(...)`, keeping whichever happens to come first in `packageMap` iteration order — not necessarily the newest version. Same applies to the `localPackageRepository` call two lines down, which reuses the *same* pre-call snapshot and so does not even dedupe against what the remote pass just added.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentAddActivityBuilder.java
```diff
@@ -138,6 +140,7 @@ public void setConcreteTemplateData(TemplateContext context) {
                     .stepOut()
                 .value(preSelected)
                 .editable(true)
+                .hidden(!preSelected.isEmpty())
```
NipunaRanasinghe, line 145: **Hiding the selector on a module-qualified activity reference makes a silent binding loss unnoticeable.**

`hidden(!preSelected.isEmpty())` hides the activity selector whenever `codedata.symbol()` is set. On the edit path, `handleOnEditDurableCapability` (FlowDiagram/index.tsx) passes `capability.values.activity` straight through as that symbol, and `CodeAnalyzer.collectDeclaredCapabilities` explicitly handles `QUALIFIED_NAME_REFERENCE` — so `activities: [{activity: mod:validate, bindings: {...}}]` arrives here as `preSelected = "mod:validate"`.

`addBindingProperties(context, preSelected)` two lines below matches on the bare name:

```java
.filter(symbol -> activityName.equals(symbol.getName().orElse("")))
```

`"mod:validate".equals("validate")` is false, so no `bindings.*` properties are generated. The values the analyser hydrated for them (CodeAnalyzer lines 1564-1576) have nowhere to land, and `toSource` rebuilds the entry from `key.startsWith(BINDING_KEY_PREFIX)` — an empty set — so saving the edit drops the declared `bindings` field. With the selector hidden there is no visible cue that the form did not understand the entry.

Strip the module prefix before the symbol lookup (and/or match against the qualified name), or keep the selector visible when `addBindingProperties` found no matching activity symbol.

packages/bi-diagram/src/components/nodes/ApiCallNode/ApiCallNodeWidget.tsx
```diff
@@ -325,6 +344,25 @@ export function ApiCallNodeWidget(props: ApiCallNodeWidgetProps) {
...
+    // rather than swallowing the click.
+    const onWorkflowClick = async (event?: React.MouseEvent<SVGElement>) => {
+        event?.stopPropagation();
+        if (readOnly) {
+            return;
+        }
+        if (workflowTargetName) {
+            const functionLocation = await project?.getFunctionLocation?.(workflowTargetName);
+            if (functionLocation) {
+                openView && openView(functionLocation);
```
NipunaRanasinghe: **The fallback is skipped when `openView` is absent, so the click is swallowed — the case the comment says is handled.**

```ts
openView && openView(functionLocation);
setMenuPos(null);
return;
```

`openView` is optional on the diagram context (`DiagramContext.tsx` defaults it to a no-op, but consumers can pass `undefined`). When it is missing, the workflow *was* located, so the `return` fires and `onConnectionClick()` is never reached — the click does nothing at all, contrary to "falls back to the connection behaviour rather than swallowing the click". Moving the `return` inside the `if (openView)` (or checking `openView` alongside `functionLocation`) fixes it.

Minor, same handler: `event?.stopPropagation()` runs *before* the `readOnly` guard, whereas the previous handler for this element (`onConnectionClick`) returned early without stopping propagation. In read-only mode a click on the side icon no longer bubbles to the parent, which is a behaviour change for read-only diagrams.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
rdulmina: Fixed with 66754085a7 and 35dd6fe086

packages/ballerina-visualizer/src/views/BI/FunctionForm/index.tsx
rdulmina: Agreed that the two rules can disagree, and no change here — fixing it properly needs the generation response to say whether it declared the default provider. The artifacts carry only path/position, and `searchNodes({kind: "MODEL_PROVIDER"})` cannot tell a WSO2 default provider from any other, so there is nothing client-side to drive the decision off. That is an LS contract change beyond this PR; tracking it separately.

The `catch → true` is deliberate rather than an accident: writing the config needs a token, so treating an unknown project state as "already configured" avoids pulling a possibly signed-out user into the sign-in/logout path (the same hazard as the `pending-artifact.ts` thread). When it is skipped the provider is still configurable from the agent's model circle.

packages/bi-diagram/src/components/nodes/ApiCallNode/ApiCallNodeWidget.tsx
rdulmina: Both fixed in 469698d90a: the `return` moved inside the `functionLocation && openView` check, so a missing `openView` falls through to `onConnectionClick()`; and `stopPropagation()` now runs after the `readOnly` guard, leaving read-only diagrams bubbling as before. Covered by a new case in `bi-diagram/src/test/WorkflowRunNavigation.test.tsx`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
rdulmina: Stale — `bindsNothingNamed()` no longer exists. The branch now splits into `bindsWildcard()` (names nothing) and `bindsNilResult()` (`error? r = …`), and the latter seeds the name into the carrier the form reads. In the golden you cite, the `logEvent` node at line 18 carries:

```json
"variable": { "value": "r", "hidden": true, "optional": true }
```

so `r` survives the round trip and the statement regenerates unchanged (`to_source/config/activity_call_no_return_unchecked_to_source.json`). The single `count` you counted was from the shape before that rework.

The related gap your repro pointed at *was* real for the checked wildcard, and is fixed in a479b0ecfc: that node also gets the branch field and carrier now, so clearing Check Error asks for a name instead of falling back to a fixed one.

packages/ballerina-extension/src/features/bi/pending-artifact.ts
rdulmina: Valid, fixed in be44564b12 — the write now happens only while the AI session is authenticated (reusing the existing `isAIAuthenticated()`), so a signed-out user creating a durable agent is no longer logged out.

On the second point: no change. This path runs only for an artifact generated into the **freshly created** package (`generatePendingArtifact`, called from the pending-artifact pointer written when the wizard creates the package), so there is no module-level provider for `resolveExistingModelProvider` to find and the generation always declares `wso2ModelProvider`.

packages/ballerina-visualizer/src/MainPanel.tsx
rdulmina: Premise does not hold — `getDefaultFunctionsFile()` (MainPanel.tsx:313-320) returns `value.documentUri` when the caller set it and only falls back to `functions.bal`, so editing does resolve the declaring file and `FunctionForm` gets that path.

The local name was what read as "always the default file", so in a479b0ecfc it is renamed to `activityFile` and the comment now names the helper.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/LSPackageLoader.java
rdulmina: Part 2 fixed in efaf2bf1a1: only the newest version of each `org/name` is resolved, and each identifier joins a local dedupe set as it goes — so one pass can no longer append two `ModuleInfo`s with the same identifier, and the merge downstream no longer picks a version by map order. That also cuts the resolve calls the scan makes, since multi-version names are now resolved once.

Two corrections. The local-repo call is not given a stale snapshot: lines 175-176 fold the remote results into `packagesList` before it runs, and `keySet()` is a live view. And on cost — I re-ran `ExpressionContextTest` uncached with this hunk in place (324/0, plus `ImportModuleCodeActionTest` and `LSPackageLoaderTest` green), so it is not affecting completion results or the suite. Scoping the scan to the project's `Dependencies.toml` is worth doing, but as a separate change.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentAddActivityBuilder.java
rdulmina: Valid — confirmed the path you describe: `collectDeclaredCapabilities` stores `item.toSourceCode()` (so `mod:validate`), and `handleOnEditDurableCapability` passes it through as the template symbol.

Fixed in 5ad7acdf3d: the qualifier is stripped before the symbol lookup, so the binding selectors are built and the entry's `bindings` field round-trips. An activity from a *different package* still will not be found, but the selector only ever lists this package's activities, so that gap is unchanged by this PR.

packages/ballerina-visualizer/src/views/BI/FunctionForm/index.tsx
rdulmina: Tracked as wso2/product-integrator#2165 (area/workflow), with the suggested direction: have the generation report whether it declared the default provider instead of the client predicting it.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1729,6 +1771,7 @@ private void populateActivityCallProperties(RemoteMethodCallActionNode remoteMet
         }
 
         if (activityParamSymbols.isEmpty()) {
+            ActivityCallBuilder.addCheckErrorProperty(nodeBuilder, isCheckedCall(remoteMethodCallActionNode));
             addNormalizedRetryPolicyProperties(rawRetryPolicyValue);
```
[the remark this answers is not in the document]
rdulmina: This is added since this feature is needed

packages/ballerina-extension/src/features/bi/pending-artifact.ts
```diff
@@ -329,6 +335,32 @@ async function generatePendingArtifact(
...
+ * package root is already known here, so the config file is targeted directly (no project
+ * quick-pick). Failures are non-fatal: the agent exists and the provider can be configured
+ * from the agent's model circle.
+ *
+ * Only attempted while the user is signed in to the AI features: fetching the default model's
+ * token signs them out when it fails, so running this for a signed-out user would end their
+ * session as a side effect of creating an agent.
+ */
+async function configureDurableAgentModelProvider(projectRoot: string): Promise<void> {
+    if (!isAIAuthenticated()) {
```
NipunaRanasinghe, line 351: This guard doesn't cover the case the commit title is about. It stops the logout for users who are *already* signed out, but a signed-in user whose token fetch fails still gets signed out: `addConfigFile` fires `AIMachineEventType.LOGOUT` when the token comes back null (`features/ai/utils.ts:275-278`) and again from its own catch on any other failure (`:291-294`). The rethrow then lands in the `catch` below and only reaches `console.error`.

So a network blip while creating a durable agent silently ends the user's AI session with no message. The guard needs to be on the failure path rather than the entry condition.

Also worth reconciling with the other path that configures the same thing: `FunctionForm` calls `configureDefaultModelProvider`, which routes through `ballerina.configureWso2DefaultModelProvider` and has a `LOGIN_REQUIRED_WARNING_FOR_DEFAULT_MODEL` branch — it warns instead of logging out. Two implementations with different guards and different UX (that one shows a quick-pick, this one doesn't).

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
```diff
@@ -252,18 +290,232 @@ private void addRestInferredReturnTypeProperty(TemplateContext context, Function
...
+                                                      String fallbackVariable) {
+        boolean checkError = isCheckError(sourceBuilder);
+        String variableName = nonBlankValue(sourceBuilder.getProperty(Property.VARIABLE_KEY));
+
+        if (declaredType == null || isNilResultType(declaredType)) {
+            if (checkError) {
+                return new ResultBinding(NIL_RESULT_TYPE, WILDCARD_RESULT_VARIABLE, true);
+            }
+            return new ResultBinding(NIL_UNCHECKED_RESULT_TYPE,
+                    variableName == null ? fallbackVariable : variableName, false);
```
NipunaRanasinghe: The fallback name isn't uniquified, unlike the template path.

`generatedNilResultName` (line 388) seeds names through `NameUtil.generateTypeName(..., getAllVisibleSymbolNames())`, so they can't collide. This falls back to the bare constant `"activityResult"`. Reading back `() _ = check ...` sets the root `variable` to `""`, so if two such calls are unchecked and submitted without a typed name, both emit `activityResult` — `redeclared symbol`, which is the bug class this PR fixes.

It's unreachable as long as the branch's Result field is enforced as required in the UI, but that puts a UI-layer defence in front of a server-side invariant. Uniquifying here (or rejecting a blank name) would close it properly.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
```diff
@@ -523,15 +780,14 @@ public Map<Path, List<TextEdit>> toSource(SourceBuilder sourceBuilder) {
...
-                .map(p -> p.value().toString())
-                .orElse(DEFAULT_RETURN_TYPE);
-        String variableName = variableProp
-                .map(p -> p.value().toString())
-                .orElse("result");
+        // An activity that returns nothing carries no result type in the form. Falling back to a fixed
+        // `anydata result` here made every such call declare the same variable, which is a duplicate
+        // symbol as soon as there are two of them.
+        String declaredType = nonBlankValue(sourceBuilder.getProperty(Property.TYPE_KEY));
+        ResultBinding binding = resolveResultBinding(sourceBuilder, declaredType, DEFAULT_NIL_RESULT_VARIABLE);
```
NipunaRanasinghe: The two paths disagree on the fallback. `toSourceBuiltin` at line 876 correctly picks per shape:

```java
hasReturnValue ? "result" : DEFAULT_NIL_RESULT_VARIABLE
```

This one passes `DEFAULT_NIL_RESULT_VARIABLE` unconditionally, so a value-returning user activity with no name in the form would bind `activityResult` rather than `result`. Probably unreachable since the form always seeds a name for value-returning activities, but the two should read the same.

packages/ballerina-side-panel/src/components/editors/utils.ts
```diff
@@ -147,6 +147,21 @@ export const getFieldKeyForAdvanceProp = (fieldKey: string, advancePropKey: stri
...
+ * node being edited. Returns the field carrying whatever the form currently holds for its key, since
+ * without it the field's editor mounts with the definition's empty value and overwrites what was loaded.
+ * Nested advanced children follow the same rule.
+ */
+export function withHeldValue(field: FormField, values: Record<string, any> | undefined): FormField {
+    const held = values?.[field.key];
+    const withChildren = field.advanceProps
+        ? { ...field, advanceProps: field.advanceProps.map((child) => withHeldValue(child, values)) }
+        : field;
+    return held === undefined || held === "" ? withChildren : { ...withChildren, value: held };
```
NipunaRanasinghe: Treating `""` the same as `undefined` means a cleared field can't stay cleared — it snaps back to the definition's value on the next render.

Harmless today because every branch definition uses `.value("")`, so both sides are empty and you can't tell the difference. But it's a trap for the first branch field that carries a non-empty default. Checking only `held === undefined` would keep "cleared" distinct from "never set".

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/LSPackageLoader.java
```diff
@@ -297,25 +317,38 @@ public List<ModuleInfo> checkAndResolvePackagesFromRepository(
...
                     return;
                 }
-                String nameComponent = components[0];
-                String version = components[1];
-                PackageOrg packageOrg = PackageOrg.from(key);
+                try {
+                    PackageVersion candidate = PackageVersion.from(components[1]);
+                    latestVersions.merge(components[0], candidate, (existing, latest) ->
+                            latest.value().greaterThan(existing.value()) ? latest : existing);
+                } catch (Throwable e) {
```
NipunaRanasinghe: `catch (Throwable)` swallows `Error` along with the bad-version exception you're after. `catch (Exception)` covers the intent.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/LSPackageLoader.java
```diff
@@ -160,6 +162,24 @@ public void loadModules(LanguageServerContext context) {
...
+
+                // The user-home repositories hold the centrally-pulled and locally-published
+                // packages that do NOT ship inside the distribution (ballerina/workflow, for
+                // one). Without loading them here, the "import module" code action and
+                // unimported-module completions only ever offer distribution packages — a
+                // module the project itself pulled would still get no import suggestion.
+                // The empty listener map skips the listener-metadata compilation pass.
+                lsClientLogger.logTrace("Loading packages from the Ballerina user home repositories");
+                BallerinaUserHome ballerinaUserHome = BallerinaUserHome.from(environment);
+                this.remoteRepoPackages.addAll(checkAndResolvePackagesFromRepository(
```
NipunaRanasinghe, line 177: Two things to confirm about resolving both user-home repositories at startup:

1. Cost. This now walks and resolves everything in `~/.ballerina` on every LS start. The newest-version-only change below keeps it from being quadratic, but for a developer with a large cache it's still new startup work. Was it measured?
2. `addAll` accumulates rather than replaces. If `loadModules` can ever run more than once, `remoteRepoPackages` and `localRepoPackages` grow with duplicates. Fine if it's one-time init — worth a guard or a comment saying so.

The reason for the change is clear and the comment explains it well; this is just about the cost side.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentAddActivityBuilder.java
```diff
@@ -237,7 +247,7 @@ private void addToolIdentityProperties() {
                 .value("")
                 .editable(true)
                 .optional(true)
-                .advanced(true)
+                .hidden(true)
```
NipunaRanasinghe: Moving these two from `advanced(true)` to `hidden(true)` removes a knob the user could set, not just a rendering detail.

The old comment called them "the name/description advertised to the model"; the new one calls them hidden carriers that exist only so a declared value round-trips. For an agent tool, the name and description are what the LLM selects on, so this changes behaviour rather than presentation.

I couldn't find it in the PR description under any of the nine issues listed, so I can't tell whether it's intended product direction or a side effect of making the bindings round-trip work. If it's intended, it should be called out in the description.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentAddActivityBuilder.java
```diff
@@ -151,11 +154,17 @@ public void setConcreteTemplateData(TemplateContext context) {
...
+     * <p>An entry declared with a module-qualified reference ({@code mod:validate}) arrives here as
+     * written, while symbols carry the bare name — so the qualifier is stripped before the lookup.
+     * Without that, no binding selector is built, the values the analysis hydrated for them have
+     * nowhere to land, and saving the edit drops the entry's {@code bindings} field.
      */
     private void addBindingProperties(TemplateContext context, String activityName) {
         if (activityName == null || activityName.isEmpty()) {
             return;
         }
+        String unqualifiedName = activityName.substring(activityName.indexOf(':') + 1);
```
NipunaRanasinghe: Third module-prefix stripper in the codebase, and the second convention: this uses `indexOf(':')`, `CodeAnalyzer.stripModulePrefix` uses `lastIndexOf(':')`, and the new TS `getWorkflowFunctionName` uses `split(":").pop()`.

They agree on every valid Ballerina name so nothing is broken. Just worth reusing one instead of adding a variant.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/index.tsx
```diff
@@ -2339,6 +2349,20 @@ export function BIFlowDiagram(props: BIFlowDiagramProps) {
...
+        // targetLineRange — the expression-editor probe position at the declaration START
+        // (see probeRangeForCapability). Submitting that range would splice the entry text
+        // into the declaration line, so restore the entry range stamped on the selected node
+        // by handleOnEditDurableCapability.
+        if (
+            updatedNode.codedata?.isNew === false &&
+            Object.values(DURABLE_CAPABILITY_NODE_KINDS).includes(updatedNode.codedata.node) &&
+            selectedNodeRef.current?.codedata?.lineRange
+        ) {
+            updatedNode.codedata.lineRange = selectedNodeRef.current.codedata.lineRange;
```
NipunaRanasinghe: The comment is honest that the form stamps the wrong range and this puts it back, but two things make the workaround fragile:

1. It trusts `selectedNodeRef.current` to still point at the edited node at submit time. This file has `beginPanelNav`/`isStaleNavigation` for exactly that race, and neither is checked here — a stale ref splices the entry into the wrong location.
2. Any capability node kind missing from `DURABLE_CAPABILITY_NODE_KINDS` silently falls through and writes into the declaration line instead of failing.

Fixing the re-stamp where it happens would remove both. If that's too big for this PR, an assertion that the restored range actually belongs to the node being edited would at least make a stale ref loud instead of silent.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/index.tsx
```diff
@@ -3538,20 +3562,16 @@ export function BIFlowDiagram(props: BIFlowDiagramProps) {
...
+        // Title the form with the chosen activity, like the edit path does.
+        activityNode.metadata = {
+            ...activityNode.metadata,
+            label: activityCodedata?.symbol || activityNode.metadata?.label,
+        } as any;
+        selectedNodeRef.current = activityNode;
+        nodeTemplateRef.current = activityNode;
+        showEditForm.current = false;
+        setSidePanelView(SidePanelView.FORM);
+        setShowSidePanel(true);
```
NipunaRanasinghe, line 3608: This drops the branch that ended with `durableAgentActivityListRef.current = false; finishCapabilityOpAfterRefresh();`, so every activity-list pick now opens the form.

On save that's fine — the post-write reset around line 2001 clears the flag. On cancel it stays `true`. That shape already existed for binding-requiring activities, so it isn't new, but it's now reachable for every activity rather than a subset. Worth confirming cancel resets it, since a stale flag changes how the next search is filtered (lines 712 and 1356 both read it).

packages/ballerina-extension/src/features/bi/pending-artifact.ts
rdulmina: You're right, and the guard was in the wrong place. Confirmed the mechanism: `addConfigFile` sends `LOGOUT` when the token comes back null (`features/ai/utils.ts:293`) and again from its own catch (`:312`), and `configureDurableAgentModelProvider` swallows the rethrow into `console.error` — so a signed-in user whose token fetch failed lost their AI session with no message. `isAIAuthenticated()` only ever spared users who were already signed out.

Fixed in f78f20b7bd by moving the guard onto the failure path: `addConfigFile` takes an `AddConfigFileOptions.signOutOnFailure`, defaulting to `true` so both interactive callers in `features/ai/activator.ts` are unchanged, and the durable agent's provider write passes `false`. A network blip while creating an agent no longer ends the session.

On the second half — reconciling with `configureDefaultModelProvider` / `LOGIN_REQUIRED_WARNING_FOR_DEFAULT_MODEL`, so the two paths share a guard and a UX: agreed that two implementations with different behaviour is the real problem, but that spans the AI and BI features and changes the interactive flow's UX, which is wider than this PR. Not done here.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
rdulmina: Valid — the fallback was the bare constant while `generatedNilResultName` seeds through `NameUtil.generateTypeName(..., getAllVisibleSymbolNames())`, so two calls reaching it would both bind `activityResult`.

Fixed in 6a297d7f81: the fallback now goes through `uniqueFallbackName`, which runs the same uniquifier against the symbols visible at the statement's position. The lookup is lazy — only the fallback path loads a semantic model — and returns an empty set on any failure, since a name that is merely not uniquified beats failing the whole source generation.

One note on the alternative you offered: rejecting a blank name would have removed the fallback that the comment on line 787 asks to bring into line with `toSourceBuiltin`, so the two suggestions can't both be taken. Keeping the fallback and uniquifying it satisfies both diagnoses — the shape-correct base name is chosen first, then made unique.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
rdulmina: Confirmed — `toSourceUserActivity` passed `DEFAULT_NIL_RESULT_VARIABLE` unconditionally, so a value-returning user activity with no name in the form would have bound `activityResult` rather than `result`.

Fixed in 6a297d7f81. Both paths now derive `hasReturnValue` and pick the same way, and the value-returning fallback is a named constant (`DEFAULT_RESULT_VARIABLE`) that both read, rather than a literal in one place and a constant in the other — so they can't drift again.

Note the fallback is also uniquified now (see the thread on line 504), which is why it stays a fallback rather than becoming a rejected-blank-name error.

packages/ballerina-side-panel/src/components/editors/utils.ts
rdulmina: Agreed. Fixed in 56628c16e1 — the check is `held === undefined` only, so a cleared field stays cleared and "never set" stays distinct from "set to empty".

Your read on why it's currently invisible matches what's there: every branch definition uses `.value("")`, so both sides are empty and the two cases are indistinguishable today. The comment now records that, so the next branch field with a real default doesn't inherit the trap.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/LSPackageLoader.java
rdulmina: Correct — `PackageVersion.from` throws an exception for an unreadable version, so `Exception` covers the intent and `Throwable` was additionally swallowing `Error`. Changed to `catch (Exception)` in f24220703f.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/LSPackageLoader.java
rdulmina: Taking the two separately.

**2. `addAll` accumulating.** It is one-time init: `loadModules` has a single caller, `BallerinaLanguageServer#initialized` (`BallerinaLanguageServer.java:232`, behind `isEnableIndexPackages`), so `remoteRepoPackages` and `localRepoPackages` are filled once and cannot accumulate duplicates as things stand. Rather than add a guard against a call pattern that doesn't exist, f24220703f records the invariant in a comment, so a second caller has to notice it. Happy to make it a guard instead if you'd rather the code enforce it than document it.

**1. Cost.** Not measured — I should be straight about that rather than imply otherwise. What I can say is what bounds it: the newest-version-only change means one resolution per org/name instead of one per version, and both repositories are resolved offline (`setOffline(true)`), so it's local disk work with no network in the loop. That is an argument that it's not quadratic, not evidence about the constant on a large cache. If you want a number before this merges, the honest thing is a timing run against a populated `~/.ballerina`; tell me if you'd like that gathered and I'll do it.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentAddActivityBuilder.java
rdulmina: You read it correctly, and the objection stands: hiding them removed a knob rather than changing a rendering detail. It was a side effect of making the round-trip work, not intended product direction.

Reverted in 989af846e8 — the two are back as form fields, `advanced(true)` and visible, in the advanced section rather than the main body since the activity's own identity is enough for the common case. They're labelled for what they are now that a user reads them: "Activity Name" and "Activity Description" (they name the activity as the model sees it, and the earlier "Tool Name" label predated the activity framing).

The round-trip that motivated the original change still holds and needs no analysis change: an entry's `name`/`description` source fields already hydrate onto exactly these property keys (`CodeAnalyzer.java:1490`) and both are in `TEXT_MODE_CAPABILITY_FIELDS`, so a declared value arrives unquoted into the text field and `quoteIfPlain` re-quotes it on save. `durable_agent_edit_activity_identity_to_source.json` covers it: an entry with both set is rewritten with both intact.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentAddActivityBuilder.java
rdulmina: Fair — nothing was broken, but a third variant of the same operation isn't worth it. Consolidated in 989af846e8: `stripModulePrefix` moves to `WorkflowUtil` as the one Java implementation (keeping the `lastIndexOf` convention), `CodeAnalyzer`'s private copy is deleted, and the binding lookup here calls it instead of doing its own `indexOf` slice.

The TS `getWorkflowFunctionName` can't share it, so that one stays as it is — `split(":").pop()` agrees with `lastIndexOf` on every valid name, so it's the same convention at least, just expressed in the other language.

(There's a fourth in `service-model-generator`'s `ExistingListenerResolver` — different Gradle module, so folding it in would mean deciding where a cross-module string util lives. Left alone here.)

packages/ballerina-visualizer/src/views/BI/FlowDiagram/index.tsx
rdulmina: Both sub-points are fair, and I've taken the fallback you offered rather than the full fix.

56628c16e1 checks that `selectedNodeRef.current` still describes the node being submitted (same capability kind, same `parentSymbol`) before adopting its range, and `console.error`s with both codedatas when it doesn't, keeping the submitted range instead. So a stale ref is loud and lands nowhere destructive, rather than splicing the entry over another node's source.

Two things I'm not claiming it fixes:

- The check discriminates on kind + agent, so it catches a ref that moved to a different agent or a different capability kind. Two entries of the same kind on the same agent would still pass it. Narrowing further needs a per-entry identity the template response doesn't carry today.
- Sub-point 2 stands as written: a capability kind absent from `DURABLE_CAPABILITY_NODE_KINDS` still falls through to the declaration line. All five kinds are registered, so it's a future footgun rather than a present bug, and the guard above doesn't help since it's inside the same condition.

The real fix is not re-stamping the range in the form, which is where you point — that's a form-layer change beyond this PR's scope, and this workaround stays until it's done.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/index.tsx
rdulmina: Confirmed, and cancel did **not** reset it: `handleOnCloseSidePanel` calls `resetNodeSelectionStates`, which clears a long list of refs but not `durableAgentActivityListRef` — so after cancelling, the flag stayed `true` and the next search was still filtered in the agent's mode (lines 792 and 1436).

Fixed in 56628c16e1 by clearing it, and `activityWizardForAgentRef` with it, in `handleOnCloseSidePanel` — `inDurableAgentActivityFlow()` ORs the two, so clearing one alone wouldn't have been enough.

Deliberately not in `resetNodeSelectionStates`: that also runs on post-write refreshes, and the flow is built to survive those — it's the situation the comment above `handleOnAddActivityFromConnection` describes, where a write inside the wizard clears the flag while the wizard stays open. Only an explicit close means cancel, so only that clears them.

Your read on save is right too — the post-write reset around line 2001 already handles it — so this is the cancel path only.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/index.tsx
```diff
@@ -2339,6 +2357,35 @@ export function BIFlowDiagram(props: BIFlowDiagramProps) {
...
+                selectedCodedata.node === updatedNode.codedata.node &&
+                selectedCodedata.parentSymbol === updatedNode.codedata.parentSymbol;
+            if (sameNode) {
+                updatedNode.codedata.lineRange = selectedCodedata.lineRange;
+            } else {
+                console.error(
+                    ">>> Stale selected node while submitting a durable agent capability; " +
+                        "keeping the submitted range",
+                    { submitted: updatedNode.codedata, selected: selectedCodedata }
+                );
```
NipunaRanasinghe: Thanks for adding the check — two gaps left in it.

**The identity test can't catch the case that actually happens.** Two activity entries on the same agent share both `node` and `parentSymbol`, so opening entry A's form, navigating, then submitting entry B passes `sameNode` and still splices over the wrong entry. That's the realistic stale-ref path; a ref pointing at a different *kind* of capability is the unlikely one. `codedata.symbol` would separate two entries of the same kind.

**The `else` branch still does the unsafe thing.** Keeping the submitted range means using the probe position at the declaration start — exactly the corruption the restore exists to prevent. So on a detected stale ref we log and then corrupt anyway. Since we know at that point that we can't place the edit, aborting the submit and surfacing it through `showConnectorError` would be safer than a `console.error` the user never sees.

Agreed that fixing the re-stamp at its source would make all of this unnecessary.

NipunaRanasinghe (dismissed):

packages/ballerina-visualizer/src/views/BI/FlowDiagram/index.tsx
rdulmina: Both gaps are real. Fixed in b3a0dc9b12 — with one divergence on the first.

**The `else` branch.** You're right, and this was the worse of the two: it detected the stale ref and then performed the corrupting write anyway, so the log was a record of the problem rather than a defence against it. It now aborts — `showConnectorError()` and return, before the progress indicator is turned on. I also folded in a case the original condition let through silently: when the selected node carried no `lineRange`, the whole block was skipped and the submit proceeded with the declaration range. That aborts too.

**The identity test.** Diagnosis right, but `codedata.symbol` won't separate two activity entries — it's constant per capability kind, exactly like `node`. `DurableAgentAddActivityBuilder.setConcreteTemplateData` calls `setConcreteConstData()`, which stamps `.symbol(REGISTER_ACTIVITY_METHOD_NAME)` unconditionally, so every activity template comes back with `symbol: "registerActivity"` whatever activity it registers. `node_template/config/durable_agent_add_activity_preselected.json` is the case that proves it — the template is requested *with* the activity function as the symbol and the output codedata is still:

```json
{"node": "DURABLE_AGENT_ADD_ACTIVITY", "org": "ballerina", "module": "workflow", "object": "AgenticWorkflowContext", "symbol": "registerActivity", "isNew": true}
```

So swapping it in would have looked like a fix while changing nothing.

What does separate two same-kind entries is the capability's name, which `handleOnEditDurableCapability` stamps as `metadata.label` from `capability.name`, so that's the discriminator now. It survives the form: `createNodeWithUpdatedLineRange` spreads the node and replaces only `codedata.lineRange`, and `updateNodeWithProperties` spreads and replaces only `properties`/`branches[0].properties` — which matters, because a false mismatch here would now block every capability edit rather than just log.

Remaining honestly: two entries registering the *same* activity twice would still be indistinguishable, and that ambiguity is inherent to matching on fields rather than a per-entry identity the template doesn't carry. And I agree the re-stamp at its source is the fix that makes all of this unnecessary — this stays a guard until that's done.

samithkavishke (approved):
LGTM

NipunaRanasinghe (approved):

PR #870, opened 2026-08-15
Fix evaluations not appearing in the test explorer and backslash paths on Windows

dan-niles:
## Purpose

Fixes two Windows-only defects in the AI evaluation flow, plus a leftover-code bug in evaluation deletion.

- Fixes wso2/product-integrator#1867 — created evaluations do not appear under **Evaluations** in Testing until the window is reloaded.
- Fixes wso2/product-integrator#1865 — auto-generated evalset paths use backslashes, producing `invalid escape sequence` (BCE0649) compile errors.
- Deleting an evaluation now also removes the `loadEvalsetData<N>` data provider generated with it, instead of leaving it orphaned in `tests.bal`.

## #1867 — evaluations missing from the Test Explorer

`testManagerService/discoverInFile` takes a **file** path, despite the request field being named `projectPath`. It was being sent `uri.path`, which on Windows is `/c:/...` — `Path.of()` rejects that with `InvalidPathException: Illegal char <:> at index 2`.

The LS catches `Throwable` and returns `TestsDiscoveryResponse.from(e)` (`result == null`), and the caller early-returns on a missing `result`, so the failure was completely silent. On macOS/Linux `uri.path === uri.fsPath`, which is why this only reproduced on Windows.

Changes:

- Send `uri.fsPath` — this is the actual root cause.
- Refresh the tree directly from `addTestFunction` / `updateTestFunction` rather than relying solely on the file watcher, so the new evaluation shows up as soon as the RPC returns.
- Debounce refreshes per file. `updateSourceCode` applies a raw edit and then a formatted edit, each triggering a save and so a watcher event; the two discovery requests had no ordering guarantee and could leave a stale tree.

## #1865 — backslashes in generated evalset paths

Already covered by 3ec3678 on this branch (normalises the generated path separators to `/`). This PR adds a follow-up commit wrapping a line that commit pushed to 121 characters, which was failing `:test-manager-service:test-manager-service-ls-extension:checkstyleMain` and therefore the whole `rush build --to ballerina`.

## Checks

- `rush build --to ballerina` gets past the language-server build (checkstyle was the blocking failure).
- `checkstyleMain` and `checkstyleTest` pass for the test-manager-service module.
- `tsc --noEmit` is clean for the three changed TypeScript files.

Not yet verified on a Windows machine — the root-cause analysis is from the LS sources rather than a live repro, so a Windows sanity check before merge would be worthwhile.

## Summary

- Fix Windows test discovery by using `uri.fsPath`.
- Refresh the Testing test tree after evaluation creation or updates.
- Sequence and debounce per-file refreshes to prevent stale test items.
- Normalize generated evalset paths to `/`.
- Remove unused generated data providers when evaluation tests are deleted.
- Validate Git SHA inputs before generating diff commands.
- Apply test-manager-service checkstyle corrections.


packages/ballerina-extension/src/features/test-explorer/activator.ts
```diff
@@ -144,8 +152,8 @@ export async function activate(ballerinaExtInstance: BallerinaExtension) {
     const fileWatcher = workspace.createFileSystemWatcher('**/tests/**/*.bal');
 
     // Handle file creation, modification, and deletion
-    fileWatcher.onDidCreate(async (uri) => await handleTestFileUpdate(ballerinaExtInstance, uri, testController));
-    fileWatcher.onDidChange(async (uri) => await handleTestFileUpdate(ballerinaExtInstance, uri, testController));
+    fileWatcher.onDidCreate((uri) => debouncedHandleFileChange(ballerinaExtInstance, uri, testController));
+    fileWatcher.onDidChange((uri) => debouncedHandleFileChange(ballerinaExtInstance, uri, testController));
     fileWatcher.onDidDelete((uri) => handleTestFileDelete(uri, testController));
```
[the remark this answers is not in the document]
dan-niles: `fileWatcher` is already registered for disposal at activator.ts:163, and disposing it stops event delivery — matching the pattern in `features/testing/discover.ts`, `evalset-tree-view.ts` and `agents-md/index.ts`.

packages/ballerina-extension/src/features/test-explorer/commands.ts
```diff
@@ -190,13 +190,17 @@ export function activateEditBiTest(ballerinaExtInstance: BallerinaExtension) {
...
         try {
             const response = await ballerinaExtInstance.langClient?.getTestFunction({
                 functionName: entry.label,
                 filePath: fileUri
             });
 
             if (response && isValidTestFunctionResponse(response) && response.function) {
+                dataProviderName = response.function.annotations
+                    ?.find((a: Annotation) => a.name === 'Config')?.fields
+                    ?.find((f: ValueProperty) => f.originalName === 'dataProvider')?.value as string;
```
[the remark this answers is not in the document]
dan-niles: Fixed in d6a50a1a2e — the data provider name is now only read when the test belongs to the evaluations group.

packages/ballerina-extension/src/features/test-explorer/commands.ts
```diff
@@ -190,16 +190,19 @@ export function activateEditBiTest(ballerinaExtInstance: BallerinaExtension) {
...
             });
 
             if (response && isValidTestFunctionResponse(response) && response.function) {
-                const isEvaluation = hasEvaluationGroup(response.function);
-                if (isEvaluation) {
+                if (hasEvaluationGroup(response.function)) {
                     testType = "AI evaluation test";
+                    dataProviderName = response.function.annotations
+                        ?.find((a: Annotation) => a.name === 'Config')?.fields
+                        ?.find((f: ValueProperty) => f.originalName === 'dataProvider')?.value as string;
```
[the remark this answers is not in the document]
dan-niles: Fixed in 88257a5d65. The cleanup now runs only when the provider name carries the generated `loadEvalsetData`/`loadQueriesData` prefix, mirroring the ownership check in `TestManagerService.replaceDataProvider`. Custom or shared providers are left untouched.

MohamedSabthar (dismissed):

packages/ballerina-extension/src/features/test-explorer/commands.ts
```diff
@@ -317,6 +322,34 @@ function hasEvaluationGroup(testFunction: any): boolean {
     return hasEvaluation;
 }
 
+/** Adds the removal of the data provider generated alongside an evaluation. */
+async function addDataProviderDeletion(edit: WorkspaceEdit, ballerinaExtInstance: BallerinaExtension,
+    fileUri: string, name?: string) {
+    // Custom providers may be shared by other tests; only generated ones are safe to delete.
```
samithkavishke, line 362: The generated data-provider deletion heuristic matches by name prefix (`loadEvalsetData`/`loadQueriesData`). If two AI-evaluation tests in the file end up referencing the same generated provider name (e.g. reused/stale reference), deleting one test via `BI_DELETE_TEST_FUNCTION` will remove the provider function via `addDataProviderDeletion`, silently breaking the remaining test's `dataProvider` reference with no warning to the user. Worth double-checking that the provider is not referenced elsewhere before deleting it.

pasindufernando1 (dismissed):
LGTM — the review suggestions are addressed. Verified against the latest commits:

- `refreshTestsForFile` is now wrapped at both call sites, so it can't leave the RPC promise unsettled.
- The per-file sequence token in `handleFileChange` prevents a superseded run from overwriting a newer tree.
- Generated data provider names are matched exactly (`/^(loadEvalsetData|loadQueriesData)\d*$/`), and `isDataProviderUsedElsewhere` guards against deleting one that's still referenced.
- `fn && isValidTestFunctionResponse(fn)` guard added.
- The eval-set path is normalized before the comparison, so no more no-op edits on Windows.

Non-blocking: the delete confirmation still names only the test function, so removing the data provider alongside it isn't visible to the user.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

packages/ballerina-extension/src/features/test-explorer/discover.ts
```diff
@@ -201,19 +205,53 @@ export async function handleFileChange(ballerinaExtInstance: BallerinaExtension,
...
+}
+
+// Only for the watcher's delete event; handleFileDelete is also reused mid-refresh by handleFileChange.
+export function cancelPendingFileChange(uri: Uri) {
+    const key = uri.fsPath;
+    const existing = fileChangeTimers.get(key);
+    if (existing) {
+        clearTimeout(existing);
+        fileChangeTimers.delete(key);
+    }
```
dulajdilshan, line 260: **[Medium] Wiring — residual of the in-flight-run fix.** `cancelPendingFileChange` clears the debounce timer but never touches `fileChangeSequence`, so a `handleFileChange` already past its `await` at `:216` is not invalidated: the delete event runs `handleFileDelete`, then the in-flight run passes the check at `:222` and re-creates test items for a file that no longer exists.

The sequence map exists precisely to express "this run is superseded", and the delete path is the one producer that doesn't use it — the cost is ghost tests that fail when run. Demoted from Major: whether the symptom is observable depends on the LS returning stale functions for a just-deleted file, which I couldn't verify; the gap in the invalidation machinery is certain either way.

Would bumping the counter here close it with the machinery already present?

```suggestion
export function cancelPendingFileChange(uri: Uri) {
    const key = uri.fsPath;
    const existing = fileChangeTimers.get(key);
    if (existing) {
        clearTimeout(existing);
        fileChangeTimers.delete(key);
    }
    // Supersede any in-flight run so it can't re-add the file.
    fileChangeSequence.set(key, (fileChangeSequence.get(key) ?? 0) + 1);
}
```

packages/ballerina-extension/src/features/test-explorer/discover.ts
```diff
@@ -27,6 +27,10 @@ import { URI } from "vscode-uri";
 import { getTestFunctionGroups } from "../../utils/test-discovery";
 
 let groups: string[] = [];
+const fileChangeTimers = new Map<string, NodeJS.Timeout>();
+// Tracks the latest handleFileChange run per file so a slower, superseded run can't overwrite a newer tree.
+const fileChangeSequence = new Map<string, number>();
+const FILE_CHANGE_DEBOUNCE_MS = 300;
```
dulajdilshan, line 33: **[Medium] Module state with no lifecycle.** `fileChangeSequence` is written on every change but never deleted — unlike `fileChangeTimers`, which self-cleans at `:239` — and neither map is cleared when the controller is disposed at `activator.ts:163`.

It grows for the life of the window, one entry per test file ever touched. Small, but it's unbounded state in a module that otherwise cleans up after itself. Deleting the key in `handleFileDelete` (or alongside the timer in `cancelPendingFileChange`) would keep it bounded — does that fit how you'd rather scope it?

packages/ballerina-extension/src/features/test-explorer/commands.ts
```diff
@@ -317,6 +321,63 @@ function hasEvaluationGroup(testFunction: any): boolean {
...
+        ?.find((f: ValueProperty) => f.originalName === 'dataProvider')?.value as string | undefined;
+}
+
+/** True if another test function in the file still references this data provider. */
+async function isDataProviderUsedElsewhere(ballerinaExtInstance: BallerinaExtension, fileUri: string,
+    excludeFunctionName: string, providerName: string): Promise<boolean> {
+    const discovery: TestsDiscoveryResponse = await ballerinaExtInstance.langClient?.getFileTestFunctions({
+        projectPath: fileUri
+    });
+    const otherNames = getTestFunctionNames(discovery).filter((n) => n !== excludeFunctionName);
```
dulajdilshan: **[Medium] Wiring — residual of the shared-provider check.** `getFileTestFunctions({ projectPath: fileUri })` returns only the functions in the deleted test's own file (same call shape as `discover.ts:209`), but Ballerina test functions in sibling files of the same test module can reference `loadEvalsetData` too — so the guard can't see them and the provider is still deleted out from under them.

I checked every producer feeding this: `getTestFunctionNames` derives purely from that one file-scoped response, so there's no wider source in the current path. `getProjectTestFunctions` (already used at `rpc-managers/test-manager/rpc-manager.ts:125`) is the module-wide equivalent — would swapping it in be acceptable, given the cost is one extra LS call on a delete?

packages/ballerina-extension/src/features/test-explorer/commands.ts
```diff
@@ -317,6 +321,63 @@ function hasEvaluationGroup(testFunction: any): boolean {
...
+
+    for (const testName of otherNames) {
+        const fn = await ballerinaExtInstance.langClient?.getTestFunction({ functionName: testName, filePath: fileUri });
+        if (fn && isValidTestFunctionResponse(fn) && getDataProviderName(fn.function) === providerName) {
```
dulajdilshan: **[Medium] Sequential LS round-trips in a loop.** One `getTestFunction` call per other test function in the file, awaited serially, inside the modal-confirmed delete path — a file with 30 evaluations means 30 serial RPCs before the edit applies.

The volume is what changes without the code changing, so the linear-scan-of-round-trips shape is worth removing now rather than when someone reports a slow delete. `Promise.all` over `otherNames` keeps the same shape; having the LS return annotations with the discovery response would remove the N+1 entirely. Which would you prefer?

packages/ballerina-extension/src/features/test-explorer/commands.ts
```diff
@@ -317,6 +321,63 @@ function hasEvaluationGroup(testFunction: any): boolean {
...
+            filePath: fileUri,
+            component: {
+                name, filePath: fileUri,
+                startLine: range.startLine.line, startColumn: range.startLine.offset,
+                endLine: range.endLine.line, endColumn: range.endLine.offset
+            }
+        });
+        for (const [file, edits] of Object.entries(res?.textEdits ?? {})) {
+            edits.forEach((e) => edit.replace(Uri.file(file), new Range(e.range.start.line,
+                e.range.start.character, e.range.end.line, e.range.end.character), e.newText));
```
dulajdilshan: **[Medium] Duplicated Code (DRY).** This `Object.entries(textEdits)` → `edit.replace(Uri.file(...), new Range(...))` loop is a re-implementation of `:252-264` in the same file, twelve lines apart.

I grepped for a shared helper (`applyTextEdits`, `toWorkspaceEdit`, `Object.entries(*.textEdits)`) — the only other instance is `features/ai/agent/utils.ts:324`, which writes to disk rather than into a `WorkspaceEdit`, so nothing existing fits. Two call sites in one file is a recurring concept rather than repeated characters; what do you think about a local `addTextEdits(edit, textEdits)` collapsing both?

packages/ballerina-extension/src/utils/git-utils.ts
```diff
@@ -94,6 +94,8 @@ export async function pinSnapshot(cwd: string, sha: string): Promise<void> {
 
 export async function getDiffStat(cwd: string, from: string, to: string): Promise<string> {
     try {
+        assertSha(from);
+        if (to !== "HEAD") assertSha(to);
```
dulajdilshan: **[Medium] Dead branch, and the throw is swallowed.** The only producer of `toSha` is `ballerina-visualizer/src/views/EvaluationHistory/RunHistoryTable.tsx:293`, which passes the literal `"HEAD"` — I grepped every `toSha` reference in `packages/` and the submodule — so `if (to !== "HEAD") assertSha(to)` never executes (same at `:109-110`). And `assertSha(from)` throws *inside* a `try` whose `catch` returns `''`, so a malformed `from` degrades to a silently empty diff rather than a reported error.

A guard that can't fire is a lie to the next reader, and one that reports nothing when it does fire doesn't buy the validation it looks like it buys. Should the `assertSha` calls sit outside the `try` so a bad SHA surfaces, and the `to` arm be dropped until a non-HEAD caller exists?

packages/ballerina-extension/src/features/test-explorer/commands.ts
```diff
@@ -263,6 +265,8 @@ export function activateEditBiTest(ballerinaExtInstance: BallerinaExtension) {
                 }
             }
 
+            await addDataProviderDeletion(edit, ballerinaExtInstance, fileUri, functionName, dataProviderName);
```
dulajdilshan: **[Question] Do the two deletions' edits ever overlap?** `addDataProviderDeletion` merges a second `deleteByComponentInfo` result into the same `WorkspaceEdit` as the test-function deletion at `:252`, both with ranges computed against the same pre-edit document.

That's safe as long as the LS returns minimal, non-overlapping deletion ranges — but if it returns an import-cleanup edit for *both* deletions (same range, same file), `workspace.applyEdit` would see overlapping edits on one resource. I couldn't find the `deleteByComponentInfo` implementation in this repo (only `DeleteComponentTest.java` in the flow-model module), so I can't rule it out. Have you tried deleting an evaluation whose provider sits adjacent to it?

packages/ballerina-extension/src/features/test-explorer/activator.ts
```diff
@@ -37,6 +38,13 @@ export let testController: TestController;
 
 export const EVALUATION_GROUP = 'evaluations';
 
+export function refreshTestsForFile(filePath: string) {
+    if (!testController) {
+        return;
+    }
+    debouncedHandleFileChange(extension.ballerinaExtInstance, Uri.file(filePath), testController);
```
dulajdilshan, line 45: **[Question] Why the global instance here?** `refreshTestsForFile` reads `extension.ballerinaExtInstance` while `activate()` two lines below already receives the instance as a parameter.

I assume it's because the RPC manager caller has no instance to hand — but stashing the activation-time instance in a module variable would keep this file's dependency on the global out of the picture, and would make the function callable before `activate()` has run without a surprise. Was the global deliberate?

packages/ballerina-extension/src/features/test-explorer/commands.ts
madushajg: Both `deleteByComponentInfo` calls are computed against the unmodified document, so each one still sees `ai:` used by the other function and neither emits the unused-import cleanup. But the combined WorkspaceEdit removes both users, leaving an orphaned `import ballerina/ai;` and a compile error in `tests.bal` when this was the last evaluation. Please either apply the test-function edits first and recompute the provider deletion against the updated document, or add an explicit unused-ai-import check here.

packages/ballerina-extension/src/features/test-explorer/discover.ts
```diff
@@ -201,19 +205,53 @@ export async function handleFileChange(ballerinaExtInstance: BallerinaExtension,
...
     }
 
+    // discoverInFile needs the native file path; uri.path is "/c:/..." on Windows, which Path.of() rejects.
     const request: TestsDiscoveryRequest = {
-        projectPath: uri.path
+        projectPath: uri.fsPath
     };
+    const fileKey = uri.fsPath;
+    const sequence = (fileChangeSequence.get(fileKey) ?? 0) + 1;
+    fileChangeSequence.set(fileKey, sequence);
```
madushajg: The sequence number is assigned after `await getCurrentProjectRoot()`, so the later-started run isn't guaranteed to get the higher sequence. The stale-result guard can pick the wrong winner. Capture the sequence at the top of the function (before any await), and re-check it after the `await handleFileDelete()` on line 226 before calling `createTests`.

packages/ballerina-extension/src/features/test-explorer/discover.ts
dan-niles: Fixed in d55af1cd. cancelPendingFileChange now also deletes the fileChangeSequence entry for the file. Any handleFileChange still awaiting the LS response gets superseded by this and skips re-adding tests for the deleted file.

packages/ballerina-extension/src/features/test-explorer/discover.ts
dan-niles: Fixed in d55af1cd, same change as the comment above. cancelPendingFileChange now deletes the fileChangeSequence entry instead of only bumping it. This keeps the map bounded to files that still exist, since the entry is removed when a file is deleted.

packages/ballerina-extension/src/features/test-explorer/commands.ts
dan-niles: Fixed in d55af1cd. isDataProviderUsedElsewhere now calls getProjectTestFunctions instead of getFileTestFunctions, and looks up each function using its own file path from the discovery result. This covers a data provider used from a sibling file in the same module.

packages/ballerina-extension/src/features/test-explorer/commands.ts
dan-niles: Fixed in d55af1cd, same change as the comment above. The getTestFunction calls now run with Promise.all instead of a sequential loop.

packages/ballerina-extension/src/features/test-explorer/commands.ts
dan-niles: Fixed in d55af1cd. Added a local addTextEdits(edit, textEdits) helper and used it at both call sites.

packages/ballerina-extension/src/utils/git-utils.ts
dan-niles: Fixed in d55af1cd. assertSha(from) and the conditional assertSha(to) now run before the try block. An invalid SHA throws and surfaces as an error, instead of being caught and turned into an empty string. Kept the `to !== "HEAD"` check as it is, since it still guards against shell injection once a non-HEAD caller exists.

packages/ballerina-extension/src/features/test-explorer/commands.ts
dan-niles: I haven't been able to verify this either way. Both deletions build their ranges against the same pre-edit document and get merged into the same WorkspaceEdit, so if deleteByComponentInfo ever returns an import-cleanup edit for both deletions in the same file, applyEdit would see overlapping ranges. I don't have visibility into the Java-side range computation, so I can't confirm whether that happens in practice. Leaving this open. If you can check the LS implementation, that would settle it. Otherwise I can add an explicit overlap check before calling applyEdit, if you'd rather not depend on the LS returning disjoint ranges.

packages/ballerina-extension/src/features/test-explorer/activator.ts
dan-niles: Not changing this. extension.ballerinaExtInstance is the pattern used across the rest of this codebase (144 call sites), so this file is consistent with the rest rather than introducing a new global usage.

dulajdilshan (dismissed):

packages/ballerina-extension/src/features/test-explorer/commands.ts
dan-niles: Fixed in f20d99b813 by applying the test-function edit first, then recomputing the data-provider deletion against the updated document.

packages/ballerina-extension/src/features/test-explorer/discover.ts
dan-niles: Fixed in f20d99b813 by capturing the sequence number before any await, and re-checking it again after handleFileDelete.

MohamedSabthar (approved):

dulajdilshan (approved):

PR #871, opened 2026-08-15
Add tooltips to all artifact cards in the Add Artifact view

### LINKED ISSUE #2147: Tool tips are missing in artifacts except worfkow artifacts
### Description

Image

Tool tips are available for workflow artifacts. Having these to other artifacts also very useful for new user

### Steps to Reproduce

See the image

### Version

5.1.0 alpha

_No response_

Sunny-Hasho:
## Purpose
Tooltips are shown only for the two Durable Workflow cards in the Add Artifact view. Every other artifact card renders with no tooltip, so new users get no explanation of what an artifact does before creating it.

Resolves wso2/product-integrator#2147

## Goals
Give every card in the Add Artifact view a tooltip, using the same mechanism the workflow cards already use.

## Approach
The tooltip plumbing already exists. `ButtonCard` wraps every card in a `Tooltip`, and `ArtifactCard` has an optional tooltip field. Only the two workflow cards set it, while the other panels passed an empty string whenever the card was enabled. The fix is therefore data and wiring, with no component changes.

**Static cards**

- Added tooltip copy to `AUTOMATION_CARD`, `AI_CHAT_AGENT_CARD`, and the three `INTEGRATION_API_CARDS`.
- `OtherArtifactCard` had no tooltip field at all, so the field was added along with copy for all eight entries.

**Dynamic cards** (event, file and MCP triggers, plus Central search results)

- These are built from `ServiceModel`, which already carries a `documentation` field, so they use that instead of invented copy.
- When the language server sends no documentation the tooltip is absent, which matches today's behaviour, so there is no regression.

**Panels**

- The enabled branch now resolves to the card's own tooltip instead of an empty string.
- The disabled-state messages are untouched and still take precedence.

### Files changed
- `views/BI/components/artifactCards.tsx`
- `views/BI/ComponentListView/AutomationPanel.tsx`
- `views/BI/ComponentListView/AIAgentPanel.tsx`
- `views/BI/ComponentListView/IntegrationApiPanel.tsx`
- `views/BI/ComponentListView/EventIntegrationPanel.tsx`
- `views/BI/ComponentListView/FileIntegrationPanel.tsx`
- `views/BI/ComponentListView/OtherArtifactsPanel.tsx`
- `views/BI/ComponentListView/CentralSearchPanel.tsx`

## UI Component Development
- [x] Added reusable UI components to the ui-toolkit: N/A, no new components. Reuses the existing `ButtonCard` and `Tooltip`.
- [x] Use ui-toolkit components wherever possible: unchanged, the existing tooltip in `ButtonCard` does the work.
- [x] Matches with the native VSCode look and feel: unchanged, no styling touched.

## User stories
As a new user browsing the Add Artifact view, I can hover any artifact card and read what it does before deciding to create it.

## Release note
Artifact cards in the Add Artifact view now show a descriptive tooltip on hover. Previously only the Durable Workflow cards did.

## Documentation
N/A. No documented behaviour changes, this only adds hover text to existing cards.

## Training
N/A

## Certification
N/A. No change to product concepts or workflows that certification exams cover.

## Automation tests
- Unit tests
  > None added. The change is static copy plus prop wiring, and there are no existing tests for the artifact panels to extend.
- Integration tests
  > None added.

## Security checks
- Ran FindSecurityBugs plugin and verified report? N/A, TypeScript and React only, no Java changed.

## Samples
N/A

## Related PRs
None

## Migrations (if applicable)
N/A

## Test environment
Not verified in a running IDE. I was unable to build the monorepo locally, since the language server build needs JDK 21, a Ballerina distribution and a GitHub PAT with `read:packages`. The change is source-only and follows the existing workflow panel pattern, but please confirm the rendered tooltips in a dev build before merging.

## Learning
Traced the existing tooltip path from the workflow panel through `ButtonCard` to the shared tooltip in the ui-toolkit, then compared it against the other panels to find where the value was being dropped.

### Note for reviewers on the tooltip wording
The copy for the static cards is proposed, not authoritative. I wrote it to match the tone of the two existing workflow tooltips. The PR template asks for UI text to be reviewed by documentation@wso2.com, so please treat these strings as a starting point and reword as needed.

| Card | Proposed tooltip |
|---|---|
| Automation | Integration logic that runs on a schedule or is triggered manually. |
| Chat Agent Service | AI chat agent, backed by tools and memory, exposed as a service. |
| HTTP Service | Expose your integration over HTTP as a REST API. |
| GraphQL Service | Expose your integration as a GraphQL API with a typed schema. |
| TCP Service | Handle raw TCP socket connections and byte streams. |
| Function | Reusable logic that can be called from anywhere in your integration. |
| Natural Function | Function whose behaviour is described in natural language and run by an LLM. |
| Data Mapper | Visually map and transform data from one type to another. |
| Type | Record, enum, or union type that shapes the data in your integration. |
| Connection | Reusable client connection to an external system or service. |
| Agent | AI agent, backed by tools and memory, that your integration can call. |
| Agent Definition | Reusable agent definition that can be shared across integrations. |
| Configuration | Configurable value that is supplied at deployment time. |

## Summary

- Added tooltips to artifact cards in the Add Artifact view.
- Added descriptive text for static artifact cards.
- Used service documentation for dynamic artifact cards.
- Preserved existing out-of-scope messages for disabled cards.
- Extended `OtherArtifactCard` with optional tooltip support.
- Updated panels to pass tooltip text when enabled.

### DIFF
--- a/packages/ballerina-visualizer/src/views/BI/components/artifactCards.tsx
+++ b/packages/ballerina-visualizer/src/views/BI/components/artifactCards.tsx
@@ -120,6 +120,7 @@ export const AUTOMATION_CARD: ArtifactCard = {
     kind: "automation",
     displayName: "Automation",
     icon: <Icon name="bi-task" />,
+    tooltip: "Integration logic that runs on a schedule or is triggered manually."
 };
 
 export const WORKFLOW_CARD: ArtifactCard = {
@@ -144,6 +145,7 @@ export const AI_CHAT_AGENT_CARD: ArtifactCard = {
     kind: "ai-agent",
     displayName: "Chat Agent Service",
     icon: <Icon name="bi-ai-agent" />,
+    tooltip: "AI chat agent, backed by tools and memory, exposed as a service."
 };
 
 /** TODO: Add the gRPC service card once gRPC support is working. */
@@ -153,6 +155,7 @@ export const INTEGRATION_API_CARDS: ArtifactCard[] = [
         kind: "service",
         displayName: "HTTP Service",
         icon: <Icon name="bi-globe" />,
+        tooltip: "Expose your integration over HTTP as a REST API.",
         artifactInfo: {
             org: "ballerina",
             packageName: "http",
@@ -164,6 +167,7 @@ export const INTEGRATION_API_CARDS: ArtifactCard[] = [
         kind: "service",
         displayName: "GraphQL Service",
         icon: <Icon name="bi-graphql" sx={{ color: "#e535ab" }} />,
+        tooltip: "Expose your integration as a GraphQL API with a typed schema.",
         isBeta: true,
         artifactInfo: {
             org: "ballerina",
@@ -176,6 +180,7 @@ export const INTEGRATION_API_CARDS: ArtifactCard[] = [
         kind: "service",
         displayName: "TCP Service",
         icon: <Icon name="bi-tcp" />,
+        tooltip: "Handle raw TCP socket connections and byte streams.",
         isBeta: true,
         artifactInfo: {
             org: "ballerina",
@@ -198,6 +203,7 @@ export interface OtherArtifactCard {
     requiresNaturalFunctions?: boolean;
     /** Shown only inside a library package. */
     requiresLibrary?: boolean;
+    tooltip?: string;
 }
 
 export const OTHER_ARTIFACT_CARDS: OtherArtifactCard[] = [
@@ -206,6 +212,7 @@ export const OTHER_ARTIFACT_CARDS: OtherArtifactCard[] = [
         displayName: "Function",
         icon: <Icon name="bi-function" />,
         directoryKey: DIRECTORY_MAP.FUNCTION,
+        tooltip: "Reusable logic that can be called from anywhere in your integration.",
     },
     {
         id: "bi-ai-function",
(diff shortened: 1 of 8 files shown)


packages/ballerina-visualizer/src/views/BI/components/artifactCards.tsx
```diff
@@ -214,42 +221,49 @@ export const OTHER_ARTIFACT_CARDS: OtherArtifactCard[] = [
...
+        tooltip: "Reusable agent definition that can be shared across integrations.",
     },
     {
         id: "configurable",
         displayName: "Configuration",
         icon: <Icon name="bi-config" />,
         directoryKey: DIRECTORY_MAP.CONFIGURABLE,
+        tooltip: "Configurable value that is supplied at deployment time.",
     },
 ];
```
madushajg, line 269: the two tooltips that existed before this PR (`WORKFLOW_CARD` `L131`, `DURABLE_AGENT_CARD` `L140`) are noun phrases (`"Long-running workflow logic with…"`), and most new copy follows that. Four break the pattern with imperative phrasing:
   - L158 HTTP Service — "Expose your integration over HTTP as a REST API."
   - L170 GraphQL Service — "Expose your integration as a GraphQL API with a typed schema."
   - L183 TCP Service — "Handle raw TCP socket connections and byte streams."
   - L231 Data Mapper — "Visually map and transform data from one type to another."

Let's use noun-phrase tooltips like: `"A REST API exposed over HTTP."`, `"A GraphQL API with a typed schema."`, `"A raw TCP socket and byte-stream listener."`, `"A visual mapping between two types."` (Note: `ARTIFACT_CATEGORY_META`'s imperative blurbs aren't a counter-precedent, those are section body text, not card tooltips.)

madushajg (changes requested):
Several of the new tooltips pipe `item.documentation` straight from the language server / Ballerina Central into the tooltip prop (`EventIntegrationPanel.tsx:91`, `FileIntegrationPanel.tsx:91`, `AIAgentPanel.tsx:113`, `CentralSearchPanel.tsx:157,180`). That field isn't a short label, it's long, sometimes multi-paragraph prose (up to ~386 chars for some inbuilt triggers, and unbounded for Central search results), and it can contain literal markdown (backticks, etc.).

ui-toolkit's `TooltipContent` is `white-space: nowrap` with no `max-width`, so a string that long renders as a single very wide line that often gets pushed off-screen rather than wrapping. Since the overflow positioning logic just clamps against window width, it doesn't handle content this large gracefully.

Can we either (a) truncate documentation to a reasonable length/first sentence for tooltip use, or (b) give the tooltip wrapping + a max-width. 

Would also help to see a screenshot of a card with long documentation (e.g. Kafka trigger) hovered, to confirm the fix.

PR #873, opened 2026-08-17
Add SAP JCo driver JAR configuration support

LakshanWeerasinghe:
## Summary

Fixes https://github.com/wso2/product-integrator/issues/2020

https://github.com/user-attachments/assets/826956dd-1c1f-4cfd-ba39-4430facdf483

## Summary

- Added project-relative file selection for proprietary driver JARs.
- Added SAP JCo driver dependency metadata and automatic `Ballerina.toml` updates.
- Added dependency detection, path handling, duplicate prevention, and TOML escaping.
- Added UI, RPC, and service model support for driver file fields.
- Added tests for file selection, path handling, dependency edits, and listener workflows.

### DIFF
--- a/packages/ballerina-core/src/rpc-types/common/interfaces.ts
+++ b/packages/ballerina-core/src/rpc-types/common/interfaces.ts
@@ -93,6 +93,18 @@ export interface FileOrDirRequest {
     allowOutsideProject?: boolean;
 }
 
+export interface ProjectFileRequest {
+    filters?: Record<string, string[]>;
+    targetDir?: string;
+    allowOutsideProject?: boolean;
+}
+
+export interface ProjectFileResponse {
+    path: string;
+    absolutePath?: string;
+    copied?: boolean;
+}
+
 export interface ShowErrorMessageRequest {
     message: string;
 }
--- a/packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/models/TriggerUISchemaModel.java
+++ b/packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/models/TriggerUISchemaModel.java
@@ -148,7 +148,8 @@ public record TriggerUISchemaModel(
      * @param formats       the data-binding formats offered, when applicable
      * @param validations   the validation rules applied to this candidate
      * @param extensions    the file extensions offered by a FILE_SELECT/PROJECT_FILE_SELECT candidate
-     *                      (e.g. {@code [".jar"]}), when applicable
+     *                      without a leading dot (e.g. {@code ["jar"]}), matching the
+     *                      {@code showOpenDialog} filter format they are passed to, when applicable
      */
     public record PropertyType(
             String fieldType,
--- a/packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/models/TriggerUISchemaModel.java
+++ b/packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/models/TriggerUISchemaModel.java
@@ -147,6 +147,8 @@ public record TriggerUISchemaModel(
      * @param template      the composition template applied to the bound element
      * @param formats       the data-binding formats offered, when applicable
      * @param validations   the validation rules applied to this candidate
+     * @param extensions    the file extensions offered by a FILE_SELECT/PROJECT_FILE_SELECT candidate
+     *                      (e.g. {@code [".jar"]}), when applicable
      */
     public record PropertyType(
             String fieldType,
@@ -156,7 +158,15 @@ public record TriggerUISchemaModel(
             List<TypeMember> typeMembers,
             Object template,
             List<PayloadFormat> formats,
-            List<ValidationRule> validations) {
+            List<ValidationRule> validations,
+            List<String> extensions) {
+
+        /** Compatibility constructor for existing call sites predating {@code extensions}. */
+        public PropertyType(String fieldType, boolean selected, String ballerinaType, List<Option> options,
+                             List<TypeMember> typeMembers, Object template, List<PayloadFormat> formats,
+                             List<ValidationRule> validations) {
+            this(fieldType, selected, ballerinaType, options, typeMembers, template, formats, validations, null);
+        }
     }
 
     /**
@@ -327,6 +337,14 @@ public record TriggerUISchemaModel(
      *                       elsewhere, so only the bound type is user-selected. Defaults to editable
      *                       when unset.
      * @param bindingGroup   the binding-group id this parameter shares with any sibling parameters
+     * @param driverDependency the Maven coordinates of a required-but-unbundled driver JAR this
+     *                         PROJECT_FILE_SELECT field registers as a
+     *                         {@code [[platform.java21.dependency]]} in Ballerina.toml. Kept as an
+     *                         open {@code Object} (like {@code modifiers}) rather than a typed
+     *                         {@code DriverDependency}, since this module is a dependency of, and so
+     *                         cannot reference a type owned by, the service-model-generator module;
+     *                         the shape survives the Gson round-trip into that module's own
+     *                         {@code Codedata}.
      */
     public record Codedata(
             String type,
@@ -354,7 +372,21 @@ public record TriggerUISchemaModel(
             String group,
             String variantLabel,
             Boolean nameEditable,
-            String bindingGroup) {
+            String bindingGroup,
+            Object driverDependency) {
+
+        /** Compatibility constructor for existing call sites predating {@code driverDependency}. */
+        public Codedata(String type, String argType, String originalName, String moduleName, String orgName,
+                         String packageName, Integer position, String path, String defaultType, String boundType,
+                         Boolean bindable, String bindingKind, String typeConstraint, String template,
+                         String modifier, List<String> supersedes, String targetParam, Object modifiers,
+                         String field, Boolean optional, String value, String valueQualifier, String group,
+                         String variantLabel, Boolean nameEditable, String bindingGroup) {
+            this(type, argType, originalName, moduleName, orgName, packageName, position, path, defaultType,
+                    boundType, bindable, bindingKind, typeConstraint, template, modifier, supersedes, targetParam,
+                    modifiers, field, optional, value, valueQualifier, group, variantLabel, nameEditable,
+                    bindingGroup, null);
+        }
 
         public static Builder builder() {
             return new Builder();
@@ -388,6 +420,7 @@ public record TriggerUISchemaModel(
             private String variantLabel;
             private Boolean nameEditable;
             private String bindingGroup;
+            private Object driverDependency;
 
             private Builder() {
             }
@@ -522,11 +555,16 @@ public record TriggerUISchemaModel(
                 return this;
             }
 
+            public Builder driverDependency(Object driverDependency) {
+                this.driverDependency = driverDependency;
+                return this;
+            }
+
             public Codedata build() {
                 return new Codedata(type, argType, originalName, moduleName, orgName, packageName, position, path,
                         defaultType, boundType, bindable, bindingKind, typeConstraint, template, modifier, supersedes,
                         targetParam, modifiers, field, optional, value, valueQualifier, group, variantLabel,
-                        nameEditable, bindingGroup);
+                        nameEditable, bindingGroup, driverDependency);
             }
         }
     }
--- a/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
+++ b/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
@@ -107,7 +107,11 @@ public final class PlatformDependencyEditUtil {
     }
 
     private static String relativize(Project project, String rawPath) {
-        Path candidate = Path.of(rawPath).toAbsolutePath().normalize();
+        Path rawCandidate = Path.of(rawPath);
+        if (!rawCandidate.isAbsolute()) {
+            return rawPath.replace('\\', '/');
+        }
+        Path candidate = rawCandidate.normalize();
         Path sourceRoot = project.sourceRoot().toAbsolutePath().normalize();
         if (candidate.startsWith(sourceRoot)) {
             return sourceRoot.relativize(candidate).toString().replace('\\', '/');
--- a/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
+++ b/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
@@ -233,7 +233,12 @@ public final class PlatformDependencyEditUtil {
         return new TextEdit(new Range(dependencyStart, dependencyStart), dependencyText.toString());
     }
 
-    /** Escapes a value for a TOML basic string ({@code "..."}), per the TOML spec's escape rules. */
+    /**
+     * Escapes a value for a TOML basic string ({@code "..."}), per the TOML spec's escape rules.
+     * Every control character other than tab (U+0009) is disallowed unescaped in a basic string
+     * (U+0000-U+0008, U+000A-U+001F, and U+007F), so any not covered by a named escape falls
+     * through to a {@code \\uXXXX} escape.
+     */
     private static String tomlString(String raw) {
         StringBuilder escaped = new StringBuilder();
         for (int i = 0; i < raw.length(); i++) {
@@ -246,7 +251,13 @@ public final class PlatformDependencyEditUtil {
                 case '\t' -> escaped.append("\\t");
                 case '\b' -> escaped.append("\\b");
                 case '\f' -> escaped.append("\\f");
-                default -> escaped.append(c);
+                default -> {
+                    if (c < 0x20 || c == 0x7F) {
+                        escaped.append(String.format("\\u%04X", (int) c));
+                    } else {
+                        escaped.append(c);
+                    }
+                }
             }
         }
         return escaped.toString();
--- a/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
+++ b/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
@@ -43,6 +43,15 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+/**
+ * Registers a connector-required driver JAR (e.g. SAP JCo) as a {@code [[platform.java21.dependency]]}
+ * in {@code Ballerina.toml}. Bridges a {@link Value} field's {@link DriverDependency} codedata
+ * (populated by {@link #overlayDriverDependencies}) to the TOML text edit that declares it
+ * (emitted by {@link #addDriverDependenciesIfPresent}/{@link #addIfMissing}), checking the existing
+ * TOML first so a dependency already declared is neither duplicated nor re-prompted for.
+ *
+ * @since 1.8.0
+ */
 public final class PlatformDependencyEditUtil {
 
     private static final String TABLE_ARRAY_NAME = "platform.java21.dependency";
--- a/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
+++ b/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
@@ -223,13 +223,33 @@ public final class PlatformDependencyEditUtil {
         Position dependencyStart = new Position(getDependencyStartLine(toml), 0);
         StringBuilder dependencyText = new StringBuilder();
         dependencyText.append(String.format("[[platform.java21.dependency]]%npath = \"%s\"%ngroupId = \"%s\"%n"
-                        + "artifactId = \"%s\"%nversion = \"%s\"%n", relativePath, dependency.getGroupId(),
-                dependency.getArtifactId(), dependency.getVersion()));
+                        + "artifactId = \"%s\"%nversion = \"%s\"%n", tomlString(relativePath),
+                tomlString(dependency.getGroupId()), tomlString(dependency.getArtifactId()),
+                tomlString(dependency.getVersion())));
         if (dependency.getScope() != null && !dependency.getScope().isBlank()) {
-            dependencyText.append(String.format("scope = \"%s\"%n", dependency.getScope()));
+            dependencyText.append(String.format("scope = \"%s\"%n", tomlString(dependency.getScope())));
         }
         dependencyText.append(String.format("%n"));
         return new TextEdit(new Range(dependencyStart, dependencyStart), dependencyText.toString());
+    }
+
+    /** Escapes a value for a TOML basic string ({@code "..."}), per the TOML spec's escape rules. */
+    private static String tomlString(String raw) {
+        StringBuilder escaped = new StringBuilder();
+        for (int i = 0; i < raw.length(); i++) {
+            char c = raw.charAt(i);
+            switch (c) {
+                case '\\' -> escaped.append("\\\\");
+                case '"' -> escaped.append("\\\"");
+                case '\n' -> escaped.append("\\n");
+                case '\r' -> escaped.append("\\r");
+                case '\t' -> escaped.append("\\t");
+                case '\b' -> escaped.append("\\b");
+                case '\f' -> escaped.append("\\f");
+                default -> escaped.append(c);
+            }
+        }
+        return escaped.toString();
     }
 
     private static int getDependencyStartLine(BallerinaToml toml) {
--- a/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
+++ b/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
@@ -30,7 +30,6 @@ import io.ballerina.toml.syntax.tree.DocumentNode;
 import io.ballerina.toml.syntax.tree.KeyValueNode;
 import io.ballerina.toml.syntax.tree.SyntaxKind;
 import io.ballerina.toml.syntax.tree.TableArrayNode;
-import io.ballerina.toml.syntax.tree.TableNode;
 import org.ballerinalang.langserver.commons.toml.common.TomlSyntaxTreeUtil;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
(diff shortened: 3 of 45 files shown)


gayaldassanayake (dismissed):

packages/ballerina-core/src/rpc-types/common/interfaces.ts
```diff
@@ -93,6 +93,18 @@ export interface FileOrDirRequest {
     allowOutsideProject?: boolean;
 }
 
+export interface ProjectFileRequest {
+    filters?: Record<string, string[]>;
+    targetDir?: string;
+    allowOutsideProject?: boolean;
+}
```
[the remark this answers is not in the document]
LakshanWeerasinghe: Fixed in [commit](https://github.com/ballerina-platform/ballerina-vscode/pull/873/commits/c3f15b631508c02f1eff39d4bf0d860766583a51).

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/ServiceModelGeneratorService.java
```diff
@@ -308,7 +313,13 @@ public CompletableFuture<CommonSourceResponse> addListener(ListenerSourceRequest
...
                 String listenerDeclaration = listener.getListenerDeclaration();
                 edits.add(new TextEdit(Utils.toRange(lineRange.endLine()), NEW_LINE + listenerDeclaration));
-                return new CommonSourceResponse(Map.of(request.filePath(), edits));
+
+                Map<String, List<TextEdit>> allEdits = new LinkedHashMap<>();
+                allEdits.put(request.filePath(), edits);
+                PlatformDependencyEditUtil.addDriverDependenciesIfPresent(allEdits, project,
+                        listener.getProperties());
+
+                return new CommonSourceResponse(allEdits);
```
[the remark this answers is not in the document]
LakshanWeerasinghe: Fixed in [commit](https://github.com/ballerina-platform/ballerina-vscode/pull/873/commits/86f3a76bd2718a3f4a8af75971ce4094766c2914).

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
```diff
@@ -0,0 +1,267 @@
...
+                case '\n' -> escaped.append("\\n");
+                case '\r' -> escaped.append("\\r");
+                case '\t' -> escaped.append("\\t");
+                case '\b' -> escaped.append("\\b");
+                case '\f' -> escaped.append("\\f");
+                default -> escaped.append(c);
+            }
+        }
+        return escaped.toString();
+    }
```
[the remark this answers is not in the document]
LakshanWeerasinghe: Fixed in [commit](https://github.com/ballerina-platform/ballerina-vscode/pull/873/commits/86f3a76bd2718a3f4a8af75971ce4094766c2914).

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/serviceInitModelUtils.ts
```diff
@@ -81,6 +81,7 @@ export function mapPropertiesToFormFields(properties: { [key: string]: PropertyM
             placeholder: property.placeholder,
             addNewButton: property.addNewButton,
             lineRange: property?.codedata?.lineRange,
+            codedata: property?.codedata,
```
NipunaRanasinghe: This attaches connector-model `codedata` to **every** service-init form field, and I can't find anything in this PR that reads `field.codedata` — the driver info round-trips through the model objects (`applyFormValuesToModel` writes into `model.properties[...]`, which already carries codedata), not through `FormField`.

Plenty of existing code does read it, though, and for these forms it was previously always `undefined`:

- `ExpressionEditor.tsx:500`, `TextEditor.tsx:44`, `AutoCompleteEditor.tsx:50` — `moduleName: field.codedata?.moduleName` goes into LS completion requests
- `useFieldDiagnostics.ts:97`, `IdentifierField.tsx:65`, `IdentifierEditor.tsx:208` — the whole codedata is sent in diagnostics / identifier-validation requests
- `ExpressionField.tsx:195` branches on `codedata.searchNodesKind`
- `EditorFactory.tsx:121` and `Form/utils.ts:109,134` branch on `codedata.kind === "PARAM_FOR_TYPE_INFER"`

So this silently changes the completion, diagnostic and validation payloads for every service-init field, not just the SAP ones. Could we drop it if it isn't needed? If it is needed, could the PR description say what consumes it, and add coverage?

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/ListenerConfigForm.tsx
```diff
@@ -231,7 +231,8 @@ function convertConfig(listener: ListenerModel): FormField[] {
             diagnostics: [],
             items: expression.items,
             placeholder: expression.placeholder,
-            lineRange: expression?.codedata?.lineRange
+            lineRange: expression?.codedata?.lineRange,
+            codedata: expression?.codedata
```
NipunaRanasinghe: Same concern as on `serviceInitModelUtils.ts:84` — this widens every listener-config form field with connector `codedata` that was previously `undefined`, which changes what the expression editor, `useFieldDiagnostics` and `IdentifierField` send to the LS.

The driver value already round-trips fine without it: `updateConfig` (below) writes back into `listener.properties[field.key]`, which keeps its original codedata. If nothing reads `field.codedata` here, this line looks droppable.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/ServiceModelGeneratorService.java
```diff
@@ -270,6 +272,9 @@ public CompletableFuture<ListenerModelResponse> getListenerModel(ListenerModelRe
                                 FTPListenerUtil.adjustFtpListenerModelForDeprecatedMode(
                                         listenerModel, request.removeDeprecated(), semanticModel.get(), document);
                             }
+                            PlatformDependencyEditUtil.overlayDriverDependencies(listenerModel,
+                                    request.codedata().getOrgName(), request.codedata().getModuleName(),
+                                    request.codedata().getVersion(), project);
```
NipunaRanasinghe, line 277: This runs unconditionally on every `getListenerModel` call, so HTTP/Kafka/every non-SAP connector now pays for a `TriggerModelReader.getSchemaDrivenServiceInitModel` lookup it can never use.

For bundled models that's a cache hit, but for a connector that isn't bundled it falls through to `getSchemaDrivenTriggerModel`, and `Resolution.UNRESOLVED` is deliberately **not** cacheable (`TriggerModelReader.java:296-298`, so a later pull isn't masked by a memoized miss). That means a connector not yet in the local repo re-runs local-repository resolution on *every* listener-model fetch.

Could we gate the overlay — e.g. skip when the model isn't bundled, or bail early when `findDriverDependencyFields` on the template is empty?

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
```diff
@@ -0,0 +1,284 @@
...
+        int lastDocumentLine = tomlSyntaxTree.lineRange().endLine().line();
+        int candidateLine = tomlSyntaxTree.members().stream()
+                .filter(member -> member.kind().equals(SyntaxKind.TABLE)
+                        && TomlSyntaxTreeUtil.toQualifiedName(((TableNode) member).identifier().value())
+                                .equals("package"))
+                .findFirst()
+                .map(member -> member.lineRange().endLine().line() + 2)
+                .orElse(lastDocumentLine);
+        return Math.min(candidateLine, lastDocumentLine);
+    }
```
NipunaRanasinghe: `getDependencyStartLine` is copied verbatim from `LocalDependencyEditUtil.java:150-160`, and `unquote` (line 219 here) from `LocalDependencyEditUtil.java:132-138`. Both classes now live in the same package — worth pulling them into a shared package-private helper so the insertion-point logic only has to be fixed once.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
```diff
@@ -0,0 +1,284 @@
...
+        String trimmed = raw.trim();
+        if (trimmed.length() >= 2 && trimmed.startsWith("\"") && trimmed.endsWith("\"")) {
+            return trimmed.substring(1, trimmed.length() - 1);
+        }
+        return trimmed;
+    }
+
+    private static TextEdit createPlatformDependencyEdit(BallerinaToml toml, DriverDependency dependency,
+                                                          String relativePath) {
+        Position dependencyStart = new Position(getDependencyStartLine(toml), 0);
```
NipunaRanasinghe: Every edit is a zero-width insert at the same computed position, so `addDriverDependenciesIfPresent` can emit two edits with an **identical** `Range`, and `SchemaDrivenServiceBuilder.addServiceInitSource` may add `LocalDependencyEditUtil`'s edit at that same position too.

On the client these are applied in a loop (`ballerina-extension/src/utils/source-utils.ts:93-95`):

```ts
for (const edit of edits) {
    await applyBallerinaTomlEdit(fileUri, edit);
}
```

but `applyBallerinaTomlEdit` (`rpc-managers/common/utils.ts:216-232`) calls `workspace.applyEdit(...).then(...)` **without returning the promise**, so the `await` resolves immediately and the edits are dispatched concurrently. Its `then` also swallows the failure case into an empty `if`/`else`.

The arithmetic happens to work out today — each emitted block ends with a blank line, so a later insert lands cleanly before an earlier one — but the ordering isn't guaranteed and a failed apply is silent. This PR is the first thing to produce more than one TOML edit per operation, so it's worth at least making `applyBallerinaTomlEdit` return the `applyEdit` promise. Did you verify the two-driver case on a real project?

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
```diff
@@ -0,0 +1,284 @@
...
+                    TomlSyntaxTreeUtil.toQualifiedName(tableArrayNode.identifier().value()))) {
+                continue;
+            }
+            Map<String, String> fields = new LinkedHashMap<>();
+            for (KeyValueNode field : tableArrayNode.fields()) {
+                fields.put(field.identifier().toSourceCode().trim(), unquote(field.value().toSourceCode()));
+            }
+            if (dependency.getGroupId().equals(fields.get("groupId"))
+                    && dependency.getArtifactId().equals(fields.get("artifactId"))) {
+                return Optional.of(fields);
```
NipunaRanasinghe, line 265: The match is on `groupId` + `artifactId` only, ignoring `version`. So a project that already declares `com.sap:com.sap.conn.jco` at some other version is treated as fully configured: the field is hidden in the init form, shown read-only in the listener form, and never updated.

`LocalDependencyEditUtil.addIfMissing` handles the equivalent drift by rewriting the version in place (`findVersionValueEdit`). Is the asymmetry intentional here — e.g. because the JAR path is what actually matters and the version is only metadata under `scope = "provided"`? If so, a comment saying so would help.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
```diff
@@ -0,0 +1,284 @@
...
+            Optional<String> declaredPath = findDeclaredPath(project, dependency);
+            displayField.setEnabled(true);
+            if (declaredPath.isPresent()) {
+                displayField.setValue(declaredPath.get());
+                displayField.setEditable(false);
+            } else {
+                displayField.setValue("");
+                displayField.setEditable(true);
+            }
+            listener.getProperties().put(entry.getKey(), displayField);
```
NipunaRanasinghe, line 225: Two things here:

1. The method guards `listener == null` and `project == null`, but line 184 then calls `listener.getProperties().put(...)` — NPE if `getProperties()` is null. `DriverDependencyFieldTest` constructs the listener with an explicit `new LinkedHashMap<>()`, so the test can't catch it. Worth guarding, or building the map when absent.

2. The already-declared UX diverges between the two entry points: `populateDriverDependencyFields` (line 137) hides the field via `setEnabled(false)`, while this one keeps it enabled and shows the path read-only. Both behaviours are individually reasonable, but the user sees a different thing depending on whether they came from create-service or edit-listener. Related: in the init form, if *both* SAP drivers are already declared, the `driverDependencies` GROUP_SECTION stays enabled with both children hidden, so I think you get an empty "SAP Driver Libraries" section — worth checking.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/models/TriggerUISchemaModel.java
```diff
@@ -147,6 +147,8 @@ public record Property(
      * @param template      the composition template applied to the bound element
      * @param formats       the data-binding formats offered, when applicable
      * @param validations   the validation rules applied to this candidate
+     * @param extensions    the file extensions offered by a FILE_SELECT/PROJECT_FILE_SELECT candidate
+     *                      (e.g. {@code [".jar"]}), when applicable
```
NipunaRanasinghe: The example says `[".jar"]`, but `sap.jco.json` ships `["jar"]` and that's what the consumer needs: `FileSelect`/`ProjectFileSelect` pass this straight through as VS Code `showOpenDialog` filters, which expect extensions **without** a leading dot. A model author following this javadoc would get a picker that silently matches nothing. Could we make it `{@code ["jar"]}`?

packages/ballerina-visualizer/src/views/BI/CreateIntegrationWizard/components/WizardRpcAdapterProvider.tsx
```diff
@@ -77,10 +77,11 @@ function createWizardRpcAdapter(wsClient: BiWsClient): BallerinaRpcClient {
 
     const commonRpcClient = withFallback("Common", {
         showErrorMessage: (params: any) => wsClient.showErrorMessage(params),
-        // FILE_SELECT pickers go over the WS bridge. `allowOutsideProject` is forced:
-        // pre-project the host would otherwise offer to copy the spec into whatever
+        // FILE_SELECT / PROJECT_FILE_SELECT pickers go over the WS bridge. `allowOutsideProject`
+        // is forced: pre-project the host would otherwise offer to copy the spec into whatever
         // project happens to be open.
         selectFileOrDirPath: (params: any) => wsClient.selectFileOrDirPath({ ...params, allowOutsideProject: true }),
+        selectProjectRelativeFile: (params: any) => wsClient.selectProjectRelativeFile({ ...params, allowOutsideProject: true }),
```
NipunaRanasinghe, line 84: Forcing `allowOutsideProject: true` makes `selectProjectRelativeFile` skip `resolveIntegrationRoot` entirely, so it returns the absolute picked path with no copy — and the LS's `relativize` then writes that absolute path verbatim into `Ballerina.toml`, giving a machine-specific, non-portable `path = "/Users/…/sapjco3.jar"`.

That's understandable in the wizard (there's no project to copy into yet), but the outcome differs from every other entry point. Is a `PROJECT_FILE_SELECT` field actually reachable from the wizard today? If not, it might be cleaner to leave it unregistered than to register a variant that produces a toml we don't want.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/model/Codedata.java
```diff
@@ -34,35 +34,23 @@ public class Codedata {
...
-    // The literal an ENUM_LITERAL choice branch emits (qualified by `valueQualifier`).
     private String value;
-    // Payload: whether the bound parameter's identifier may be renamed in the edit UI (unset
-    // defaults to editable). False for connectors that bind to a fixed, structural identifier.
     private Boolean nameEditable;
-    // LISTENER_VAR_NAME: the node's shipped value is a connector-curated default the creation flow
-    // must keep as the base name (still made project-unique) instead of deriving a protocol-based one.
     private Boolean preserveValue;
     private String castType;
+    private DriverDependency driverDependency;
```
NipunaRanasinghe, line 53: The `driverDependency` field itself is fine, but this hunk also deletes several unrelated explanatory comments above it — the "Source-generation hints carried by connector-shipped models (phase2 schema)" note, the payload/annotation-hints block explaining `template`/`defaultType`/`boundType`/`bindable`/`modifier`/`targetParam`/`field`/`optional`, the `typeIdentifier` example, the `value` note, and the `nameEditable` and `preserveValue` rationale.

None of that is related to driver dependencies. Could you restore them?

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/model/Value.java
```diff
@@ -125,10 +120,9 @@ public enum FieldType {
         RAW_TEMPLATE,
         SINGLE_SELECT_LISTENER,
         MULTIPLE_SELECT_LISTENER,
-        // The spelling connector-shipped trigger models use (phase-6 schema); without the constant
-        // Gson silently nulls the fieldType on parse.
         MULTI_SELECT_LISTENER,
         FILE_SELECT,
+        PROJECT_FILE_SELECT,
```
NipunaRanasinghe, line 125: Adding `PROJECT_FILE_SELECT` is right, but the same hunk drops the comments on `DOC_TEXT` and `MULTI_SELECT_LISTENER`. Those aren't decorative — they record *why* each constant has to exist (without `DOC_TEXT`, `wireFieldType`'s `valueOf` throws and silently falls back to `EXPRESSION`; without `MULTI_SELECT_LISTENER`, Gson silently nulls the `fieldType` on parse). That's exactly the kind of thing someone tidying the enum later would need. Please restore.

packages/ballerina-side-panel/src/components/Form/validationRules.ts
```diff
@@ -658,5 +658,6 @@ export async function evaluateAsyncClientRules(
             });
         }
     }
+
```
NipunaRanasinghe: Stray blank line — this is the only change in the file. Worth dropping so `validationRules.ts` stays out of the diff.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger-models/sap.jco.json
```diff
@@ -751,6 +751,86 @@
...
+          "value": "",
+          "enabled": true,
+          "editable": true,
+          "optional": false,
+          "advanced": false,
+          "codedata": {
+            "driverDependency": {
+              "groupId": "com.sap",
+              "artifactId": "com.sap.conn.idoc",
+              "version": "3.1.*",
```
NipunaRanasinghe, line 791: Has `version = "3.1.*"` been verified against a real `bal build`? A wildcard is fine as inert metadata when `path` is set and `scope = "provided"`, but I'd want to confirm the manifest parser doesn't warn or reject it, since this value goes straight into the user's `Ballerina.toml`.

LakshanWeerasinghe: this works fine.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/model/Value.java
LakshanWeerasinghe: In line comments are the correct place. Will add them later for all the enum values.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/model/Codedata.java
LakshanWeerasinghe: This was intentional to remove unwanted inline comments.

packages/ballerina-visualizer/src/views/BI/CreateIntegrationWizard/components/WizardRpcAdapterProvider.tsx
LakshanWeerasinghe: this is not reachable.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/serviceInitModelUtils.ts
LakshanWeerasinghe: Good catch — dropped.

Nothing in this PR reads `field.codedata`: `ProjectFileSelect` takes its extensions from `field.types` via `getPrimaryInputType`, and the driver value round-trips through `model.properties[...]` in `applyFormValuesToModel`, which keeps its own codedata. Removed the line so the completion/diagnostic payloads for service-init fields stay exactly as they were.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/ListenerConfigForm.tsx
LakshanWeerasinghe: Dropped here too, same reasoning — `updateConfig` writes back into `listener.properties[field.key]`, which keeps its original codedata, so the driver path round-trips fine without it.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/ServiceModelGeneratorService.java
LakshanWeerasinghe: Gated. `overlayDriverDependencies` now returns early unless `TriggerModelReader.hasBundledTriggerModel(moduleName)`, so http/grpc and every other non-bundled connector never reaches `LibraryMetadataReader` on this path.

The cost was slightly worse than the `UNRESOLVED` case alone: `schemaDrivenTriggerCache` is `Caffeine.maximumSize(2)`, so even cacheable `ABSENT` verdicts thrash once more than two connectors are in play.

The tradeoff is documented in a comment — driver dependencies are declared only by bundled models today, so a connector-*shipped* model declaring one would now be skipped here. Doing better needs a cheap "does this connector declare drivers" probe that `TriggerModelReader` doesn't expose yet. Added `testOverlaySkipsConnectorsWithoutABundledModel`.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
LakshanWeerasinghe: Extracted both into a new package-private `TomlDependencyUtil` in the same package. `LocalDependencyEditUtil` and `PlatformDependencyEditUtil` now call through it, and their newly-unused imports are gone.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
LakshanWeerasinghe: Valid — and the client side was the real problem, now fixed.

`applyBallerinaTomlEdit` was `async` but dropped the `applyEdit` promise, so the `await` in `source-utils.ts` resolved immediately and the edits were dispatched concurrently. It now awaits and returns `Promise<boolean>`, and logs on failure instead of the empty `if`/`else`. With the loop actually sequential, the ordering is guaranteed rather than incidental.

You're right that this PR is the first to emit more than one — `addServiceInitSource` can produce three at the same position (the local `[[dependency]]` plus both SAP drivers). On the LS side the two-driver case is covered by `testFilledFieldsProduceOneTomlEditEach`.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
LakshanWeerasinghe: Intentional — added a javadoc on `findEntry` saying so.

For a `scope = "provided"` dependency the JAR at `path` is what actually resolves and the version is inert metadata, so a project already pointing at its own SAP JCo build is treated as configured rather than having its manifest silently rewritten. That's the asymmetry with `LocalDependencyEditUtil`, where the version *is* the thing that resolves.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
LakshanWeerasinghe: Point 3 confirmed and fixed — that was a real bug, thanks.

`populateDriverDependencyFields` now calls a new `collapseEmptyDriverGroups`, which disables a group left holding nothing but drivers that were just hidden, so a project declaring both no longer renders an empty "SAP Driver Libraries" header. Display-only: `SchemaDrivenSourceGenerator.collect` doesn't gate on `isEnabled()`, and `placeArg` no-ops for these fields since their codedata carries no `argType`. Covered by `testGroupIsCollapsedWhenEveryDriverIsAlreadyDeclared` and `testGroupStaysEnabledWhileAnyDriverIsStillNeeded`, with a new `both_declared` fixture.

Point 1 I've left alone: `createBaseListenerModel` always calls `.setProperties(...)` with a constructed map, and `Listener` already dereferences `properties` unguarded at lines 88 and 137 — a null map would fail well before this line. Happy to add the guard if you'd still prefer one.

Point 2 also left as-is, deliberately: create-service has nothing actionable to show once a driver is declared, whereas edit-listener showing the resolved path read-only is informative. Leaving this thread open in case you disagree on either.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/models/TriggerUISchemaModel.java
LakshanWeerasinghe: Fixed — now `{@code ["jar"]}`, with a note that it matches the `showOpenDialog` filter format it gets passed to (consistent with the existing `['yaml', 'json', ...]` default in `askFilePath`).

packages/ballerina-side-panel/src/components/Form/validationRules.ts
LakshanWeerasinghe: Removed — `validationRules.ts` is now out of the diff entirely.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/PlatformDependencyEditUtil.java
NipunaRanasinghe: Correcting myself here — my premise was wrong, so please don't bank on the race being fixed.

I traced `applyBallerinaTomlEdit` to `rpc-managers/common/utils.ts:117`, but `source-utils.ts:29` imports it from **`rpc-managers/bi-diagram/utils`**. There are two divergent copies of that function, and the one actually on the TOML path already awaited:

```ts
// rpc-managers/bi-diagram/utils.ts:66
await workspace.applyEdit(workspaceEdit);
```

So the `for … await` loop in `source-utils.ts:93-95` was already sequential and there was no race. Sorry for the bad steer.

Two things that follow:

1. Your fix landed on the `common/utils.ts` copy, which `grep -rn applyBallerinaTomlEdit packages/` shows has **no importers at all**. It's a dead duplicate of the `bi-diagram/utils.ts` one (minus the latter's `adjustRangeForHeaderComments` handling). Worth deleting rather than maintaining two copies — that divergence is what misled me. Happy for that to be a separate cleanup if you'd rather keep this PR scoped.

2. The multiple-inserts-at-one-position concern still stands on its own, just for a benign reason: `bi-diagram/utils.ts` re-reads the document per call, and each emitted block ends with a blank line, so a later insert landing before an earlier one still produces valid stanzas. Did you get to eyeball the resulting `Ballerina.toml` for the both-drivers case (and the local-`[[dependency]]` + two-driver case from `addServiceInitSource`)? That's the one bit I can't verify from the diff.

NipunaRanasinghe (approved):
Approving. 

One non-blocking item, from the thread on `PlatformDependencyEditUtil.java`: `testFilledFieldsProduceOneTomlEditEach` proves two edits are *produced*, but nothing proves the *applied* result is valid TOML — the applier is extension-side, so the LS tests stop at the edit objects. The arithmetic works out on paper (each block ends with a blank line, so a later insert landing before an earlier one still yields well-formed stanzas), and the failure mode would be a loud malformed manifest rather than anything silent, so I don't want to hold the PR on it. Worth a manual look at the both-drivers case and the local-`[[dependency]]` + two-drivers case from `addServiceInitSource` before merge.

Also flagged there: the `applyBallerinaTomlEdit` you fixed in `rpc-managers/common/utils.ts` has no importers — `source-utils.ts` uses the `bi-diagram/utils.ts` copy, which already awaited. My original comment misattributed it; the dead duplicate is worth deleting as separate cleanup.

Leaving the `Codedata.java` / `Value.java` comment removals to your judgment — you've reaffirmed they were deliberate, so that's your call to make.

PR #874, opened 2026-08-17
Fix onUpdate function handler UX and other minor issues

LakshanWeerasinghe:
## Purpose
$subject

This PR include fixes for multiple issues.

1. Group data binding params
Fixes https://github.com/wso2/product-integrator/issues/2044
Screenshot 2026-08-17 at 10 10 12

2. Adding duplicated imports
Fixes https://github.com/wso2/product-integrator/issues/2046

3. Improve handler form titles
Screenshot 2026-08-17 at 10 11 49

4. Add tooltip description configuring support for data binding input
Screenshot 2026-08-17 at 10 11 26

5. Add tooltip description configuring support for grouped handlers
Screenshot 2026-08-17 at 10 11 08

Fixes part of https://github.com/wso2/product-integrator/issues/2059

## Summary

- Group related data-binding parameters in handler payload editors.
- Preserve parameter identity while applying schema and type changes across grouped parameters.
- Prevent duplicate imports and preserve aliases in generated trigger code.
- Improve handler configuration titles for repeatable and non-repeatable handlers.
- Add descriptions and tooltips for data-binding inputs and grouped handlers.
- Extend CDC trigger models with `bindingGroup` and `addDescription` metadata.
- Add tests for CDC payload grouping, handler behavior, and import generation.

### DIFF
--- a/packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/models/TriggerUISchemaModel.java
+++ b/packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/models/TriggerUISchemaModel.java
@@ -326,6 +326,7 @@ public record TriggerUISchemaModel(
      *                       some connectors bind to a fixed, structural identifier referred to by name
      *                       elsewhere, so only the bound type is user-selected. Defaults to editable
      *                       when unset.
+     * @param bindingGroup   the binding-group id this parameter shares with any sibling parameters
      */
     public record Codedata(
             String type,
@@ -352,7 +353,8 @@ public record TriggerUISchemaModel(
             String valueQualifier,
             String group,
             String variantLabel,
-            Boolean nameEditable) {
+            Boolean nameEditable,
+            String bindingGroup) {
 
         public static Builder builder() {
             return new Builder();
@@ -385,6 +387,7 @@ public record TriggerUISchemaModel(
             private String group;
             private String variantLabel;
             private Boolean nameEditable;
+            private String bindingGroup;
 
             private Builder() {
             }
@@ -514,11 +517,16 @@ public record TriggerUISchemaModel(
                 return this;
             }
 
+            public Builder bindingGroup(String bindingGroup) {
+                this.bindingGroup = bindingGroup;
+                return this;
+            }
+
             public Codedata build() {
                 return new Codedata(type, argType, originalName, moduleName, orgName, packageName, position, path,
                         defaultType, boundType, bindable, bindingKind, typeConstraint, template, modifier, supersedes,
                         targetParam, modifiers, field, optional, value, valueQualifier, group, variantLabel,
-                        nameEditable);
+                        nameEditable, bindingGroup);
             }
         }
     }
--- a/packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/models/TriggerUISchemaModel.java
+++ b/packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/models/TriggerUISchemaModel.java
@@ -617,6 +617,7 @@ public record TriggerUISchemaModel(
      * @param addLabel    the label used when offering to add this node
      * @param groupName   the logical group this node is listed under
      * @param badge       a short badge tag shown on the node
+     * @param addDescription the description used when offering to add this node
      */
     public record Metadata(
             String label,
@@ -626,6 +627,7 @@ public record TriggerUISchemaModel(
             String subLabel,
             String addLabel,
             String groupName,
-            String badge) {
+            String badge,
+            String addDescription) {
     }
 }
--- a/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/SchemaDrivenSourceGenerator.java
+++ b/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/SchemaDrivenSourceGenerator.java
@@ -161,9 +161,17 @@ public final class SchemaDrivenSourceGenerator {
                     continue;
                 }
                 String org = moduleRef.substring(0, slash);
-                String module = moduleRef.substring(slash + 1);
+                String rest = moduleRef.substring(slash + 1).trim();
+                String module = rest;
+                String alias = null;
+                int asIndex = rest.lastIndexOf(" as ");
+                if (asIndex > 0) {
+                    module = rest.substring(0, asIndex).trim();
+                    alias = rest.substring(asIndex + 4).trim();
+                }
                 if (!Utils.importExists(rootNode, org, module)) {
-                    imports.append(Utils.getImportStmt(org, module));
+                    imports.append(alias == null ? Utils.getImportStmt(org, module)
+                            : Utils.getImportStmt(org, module, alias));
                 }
             }
         }
--- a/packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
+++ b/packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
@@ -882,7 +882,8 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
                                     {hasDefaultPayload(param) ? (
                                         <AddButtonWrapper>
                                             <Tooltip
-                                                content={`Define ${label} for easier access in the flow diagram`}
+                                                content={param.metadata?.description
+                                                    || `Define ${label} for easier access in the flow diagram`}
                                                 position="bottom"
                                             >
                                                 <LinkButton onClick={() => setTypeEditorParamName(param.name?.value ?? null)}>
--- a/packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
+++ b/packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
@@ -65,10 +65,12 @@ import {
     CODEDATA_PAYLOAD_MODIFIER,
     CODEDATA_PAYLOAD_TYPE_INCLUDED_RECORD,
     addableCatalogOf,
-    bindingGroupSiblingsOf,
+    bindingGroupDiverges,
+    bindingGroupOf,
     boundRepresentativeOf,
     composePayloadType,
     decomposePayloadType,
+    editTargetsOf,
     functionSignatureKey,
     groupedPayloadParametersOf,
     handlerGroupId,
@@ -456,10 +458,7 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
 
     const metadataFlags = functionModel ? propertiesOfRole(functionModel, CODEDATA_METADATA_FLAG) : [];
     const modifierFlags = functionModel ? propertiesOfRole(functionModel, CODEDATA_PAYLOAD_MODIFIER) : [];
-    // All bindable payload params — a handler may expose more than one (e.g. CDC onUpdate's
-    // before/after), each configured independently below.
     const payloadParams = functionModel ? payloadParametersOf(functionModel) : [];
-    // The first payload param still drives shared UI bits (e.g. the variant dropdown's label).
     const payloadParam = payloadParams[0];
     const groupedPayloadParams = functionModel ? groupedPayloadParametersOf(functionModel) : [];
 
@@ -544,11 +543,9 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
         const typeName = typeof type === "string" ? type : type.name;
         const target = payloadParams.find((p) => p.name?.value === targetName);
         // The parameter keeps its schema-shipped name — only the bound shape changes.
-        const groupNames = new Set(
-            (target ? bindingGroupSiblingsOf(functionModel, target) : []).map((p) => p.name?.value)
-        );
+        const editTargets = new Set(target ? editTargetsOf(functionModel, target) : []);
         const parameters = functionModel.parameters.map((p) => {
-            if (!isPayloadParameter(p) || !groupNames.has(p.name?.value)) {
+            if (!isPayloadParameter(p) || !editTargets.has(p)) {
                 return p;
             }
             const updatedType: PropertyModel = {
@@ -567,9 +564,9 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
         if (!functionModel) {
             return;
         }
-        const groupNames = new Set(bindingGroupSiblingsOf(functionModel, target).map((p) => p.name?.value));
+        const editTargets = new Set(editTargetsOf(functionModel, target));
         const parameters = functionModel.parameters.map((p) =>
-            isPayloadParameter(p) && groupNames.has(p.name?.value)
+            isPayloadParameter(p) && editTargets.has(p)
                 ? { ...p, type: { ...p.type, codedata: { ...p.type.codedata, boundType: undefined } } }
                 : p
         );
@@ -872,15 +869,18 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
                         </FlagsColumn>
                     )}
 
-                    {/* Payload schema — one section per bindable DATA_BINDING param (a handler such
-                        as CDC onUpdate exposes both a before- and an after-image). */}
+                    {/* Payload schema — one section per binding group (CDC onUpdate's before/after
+                        share a group and render as a single section; see groupedPayloadParams). */}
                     {groupedPayloadParams
-                        .filter((param) => param.type?.codedata?.bindable === true)
-                        .map((param) => {
-                            const displayParam = boundRepresentativeOf(functionModel, param);
+                        .map((param) => boundRepresentativeOf(functionModel, param))
+                        .filter((displayParam) => displayParam.type?.codedata?.bindable === true)
+                        .map((displayParam) => {
                             const label = displayLabelOf(displayParam);
+                            const key = bindingGroupDiverges(functionModel, displayParam)
+                                ? (displayParam.name?.value ?? label)
+                                : (bindingGroupOf(displayParam) ?? displayParam.name?.value ?? label);
                             return (
-                                <Fragment key={displayParam.name?.value ?? label}>
+                                <Fragment key={key}>
                                     {hasDefaultPayload(displayParam) ? (
                                         <AddButtonWrapper>
                                             <Tooltip
@@ -936,15 +936,14 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
                                                     const editedElement = decomposePayloadType(
                                                         functionModel, displayParam, editedPayload.type?.value ?? "");
                                                     const editedName = editedPayload.name?.value;
-                                                    const groupNames = new Set(
-                                                        bindingGroupSiblingsOf(functionModel, displayParam)
-                                                            .map((p) => p.name?.value)
+                                                    const editTargets = new Set(
+                                                        editTargetsOf(functionModel, displayParam)
                                                     );
                                                     const parameters = functionModel.parameters.map((p) => {
-                                                        if (!isPayloadParameter(p) || !groupNames.has(p.name?.value)) {
+                                                        if (!isPayloadParameter(p) || !editTargets.has(p)) {
                                                             return p;
                                                         }
-                                                        const isTarget = p.name?.value === displayParam.name?.value;
+                                                        const isTarget = p === displayParam;
                                                         return {
                                                             ...p,
                                                             name: isTarget && editedName !== undefined
--- a/packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
+++ b/packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
@@ -66,6 +66,7 @@ import {
     CODEDATA_PAYLOAD_TYPE_INCLUDED_RECORD,
     addableCatalogOf,
     bindingGroupSiblingsOf,
+    boundRepresentativeOf,
     composePayloadType,
     decomposePayloadType,
     functionSignatureKey,
@@ -876,17 +877,18 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
                     {groupedPayloadParams
                         .filter((param) => param.type?.codedata?.bindable === true)
                         .map((param) => {
-                            const label = displayLabelOf(param);
+                            const displayParam = boundRepresentativeOf(functionModel, param);
+                            const label = displayLabelOf(displayParam);
                             return (
-                                <Fragment key={param.name?.value ?? label}>
-                                    {hasDefaultPayload(param) ? (
+                                <Fragment key={displayParam.name?.value ?? label}>
+                                    {hasDefaultPayload(displayParam) ? (
                                         <AddButtonWrapper>
                                             <Tooltip
-                                                content={param.metadata?.description
+                                                content={displayParam.metadata?.description
                                                     || `Define ${label} for easier access in the flow diagram`}
                                                 position="bottom"
                                             >
-                                                <LinkButton onClick={() => setTypeEditorParamName(param.name?.value ?? null)}>
+                                                <LinkButton onClick={() => setTypeEditorParamName(displayParam.name?.value ?? null)}>
                                                     <Codicon name="add" />
                                                     Define {label}
                                                 </LinkButton>
@@ -917,32 +919,32 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
                                                 name+type editor. */}
                                             <Parameters
                                                 parameters={[{
-                                                    ...param,
+                                                    ...displayParam,
                                                     editable: true,
                                                     type: {
-                                                        ...param.type,
-                                                        value: composePayloadType(functionModel, param),
+                                                        ...displayParam.type,
+                                                        value: composePayloadType(functionModel, displayParam),
                                                     },
                                                 }]}
-                                                hideName={param.type?.codedata?.nameEditable === false}
+                                                hideName={displayParam.type?.codedata?.nameEditable === false}
                                                 onChange={(edited) => {
                                                     if (edited.length === 0) {
-                                                        handleDeletePayloadSchema(param);
+                                                        handleDeletePayloadSchema(displayParam);
                                                         return;
                                                     }
                                                     const [editedPayload] = edited;
                                                     const editedElement = decomposePayloadType(
-                                                        functionModel, param, editedPayload.type?.value ?? "");
+                                                        functionModel, displayParam, editedPayload.type?.value ?? "");
                                                     const editedName = editedPayload.name?.value;
                                                     const groupNames = new Set(
-                                                        bindingGroupSiblingsOf(functionModel, param)
+                                                        bindingGroupSiblingsOf(functionModel, displayParam)
                                                             .map((p) => p.name?.value)
                                                     );
                                                     const parameters = functionModel.parameters.map((p) => {
                                                         if (!isPayloadParameter(p) || !groupNames.has(p.name?.value)) {
                                                             return p;
                                                         }
-                                                        const isTarget = p.name?.value === param.name?.value;
+                                                        const isTarget = p.name?.value === displayParam.name?.value;
                                                         return {
                                                             ...p,
                                                             name: isTarget && editedName !== undefined
--- a/packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
+++ b/packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
@@ -65,9 +65,11 @@ import {
     CODEDATA_PAYLOAD_MODIFIER,
     CODEDATA_PAYLOAD_TYPE_INCLUDED_RECORD,
     addableCatalogOf,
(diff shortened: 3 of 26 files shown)


gayaldassanayake (dismissed):

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
```diff
@@ -874,7 +882,8 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
                                     {hasDefaultPayload(param) ? (
```
NipunaRanasinghe: Collapsing to one representative means every decision below is made from the *first* param of the group only — including `hasDefaultPayload(param)` here. That hides an asymmetric pre-existing binding.

`boundType` is populated per-param from source (`onRead`'s `afterEntry` reads back as `AfterEntrySchema` in `get_sm_from_source/config/cdc_service_model.json`), and the previous release rendered the two payloads as independent sections — so `onUpdate(record {} before, AfterSchema after)` is a state real users can already be in, having bound only the after-image.

On that source the representative is `before`, `hasDefaultPayload` is `true`, and the form renders the "Define Database Entry" add-link as if nothing were bound — `after`'s existing `AfterSchema` is invisible. See the follow-up comment on `handleTypeCreated` for what happens when the user then clicks it.

Worth deciding what the grouped section should show when siblings disagree: treat the group as bound if *any* sibling is bound, or reconcile the siblings when the model is first loaded.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
```diff
@@ -538,9 +541,13 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
...
         }
         const typeName = typeof type === "string" ? type : type.name;
+        const target = payloadParams.find((p) => p.name?.value === targetName);
         // The parameter keeps its schema-shipped name — only the bound shape changes.
+        const groupNames = new Set(
+            (target ? bindingGroupSiblingsOf(functionModel, target) : []).map((p) => p.name?.value)
+        );
         const parameters = functionModel.parameters.map((p) => {
-            if (!isPayloadParameter(p) || p.name?.value !== targetName) {
+            if (!isPayloadParameter(p) || !groupNames.has(p.name?.value)) {
```
NipunaRanasinghe: Second half of the issue flagged on the payload-section block above: this now writes the new `boundType` to *every* group sibling, not just the target.

Combined with the representative-only `hasDefaultPayload` check, the sequence on `onUpdate(record {} before, AfterSchema after)` is:

1. form shows "Define Database Entry" (decided from `before`, which is unbound)
2. user defines a type `Foo`
3. `groupNames` = `{before, after}`, so both params get `boundType: "Foo"`
4. `after`'s existing `AfterSchema` is replaced — the user was never shown that it existed

Same applies to the inline-editor `onChange` further down, which also fans the edited element out across the group.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
```diff
@@ -520,7 +523,7 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
     // which. A single-payload handler keeps the plain label.
     const displayLabelOf = (param?: ParameterModel) => {
         const base = labelOfPayload(param);
-        return payloadParams.length > 1 && param?.name?.value ? `${base} (${param.name.value})` : base;
+        return groupedPayloadParams.length > 1 && param?.name?.value ? `${base} (${param.name.value})` : base;
```
NipunaRanasinghe, line 526: This comment now describes the opposite of what the code does. It cites CDC `onUpdate`'s before/after as the case where the param name gets appended, but with the switch to `groupedPayloadParams` that handler has `length === 1`, so `onUpdate` is now precisely the case that *keeps* the plain label.

The suffix now only appears when a handler exposes more than one distinct binding *group*, which needs a different example (or none).

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
```diff
@@ -865,7 +873,7 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
 
                     {/* Payload schema — one section per bindable DATA_BINDING param (a handler such
                         as CDC onUpdate exposes both a before- and an after-image). */}
-                    {payloadParams
+                    {groupedPayloadParams
```
NipunaRanasinghe: Stale after this change: it's now one section per binding *group*, and CDC `onUpdate` renders a single section rather than "both a before- and an after-image".

Same problem at lines 457-458 (outside this diff), where `payloadParams` is still documented as "each configured independently below" — grouped params are no longer configured independently.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
LakshanWeerasinghe: Fixed in 87f02ee8e3 — the grouped section now renders whichever binding-group sibling actually holds the bound schema (via a new `boundRepresentativeOf` helper), instead of always deciding from the group's first-listed member.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
LakshanWeerasinghe: This is intentional — a binding group's siblings (CDC's before/after) are meant to share one type, since they're the same row shape before/after the update, so a single editor writing the whole group is correct by design.

The concrete overwrite scenario you walked through no longer applies after the fix in 87f02ee8e3: `hasDefaultPayload`/the "Define" link now reflect whichever sibling actually holds a bound schema, so the "Define" flow only fires when the *whole* group is still unbound, and the inline editor always previews the real bound value first rather than hiding it. Added a short comment at both fan-out sites (8c1ff09b70) to make the shared-type intent explicit so this doesn't read as an oversight again.

LakshanWeerasinghe: Intentional — before/after share one row shape by design, so one editor writing the whole group is correct. After 87f02ee8e3, hasDefaultPayload reflects whichever sibling is actually bound, so "Define" only appears when the whole group is unbound and the inline editor always shows the real value first — the overwrite scenario above can't happen anymore.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
LakshanWeerasinghe: Checked every shipped trigger model; onUpdate's before/after are the only multi-payload case anywhere, and they always share one bindingGroup, so no handler today hits length > 1. Reworded the comment in ac7fff1c85 to describe the real condition instead of the now-wrong onUpdate example.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
LakshanWeerasinghe: Fixed in [e8d6718320](https://github.com/ballerina-platform/ballerina-vscode/pull/874/commits/f68841a8fcc3a30e1d077bf868e4b2e8d6718320)

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
NipunaRanasinghe: This one is still open — `ac7fff1c85` is not on the branch (`gh api repos/ballerina-platform/ballerina-vscode/commits/ac7fff1c85` returns 422, and the branch only picked up `87f02ee8e3` and `f68841a8fc`). At `f68841a8` the comment still points at CDC onUpdate as the case that gets the name appended, which is now the one case that does not. Looks like a local commit that never got pushed.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
NipunaRanasinghe: Thanks, `boundRepresentativeOf` does the job — I traced the section and `displayParam` is used consistently everywhere it needs to be, so the add-link (and the overwrite behind it) can no longer show up while a sibling is bound.

One thing left: `boundRepresentativeOf` has no test, even though this PR added a describe block for its three sibling helpers at `payloadComposer.test.ts:422`. The asymmetric case is the whole reason it exists, so it would be worth two cases there — one where a later sibling is the bound one, and one where none are bound and it falls back to the first.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
NipunaRanasinghe: New wording is accurate, thanks.

Small thing: `87f02ee8e3` also dropped `// The first payload param still drives shared UI bits (e.g. the variant dropdown's label).` at line 457. Only the other one was stale — `payloadParam` is still live at line 460 and used at line 810 for the Format/Variant label, so that comment was doing useful work. Worth putting back.

LakshanWeerasinghe: The param already carries the semantic meaning. No need for separate inline comment.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
LakshanWeerasinghe: yes. Removed that local commit. For a different group there will be different param label. Currently there are no such cases.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/payloadComposer.ts
```diff
@@ -383,6 +415,13 @@ export function hasDefaultPayload(param: ParameterModel): boolean {
     return !param.type?.codedata?.boundType;
 }
 
+/** The group's already-bound sibling, if any, so an asymmetric pre-existing binding isn't hidden
+ * behind the always-first-listed, possibly-still-unbound representative (e.g. CDC's `before`). */
+export function boundRepresentativeOf(fn: FunctionModel, param: ParameterModel): ParameterModel {
```
NipunaRanasinghe, line 445: `boundRepresentativeOf` picks the first *bound* sibling, and every edit path (`handleTypeCreated`, the `Parameters` `onChange`, `handleDeletePayloadSchema`) writes `boundType` to *all* siblings. That means a pre-existing `onUpdate(before: Employee, after: EmployeeV2)` renders as a single section showing only `before`, gives no indication `after` differs, and the first schema edit silently overwrites `after`.

Forcing the two to match is defensible — for CDC, before/after *are* the same row shape — but hiding the divergence isn't: the user can neither see nor selectively fix `after`.

Cheapest mitigation: when the siblings' `boundType`s disagree, don't collapse the group — fall back to rendering one section per parameter (the pre-PR behaviour). Grouping then only applies where the invariant actually holds.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
```diff
@@ -863,21 +869,23 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
...
-                    {/* Payload schema — one section per bindable DATA_BINDING param (a handler such
-                        as CDC onUpdate exposes both a before- and an after-image). */}
-                    {payloadParams
+                    {/* Payload schema — one section per binding group (CDC onUpdate's before/after
+                        share a group and render as a single section; see groupedPayloadParams). */}
+                    {groupedPayloadParams
                         .filter((param) => param.type?.codedata?.bindable === true)
                         .map((param) => {
-                            const label = displayLabelOf(param);
+                            const displayParam = boundRepresentativeOf(functionModel, param);
```
NipunaRanasinghe: The `bindable` filter on line 875 runs against the group's *first* member, but the parameter actually rendered is `displayParam`, which may be a different sibling. So:

- a group whose first member isn't `bindable` disappears from the form entirely, even when a sibling is bindable;
- conversely the filter can admit a group whose rendered `displayParam` isn't the one that was checked.

Both CDC params are bindable today, so this is latent — but the filter and the render subject should be the same parameter. Suggest computing `displayParam` first and filtering on it (or on "any sibling is bindable"), e.g. map to `{ param, displayParam }` before the `.filter`.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
```diff
@@ -538,9 +539,13 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
             return;
         }
         const typeName = typeof type === "string" ? type : type.name;
+        const target = payloadParams.find((p) => p.name?.value === targetName);
         // The parameter keeps its schema-shipped name — only the bound shape changes.
+        const groupNames = new Set(
+            (target ? bindingGroupSiblingsOf(functionModel, target) : []).map((p) => p.name?.value)
+        );
```
NipunaRanasinghe: This name-set indirection is a correctness hazard, and it's repeated in three places (here, line 567, and line 936).

`siblings.map((p) => p.name?.value)` puts `undefined` in the set whenever a payload param has no `name.value`; the subsequent `groupNames.has(p.name?.value)` then matches *every* unnamed payload param and rewrites its `boundType`.

Since `bindingGroupSiblingsOf` returns objects out of the same `fn.parameters` array, identity matching is both simpler and safe:

```js
const siblings = new Set(bindingGroupSiblingsOf(functionModel, target));
const parameters = functionModel.parameters.map((p) => siblings.has(p) ? { ...p, /* ... */ } : p);
```

Worth a short comment noting why the obvious-looking alternative — comparing `bindingGroupOf(p) === bindingGroupOf(target)` — is wrong: for ungrouped params `undefined === undefined` would collapse all of them into one group.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
```diff
@@ -863,21 +869,23 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
...
+                    {groupedPayloadParams
                         .filter((param) => param.type?.codedata?.bindable === true)
                         .map((param) => {
-                            const label = displayLabelOf(param);
+                            const displayParam = boundRepresentativeOf(functionModel, param);
+                            const label = displayLabelOf(displayParam);
                             return (
-                                <Fragment key={param.name?.value ?? label}>
-                                    {hasDefaultPayload(param) ? (
+                                <Fragment key={displayParam.name?.value ?? label}>
```
NipunaRanasinghe: `displayParam` is derived from which sibling is currently bound, so this key changes as a *result* of editing inside its own subtree: bind the schema on a group where only `after` was bound, both siblings get `boundType`, `boundRepresentativeOf` flips to `before`, the key goes `"after"` → `"before"`, and the `Parameters` subtree unmounts mid-edit.

Key on the stable group identity instead:

```jsx
<Fragment key={bindingGroupOf(param) ?? param.name?.value ?? label}>
```

(Note this is the same reasoning as the `AnnotationConfigSection` key below, which is deliberately keyed on variant identity rather than the editable name value.)

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/payloadComposer.ts
LakshanWeerasinghe: Fixed in [efa146cb3f5394e7e](https://github.com/ballerina-platform/ballerina-vscode/pull/874/commits/34cc2419a4b0f8f2f6f68f3efa146cb3f5394e7e).

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
LakshanWeerasinghe: Fixed in [efa146cb3f5394e7e](https://github.com/ballerina-platform/ballerina-vscode/pull/874/commits/34cc2419a4b0f8f2f6f68f3efa146cb3f5394e7e).

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
LakshanWeerasinghe: Fixed in [efa146cb3f5394e7e](https://github.com/ballerina-platform/ballerina-vscode/pull/874/commits/34cc2419a4b0f8f2f6f68f3efa146cb3f5394e7e).

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
LakshanWeerasinghe: Fixed in [efa146cb3f5394e7e](https://github.com/ballerina-platform/ballerina-vscode/pull/874/commits/34cc2419a4b0f8f2f6f68f3efa146cb3f5394e7e).

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/payloadComposer.ts
```diff
@@ -280,6 +333,10 @@ export function computeHandlerGroups(serviceModel: ServiceModel): HandlerGroup[]
         if (!group.needsForm) {
             group.quickAddFunction = members[0];
         }
+        const addDescription = members.find((member) => member.addDescription)?.addDescription;
+        if (addDescription) {
+            group.description = addDescription;
```
NipunaRanasinghe, line 338: `members.find(...)` makes the group's tooltip order-dependent: if two variants in a group ship different `addDescription` values, whichever the catalog lists first silently wins and the other is never shown. Today every variant in a group carries the identical string (all five FTP/azure/smb variants say "Invoked when a file is created in the monitored directory."), so it's consistent by luck of the data rather than by construction.

Two things worth doing:

1. A unit test for this override — `computeHandlerGroups` currently has coverage for grouping and `needsForm`, but nothing asserts that `addDescription` replaces `metadata.description`, nor what happens when only some members carry it.
2. Consider whether disagreement should be a defined outcome rather than first-wins. If `addDescription` is conceptually a property of the *group*, the group's members disagreeing is a model authoring error; picking one silently makes it invisible.

Not blocking — the shipped data is uniform.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/index.tsx
```diff
@@ -504,8 +506,11 @@ export function ServiceDesigner(props: ServiceDesignerProps) {
...
         const groupId = selectedTriggerGroup ?? (functionModel ? handlerGroupId(functionModel) : undefined);
         const groupMembers = [...(serviceModel ? catalogFunctionsOf(serviceModel) : []), ...(serviceModel?.functions ?? [])];
-        const groupLabel = groupMembers.find(fn => handlerGroupId(fn) === groupId)?.metadata?.label
-            ?? "Handler";
+        const matchedFunction = groupMembers.find(fn => handlerGroupId(fn) === groupId);
+        const groupLabel = matchedFunction?.metadata?.label ?? "Handler";
+        if (matchedFunction && repeatBehaviorOf(matchedFunction) !== RepeatBehavior.TRUE) {
+            return `Configure ${groupLabel} ${isMcpService ? "Tool" : "Handler"}`;
+        }
         return `${isNew ? "New " : ""}${groupLabel} Configuration`;
```
NipunaRanasinghe, line 514: Two things here:

**The `"Tool"` arm looks unreachable.** This branch requires `repeatBehaviorOf(matchedFunction) !== RepeatBehavior.TRUE`, but MCP tool groups are exactly the repeat-always case — `isSoleRepeatableGroup` identifies MCP's `Tool` by `repeatable === RepeatBehavior.TRUE`. So an MCP service should always fall through to the line below, and `isMcpService ? "Tool" : "Handler"` can only ever produce `"Handler"`. Either the condition is wrong or the ternary is dead code; worth confirming which.

**Two title shapes now coexist.** Non-repeatable groups get `Configure On Create Handler`, repeatable ones keep `New Handler Configuration` / `Handler Configuration`. If that's the intended end state, fine — but it reads like a half-finished migration, and the non-repeatable branch also drops `isNew`, so creating and editing a handler now show an identical title.

Also note `groupLabel` falls back to `matchedFunction.metadata.label`, which for authoring-synthesized triggers is `option.name()` (see `TriggerModelSynthesizer.buildFunctionFromAuthoring`, where `label = many ? "Handler" : option.name()`). For a non-repeatable synthesized handler that yields a camelCase title like `Configure onMessage Handler`. The bundled models all ship a humanized `label`, so this only bites the synthesized path.

Both this branch and the `addDescription` override would be cheap to pin down with an L2 render test — CodeRabbit flagged the same gap.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/SchemaDrivenSourceGenerator.java
```diff
@@ -161,9 +161,17 @@ private static String buildImports(ServiceInitModel filledInitForm, TriggerUISch
...
                 }
                 String org = moduleRef.substring(0, slash);
-                String module = moduleRef.substring(slash + 1);
+                String rest = moduleRef.substring(slash + 1).trim();
+                String module = rest;
+                String alias = null;
+                int asIndex = rest.lastIndexOf(" as ");
+                if (asIndex > 0) {
+                    module = rest.substring(0, asIndex).trim();
+                    alias = rest.substring(asIndex + 4).trim();
```
NipunaRanasinghe, line 170: The parsing is correct — I checked `getImportStmt(org, module, alias)` drops the `as` clause only when the alias equals the module's natural last-segment prefix, so `ballerinax/mssql.cdc.driver as _` keeps its `_` and the duplicate-emission bug is genuinely fixed.

Two follow-ups:

**The javadoc above is now stale.** Line 145 still says the additional imports are "each an `org/module` reference"; they can now carry `as <alias>`. Worth stating the accepted form, since the trigger-model JSON is hand-authored against this contract.

**Consider moving the parse into `Utils`.** It's the exact inverse of `Utils.getImportStmt(org, module, alias)` and sits next to `existingImportPrefix` / `importPrefixOf`, which already own the "what prefix is this module bound to" logic. Keeping the split-on-`" as "` next to the method that joins on it makes them easier to keep consistent than having one in the generator and one in `Utils`.

One behavioral note, no change requested: `importExists(rootNode, org, module)` ignores the alias, so if the file already imports the module under a *different* prefix, the aliased import is skipped. That's the right call — Ballerina rejects the same module imported twice — but it does mean the generated service body assumes a prefix the file may not bind. Only reachable if a user hand-edits the driver import, so I'd leave it.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/models/TriggerUISchemaModel.java
```diff
@@ -618,6 +627,7 @@ public record Metadata(
             String subLabel,
             String addLabel,
             String groupName,
-            String badge) {
+            String badge,
+            String addDescription) {
```
NipunaRanasinghe, line 631: Adding one component to this positional record cost ~20 mechanical call-site edits in this PR — every `new Metadata(label, description, null, null, null, null, null, null)` in `TriggerModelSynthesizer` and `PropertyValueAdapter` grew another `null`, and `PropertyValueAdapterTest` now constructs a `Codedata` with 26 positional args where the only meaningful one is `nameEditable`.

`Codedata` already has a builder, and this PR correctly extended it (`bindingGroup`) with a one-line change and zero call-site churn. `Metadata` doesn't have one, which is the whole difference in blast radius between the two new fields.

Suggest a builder for `Metadata` as a follow-up before the next field lands. It would also make the trailing-`null` call sites self-documenting — right now `new Metadata("Listener", "The listener this service attaches to", null, null, null, null, null, null, null)` gives no clue which of those nulls is `badge` versus `addDescription`, and a mis-ordered argument is silently type-correct since every component is a `String`.

Nothing wrong with the change as written — flagging the trend, not this diff.

LakshanWeerasinghe: I'll introduce a builder in a later PR.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/SchemaDrivenSourceGenerator.java
LakshanWeerasinghe: I'll do this refactor suggestion in a follow up PR.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/index.tsx
LakshanWeerasinghe: MCP has use case if the UI model default type set to AdvanceService or AdvanceStreamableService.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/payloadComposer.ts
LakshanWeerasinghe: The json models will be updated later. The UI is just a generator that should work for any case. Since I see no issue.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/index.tsx
NipunaRanasinghe: You're right, the arm is reachable — my "unreachable" claim was wrong. It's actually reachable more broadly than just the advanced types: `TriggerModelSynthesizer.buildFunctionFromAuthoring` never populates `repeatable` (it passes `null` for that component, `TriggerModelSynthesizer.java:534`), and `Repeatable.orDefault(null)` returns `FALSE`. MCP has no bundled `trigger-models/mcp.json` — it's synthesized from `trigger-metadata-models/mcp/trigger-metadata.json` — so *every* MCP handler resolves to `FALSE` and enters this branch, regardless of service type.

That does surface two titles worth fixing, though, both from `label = many ? "Handler" : option.name()` (`TriggerModelSynthesizer.java:530`):

| MCP service type | `addMode` | `metadata.label` | Resulting title |
|---|---|---|---|
| `service`, `streamableHttpService` | `many` | `"Handler"` | `Configure Handler Tool` |
| `streamableHttpAdvancedService` | `subset` | `"onListTools"` | `Configure onListTools Tool` |

The first is a regression from this PR — it read `New Handler Configuration` before. The second is the camelCase case I mentioned above; it turns out to be live rather than hypothetical.

Suggested fix, both small:
- Run the synthesized label through `humanize(...)` as `buildServiceType`/`buildParameterFromFacts` already do, so `onListTools` → `On List Tools`.
- For the `many` case the label is literally `"Handler"`, so `Configure ${groupLabel} ${noun}` composes to `Configure Handler Tool`. Either use the noun alone when the label already *is* the generic noun, or key the title off the noun rather than concatenating both.

I confirmed MCP reaches this panel: `isSchemaTriggerService` is true for it (synthesized `schemaFunctions`), and it's the `PanelContainer` at `ServiceDesigner/index.tsx:1341` that calls `getTriggerHandlerTitle()`.

Unrelated and pre-existing, but it falls out of the same `orDefault(null) → FALSE` fact, so flagging for whoever owns the MCP flow: `isSoleRepeatableGroup` requires `repeatable === RepeatBehavior.TRUE` and is documented as existing for "MCP's `Tool`" — that shortcut can never fire today. Not this PR's problem.

NipunaRanasinghe (approved):
Approving. 

One thing I'd like addressed before merge, or in the immediate follow-up — see the MCP thread: because `TriggerModelSynthesizer` never populates `repeatable`, every MCP handler takes the new non-repeatable title branch, which renders `Configure Handler Tool` for the plain/streamable service types (previously `New Handler Configuration`) and `Configure onListTools Tool` for the advanced one. Cosmetic, but it is a regression rather than a pre-existing wart.

Remaining non-blocking items, all acknowledged as follow-ups: `Metadata` builder, moving the import-alias parse into `Utils`, the now-stale `buildImports` javadoc at `SchemaDrivenSourceGenerator.java:145` (a one-line fix independent of that refactor), and a unit test for the `addDescription` override in `computeHandlerGroups`.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/index.tsx
LakshanWeerasinghe: Getting Configure here is correct. Only the Tool schema function with repeatable set to "TRUE" should get the "New" prefix.

PR #875, opened 2026-08-17
Sync language server hotfixes from ballerina-language-server

pasindufernando1:
## Purpose

Move eight merged hotfixes of the old language server repo,
[ballerina-platform/ballerina-language-server](https://github.com/ballerina-platform/ballerina-language-server),
into `packages/ballerina-language-server` of this repo. They are carried as a single PR rather than eight,
since several of them touch the same files and the intermediate states would not build.

## PRs included

| Old repo PR | Title |
| --- | --- |
| [#972](https://github.com/ballerina-platform/ballerina-language-server/pull/972) | Fix missing method name and argument types in resource call interactions |
| [#973](https://github.com/ballerina-platform/ballerina-language-server/pull/973) | Resolve types from the module when the requested file is absent |
| [#977](https://github.com/ballerina-platform/ballerina-language-server/pull/977) | Fix visualizer crash when searching connections |
| [#979](https://github.com/ballerina-platform/ballerina-language-server/pull/979) | Report expression bodied resource accessors as resources |
| [#980](https://github.com/ballerina-platform/ballerina-language-server/pull/980) | Fix empty type list when the file is absent |
| [#981](https://github.com/ballerina-platform/ballerina-language-server/pull/981) | Handle empty packageNames in getNodeTemplate API |
| [#982](https://github.com/ballerina-platform/ballerina-language-server/pull/982) | Resolve types.bal when the package does not hold it |
| [#983](https://github.com/ballerina-platform/ballerina-language-server/pull/983) | Withhold the visualize code lens from a submodule construct |

They are applied in the order they were merged upstream, so that the later ones build on the earlier ones
(#977 extends the `FileSystemUtils` that #973 introduced, and #982 moves that same file).

## Deviations from the upstream diffs

Most of the eight applied unchanged. Three needed a hand merge, since the file has moved on in this repo:

- **#977 — `FlowModelGeneratorService.search()`.** Upstream replaces `workspaceManager.loadProject()` with
  `FileSystemUtils.resolveProject()` and takes the project path from `project.sourceRoot()` instead of
  `WorkspaceManager.projectRoot()`, which cannot resolve a file that is absent on disk. This repo has since added a
  `functions.bal` preload ahead of the project load, which needs the project path and has to run before the project
  is read. The project is therefore resolved first for its source root, and resolved again after the preload so that
  the `didOpen` is reflected in what the search sees. Both intents are kept.
- **#979 — `ModuleNodeTransformer.transform()`.** Upstream moves the `EXPRESSION_FUNCTION_BODY` branch below the
  resource and remote branches, so that an expression bodied resource accessor is reported as a resource rather than
  as a data mapper. This repo has since added workflow and activity branches after the remote one, so the branch is
  placed below those as well. Only the resource and remote cases change; every other classification of this repo is
  left as it is.
- **#982 — the move of `FileSystemUtils` to `model-generator-commons`.** The upstream diff swaps the import in the
  callers it knew of. This repo has gained sixteen more callers since (the workflow, activity and durable agent
  builders, `WorkflowManagerService`, and `WorkflowUtil`, which used the class without an import as it shared its
  package). All of them are swapped over.

One change is left out: the `stdlibJsonDataVersion` bump of #977 from `1.1.3` to `1.1.4`. The test expectations of
this repo already carry `1.1.4`, and the bump changes nothing here.

The other rejected hunks of #977 were the `1.1.3` to `1.1.4` version strings of four test configs, which this repo
already holds at `1.1.4`.

## Tests

`compileJava`, `compileTestJava`, `checkstyleMain` and `checkstyleTest` pass over the whole project.

Suites run, against this branch and against `staging/5.14.0` for a baseline:

| Suite | Baseline | This branch |
| --- | --- | --- |
| `langserver-core` `CodeLensTest` | pass | pass |
| `architecture-model-generator-ls-extension` | pass | pass |
| `sequence-model-generator-ls-extension` | pass | pass |
| `flow-model-generator-ls-extension` | 1965 tests, 1 failed | 1969 tests, 3 failed |

The three failures of `flow-model-generator-ls-extension`:

- `diagram_generator/config/diagnostics4.json` — **fails on `staging/5.14.0` as well**, with the same message
  (`nodes[5].properties.indentation.defaultValue`, actual `4`, expected `0`). Pre-existing, not from this PR.
- `search/config/connectors/caller_with_missing_connections_bal.json` — the test config that #977 adds. The API
  under test works: the search returns connectors for a package whose `connections.bal` is absent, which is the
  crash that #977 fixes. The expected list was recorded against the connector index of the old repo, and this repo
  carries a different one (`googleapis.chat` callers). The expectation needs to be regenerated here.
- `node_template/config/resource_action_call-customclient-get.json` — a regression to be investigated.

The last two are being followed up on separately; this PR is opened first so the port itself can be reviewed.

## Summary

- Ported eight language server hotfixes.
- Improved resource calls, missing-file handling, project resolution, type lookup, and package-name resolution.
- Moved `FileSystemUtils` to `model-generator-commons`.
- Added support for missing `types.bal` files and single-file project paths.
- Limited visualization code lenses to default modules.
- Improved agent-call return-type handling and test cleanup.
- Updated flow, sequence, and service model tests.
- Added error details to `GetTypesResponse` and client-side logging.

Compilation and checkstyle checks pass. Most test suites pass. Three flow-model test issues remain: one pre-existing failure, one requiring regenerated expectations, and one under investigation.


packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/main/java/io/ballerina/flowmodelgenerator/extension/TypesManagerService.java
```diff
@@ -134,17 +135,17 @@ public CompletableFuture<TypeListResponse> getTypes(FilePathRequest request) {
...
                     return response;
                 }
-                TypesManager typesManager = new TypesManager(document.get());
-                JsonElement allTypes = typesManager.getAllTypes(semanticModel.get());
+                TypesManager typesManager = new TypesManager(moduleModel.get().document());
+                JsonElement allTypes = typesManager.getAllTypes(moduleModel.get().semanticModel());
                 response.setTypes(allTypes);
             } catch (Throwable e) {
-                throw new RuntimeException(e);
+                response.setError(e);
```
NipunaRanasinghe, line 149: This makes failures invisible. `types` stays null here, and the consumers only guard on it — `TypeDiagram/index.tsx:333` does `if (response?.types)` and `ReadonlyTypeDiagram.tsx:64` the same. So an error now renders an empty type diagram with nothing logged. Before, the throw reached the `.catch` in `rpc-manager.ts:1957` which logged it. Nobody reads `errorMsg`.

Suggest also setting `response.setTypes(new JsonArray())` on this path, like the `moduleModel.isEmpty()` branch above, and narrowing `Throwable` to `Exception`.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/FileSystemUtils.java
```diff
@@ -123,6 +145,87 @@ public static SemanticModel getSemanticModel(WorkspaceManager workspaceManager,
...
+        }
+
+        Path parentPath = filePath.getParent();
+        if (parentPath == null) {
+            return Optional.empty();
+        }
+
+        Project project = resolveProject(workspaceManager, filePath);
+        Package currentPackage = project.currentPackage();
+        Module module = workspaceManager.module(parentPath).orElseGet(currentPackage::getDefaultModule);
```
NipunaRanasinghe: This falls back to the default module for a submodule file, which gives a wrong answer rather than no answer.

`BallerinaWorkspaceManager.module(Path)` (langserver-core, line 265) only returns the default module when the path equals the project root. For a submodule directory like `modules/utils` the document lookup is empty and the path is not the project root, so it returns empty. So for an absent `modules/utils/types.bal`, `parentPath` is `modules/utils`, `module()` is empty, and this hands back the default module — `getTypes` then returns the default module's types.

The new `test_pack4` is single-module, so this path isn't covered. If submodules are out of scope for now (#983 disables the visualizer for them anyway), that's fine, but it would be better to return empty here than to silently answer with the wrong module.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/FileSystemUtils.java
```diff
@@ -196,5 +298,8 @@ public static void deleteCreatedFiles() {
                 throw new RuntimeException("Error occurred while deleting the file: " + path, e);
             }
         });
+        // The list is cleared so that it holds the files of the current test class alone, and so that it does not
+        // grow without a bound in a language server that is long lived.
```
NipunaRanasinghe: The second half of this comment isn't accurate. `deleteCreatedFiles()` has only two callers, both tests (`AbstractLSTest.shutDownLanguageServer` and `SourceGeneratorTest`). Nothing in the running server calls it, so `CREATED_FILES` still grows for the life of the process. The clear only helps the test case.

The growth itself is pre-existing and small (just `Path` objects), so no change needed — just drop the "language server that is long lived" claim so the comment doesn't read as if production is covered.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/artifactsgenerator/ModuleNodeTransformer.java
```diff
@@ -132,6 +121,17 @@ public Optional<Artifact> transform(FunctionDefinitionNode functionDefinitionNod
             functionBuilder
                     .name(functionName)
                     .type(Artifact.Type.ACTIVITY);
+        } else if (functionDefinitionNode.functionBody().kind() == SyntaxKind.EXPRESSION_FUNCTION_BODY) {
```
NipunaRanasinghe, line 124: The PR description says "Only the resource and remote cases change", but moving this branch below the workflow and activity branches too means an expression-bodied workflow or activity function now comes out as `WORKFLOW`/`ACTIVITY` instead of `DATA_MAPPER`/`NP_FUNCTION`.

If that's intended, or if workflow/activity functions can't have expression bodies in practice, worth saying so in the description so the next person reading it doesn't have to work it out.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/DatabindUtil.java
```diff
@@ -1040,23 +1044,45 @@ public static Map<String, List<TextEdit>> createTypeDefinitionEdits(Project proj
...
+     * @param workspaceManager The workspace manager for document retrieval
+     * @return The types.bal Document, or null if the path cannot be resolved
+     */
+    private static Document getOrCreateTypesDocument(String contextFilePath,
+                                                     WorkspaceManager workspaceManager) {
+        Path typesFilePath = getFilePathForFile(contextFilePath, workspaceManager, TYPES_BAL);
+        if (typesFilePath == null) {
+            return null;
+        }
+        return FileSystemUtils.getDocument(workspaceManager, typesFilePath);
```
NipunaRanasinghe, line 1077: `FileSystemUtils.getDocument` writes the file to disk, so this creates `types.bal` in the user's project while only *computing* the edits. If the user then cancels the data-binding operation, they're left with an empty `types.bal`, and nothing in the server removes it (`deleteCreatedFiles()` is test-only).

I think this is needed for the returned TextEdits to apply, so probably intended — just confirming that's the case.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/expressioneditor/DocumentContext.java
```diff
@@ -193,7 +202,7 @@ private void initialize() {
         }
         DocumentId documentId = documentIds.iterator().next();
         document = module.document(documentId);
-        filePath = inputFilePath.resolve(document.name());
+        filePath = modulePath.resolve(document.name());
```
NipunaRanasinghe, line 209: Same submodule problem as in `FileSystemUtils.resolveModuleModel`, and it's worse here. For an absent file inside a submodule, `workspaceManager().module(modulePath)` returns empty (see `BallerinaWorkspaceManager.module` line 265), so `module` becomes the default module. This line then builds `modules/utils/<default-module-doc>.bal`, a path that doesn't exist, and `module()`/`semanticModel()` go back to returning empty — so the fix doesn't apply for submodules.

For a file in the default module it works, which is the case the PR targets. Worth noting the limitation somewhere.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/expressioneditor/DocumentContext.java
```diff
@@ -152,8 +155,13 @@ private void initialize() {
         if (initialized) {
             return;
         }
+
+        Optional<Project> optProject = project();
+        if (optProject.isEmpty()) {
+            throw new IllegalStateException("Project not found for the file: " + inputFilePath);
```
NipunaRanasinghe, line 161: Minor: this used to fail as a `ProjectException` (or the `RuntimeException` from `PackageUtil.loadProject`) and now fails as `IllegalStateException`. Most callers catch broadly so it's probably fine, but worth a quick check that nothing catches the old type specifically.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/AbstractLSTest.java
```diff
@@ -446,6 +446,7 @@ protected void startLanguageServer() {
 
     @AfterClass
     public void shutDownLanguageServer() {
+        FileSystemUtils.deleteCreatedFiles();
```
NipunaRanasinghe, line 450: This puts shared static state into a base class that every LS test extends. It's safe today — the flow-model and service-model suites are `parallel="none"`/`"false"` and Gradle forks a JVM per module — but if anyone turns on parallel classes, one class's `@AfterClass` will delete files another class is still using. A short comment saying it depends on serial execution would save someone a confusing failure later.

packages/ballerina-language-server/sequence-model-generator/modules/sequence-model-generator-core/src/main/java/io/ballerina/sequencemodelgenerator/core/ParticipantBodyAnalyzer.java
```diff
@@ -100,14 +102,16 @@ public void visit(ClientResourceAccessActionNode resourceAccessActionNode) {
...
+        resourceAccessActionNode.methodName().ifPresentOrElse(
+                name -> nodeBuilder.property(Interaction.NAME_LABEL, Expression.Factory.createStringType(name)),
+                () -> nodeBuilder.property(Interaction.NAME_LABEL,
+                        Expression.Factory.createStringType(Constants.DEFAULT_RESOURCE_METHOD)));
 
         SeparatedNodeList<Node> nodes = resourceAccessActionNode.resourceAccessPath();
         String resourcePath = nodes.stream()
                 .map(Node::toSourceCode)
-                .collect(Collectors.joining("/"));
+                .collect(Collectors.joining("/", "/", ""));
```
NipunaRanasinghe, line 114: Just noting: nothing reads `resourcePath`. `packages/sequence-diagram` builds the interaction label from `properties.name` and `properties.params` only (`ElementFactoryVisitor.ts:155`). So the leading slash (and `""` becoming `"/"`) is a model-only change with no visible effect.

The `get` default a few lines up is a real fix though — the label was previously rendering as `undefined (params)`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/main/java/io/ballerina/flowmodelgenerator/extension/TypesManagerService.java
pasindufernando1: Good catch — the visible half is fixed in `b0aab6ca`: the catch now sets an empty `JsonArray`, so a failure renders an empty diagram instead of one built off a null `types` field.

I also made the error actually reachable rather than just stored. `GetTypesResponse` in `ballerina-core` now declares `errorMsg`/`stacktrace` (the LS was already sending them, TS just didn't model them), and `rpc-manager.ts` logs them when present. That restores the output-channel logging the old `throw` gave us without going back to rejecting the promise.

On narrowing `Throwable` to `Exception` — I've left it as `Throwable`. This file catches `Throwable` in 15 places, and two of them (lines 478 and 576) catch `OutOfMemoryError` deliberately and turn it into a user-facing message; deep record/type expansion in this service does hit OOM and `StackOverflowError` in practice. Narrowing just this one site would be inconsistent with the rest of the file and would let an `Error` kill the async worker instead of being reported. Happy to change it if you'd rather we take the `Throwable` usage across the whole file as a separate cleanup.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/FileSystemUtils.java
pasindufernando1: Agreed, and I took the option you suggested — `b0aab6ca` returns `Optional.empty()` when `workspaceManager.module(parentPath)` is empty, instead of falling back to the default module.

Confirmed the mechanism you described: `BallerinaWorkspaceManager.module(Path)` only returns the default module when the path equals the project root (line 275), so the old `orElseGet` was only ever exercised in the wrong-module case. The default-module case is unaffected — for `test_pack4/types.bal` the parent *is* the project root, so `module()` resolves normally and `GetAllTypesTest` still passes.

`resolveModuleModel` has exactly one caller (`getTypes`), so nothing else changes behaviour. Submodules stay out of scope, but they now return nothing rather than the wrong module.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/FileSystemUtils.java
pasindufernando1: You're right — checked, and the only callers are `AbstractLSTest.shutDownLanguageServer` and `SourceGeneratorTest`, both tests. Dropped the "language server that is long lived" clause in `b0aab6ca` and kept the accurate half.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/artifactsgenerator/ModuleNodeTransformer.java
pasindufernando1: Intended. The ordering is meant to express that what a function *is* (resource, remote, workflow, activity) outranks the shape of its body — an expression-bodied function carrying `@workflow:Workflow` is a workflow, not a data mapper. Workflow/activity detection is annotation-based, so the combination you describe is syntactically possible and the behaviour change is real; we just consider the new classification the correct one.

You're right that the description is wrong, though — "Only the resource and remote cases change" doesn't cover this. Updating it to call out the workflow/activity cases too.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/DatabindUtil.java
pasindufernando1: Yes, intended. Traced the chain to be sure:

```
SchemaDrivenFunctionBuilder.addModel/updateModel
  -> IncludedRecordBinder.forAdd/forUpdate
    -> DatabindUtil.createTypeDefinitionEdits
      -> prepareTypeDefinitionEditContext -> getOrCreateTypesDocument
```

Those are the mutation entry points, not a preview path — the user has already committed and the returned TextEdits are applied straight away. A plain LSP TextEdit can't bring the file into existence either (that needs a `CreateFile` resource operation), and the edit positions are anchored to this document, so it has to exist first. The cancel case isn't reachable from these callers.

Added that reasoning to the javadoc in `b0aab6ca` so it doesn't have to be re-derived.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/expressioneditor/DocumentContext.java
pasindufernando1: Fixed rather than just noted, in `b0aab6ca`. The underlying problem was that `modulePath` was doing two jobs — the lookup key *and* the directory the reserved file is placed in — and those diverge exactly when the lookup fails. Split into `lookupPath` (what we ask `module()` about) and `modulePath` (where the resolved module's documents actually live), so the default-module fallback now resolves against `project.sourceRoot()` instead of the submodule directory. Same idiom as `ModelGenerator:384` and `CodeAnalyzer:873`.

A submodule file now yields a real document with a matching path instead of `modules/utils/<default-module-doc>.bal`, and `module()`/`semanticModel()` no longer come back empty. It still evaluates in the default module's scope, which is a genuine remaining limitation for submodules — but it's no longer producing a path that can't exist.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/expressioneditor/DocumentContext.java
pasindufernando1: Checked — nothing catches the old type specifically. `io.ballerina.projects.ProjectException` extends `RuntimeException` (verified against the class in the jar), and every catch downstream of `DocumentContext.initialize()` is on `Throwable`, `Exception` or `RuntimeException`: `ExpressionEditorService` lines 113, 128, 222, 238, 275 and `ExpressionEditorContext:152`. The only `catch (ProjectException)` in the flow-model module is `DataMappingModelConverterUtils:143`, which isn't on this path. `IllegalStateException` is a `RuntimeException`, so every existing handler still catches it.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/AbstractLSTest.java
pasindufernando1: Fair point — added the comment in `b0aab6ca` noting that the tracked list is static and JVM-wide, that this depends on the test classes running serially, and what would break if parallel classes were turned on.

packages/ballerina-language-server/sequence-model-generator/modules/sequence-model-generator-core/src/main/java/io/ballerina/sequencemodelgenerator/core/ParticipantBodyAnalyzer.java
pasindufernando1: Confirmed both halves — `ElementFactoryVisitor.ts` builds the `ENDPOINT_CALL` label from `properties.name` alone, so `resourcePath` (and `params`, on that branch) aren't read today. Leaving the leading slash in: it makes the emitted path correct as a path, it's harmless, and `endpoint_call2.json` was already rebaselined for it. The `get` default is the one that fixes the visible `undefined (params)` label.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/expressioneditor/DocumentContext.java
```diff
@@ -169,17 +177,22 @@ private void initialize() {
...
+            modulePath = lookupPath;
         } else {
-            // Get the default module if not exists
-            Optional<Project> project = project();
-            if (project.isEmpty()) {
-                throw new IllegalStateException("Project not found for the file: " + inputFilePath);
-            }
-            module = project.get().currentPackage().getDefaultModule();
+            module = optProject.get().currentPackage().getDefaultModule();
+            modulePath = optProject.get().sourceRoot();
```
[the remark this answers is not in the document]
pasindufernando1: Applied the first half in 63a7886. Declining the second, and the investigation turned up something worth recording: the submodule case never reaches the branch this comment is about.

**1. Resolving the module directory — applied**

`modulePath` was doing double duty as the lookup key and as the directory the reserved file path is built from. Both branches now go through a `moduleRootOf()` helper that takes a file as its parent.

To be accurate about what this fixes: I could not construct a reachable case for the `main.bal/other.bal` path. It needs `module()` to return present while `lookupPath` denotes a file, and `BallerinaWorkspaceManager.module(Path)` (:265-281) returns present only when the document lookup succeeds — excluded, since we only reach line 182 because `document(inputFilePath)` came back empty and `module()` repeats that same lookup — or when `filePath.equals(projectRoot(filePath))`, which for a file requires `ProjectPaths.isStandaloneBalFile` (`computeProjectRoot:1514`), i.e. a single-file project, and those resolve their document so the early return at 172-178 fires first.

The shape actually exercised is a directory: the existing `types/config/non_existence.json` passes `data_mapper`, the package root, where `projectRoot(dir) == dir`. `Files.isDirectory` preserves that. So this is hardening plus an explicit invariant. It also covers a half the comment did not mention: `Project.sourceRoot()` is the `.bal` file for a single-file project, so the `else` branch carried the same asymmetry.

**2. Failing instead of the default module — declining, and the premise does not hold**

I measured this instead of reasoning about it, with a throwaway fixture package (default module plus `modules/utils`) through `expressionEditor/types`:

| input | result |
|---|---|
| absent file, default module | 25 completions, default module's types — works |
| package directory | 25 completions — works |
| absent file in submodule | throws |
| existing file in submodule | throws on a cold workspace |

The submodule failure is not the wrong module. It is an exception from line 189, before the `else` branch is evaluated:

```
at ...DocumentContext.initialize(DocumentContext.java:189)
at ...DocumentContext.semanticModel(DocumentContext.java:115)
at ...TypesGenerator.getTypes(TypesGenerator.java:117)
Caused by: io.ballerina.projects.ProjectException:
  '.../submodule_proj/modules/utils' does not belong to the current project
```

`workspaceManager().module(submoduleDir)` throws rather than returning empty. In every submodule shape I probed the default-module fallback was never reached, so "this returns a document from the wrong module" does not occur, and the proposed throw is already the de facto behaviour there.

For the shapes that do reach the fallback I am still declining, because `ExpressionEditorService.types:128-129` catches `Throwable` and returns an empty `CompletionList`, and `completion`/`diagnostics` route through `Debouncer`. A throw surfaces as silently no completions and no diagnostics rather than as a diagnosable error — the same symptom, moved.

**3. Submodules are out of scope**

Both submodule failures predate this PR: before it, line 189 was `module(inputFilePath)` on a path that does not exist, which `ProjectPaths.packageRoot` rejects outright. This PR fixed the default-module case, which is what `config14.json` covers, and left the submodule case failing by a different route.

WSO2 Integrator does not support submodules today, and #983 disables the visualizer for non-default modules, so we are not fixing submodule resolution here. Recording the limitation for whoever picks it up: an absent file in a submodule throws as above, and an existing file in a submodule resolves only if the package was already loaded through its root — cold, it throws too.

**4. Tests**

No fixture shipped for the submodule cases: neither can be made green, which is the finding above rather than an omission. I removed the temporary fixture package instead of committing a failing test. The default-module absent-file case remains covered by `config14.json`.

NipunaRanasinghe (dismissed):
Approving.

All nine review comments are answered — six fixed, three declined with reasoning I accept (the classification ordering, the `types.bal` creation on the mutation path, and the leading slash on `resourcePath`).

I verified the three substantive fixes in the code: `getTypes` now returns an empty array and surfaces `errorMsg`/`stacktrace` through to the client log; `resolveModuleModel` returns empty instead of answering with the wrong module; and `DocumentContext` splits the module lookup key from the directory the reserved file path is built from, which fixes the cause rather than the symptom I had asked to be documented.

NipunaRanasinghe (dismissed):

NipunaRanasinghe (approved):

PR #876, opened 2026-08-17
Add connector categorization and agent tool return types to the LS

dan-niles:
## Summary

Language-server half of #809, split out so LS and client can be reviewed and merged separately. **Merge this before the client PR.**

Resolves the LS side of https://github.com/wso2/product-integrator/issues/2152

- Resolve a category for every connector (`ConnectorCategoryResolver`) and expose a `GROUPED` connector set on the connector search, so clients can browse connectors by category instead of a shipped list.
- Index connector keywords and descriptions in `SearchDatabaseManager` so search matches on them.
- Resolve a usable return type for agent tools in `AgentToolBuilder`.
- Expand `connector_name_correction.json` so connector display names are consistent across all search results. This is an intentional product-wide relabel, not scoped to the new browser — 13 golden files are updated as a result.
- New tests: `ConnectorCategoryResolverTest`, `ConnectorIndexKeywordTest` (both registered in `testng.xml`).

## Related

- Client half: #877
- Original combined PR (kept open for reference): #809


packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
```diff
@@ -426,6 +454,10 @@ List<ToolParam> resolveParams(ToolGenContext ctx) {
 
             @Override
             ReturnInfo resolveReturn(ToolGenContext ctx) {
+                String chosen = dataString(ctx.data, RETURN_TYPE_KEY, "").trim();
+                if (!chosen.isEmpty()) {
+                    return new ReturnInfo(chosen, true, "The response from the " + ctx.agentVarName + " agent.");
+                }
```
NipunaRanasinghe, line 465: This early return hands back `chosen` before reaching `resolveAgentRunReturnType`, which is the only call site of `acceptTypeImports` on this path. A qualified cross-module type arriving via `returnType` is emitted into the generated tool verbatim with no matching `import`, producing source that does not compile.

The sibling branch immediately below does register imports:

https://github.com/ballerina-platform/ballerina-vscode/blob/724003c3959a85a2fe2882c00126e2e858058124/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java#L952-L955

Not reachable on this branch as far as I can tell: nothing populates `codedata.data.returnType` for an `AGENT_CALL` node today. The `RETURN_TYPE_KEY` in `NewActivityFromConnection.tsx` is a different key (`"type"`) that feeds `properties`, and `AgentDefinitionDesigner.handleOutputTypeSave` rewrites the agent's `run` signature in source with imports passed explicitly through `getImportsForProperty`. So this becomes live when the client half that sends `returnType` lands.

The same concern was raised on #809 and is still open there.

Suggest resolving the chosen type through its symbol and applying `acceptTypeImports` before returning, rather than trusting the raw string.

NipunaRanasinghe (commented):
### Code review

Found 1 issue. See the inline comment on `AgentToolBuilder.java`.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentToolBuilder.java
dan-niles: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/876/commits/c059d363ddfafc2666144fb96e159cc517625faf

NipunaRanasinghe (approved):
LGTM

MohamedSabthar (approved):

PR #877, opened 2026-08-17
Revamp `Use Connection` agent tool creation UX

dan-niles:
## Summary

Client half of #809, split out from the language-server changes. Revamps the `Use Connection` agent tool creation flow so users can discover connector capabilities before creating a connection.

Resolves https://github.com/wso2/product-integrator/issues/2152

**Depends on #876** for connector categories (`connectorSet: "GROUPED"`), corrected connector names, and agent tool return types. This builds and tests on its own, but category grouping and result-type resolution need the LS change to be in place.

https://github.com/user-attachments/assets/dae11089-d29a-4e85-8489-801f0ecf1cf7

- Replace the connection-first flow with a connector → action → tool form.
- Add a browsable connector catalogue with category navigation, sticky headers, curated Popular entries, and instant local filtering.
- Search Ballerina Central for connectors outside the catalogue, without duplicating catalogue results.
- Discover remote and resource actions from Central documentation, including OpenAPI-generated resource actions.
- Prefill and organize tool forms: suggested names/descriptions, type-filtered connection selection, collapsible Inputs, OAuth, and Result Type sections.
- Improve related action labels, connection-select behavior, form loading, navigation, and deletion behavior.
- Note one intentional product-wide relabel, not scoped to the new browser: `formatMethodName` now preserves acronyms (`getUserId` → "Get User ID"), which also affects its existing `GroupList` caller.

## Related

- Language-server half: #876
- Original combined PR (kept open for reference): #809

## Summary

- Reworked agent connection tool creation into a connector-first workflow.
- Added connector browsing, category filtering, local search, and action selection.
- Added discovery for remote, resource, and OpenAPI-generated actions.
- Improved tool forms with prefilled values, grouped collapsible sections, OAuth fields, result types, and type-filtered connection selection.
- Improved connection creation, navigation, loading states, error handling, deletion behavior, and user feedback.
- Added acronym-aware method-name formatting, including `getUserId` → `Get User ID`.
- Extended agent tool metadata with optional return types and connection names.
- Added shared form utilities and unit tests for connector actions, tool forms, validation, and method-name formatting.

### DIFF
--- a/packages/ballerina-side-panel/src/components/Form/index.tsx
+++ b/packages/ballerina-side-panel/src/components/Form/index.tsx
@@ -33,7 +33,7 @@ import {
 } from "@wso2/ui-toolkit";
 import styled from "@emotion/styled";
 
-import { ExpressionFormField, FieldDerivation, FormExpressionEditorProps, FormField, FormImports, FormValues } from "./types";
+import { ExpressionFormField, FieldDerivation, FieldGroup, FormExpressionEditorProps, FormField, FormImports, FormValues } from "./types";
 import { FieldFactory } from "../editors/FieldFactory";
 import { InputMode } from "../editors/MultiModeExpressionEditor/ChipExpressionEditor/types";
 import { getValueForDropdown, isDropdownField } from "../editors/utils";
@@ -66,6 +66,7 @@ import {
     formatJSONLikeString,
     updateFormFieldWithImports,
     hasIncompleteRequiredFormFields,
+    groupHasBlockingIssue,
     shouldRunExternalFormValidation,
     isPrioritizedField,
     hasRequiredParameters,
@@ -143,6 +144,85 @@ namespace S {
         border-top: ${({ topBorder }) => (topBorder ? `1px solid ${ThemeColors.OUTLINE_VARIANT}` : "none")};
     `;
 
+    export const GroupCard = styled.div`
+        width: 100%;
+        border: 1px solid ${ThemeColors.OUTLINE_VARIANT};
+        border-radius: 4px;
+        background-color: ${ThemeColors.SURFACE_DIM};
+        transition: background-color 120ms ease, border-color 120ms ease;
+
+        &:hover {
+            border-color: ${ThemeColors.PRIMARY};
+        }
+    `;
+
+    export const GroupHeader = styled.button`
+        display: flex;
+        align-items: center;
+        justify-content: space-between;
+        gap: 10px;
+        width: 100%;
+        padding: 10px 12px;
+        background: none;
+        border: none;
+        cursor: pointer;
+        text-align: left;
+        color: var(--vscode-editor-foreground);
+        font-family: var(--vscode-font-family);
+
+        &:focus-visible {
+            outline: 1px solid ${ThemeColors.PRIMARY};
+            outline-offset: -1px;
+            border-radius: 4px;
+        }
+    `;
+
+    export const GroupTitle = styled.span`
+        font-family: var(--vscode-font-family);
+        font-size: 13px;
+        color: var(--vscode-editor-foreground);
+    `;
+
+    export const GroupIssueIcon = styled.span`
+        display: flex;
+        align-items: center;
+        flex-shrink: 0;
+        margin-left: auto;
+        color: ${ThemeColors.ERROR};
+    `;
+
+    export const GroupChevron = styled.span<{ expanded?: boolean }>`
+        display: flex;
+        align-items: center;
+        flex-shrink: 0;
+        color: var(--vscode-descriptionForeground);
+        transform: rotate(${({ expanded }) => (expanded ? "180deg" : "0deg")});
+        transition: transform 150ms ease;
+    `;
+
+    export const GroupSection = styled.div`
+        display: flex;
+        flex-direction: column;
+        gap: 16px;
+        width: 100%;
+    `;
+
+    export const GroupDivider = styled.hr`
+        width: 100%;
+        margin: 0 0 8px;
+        border: 0;
+        border-top: 1px solid ${ThemeColors.OUTLINE_VARIANT};
+    `;
+
+    export const GroupBody = styled.div`
+        display: flex;
+        flex-direction: column;
+        gap: 20px;
+        padding: 16px 14px;
+        border-top: 1px solid ${ThemeColors.OUTLINE_VARIANT};
+        margin-top: -1px;
+    `;
+
     export const CheckboxRow = styled.div<{}>`
         display: flex;
         flex-direction: row;
@@ -171,7 +251,6 @@ namespace S {
         z-index: 10;
         width: 100%;
         padding: 16px 0 0;
-        background: var(--vscode-editor-background);
         border-top: 1px solid var(--vscode-panel-border);
     `;
 
@@ -364,6 +443,8 @@ export interface FormProps {
     // its `propertyPath` resolves to; anything unresolvable falls back to the form-level banner.
     serverValidationErrors?: ValidationResult[];
     preserveOrder?: boolean;
+    groups?: FieldGroup[];
+    opensPrefilled?: boolean;
     handleSelectedTypeChange?: (type: string | CompletionItem) => void;
     scopeFieldAddon?: React.ReactNode;
     onChange?: (fieldKey: string, value: any, allValues: FormValues) => void;
@@ -426,6 +507,8 @@ export const Form = forwardRef((props: FormProps, _ref) => {
         concertMessage,
         formImports,
         preserveOrder = false,
+        groups,
+        opensPrefilled = false,
         bottomFields = [],
         handleSelectedTypeChange,
         scopeFieldAddon,
@@ -503,6 +586,13 @@ export const Form = forwardRef((props: FormProps, _ref) => {
     }, [serverValidationErrors, setError]);
 
     const [showAdvancedOptions, setShowAdvancedOptions] = useState(props.defaultExpandAdvanced ?? false);
+    const [expandedGroups, setExpandedGroups] = useState<Record<string, boolean>>({});
+    const toggleGroup = (groupId: string) =>
+        setExpandedGroups((prev) => {
+            const group = (props.groups ?? []).find((candidate) => candidate.id === groupId);
+            const current = prev[groupId] ?? !(group?.defaultCollapsed ?? true);
+            return { ...prev, [groupId]: !current };
+        });
     const [activeFormField, setActiveFormField] = useState<string | undefined>(undefined);
     const [diagnosticsInfo, setDiagnosticsInfo] = useState<FormDiagnostics[] | undefined>(undefined);
     const [isMarkdownExpanded, setIsMarkdownExpanded] = useState(false);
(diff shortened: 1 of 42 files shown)


packages/ballerina-visualizer/src/views/BI/AIChatAgent/AIAgentSidePanel.tsx
```diff
@@ -762,7 +770,28 @@ export function AIAgentSidePanel(props: BIFlowDiagramProps) {
...
+            ...toolInputFields,
+            ...parameterFields.map((field) => ({
+                ...field,
+                value: typeof field.value === 'string' ? field.value.replace(/^\$/, '') : field.value,
+            })),
+        ].map((field) => ({
+            ...field,
+            group: isResultTypeField(field) ? RESULT_TYPE_GROUP : TOOL_INPUT_GROUP,
+            advanced: false,
+        }));
```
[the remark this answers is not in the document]
dan-niles: Not changing this. The `$` stripping is pre-existing upstream behaviour (7bb3a2d, already on `staging/5.14.0` and `main`) — staging has the identical line in both loaders; this PR only factored the two copies into `buildGroupedInputFields`.

Also, the `$` is stripped from the mapping **value**, not the identifier. That value has to reference the generated tool function's parameter, which `createToolInputFields` declares from the same `cleanKey = field.key.replace(/^\$/, '')`. Emitting `'foo` on only one side would desync the pair. Identifier escaping is done server-side by `SourceBuilder.param()` / `namedArg()`.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/UseAgentToolForm.tsx
```diff
@@ -73,27 +59,82 @@ export function UseAgentToolForm(props: UseAgentToolFormProps): JSX.Element {
...
+                : (await rpcClient.getVisualizerRpcClient().joinProjectPath({
+                    segments: [agentNode?.codedata?.lineRange?.fileName ?? "agents.bal"],
+                })).filePath;
             setAgentFilePath(filePath);
+            setOauthProperties(await fetchOAuthConfigProperties(rpcClient, filePath));
+            setDefaultReturnType(await fetchAgentRunReturnType(rpcClient, filePath, agentVarName,
+                hostClass?.className));
             setReady(true);
         })();
     }, [agentNode]);
```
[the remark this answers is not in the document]
dan-niles: Not changing this. `joinProjectPath` returns an empty `filePath` only when no project is open (`rpc-manager.ts:255`), where this panel is unreachable; the `hostClass` path skips the RPC entirely. Staging has the same unchecked usage, so it isn't new here.

The stated failure mode doesn't occur either: `fetchOAuthConfigProperties` and `fetchAgentRunReturnType` are both internally try/caught and return `[]` / `""`, so they never reject and `ready` can't get stuck. The suggested early `return` would leave a permanent spinner in exactly the case it's meant to protect.

packages/statement-editor/src/components/LibraryBrowser/ModuleElement/index.tsx
```diff
@@ -65,16 +65,22 @@ export function ModuleElement(props: ModuleElementProps) {
...
+                // tslint:disable-next-line: no-console
+                console.error("Failed to fetch library data", { moduleOrgName, moduleId, moduleVersion, error });
+            }
 
             let functionProperties: LibraryFunction = null;
-            response.docsData.modules[0].functions.map((libFunction: LibraryFunction) => {
+            response?.docsData?.modules?.[0]?.functions?.map((libFunction: LibraryFunction) => {
                 if (libFunction.name === id) {
-                    functionProperties =  libFunction;
+                    functionProperties = libFunction;
```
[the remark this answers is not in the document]
dan-niles: Intentional fallback. The `if (functionProperties)` guard is pre-existing — a 200 response that doesn't contain the function already falls through to the bare `moduleName:id`, which is also what every non-function element inserts.

Before this PR, `getLibraryData` returned `null` from inside `res.on('end')` on a non-200, so the promise never settled: clicking a function whose docs aren't on Central left the spinner up forever and inserted nothing. Terminating with the bare name is the fix; returning early would restore the silent no-op, this time without an error surface in this component.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/NewTool.tsx
```diff
@@ -184,14 +199,16 @@ export function NewTool(props: NewToolProps): JSX.Element {
...
+                console.error("Tool generation failed", { error: toolResponse?.error });
+                setError("The tool could not be generated. Please try again.");
                 return;
             }
 
             const updatedAgentNode = await addToolToAgentNode(agentNode, data.toolName);
             if (!updatedAgentNode) {
                 console.error("Failed to add tool to agent node");
+                setError("The tool was created but could not be attached to the agent.");
                 return;
```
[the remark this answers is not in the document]
dan-niles: Fixed in 804195b — `setError("")` now runs before `setSavingForm(true)`, so a retry starts clean.

packages/ballerina-side-panel/src/components/Form/index.tsx
```diff
@@ -171,7 +251,6 @@ namespace S {
         z-index: 10;
         width: 100%;
         padding: 16px 0 0;
-        background: var(--vscode-editor-background);
```
gigara, line 174: This container is `position: sticky; bottom: 0; z-index: 10` precisely because the form body scrolls underneath it. Dropping the background makes it transparent, so form content will scroll visibly through the Save button.

The blast radius is much wider than the new tool flow: `ArtifactForm` passes `footerActionButton` unconditionally, and `EditConnectionPopup`, `APIConnectionPopup`, `CreateAgentForm`, `CreateMemoryForm`, `AddAgentPopupContent`, `KnowledgeBaseForm`, `ServiceConfigureForm` and `FunctionConfigureForm` all opt in too.

If this was removed to fix a seam against the new group cards, a targeted swap (e.g. `var(--vscode-sideBar-background)`) would be safer than removing the paint entirely.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/NodeReferenceSelectEditor/NodeReferenceSelectEditor.tsx
```diff
@@ -146,9 +164,13 @@ export const NodeReferenceSelectEditor: React.FC<NodeReferenceSelectEditorProps>
...
                     };
                 });
-            setSelectItems(ensureValueInItems(
+            const resolved = ensureValueInItems(
                 applyNodeReferenceFilter([...staticItems, ...items]), value, searchNodesKind
-            ));
+            );
+            setSelectItems(resolved);
+            if (!value && resolved.length > 0) {
+                onChange(resolved[0].value, resolved[0].value.length);
```
gigara, line 173: This auto-select runs for **every** `searchNodesKind` field, not just the new `NEW_CONNECTION` picker — memory stores, model providers, agent references all go through this editor. Two consequences:

1. It doesn't check `field.optional`, unlike `showEmptyPrompt` a few lines below. An optional reference field will silently acquire a value the user never chose, and that value gets written to source on save.
2. It overlaps the existing `useEffect` just below that already auto-selects from `staticItems` when `!value`.

Could this be gated on `!field.optional` at minimum, or on an explicit opt-in prop so it only applies to the connection picker this PR introduces?

packages/ballerina-visualizer/src/views/BI/Connection/ConnectorBrowser/ConnectorActionList.tsx
```diff
@@ -0,0 +1,288 @@
...
+            return (
+                label.includes(query) ||
+                description.includes(query) ||
+                symbol.includes(query) ||
+                resourcePath.includes(query)
+            );
+        });
+    }, [actions, searchText]);
+
+    const searchPlaceholder = actions.length > 0 ? `Search ${actions.length} actions` : "Search actions";
```
gigara: Reads "Search 1 actions" for any connector with a single action.

`connectionRowsOf` in `ConnectorList.tsx` already gets this right — worth reusing the same shape:

```ts
const searchPlaceholder = actions.length > 0
    ? `Search ${actions.length} action${actions.length === 1 ? "" : "s"}`
    : "Search actions";
```

Related, one line down in the empty state: when `actions` is genuinely empty (the existing-connection path with no actions), it renders `No actions match "".` with empty quotes. Worth guarding on `searchText.trim()`.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AddTool.tsx
```diff
@@ -129,11 +142,11 @@ export function AddTool(props: AddToolProps): JSX.Element {
...
                             <Icon name="bi-connection" />
                         </OptionIcon>
-                        <OptionTitle>Use Connection</OptionTitle>
+                        <OptionTitle>{TOOL_OPTION_LABELS.CONNECTION}</OptionTitle>
                     </OptionHeader>
                     <OptionDescription>
-                        Turn an existing connection (HTTP client, database, message broker) into an agent tool.
-                        Your agent will be able to make requests and interact with these services.
+                        Call an action on an HTTP client, database, or message broker. Pick the action
+                        first, then the connection it runs on.
```
gigara: "Pick the action first, then the connection it runs on" describes the reverse of what the new flow actually does — `ConnectorBrowser` opens on the connector/connection list, you choose a connector, *then* its action, and the connection is confirmed on the tool form afterwards.

This card is the entry point to the feature, so it's the first thing users read about it. Something like:

> Call an action on an HTTP client, database, or message broker. Pick a connector, choose the action you want, then point it at a connection.

While here — `NodeList`'s `listDescription` for the FUNCTION mode and the `ConnectorBrowser` `description` prop both read well; this one just got out of sync with the implemented order.

gigara (commented):
Reviewed the full diff. The structure of the refactor is solid and the unit coverage on `toolForm` / `connectorActions` / `formatMethodName` is genuinely good — the table-driven `resourceToolNameSeed` cases and the five-case `groupHasBlockingIssue` suite are the kind of invariant tests that will catch regressions later.

Four things I'd want addressed before merge, left inline. Two of them (the footer background and the node-reference auto-select) reach well outside the new tool flow, so they're the ones I'd look at first.

A handful of lower-priority items — a request-id guard for `handleSelectConnector`, a cancellation guard in `UseAgentToolForm`'s effect, a duplicated `CreatedBuiltInAgent` declaration in `AIChatAgent/utils.ts`, and some copy polish — I'll leave out of the blocking set; happy to write them up separately if useful.

packages/ballerina-side-panel/src/components/Form/index.tsx
dan-niles: Thanks for flagging this, and for mapping out the consumers.

I looked into it, and I don't think the paint was doing any work here. When `footerActionButton` is set, `S.Container` is `overflow: hidden` and the scrolling happens inside `S.ScrollableContent` (`flex: 1; overflow-y: auto`). The footer is a **sibling** of that scroller, not inside it, so the body is clipped at the scroller's bottom edge and never reaches the button's strip. `position: sticky` is effectively inert in that arrangement, since its nearest scrolling ancestor (`Container`) doesn't scroll.

`nestedForm` would make `Container` scrollable, but it's never combined with `footerActionButton` — `KnowledgeBaseForm` explicitly does `nestedForm={!footerActionButton}`.

Verified by scrolling the tool form, Edit Connection, and Create Agent in both light and dark themes — content clips cleanly above the divider, nothing shows through (screenshots below).

So I'd leave it as is, but happy to put the background back if you spot a case where it does show through.

image

image

image

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/NodeReferenceSelectEditor/NodeReferenceSelectEditor.tsx
dan-niles: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/877/commits/2ceb5117ef — gated on `!field.optional` so an optional reference is never auto-filled, and on `staticItems.length === 0` so a fetched node cannot land on top of the static default the mount-time preselect picks.

packages/ballerina-visualizer/src/views/BI/Connection/ConnectorBrowser/ConnectorActionList.tsx
dan-niles: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/877/commits/a56fd61a41 — pluralized the count, and the empty state now reads "No actions available." when nothing is filtered.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AddTool.tsx
dan-niles: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/877/commits/a56fd61a41 — reworded to follow the order the wizard walks. Thanks for the suggestion.

gigara (approved):

MohamedSabthar (approved):

PR #882, opened 2026-08-18
Treat a browsed existing project as the project, not a parent

gigara:
Fixes wso2/product-integrator#2115

## Problem

Browse in the project-creation flow read every pick as the **parent location** and appended a name-derived folder. Selecting an existing project `projectexp` therefore targeted `projectexp/default` — the reporter's "asking for project name again or creating a default project inside it".

The UI and the submit also disagreed. `validateProjectPath` returns `existingWorkspace: true` here (via its parent-is-a-workspace check), so the status strip read *"Existing project · your new integration will be added here"* — but the submit sent `projectPath: projectexp/default`, and `resolveExistingWorkspaceTarget` matched neither that path nor its child. The package was scaffolded at `projectexp/default/untitled` and never registered in `projectexp`'s workspace toml, leaving an orphan.

## Change

A pick that is **itself** a project is held as the project: its own folder becomes the path's last segment, and its `Ballerina.toml` title fills the Project name field. That is exactly what typing the same path into the field already did, so Browse and typing now agree. The strip becomes truthful, `resolveExistingWorkspaceTarget` case (a) matches, and the integration lands in the project it was pointed at.

Picks are also **reversible**, which the first cut of this got wrong. Adopting a project's title and pinning its folder is now remembered, so browsing on to an empty folder restores the placeholder name and re-derives the folder instead of carrying `projectexp` over in both fields. The two undos are independent, because the user can claim back one without the other — typing a name recouples the folder, editing the path's last segment claims only the folder. The default-project seed records the same memory, since it adopts a name the same way.

`ProjectCreationView` gets the snap without the name adoption: it only ever creates new projects, so an existing project at the pick is a conflict for the path validation to report, not a name to take on.

The decision lives in a pure reducer, `applyBrowsePick`, so the multi-step sequences are testable.

### Behaviour

| Sequence | Path | Project name |
|---|---|---|
| pick `projectexp` | `…/projectexp` | `projectexp` |
| then pick an empty folder | `…/empty/default` | `Default` |
| typed "My Thing" → project → empty | `…/empty/my_thing` | `My Thing` |
| project → other project → empty | `…/empty/default` | `Default` (not `projectexp`) |
| project → type a name → empty | `…/empty/my_thing` | `My Thing` |

## Notes on the design discussion

The issue carries two proposals — auto-populate the name (reporter) and a *"Existing project — Open instead"* warning with an Open action (@rdulmina). This implements the first, and removes the nesting the second was aimed at.

An **Open** button is deliberately not included. It would need to be additive rather than a disabled form: the chooser passes `createAsWorkspace: false` on purpose so that adding into an existing project is a *supported* flow, and disabling on detection would break it — and the Location field can't be disabled without trapping the user. It also needs a new `openProject` RPC plus an `isAlreadyOpenFolder` guard, since the default location is the open workspace root, so the detected project is often already open and `openInVSCode` degrades to a window reload. Happy to add it as an action inside the strip in a follow-up.

## Testing

- 19 tests in `utils.test.ts`, threading the reducer's output back into the next pick's input so the sequences above are what's asserted. Covers trailing separators, root-level and Windows paths, title ≠ folder, untitled projects, and an uninspectable folder.
- Full `ballerina-visualizer` suite: 19 suites / 247 tests pass.
- `tsc --noEmit -p tsconfig.spec.json` diffed against a stashed baseline: identical, zero added errors.

Not yet exercised in a real VS Code window — the reducer is covered, the wiring into `dirCoupling` and the refs is not. Worth walking the video's exact steps before merge.

## Summary

- Fix project creation when Browse selects an existing Ballerina project.
- Use the selected project folder as the target project.
- Populate the Project name field from `Ballerina.toml`.
- Restore default name and folder behavior when users browse to another location.
- Support independent name and folder edits.
- Apply consistent path handling in `ProjectCreationView` while preserving existing-project validation.
- Add the pure `applyBrowsePick` reducer and comprehensive tests for browsing, path normalization, project titles, and state transitions.


packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/ProjectCreationView.tsx
```diff
@@ -108,7 +116,7 @@ export function ProjectCreationView({ onBack, ballerinaUnavailable }: { onBack?:
                 // render ever pairs a real path with the un-indexed default name (which is
                 // what triggers the realtime "already exists" diagnostic flash).
                 // Don't clobber a name the user typed while the seed was resolving.
-                if (!projectNameTouchedRef.current) {
+                if (!projectNameTouchedRef.current && !pathPickedRef.current) {
                     dirCoupling.setDirectoryName(dirName);
                 }
                 setDefaultPath(dp);
```
samithkavishke, line 136: **The seed can still overwrite a browsed path.**

The `pathPickedRef` guard added just above covers `setDirectoryName(dirName)` only — `setDefaultPath(dp)` and `setEditablePath(dp)` run unconditionally. If a Browse resolves while this seed is still awaiting `getWorkspaceRoot` / `getDefaultCreationPath` / `getProjectComponentNames`, the seed replaces the picked base with the default location while the pinned project folder survives as the segment, producing `<defaultPath>/<other project's folder>` — the same class of wrong target this PR is fixing.

Suggest putting the whole tail of the seed under the guard:

```ts
if (pathPickedRef.current) return;
if (!projectNameTouchedRef.current) {
    dirCoupling.setDirectoryName(dirName);
}
setDefaultPath(dp);
setEditablePath(dp);
```

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
```diff
@@ -280,10 +290,15 @@ export function CreateProjectChooser({
                 // Browse does. The folder stays "default"; only the display name changes.
                 const defaultProjectPath = joinPath(dp, directoryName);
                 const info = await wsClient.getExistingProjectInfo({ projectPath: defaultProjectPath });
-                if (!mounted) return;
+                if (!mounted || pathPickedRef.current) return;
```
samithkavishke: **Same asymmetric guard as in `ProjectCreationView`.**

This `pathPickedRef.current` check is the right idea, but it lands after the state has already been written: `defaultPathInitialized.current = true`, `setDefaultPath(dp)` and `setEditablePath(dp)` (lines 284-286) all run before it. A pick that completes during the first two RPCs is therefore silently reset to the default path.

Checking the ref before any state write — rather than partway through the effect — fixes both this and the `ProjectCreationView` case in the same shape.

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
```diff
@@ -427,30 +442,77 @@ export function CreateProjectChooser({
...
+            setEditablePath(next.base);
+            if (next.folder.action === "pin") {
+                // Hold the project's OWN folder rather than re-deriving it from the name:
+                // a project's title and its directory need not match.
+                dirCoupling.setDirectoryName(next.folder.directoryName);
+                dirCoupling.setDirTouched(true);
+            } else if (next.folder.action === "recouple") {
+                dirCoupling.handleDisplayNameChange(next.folder.displayName, { recouple: true });
+            }
+            setProjectName(next.projectName);
```
samithkavishke: **This writes back a stale closure value and can clobber typing.**

`applyBrowsePick` is fed `projectName` from the render closure captured at click time, but it now runs *after* the `getExistingProjectInfo` round-trip, during which the webview stays interactive. If the user types in the Project name field while the folder is being inspected, `handleNameChange` sets `projectNameTouchedRef = true` and clears both memories, `applyBrowsePick` takes the early `keep` branch and returns `current.projectName` (the pre-typing value), and this line reverts what they just typed.

The asymmetry that makes it easy to miss: `projectNameTouchedRef` is read fresh (a ref) while `projectName` is stale (state). Mirroring the name in a ref alongside the touched flag would make both consistent across the await.

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
```diff
@@ -427,30 +442,77 @@ export function CreateProjectChooser({
...
     };
 
     const handlePathChange = (value: string) => {
         const { base, name } = splitPath(value);
         setPathTouched(true);
         setEditablePath(base);
         dirCoupling.handleDirectoryNameEdit(name, autoDirectoryName);
+        // The segment is hand-edited now — a later Browse must leave it alone rather than
+        // recoupling it. Any adopted NAME is still the pick's, so that memory stands.
+        folderPinnedByPickRef.current = false;
```
samithkavishke: **The pin is released on *any* path edit, not just an edit to the segment.**

The comment says "The segment is hand-edited now", but `folderPinnedByPickRef` is cleared even when `splitPath(value).name` is unchanged and only the parent portion was retyped.

Repro: Browse onto project `…/payments-svc` (name adopts "Payments", segment pinned) → type one character in the parent portion of the path field → Browse to a plain folder `X`. `applyBrowsePick` now sees `folderPinnedByPick === false` with `displacedName` still set, so it returns `folder: keep` — the target becomes `X/payments-svc` with the name reverted to "Default", carrying the old project's folder to a new location.

Suggest clearing the ref only when the extracted `name` actually differs from the pinned `directoryName`. (Same for `ProjectCreationView.tsx:198`.)

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
```diff
@@ -427,30 +442,77 @@ export function CreateProjectChooser({
...
     const handlePathSelection = async () => {
         try {
             const result = await wsClient.selectFileOrDirPath({ startPath: editablePath || defaultPath });
             if (!result.path) return;
-            setEditablePath(result.path);
             setPathTouched(true);
-            // Pin the name: the one-shot default-project lookup can still be in flight,
-            // and its result no longer describes the location just picked.
-            projectNameTouchedRef.current = true;
+            // The one-shot default-project lookup can still be in flight, and its result
```
samithkavishke: **Minor: the picked path is no longer reflected until an extra RPC resolves.**

`setEditablePath(result.path)` used to be synchronous with the dialog closing; the field now updates only once `getExistingProjectInfo` returns. The `.catch` covers rejection but not a slow or hung RPC, so a stalled extension host leaves the user looking at the old path with no sign the pick registered. Worth considering an optimistic `setEditablePath(result.path)` before the await, or a pending indicator. (Same at `ProjectCreationView.tsx:214`.)

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/ProjectCreationView.tsx
```diff
@@ -179,14 +193,52 @@ export function ProjectCreationView({ onBack, ballerinaUnavailable }: { onBack?:
...
     const handlePathSelection = async () => {
         try {
             const result = await wsClient.selectFileOrDirPath({ startPath: editablePath || defaultPath });
             if (!result.path) return;
             setPathTouched(true);
-            setEditablePath(result.path);
+            pathPickedRef.current = true;
+
+            // Best effort — an unreadable folder simply keeps the parent-location reading,
+            // which is what every pick did before.
```
samithkavishke: **Question on scoping: only `[workspace]` folders are treated as projects.**

`getExistingProjectInfo` (`packages/ballerina-extension/src/utils/bi.ts:231`) returns `isProject: false` for a `Ballerina.toml` without a `[workspace]` table. Browsing onto a plain Ballerina *package* directory therefore keeps the parent-location reading and targets `<package>/default` — a project scaffolded inside a package, which looks as unsupported as the case this PR fixes.

Was that deliberate scoping, or should the pick be treated as a project root whenever a `Ballerina.toml` is present?

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/utils.ts
```diff
@@ -167,6 +167,135 @@ export const splitPath = (fullPath: string): { base: string; name: string } => {
...
+): ResolvedBrowsePick => {
+    if (!existingProject?.isProject) {
+        return { base: pickedPath };
+    }
+    // A trailing separator would split into an empty segment and collapse the target onto
+    // the parent — drop it so the project's own folder survives as the last segment.
+    const trimmed = pickedPath.replace(/[/\\]+$/, '');
+    const { base, name } = splitPath(trimmed);
+    if (!base || !name) {
+        // Nothing to split (a bare or root-level pick): keep it as the location and leave
```
samithkavishke: **The "nothing to split" fallback re-introduces the nesting this function documents as impossible.**

When `splitPath` yields no base, `resolveBrowsePick` returns `{ base: pickedPath }` with no `directoryName`, so `applyBrowsePick` treats a *project* pick as a plain parent location and the form targets `<project>/<name-derived>` — breaking the stated invariant that `joinPath(base, directoryName)` is the pick itself. The test "falls back to the parent-location reading when there is nothing to split" locks the behavior in.

Only reachable if the dialog ever returns a relative path, so low severity — but since the code and tests explicitly model the case, they should model it correctly. `{ base: '', directoryName: trimmed }`, or an explicit no-op, would be safer.

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/utils.ts
```diff
@@ -167,6 +167,135 @@ export const splitPath = (fullPath: string): { base: string; name: string } => {
...
+export const applyBrowsePick = (
+    pickedPath: string,
+    existingProject: { isProject?: boolean; name?: string } | null | undefined,
+    current: BrowsePickState,
+    options: { adoptProjectName?: boolean } = {}
+): AppliedBrowsePick => {
+    const pick = resolveBrowsePick(pickedPath, existingProject);
+
+    if (pick.directoryName) {
+        const adoptedName = options.adoptProjectName ? pick.projectName : undefined;
```
samithkavishke: **A titled adoption isn't cleared when a later project pick has no title.**

In this `pin` branch `adoptedName` is `undefined` when the picked project reports no `name`, so `projectName` keeps the *previous* project's adopted title while the path moves to the new project — the field then labels a project that isn't the target.

Defensive-only today (the backend's `getExistingProjectInfo` falls back to `path.basename(dir)`, so `name` is effectively always set), but since the untitled case is modelled here and in the tests, falling back to `pick.directoryName` would avoid leaving a stale adoption.

samithkavishke (approved):
LGTM

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/ProjectCreationView.tsx
gigara: Fixed in dbe1940. You are right that guarding only `setDirectoryName` left the two path writes exposed — a pick landing during the earlier round-trips got reset to the default location while the pinned folder survived as the segment.

Rather than guarding the tail, every resumption point now re-checks, via a `superseded()` helper that folds in the `mounted` check:

```ts
const superseded = () => !mounted || browsePick.pathPickedRef.current;
const { path: workspacePath } = await wsClient.getWorkspaceRoot();
if (superseded()) return;
const dp = workspacePath || (await wsClient.getDefaultCreationPath()).path;
if (superseded()) return;
```

Same shape in both forms.

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
gigara: Fixed in dbe1940, same shape in both views — the check now runs before every state write rather than partway through, so a pick that completes during the first two RPCs is no longer reset.

The ref itself moved into the shared `useBrowsePick` hook (see the reply on the duplication point), so the seed reads `browsePick.pathPickedRef` and there is one definition of what "a pick has superseded this" means.

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
gigara: Fixed in dbe1940. The read/write asymmetry was exactly as you describe, and it was the part I would not have found by inspection — thanks.

The hook now mirrors the name into a ref alongside the touched flag, so both sides of the await are consistent:

```ts
const projectNameRef = useRef(projectName);
projectNameRef.current = projectName;
```

`applyBrowsePick` is fed `projectNameRef.current`, so a name typed during the inspection is what gets read back.

Your point also generalises to two Browse clicks racing each other, so the hook now carries a monotonic request token and drops a reading that a later pick has superseded — which is CodeRabbit's suggestion on the same lines.

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
gigara: Fixed in dbe1940 — I reproduced your `X/payments-svc` case before changing anything, and it behaved exactly as you predicted. The comment claiming "the segment is hand-edited now" was straightforwardly wrong when only the parent portion changed.

The hook now remembers the segment a pick pinned and releases the hold only when an edit actually changes it:

```ts
const releaseFolderIfSegmentEdited = useCallback((editedDirectoryName: string) => {
    const pinned = pinnedDirectoryNameRef.current;
    if (pinned === null || editedDirectoryName === pinned) return;
    folderPinnedByPickRef.current = false;
    pinnedDirectoryNameRef.current = null;
}, []);
```

Both views call it, so `ProjectCreationView` is covered by the same change.

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
gigara: Not taken, and I want to be explicit about why rather than leave it looking overlooked.

The field renders `<base>/<segment>`, so writing the pick before the inspection composes it with the **previous** segment. Browsing onto a project would flash `/Users/me/projectexp/default` — precisely the wrong target this PR exists to remove — for as long as the stat takes. I tried it, and it is a visible paint, not a sub-frame blip: the write sits after an await, so React flushes it.

On the risk it guards against: a hang in `getExistingProjectInfo` means the same transport that had just returned the dialog result has wedged in the interim, and the call is a local `fs` stat with no network. Trading a certain flash of a wrong path for that seemed the wrong way round.

The reasoning is in a comment at the call site so the omission is not mistaken for an oversight later. Happy to revisit if you would rather have the optimistic write with a pending state on the segment.

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/utils.ts
gigara: Fixed in dbe1940, though not with the suggested shape — `{ base: '', directoryName: trimmed }` does not work here. `joinPath` returns `''` when `base` is falsy, so the composed target would be empty rather than the pick. My first attempt used it and the invariant test caught it.

What works is the mirror image, since `joinPath` returns `base` unchanged when `name` is empty:

```ts
return { base: trimmed, directoryName: '', projectName: existingProject.name || undefined };
```

That required changing the discriminator in `applyBrowsePick` from truthiness to `pick.directoryName !== undefined`, because an empty segment still means "the pick IS the target".

Also fixed the untrimmed `base` you spotted, and split the root-only case out — `/` genuinely has no folder of its own, so it stays a plain location. The test now asserts the pick composes back to itself.

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/utils.ts
gigara: Fixed in dbe1940 — `adoptedName` falls back to `pick.directoryName` when a project reports no title, so the field can never keep labelling a project that is no longer the target.

Agreed it is defensive today given the `path.basename(dir)` fallback in `getExistingProjectInfo`, but as you say the untitled case is modelled here and in the tests, so it should be modelled correctly. Added a test that picks a titled project and then an untitled one, asserting the name follows the second.

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/useBrowsePick.ts
```diff
@@ -0,0 +1,209 @@
...
+    adoptProjectName,
+    selectFailedMessage,
+}: UseBrowsePickOptions): BrowsePick {
+    const displacedNameRef = useRef<BrowsePickState["displacedName"]>(null);
+    const folderPinnedByPickRef = useRef(false);
+    /** The segment a pick pinned, so a path edit can tell an edit OF it from an edit AROUND it. */
+    const pinnedDirectoryNameRef = useRef<string | null>(null);
+    const pathPickedRef = useRef(false);
+    /** Latest pick wins: an earlier, slower inspection must not write over a later one. */
+    const pickRequestIdRef = useRef(0);
```
NipunaRanasinghe, line 105: These four refs, and the ordering logic in `selectPath` that reads them, have no test coverage.

The 21 tests in `utils.test.ts` all target `applyBrowsePick`, which is pure. Everything the last review round actually fixed lives here instead:

- `pathPickedRef` — the seed-supersede guard both forms call before every write
- `pickRequestIdRef` — dropping a superseded inspection when two Browse clicks interleave
- `pinnedDirectoryNameRef` — distinguishing an edit *of* the pinned segment from an edit *around* it
- `projectNameRef` — reading the name back after the await instead of from the render closure

Each of those was a real bug in the first commit, and none of them can regress into a failing test today. Reading the suite, it looks like the browse flow is covered; it isn't — only the reducer is.

Two things would close this, and the second is the one I'd insist on:

1. Tests for the hook via `renderHook` with a stubbed `wsClient`. Note there is no `renderHook` precedent anywhere in `ballerina-visualizer`, so this would be new convention for the package — reasonable to skip if you'd rather not introduce it in a fix PR.
2. Walking the video's exact steps in a real VS Code window before merge, given the PR notes the wiring into `dirCoupling` and the refs was never exercised there. Worth covering at least: pick project → pick empty folder; pick project → type a name; pick project → retype the parent portion of the path (must keep the pinned segment); and clicking Browse while the initial seed is still resolving.

I verified the extension side separately and the fix is correct — `resolveExistingWorkspaceTarget` matches case (b) and the package lands in the picked project's workspace toml. This is about the stateful layer being the untested part, not about the decision it implements.

NipunaRanasinghe (approved):
Approving with a non-blocking comment

PR #884, opened 2026-08-18
Update the Durable Workflow designer for ballerina/workflow 0.9.0

hasithaa:
## Purpose

Update the Durable Workflow designer for the released `ballerina/workflow` 0.9.0, which moves the human task definition and the Human Review policy into records, adds `stepId` to every step, and moves the management REST API to `ballerina/workflow.management.rest`.

Fixes wso2/product-integrator#2372

## What changed

- **Human task**: the form is derived from the resolved `awaitHumanTask` signature, so it follows the module (`taskInput`, the `HumanTaskDefinition` fields). `stepId` is kept out of the form — hidden, not removed, so a program that names its step keeps that name through an edit. The form stays flat and in its pre-0.9 order (name, roles, input, details, completion type); the record's `resultType` (a duplicate of the Completion Type selector) and `taskInputType` are not offered. Both the pre-0.9.0 argument layout and the 0.9.0 one are read; the 0.9.0 one is written.
- **Human Review**: the retry policy is read and written as a `ReviewTaskDefinition` record with `title`, `description` and `timeout` alongside the roles; the pre-0.9.0 role-string and sentinel forms are still read.
- **Round-trip fixes from review**: the record literal is read at its own nesting level (a role list or a title with a comma survives), an omitted task input becomes the explicit `{}` and a cleared required input is refused rather than silently dropped, and the options overlay reads only the options argument (a payload's `title` is not the task's title). Plain-text fields go through one string-literal encoder.
- **Overview → Workflow**: the checkbox becomes *Enable Workflow Management REST API*. It imports `ballerina/workflow.management.rest as _` into `main.bal` (created when absent), only that import counts as enabled, disabling also removes the unused legacy `workflow.management as _` import, and enabling ICP no longer enables it as a side effect. Port/TLS/CORS remain configuration-editor settings.
- **LS test pin** moved to workflow 0.9.0 (now on central); goldens regenerated for the record-derived forms and `stepId`.

## Tests

- L1 (flow-model-generator-core): `WorkflowUtilLiteralTest`, `HumanReviewLiteralTest`, `HumanTaskFormTest` — record reader, literal encoder, review record, stepId/input/overlay rules, form order.
- L3 (flow-model-generator-ls-extension): human task and retry goldens at the 0.9.0 pin; `AddWorkflowManagementTest`, `WorkflowManagementEnablerTests` reworked for the REST import; new `DisableWorkflowManagementTest`; the ICP default-enable tests removed with the behaviour.


packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -2157,45 +2189,55 @@ private void populateBuiltinActivityProperties(RemoteMethodCallActionNode callNo
...
-                retryBackoff = fields.getOrDefault(ActivityCallBuilder.RETRY_BACKOFF_KEY, "");
-                maxRetryDelay = fields.getOrDefault(ActivityCallBuilder.MAX_RETRY_DELAY_KEY, "");
+                if (fields.containsKey(USER_ROLES_FIELD)) {
+                    dropdownValue = ActivityCallBuilder.MANUAL_RETRY_VALUE;
+                    review = new ActivityCallBuilder.ReviewFormValues(
+                            fields.getOrDefault(USER_ROLES_FIELD, ""),
+                            unquoted(fields.get("taskName")),
+                            unquoted(fields.get("title")),
+                            unquoted(fields.get("description")),
+                            fields.getOrDefault("timeout", ""));
```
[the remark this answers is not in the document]
hasithaa: Fixed in c6702f073b. The record is now read at its own nesting level only (`WorkflowUtil.parseRecordLiteral`): a comma or colon inside a nested list or record, a string literal or a template stays with its value, and an escaped quote does not end the string. Covered by `WorkflowUtilLiteralTest` — role list, nested duration record, comma/colon/escaped quote inside a string, template, quoted key, and the AutoRetry record reading as before.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
```diff
@@ -364,17 +370,50 @@ public static ConnectionSelectorData resolveUserActivityConnectionSelector(Class
...
+    public record ReviewFormValues(String userRoles, String title, String description, String timeout) {
+
+        /** A review with nothing declared — the form's starting state. */
+        public static ReviewFormValues empty() {
+            return new ReviewFormValues("", "", "", "");
+        }
+
+        /** Only the roles were read, as the pre-record form could express. */
+        public static ReviewFormValues ofRoles(String userRoles) {
+            return new ReviewFormValues(userRoles == null ? "" : userRoles, "", "", "");
```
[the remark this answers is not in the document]
hasithaa: Added `HumanReviewLiteralTest` (L1, TestNG) against `humanReviewRecordLiteral` in c6702f073b: empty form (`{userRoles: []}`), roles only (bare role quoted, list passed through), all fields populated in record order, escaping (quote, backslash, line break — via one string-literal encoder that now escapes more than the quote), source-shaped values passed through (string literal, template), and unrelated properties ignored, including a stray `retryTaskName`. On that last point: no `retryTaskName`/`RETRY_TASK_NAME` exists anywhere in the LS at this branch, its merge base, or any fixture, so there was no exclusion to preserve.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/HumanTaskBuilder.java
```diff
@@ -249,10 +261,14 @@ public static void addFallbackHumanTaskParameters(NodeBuilder nodeBuilder, Map<S
...
                     .stepOut()
                 .type().fieldType(Property.ValueType.EXPRESSION).ballerinaType("map<json>").selected(true).stepOut()
-                .codedata().kind(ParameterData.Kind.DEFAULTABLE.name()).originalName(PAYLOAD_KEY).stepOut()
-                .value(values.get(PAYLOAD_KEY))
+                .codedata().kind(ParameterData.Kind.REQUIRED.name()).originalName(PAYLOAD_KEY).stepOut()
+                // The payload is a required argument of awaitHumanTask now, not a field of the
+                // definition record: a task with nothing to show says so with `{}` rather than
+                // by omission. The form defaults to `{}` for the same reason — leaving the
+                // field blank would emit a call that does not compile.
+                .value(values.getOrDefault(PAYLOAD_KEY, EMPTY_PAYLOAD))
```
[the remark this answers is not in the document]
hasithaa: Fixed in c6702f073b. The fallback form treats a mapped null as the omitted input it is and fills `{}`, and `toSource` now refuses to emit when an input the module declares REQUIRED has been cleared in the form (a defaultable input, as older releases declare, is left alone). `HumanTaskFormTest` covers the mapped null, the absent key and an explicit value. The same commit also addresses the two outside-diff findings: `stepId` is hidden and non-editable instead of removed, so a save re-emits it; and the options overlay reads only the options argument (skips the required positional arguments, fills only INCLUDED_FIELD properties), so a payload's `title` no longer becomes the task's title.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -2253,147 +2326,84 @@ private void addNormalizedRetryPolicyProperties(String rawValue) {
...
-                inTemplate = true;
-            } else if (c == '{' || c == '[' || c == '(') {
-                depth++;
-            } else if (c == '}' || c == ']' || c == ')') {
-                depth--;
-            } else if (c == terminator && depth == 0) {
-                return i;
-            }
+    /** A string literal as the form shows it — the quotes belong to the source, not the value. */
+    private static String unquoted(String literal) {
```
NipunaRanasinghe, line 2459: **Medium — CONFIRMED. The review's text fields are encoded on write but not decoded on read, so escapes accumulate.**

`unquoted` strips the surrounding quotes and nothing else, while the write side (`ActivityCallBuilder.addQuotedRecordField` → `WorkflowUtil.stringLiteral`) escapes `\`, `"`, `\n`, `\t` and `\r`. A review titled `He said "hi"` is written as `title: "He said \"hi\""`, read back into the form as `He said \"hi\"` — already wrong in the UI — and re-emitted as `title: "He said \\\"hi\\\""`, gaining a backslash pair on every save.

`stringLiteral` needs an inverse applied here and in the `description` read on the next line. `HumanReviewLiteralTest` covers the encoder but there is no read-then-write test, which is why this got through; a round-trip assertion over a title with a quote and a newline would pin it.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -2066,6 +2063,78 @@ private void populateFallbackHumanTaskProperties(RemoteMethodCallActionNode call
...
+                ? namedArgs.get(HumanTaskBuilder.TASK_INPUT_KEY) : namedArgs.get(HumanTaskBuilder.PAYLOAD_KEY);
+        String userRolesValue = namedArgs.getOrDefault(HumanTaskBuilder.USER_ROLES_KEY, "");
+        String firstArg = args.size() > 1 && args.get(1) instanceof PositionalArgumentNode posArg1
+                ? posArg1.expression().toSourceCode().strip() : null;
+        if (firstArg != null) {
+            if (firstArg.startsWith("{")) {
+                // A record literal in the roles' position: 0.9.0's task input.
+                taskInputValue = firstArg;
+            } else if (userRolesValue.isEmpty()) {
+                userRolesValue = firstArg;
```
NipunaRanasinghe, line 2174: **Medium — CONFIRMED. In the fallback read, a task input that is not a record literal is dropped when the roles are named, and the form overwrites it with `{}`.**

For `ctx->awaitHumanTask("approve", requestData, userRoles = "manager")`, `userRolesValue` is already `"manager"`, so `firstArg` (`requestData`) matches neither branch and is discarded; positional index 2 is a named argument, so the later positional check does not pick it up either. `taskInputValue` stays null, and `addFallbackHumanTaskParameters` fills the REQUIRED task input with `EMPTY_TASK_INPUT`, so the next save emits `taskInput = {}` and the user's argument is gone.

A named `userRoles` settles the layout as 0.9.0 on its own, so positional 1 is the task input whatever its shape. Testing `!userRolesValue.isEmpty()` before the `startsWith("{")` shape guess would close it. `HumanTaskFallbackArgumentsTest` only ever passes a record literal or a role string in that position.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/HumanTaskBuilder.java
```diff
@@ -312,9 +349,37 @@ public static void addFallbackHumanTaskParameters(NodeBuilder nodeBuilder, Map<S
...
+        // in this form: resultType is what the Completion Type selector already edits, and
+        // taskInputType is derived from the input. Both are hidden the same way as stepId, so a
+        // program that states them keeps them through an edit.
+        hide(properties, RESULT_TYPE_KEY);
+        hide(properties, TASK_INPUT_TYPE_KEY);
+        // The record's fields arrive after the positional parameters and the inferred type, which
+        // puts the required roles below the type selector. The form is what it was before the
+        // record: name, roles, input, then the optional details, then the completion type. Keys
+        // the module does not declare are simply absent; anything else keeps its place after.
+        reorder(properties, TASK_NAME_KEY, USER_ROLES_KEY, TASK_INPUT_KEY, PAYLOAD_KEY,
```
NipunaRanasinghe, line 380: **Low — CONFIRMED. `reorder` handles `$description`, but the `relabel` below it does not, so that field keeps the module's own label.**

`description` is in `Property.RESERVED_PROPERTY_KEYS`, so `FlowNodeUtil.getPropertyKey` gives the resolved-signature property the key `$description` — which is exactly why it appears in this list. `relabel(properties, DESCRIPTION_KEY, ...)` a few lines down looks up `description` and finds nothing, so in the real 0.9.0 form the Description field falls back to `paramData.label()` while its neighbours Title and Timeout get this file's wording.

Since `relabelHumanTaskFormProperties` is only reached from the resolved path, the `DESCRIPTION_KEY` relabel never fires at all. `HumanTaskFormTest` asserts the ordering of `$description` but not its label.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/main/java/io/ballerina/flowmodelgenerator/extension/WorkflowManagementService.java
```diff
@@ -193,7 +151,7 @@ public CompletableFuture<CommonSourceResponse> disableWorkflowManagement(CreateF
                     Document document = defaultModule.document(documentId);
                     ModulePartNode root = document.syntaxTree().rootNode();
                     for (ImportDeclarationNode importNode : root.imports()) {
-                        if (validOrg(importNode) && validModuleName(importNode)) {
+                        if (validOrg(importNode) && (validModuleName(importNode) || unusedLegacyImport(importNode))) {
```
NipunaRanasinghe: **Low — CONFIRMED. Disabling removes a `workflow.management.rest` import the program uses by name, leaving unresolved references.**

`unusedLegacyImport` deliberately requires the `_` prefix so that a legacy import a program actually uses survives. `validModuleName` applies no such guard, and the comment above it gives the reason it should: the import counts as enabled either way because "a program may well refer to its configurables by name". So `import ballerina/workflow.management.rest;` (or `as rest`) is deleted here, and every reference through that prefix stops compiling.

The old code had the same shape, so this is carried forward rather than introduced — but the PR reasons about the used form explicitly and guards the legacy module against it, which makes the asymmetry worth closing while it is in view. `DisableWorkflowManagementTest` only exercises the `as _` form.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -792,7 +793,166 @@ public static String quoteIfPlain(String value) {
...
+                continue;
+            }
+            out.append(c);
+        }
+        return out.toString();
+    }
+
+    // The comma-separated pieces of a literal's interior, splitting only where a comma is not
+    // inside brackets, braces, parentheses, a string or a template.
+    private static List<String> splitTopLevel(String text) {
```
dulajdilshan, line 999: **[Medium] The same quote/depth scanner is written 3x in this file.**

`splitTopLevel:897` and `topLevelIndexOf:932` share a byte-identical escape block:

```java
if (quote != 0) {
    if (c == '\\' && quote == '"') { i++; }
    else if (c == quote)             { quote = 0; }
    continue;
}
```

and identical depth tracking — `splitTopLevel` writes it as a `switch`, `topLevelIndexOf` as an `if/else` chain. They differ only in the action at the target char: split vs. return index. `stripLineComments:864` is a third pass over the same states.

So the escaping rules for Ballerina literals now live in 3 places. Fix a quoting bug in one and `parseRecordLiteral` can still split correctly while finding the wrong colon.

One private scanner yielding top-level positions, consumed by both, removes two copies.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -792,7 +793,166 @@ public static String quoteIfPlain(String value) {
...
+
+    /**
+     * The Ballerina string literal for a plain text value: quoted, with every character the
+     * literal syntax would otherwise interpret escaped — the backslash and quote, and the line
+     * break, tab and carriage return, which a bare {@code "..."} literal cannot carry.
+     *
+     * @param text the value as the form holds it
+     * @return the literal source, quotes included
+     */
+    public static String stringLiteral(String text) {
```
dulajdilshan, line 807: **[Medium] This is the correct encoder — but two older copies survive and still emit source that won't compile.**

`stringLiteral` escapes `\` `"` `\n` `\t` `\r`. Two hand-rolled copies escape only the first two, both ending:

```java
return "\"" + trimmed.replace("\\", "\\\\").replace("\"", "\\\"") + "\"";
```

- `SendDataBuilder.toStringLiteral:201`
- `DurableAgentUpdateBuilder.toStringLiteral:205`

A value containing a newline through either path yields a literal with a raw line break inside it — the exact defect this method was added to fix. `ChildWorkflowSendDataBuilder:171` calls the `SendDataBuilder` copy, so it inherits the gap.

3 encoders, 1 correct. The PR description's "plain-text fields go through one string-literal encoder" holds for the human-task/review paths but not these two.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
```diff
@@ -706,10 +745,22 @@ public static void addRetryPolicyFormProperties(NodeBuilder nodeBuilder, String
...
         // own default.
         dynamicFields.put(AUTO_RETRY_VALUE, autoRetryFields);
+        // Human Review renders the ReviewTaskDefinition record — the same shape a human task is
+        // declared with. Only the roles are required; the rest default to wording derived
+        // from the activity being reviewed, which is why each says so in its description.
         Map<String, Property> manualRetryFields = new LinkedHashMap<>();
         manualRetryFields.put(RETRY_USER_ROLES_KEY, buildReviewerRolesSubProperty("Reviewer Roles",
                 "Role(s) permitted to decide the human review, e.g. \"manager\" or "
                         + "[\"finance\", \"manager\"]. Leave empty to allow any role."));
+        manualRetryFields.put(RETRY_TITLE_KEY, buildRetrySubProperty("Title",
```
dulajdilshan: **[Medium] Each review field's label + description is written twice, and the two copies already disagree.**

Visible form (`:751-763`) vs hidden storage (`:792-800`) — same 4 fields, 8 inline literals:

| field | `:751-763` | `:792-800` |
|---|---|---|
| Reviewer Roles | "Role(s) permitted to decide the **human** review, e.g. …" | "Role(s) permitted to decide the **retry** review" |
| Title | "Short summary shown in the reviewer's inbox. **Defaults to a phrase naming the activity being reviewed.**" | "Short summary shown in the reviewer's inbox" |
| Description | "Context shown with the decision. **Defaults to a description of the failure and the outcomes available.**" | "Context shown with the decision" |
| Timeout | "How long to wait for a decision, **e.g. {hours: 4}. Empty waits indefinitely.**" | "How long to wait for a decision" |

"human review" vs "retry review" is already a drift, and whichever copy the UI shows depends on whether the field is visible or hidden. `HumanTaskBuilder` keeps its equivalents as constants (`TITLE_LABEL` / `TITLE_DOC`); 4 label/doc constant pairs passed to both call sites would do the same here.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
```diff
@@ -658,17 +664,50 @@ public static ConnectionSelectorData resolveUserActivityConnectionSelector(Class
...
         addRetryPolicyFormProperties(nodeBuilder, retryPolicyValue, maxRetries, retryDelay,
-                retryBackoff, maxRetryDelay, "");
+                retryBackoff, maxRetryDelay, ReviewFormValues.empty());
     }
 
     public static void addRetryPolicyFormProperties(NodeBuilder nodeBuilder, String retryPolicyValue,
                                                     String maxRetries, String retryDelay,
                                                     String retryBackoff, String maxRetryDelay,
                                                     String retryUserRoles) {
+        addRetryPolicyFormProperties(nodeBuilder, retryPolicyValue, maxRetries, retryDelay,
```
dulajdilshan: **[Minor] This overload has no callers.**

Three overloads of `addRetryPolicyFormProperties`:

| overload | callers |
|---|---|
| 6-arg `:691` | `:222`, `:263`, `DurableAgentAddActivityBuilder:279` |
| **7-arg `String retryUserRoles` `:698`** | **none** |
| 7-arg `ReviewFormValues` `:706` | `CodeAnalyzer:2377` |

Its whole body is `ReviewFormValues.ofRoles(retryUserRoles)` forwarded to `:706`. `ofRoles` itself is still needed (`CodeAnalyzer:2427`) — only this wrapper is dead.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/HumanTaskBuilder.java
```diff
@@ -312,9 +350,37 @@ public static void addFallbackHumanTaskParameters(NodeBuilder nodeBuilder, Map<S
...
+                TITLE_KEY, DESCRIPTION_KEY, "$description", TIMEOUT_KEY, DATABINDING_TYPE_KEY);
+
         relabel(properties, TASK_NAME_KEY, TASK_NAME_LABEL, TASK_NAME_DOC);
         relabel(properties, USER_ROLES_KEY, USER_ROLES_LABEL, USER_ROLES_DOC);
-        relabel(properties, PAYLOAD_KEY, PAYLOAD_LABEL, PAYLOAD_DOC);
+        relabel(properties, TASK_INPUT_KEY, TASK_INPUT_LABEL, TASK_INPUT_DOC);
+        // The pinned workflow bala still names this parameter `payload`; a resolved-signature
+        // form therefore carries that key, and the emitted argument keeps the module's own
+        // name, so the label must not pretend otherwise. Dies with the pre-rename pin.
+        relabel(properties, PAYLOAD_KEY, "Payload",
```
dulajdilshan: **[Minor] Label + doc inlined where every neighbour uses a constant, and the doc is `TASK_INPUT_DOC` with one word changed.**

```java
relabel(properties, TASK_INPUT_KEY, TASK_INPUT_LABEL, TASK_INPUT_DOC);
relabel(properties, PAYLOAD_KEY, "Payload",
        "Read-only JSON object shown alongside the form. Required — …payloadType");   // <- inlined
relabel(properties, TITLE_KEY,       TITLE_LABEL,       TITLE_DOC);
relabel(properties, DESCRIPTION_KEY, DESCRIPTION_LABEL, DESCRIPTION_DOC);
```

This PR deleted `PAYLOAD_LABEL`/`PAYLOAD_DOC` and inlined their replacements, so this one call breaks the file's pattern. The doc string is also `TASK_INPUT_DOC:127` verbatim except for the trailing `payloadType` vs `taskInputType` — two sentences that must now be edited together for as long as the pre-rename pin lives.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
rdulmina: Fixed in 65e664b9.

The read now goes through the inverse of the encoder that writes these fields — `WorkflowUtil.stringLiteralText`, next to `stringLiteral` — so the quotes *and* the escapes stay with the source:

```java
private static String unquoted(String literal) {
    return WorkflowUtil.stringLiteralText(literal);
}
```

It decodes `\\`, `\"`, `\n`, `\t`, `\r` and `\u{...}` (that last one because the re-encode can only reproduce the character, not the escape, so leaving it would double its backslash the same way). Anything that is not one string literal — a variable reference, a template, an expression that merely begins and ends with a quote — is returned as written, since the form holds those as source.

New L1 `RetryPolicyRoundTripTest` pins the read-then-write path you identified as the missing test, over a title with a quote and a description with a backslash and a newline:

```java
String source = "{userRoles: \"manager\", title: \"He said \\\"hi\\\"\", "
        + "description: \"path C:\\\\x\\nnext\", timeout: {hours: 4}}";
ActivityCallBuilder.ReviewFormValues form = CodeAnalyzer.normalizeRetryPolicy(source).review();
Assert.assertEquals(form.title(), "He said \"hi\"");
Assert.assertEquals(rewrite(form), source);                                  // one save
Assert.assertEquals(rewrite(CodeAnalyzer.normalizeRetryPolicy(rewrite(form)).review()), source); // and the next
```

`WorkflowUtilLiteralTest` also covers the decode, the source pass-through and idempotence over repeated saves.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
rdulmina: Fixed in 65e664b9 — your reading of the branch was right, and the guard is the one you suggested:

```java
if (firstArg != null) {
    if (!userRolesValue.isEmpty() || firstArg.startsWith("{")) {
        // 0.9.0's task input: named roles settle the layout on their own, so this argument is
        // the input whatever its shape — a variable holding it is not a role list. Without
        // them, a record literal in the roles' position is the input.
        taskInputValue = firstArg;
    } else {
        userRolesValue = firstArg;
    }
}
```

`awaitHumanTask("approve", requestData, userRoles = "manager")` now reads `requestData` into the task input instead of leaving it null for the form to fill with `{}`.

`HumanTaskFallbackArgumentsTest` gained the case, with both a variable and a call in that position:

```java
Map<String, String> values = CodeAnalyzer.fallbackHumanTaskArgumentValues(arguments(
        "ctx->awaitHumanTask(\"approve\", requestData, userRoles = \"manager\")"));
Assert.assertEquals(values.get(HumanTaskBuilder.TASK_INPUT_KEY), "requestData",
        "the input must survive the read, or the next save replaces it with {}");
```

The shape-based branch stays for a call that names neither, which is still ambiguous without a resolved signature.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/HumanTaskBuilder.java
rdulmina: The premise no longer holds on this branch: `relabel` does not look up the plain key. It resolves through `presentKey`, which tries the plain name first and the reserved-escaped one second — `HumanTaskBuilder.java:470`:

```java
private static void relabel(Map<String, Property> properties, String key, String label, String description) {
    String actualKey = presentKey(properties, key);
    ...
}

static String presentKey(Map<String, Property> properties, String key) {
    return properties.containsKey(key) ? key : FlowNodeUtil.getPropertyKey(key);
}
```

That went in with dcef9f1b8c (Sep 7), for exactly the reason you describe — the comment it replaced said the plain-key relabel had left the field showing the compiler-derived doc. So on the resolved 0.9.0 path the Description field does get this file's wording, and `$description` is in the `reorder` list because that is the key the property sits under, not because the relabel misses it.

The test gap was real, though, and is closed in 65e664b9 — `HumanTaskFormTest` now asserts the label and the doc on the escaped key, not just its position:

```java
Assert.assertEquals(properties.get("$description").metadata().label(), "Description",
        "the escaped key is relabelled like every other field");
Assert.assertEquals(properties.get("$description").metadata().description(),
        "Additional context shown alongside the form");
```

The same commit also drops the literal from the reorder list, so the escaping rule is named once:

```suggestion
        reorder(properties, TASK_NAME_KEY, USER_ROLES_KEY, TASK_INPUT_KEY, PAYLOAD_KEY,
                TITLE_KEY, DESCRIPTION_KEY, FlowNodeUtil.getPropertyKey(DESCRIPTION_KEY),
                TIMEOUT_KEY, DATABINDING_TYPE_KEY);
```

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/main/java/io/ballerina/flowmodelgenerator/extension/WorkflowManagementService.java
rdulmina: Fixed in 65e664b9, though with a different remedy than the `_` guard — applying `unusedLegacyImport`'s rule here would make the checkbox unable to disable `import ballerina/workflow.management.rest;`, which the existing `config2` covers: nothing would be removed and the checkbox would stay on.

So the REST import is removable under any prefix, but only where the file does not name the module through it:

```java
// The REST import in a form removing it cannot break the file. `validModuleName` counts the
// import as enabled under any prefix, but disable is not free to delete one the file uses: an
// `as _` import cannot be referred to at all, while any other prefix goes only when nothing in
// the file names it — deleting that one would leave its references unresolved.
private static boolean removableRestImport(ImportDeclarationNode importNode, ModulePartNode root) {
    return validModuleName(importNode) && !isPrefixReferenced(root, importPrefix(importNode));
}
```

`isPrefixReferenced` walks the file's members for a `QualifiedNameReferenceNode` under that prefix (import prefixes are file-scoped, so the check is against the same document the import sits in), and `importPrefix` falls back to the last module-name segment when the import states none. `unusedLegacyImport` keeps its `as _`-only rule deliberately: a legacy import written without it is the management library the program means to use, whether or not this file happens to reference it yet.

New `config6`/`proj6` covers the used form:

```ballerina
import ballerina/io;
import ballerina/workflow.management.rest;

public function main() {
    io:println(rest:PORT);
}
```

with `"output": {}` — nothing removed. `DisableWorkflowManagementTest` is 6/6, `config2` still removing the unreferenced prefixed import.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
rdulmina: Fixed in ae6578d1 — one scanner now owns the rules, and both readers consume it:

```java
private static List<String> splitTopLevel(String text) {
    List<String> parts = new ArrayList<>();
    int start = 0;
    for (int comma : topLevelPositions(text, ',', false)) {
        parts.add(text.substring(start, comma));
        start = comma + 1;
    }
    ...
}

private static int topLevelIndexOf(String text, char target) {
    List<Integer> positions = topLevelPositions(text, target, true);
    return positions.isEmpty() ? -1 : positions.get(0);
}
```

`topLevelPositions(text, target, firstOnly)` holds the escape block and the depth tracking once, so a quoting fix lands in both the comma split and the key/value cut — the drift you describe is no longer possible.

`stripLineComments` stays separate: it tracks no depth, and it copies text rather than reporting positions (it has to emit the characters it passes and keep the line break that ends a comment). Folding it in would mean a per-character callback for one caller, so it keeps its own loop — the two that could disagree are the two that are now one.

`WorkflowUtilLiteralTest` covers the shared scanner through both readers: role lists, nested duration records, comma- and colon-carrying strings, templates, and comments.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
rdulmina: Fixed in ae6578d1 — both copies are gone. `WorkflowUtil.eventNameLiteral` is the one entry point, and it encodes through `stringLiteral`:

```java
public static String eventNameLiteral(String value) {
    String trimmed = value == null ? "" : value.trim();
    // Already a string literal: needs a distinct pair of quotes (a lone quote does not qualify).
    return trimmed.length() >= 2 && trimmed.startsWith("\"") && trimmed.endsWith("\"")
            ? trimmed : stringLiteral(trimmed);
}
```

`SendDataBuilder:178`, `ChildWorkflowSendDataBuilder:171` and `DurableAgentUpdateBuilder:168` all call it; the two `toStringLiteral` methods are deleted. The already-quoted pass-through is kept as it was, so only the escaping changes — a name with a line break now emits `"line\nbreak"` instead of a literal broken across two lines.

Pinned in `WorkflowUtilLiteralTest`:

```java
Assert.assertEquals(WorkflowUtil.eventNameLiteral("line\nbreak"), "\"line\\nbreak\"");
Assert.assertEquals(WorkflowUtil.eventNameLiteral("say \"hi\""), "\"say \\\"hi\\\"\"");
Assert.assertEquals(WorkflowUtil.eventNameLiteral("\""), "\"\\\"\"");   // a lone quote is not a pair
```

Worth noting the reader side is deliberately narrower: `unescapeLiteralBody`, which decodes a declared channel name out of an agent declaration, decodes only `\\` and `\"` on purpose (its own doc says why). That pairing is unchanged here.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
rdulmina: Fixed in ae6578d1 — four label/doc constant pairs, used by both call sites, with the visible (richer) wording winning the drift:

```java
private static final String RETRY_USER_ROLES_LABEL = "Reviewer Roles";
private static final String RETRY_USER_ROLES_DOC =
        "Role(s) permitted to decide the human review, e.g. \"manager\" or [\"finance\", "
                + "\"manager\"]. Leave empty to allow any role.";
private static final String RETRY_TITLE_LABEL = "Title";
private static final String RETRY_TITLE_DOC = "Short summary shown in the reviewer's inbox. Defaults to a "
        + "phrase naming the activity being reviewed.";
...
```

```java
manualRetryFields.put(RETRY_TITLE_KEY,
        buildRetrySubProperty(RETRY_TITLE_LABEL, RETRY_TITLE_DOC, "string", true));
...
addHiddenRetrySubFieldProperty(nodeBuilder, RETRY_TITLE_KEY,
        RETRY_TITLE_LABEL, RETRY_TITLE_DOC, "string", review.title());
```

"retry review" is gone with the rest of the short forms. The hidden properties are serialized into the flow model, so 14 L3 goldens carried the old wording and are regenerated in the same commit — `ModelGeneratorTest` (188), `NodeTemplateTest` (123), `WorkflowManagerTest`, `WorkflowNodeTemplateTest`, `AnalyzeActivityActionTest`, `GenActivityTest` all green afterwards.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
rdulmina: The overload is gone in ae6578d1, but the table is off by one caller — it was not dead. `DurableAgentAddActivityBuilder:279` passes **seven** arguments, so it bound to this overload, not the six-arg one:

```java
ActivityCallBuilder.addRetryPolicyFormProperties(this, ActivityCallBuilder.NO_RETRY_VALUE,
        "", "", "", "", "");   // <- seven: the last "" is retryUserRoles
```

Deleting the overload on its own therefore broke the build (`incompatible types: String cannot be converted to ReviewFormValues`).

What that call site actually says is "no roles", which the six-arg overload already expresses — `ReviewFormValues.ofRoles("")` and `ReviewFormValues.empty()` are the same value. So the call site drops the trailing argument and the overload goes with it:

```suggestion
        ActivityCallBuilder.addRetryPolicyFormProperties(this, ActivityCallBuilder.NO_RETRY_VALUE,
                "", "", "", "");
```

Two overloads remain: the six-arg default and the `ReviewFormValues` one. `ofRoles` stays, still used by `CodeAnalyzer` for the pre-record `string|string[]` review. `durable_agent_add_activity.json` and `durable_agent_add_activity_preselected.json` are unchanged by it beyond the wording regeneration.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/HumanTaskBuilder.java
rdulmina: Fixed in ae6578d1. The label is a constant again, and the two docs come from one template so they cannot drift apart while the pre-rename pin lives:

```java
// The input field's documentation. The module renamed the parameter — `payload` before 0.9.0,
// `taskInput` after — and a resolved form carries whichever key the pinned bala declares, so both
// spellings are needed. One template, so the wording cannot drift while the pre-rename pin lives.
private static final String INPUT_DOC_TEMPLATE =
        "Read-only JSON object shown alongside the form. Required — a task with nothing to show "
                + "says so with {}, and the runtime checks this against the task's %s";
private static final String TASK_INPUT_DOC = INPUT_DOC_TEMPLATE.formatted("taskInputType");
private static final String PAYLOAD_DOC = INPUT_DOC_TEMPLATE.formatted("payloadType");
```

so the call site matches its neighbours:

```suggestion
        relabel(properties, PAYLOAD_KEY, PAYLOAD_LABEL, PAYLOAD_DOC);
```

`TASK_INPUT_DOC` is byte-identical to what it was, so no golden moves for this one — the type field named is the only difference between the two, which is now the only thing stated twice.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
```diff
@@ -1116,16 +1178,55 @@ private static String retryPolicyExpression(Property retryPolicy, Map<String, Pr
...
+    }
+
+    /** Adds {@code name: <value>} when the form holds one, quoting a bare word as a string. */
+    private static void addQuotedRecordField(List<String> fields, Map<String, Property> properties,
+                                             String key, String name) {
+        String value = trimmedValue(properties, key);
+        if (value.isBlank()) {
+            return;
+        }
+        boolean quoted = value.startsWith("\"") || value.startsWith("string `");
```
NipunaRanasinghe, line 1221: **Medium — CONFIRMED. The decoder and the encoder still disagree for a title that is not a literal.**

`stringLiteralText` deliberately returns a non-literal expression as source — its javadoc says "a variable reference, a template … is returned as written" — but this check only recognises `"` and `string \`` as already-source. So `title: reviewTitle` reads as `reviewTitle` and writes back as `title: "reviewTitle"`; a const or `getTitle()` goes the same way. The field is `ValueType.EXPRESSION`, so this is its primary editing mode, not just a source round trip.

`testExpressionTitlePassesThrough` covers the template and stops one step short of a bare reference. The roles field a few lines up already has the pattern — `roleSource` reads either mode back as source.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/main/java/io/ballerina/flowmodelgenerator/extension/WorkflowManagementService.java
```diff
@@ -223,34 +189,85 @@ private static boolean hasManagementImport(Package pkg) {
...
+    // brings the listener in, and a program may well refer to its configurables by name.
+    private static boolean validModuleName(ImportDeclarationNode importNode) {
+        return moduleName(importNode).equals(MODULE_NAME);
+    }
+
+    // The REST import in a form removing it cannot break the file. `validModuleName` counts the
+    // import as enabled under any prefix, but disable is not free to delete one the file uses: an
+    // `as _` import cannot be referred to at all, while any other prefix goes only when nothing in
+    // the file names it — deleting that one would leave its references unresolved.
+    private static boolean removableRestImport(ImportDeclarationNode importNode, ModulePartNode root) {
```
NipunaRanasinghe, line 207: **Low — CONFIRMED. Correctly skipping a referenced import leaves the checkbox lying about it.**

When the file names the module through its prefix, disable is a no-op — right call — but nothing says so. `disableWorkflowManagement` returns `CommonSourceResponse`, which has no `enabled` field, so `PackageOverview`'s `res.enabled ?? false` unchecks the box while the import, and the REST API, stay on. It re-checks itself on the next load.

Returning whether the import was actually removed would let the checkbox tell the truth.

NipunaRanasinghe (approved):
Approving with non-blocking suggestions.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
rdulmina: Confirmed and fixed in 61d6928189 (on #1062, which now carries the staging merge — happy to move it if you would rather it landed separately).

The premise holds exactly as described. `stringLiteralText` returns a non-literal as written, and the writer's test was a prefix sniff, so the round trip was:

```
title: reviewTitle   →  form holds `reviewTitle`  →  saved as title: "reviewTitle"
```

A const, `getTitle()` and `"Order " + id` went the same way.

Following the roles field as you suggested took one step more than reusing `roleSource`, because a mode has to exist before it can be read: title and description were declared `EXPRESSION`-only, so there was nothing to check. Decoding a literal is also what destroys the distinction — once the quotes are off, `reviewTitle` could be either — so the mode is now captured at the point of the read, where both spellings are still in hand, and carried:

```java
public record ReviewText(String value, boolean expression) { ... }
```

The hidden property that stores each field now carries TEXT and EXPRESSION with that one selected, `addQuotedRecordField` quotes only in text mode, and the sub-field offers a text box alongside the expression editor — so a wording typed as text is still quoted, and a reference is written as it stands.

`testReferenceTitleIsNotQuoted` covers a bare reference, a qualified one, a call and a concatenation; `testTypedTitleIsQuoted` covers the text-mode side. The activity goldens are regenerated for the extra type — none of them holds a title, which is why the L3 suite was silent on this.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/main/java/io/ballerina/flowmodelgenerator/extension/WorkflowManagementService.java
rdulmina: Premise does not hold on the current code — no change made, and here is why.

The response from `disableWorkflowManagement` never reaches the checkbox. The visualizer calls the RPC client, which goes to the extension's rpc-manager, and that method discards the LS response after applying its edits and re-queries the real state:

```ts
// rpc-managers/workflow-management-service/rpc-manager.ts:117
const res: any = await context.langClient.disableWorkflowManagement({ projectPath });
if (res?.textEdits && Object.keys(res.textEdits).length > 0) {
    await updateSourceCode({ textEdits: res.textEdits, ... });
}
removeWorkflowManagementConfigToml(projectPath);
const result = await context.langClient.isWorkflowManagementEnabled({ projectPath });
return result as WorkflowManagementResponse;   // ← this is what PackageOverview reads
```

So `res.enabled` at `PackageOverview/index.tsx:1122` comes from `isWorkflowManagementEnabled`, which sets it from `hasManagementImport` (`WorkflowManagementService.java:104`) — and `hasManagementImport` counts an import under any prefix. In the case you describe, the referenced prefixed import is skipped by `removableRestImport`, stays in the file, and `hasManagementImport` therefore returns true, so the response carries `enabled: true` and the box stays checked. The `?? false` fallback only fires when the field is absent, which it is not on that response.

`CommonSourceResponse` having no `enabled` field is real but not observable here, since nothing reads that response's `enabled`.

One adjacent thing the trace did turn up, which is the opposite ordering hazard rather than this one: `isWorkflowManagementEnabled` re-loads the project immediately after `updateSourceCode`, so a genuinely successful removal reports stale `true` if the edit has not landed in the LS's snapshot yet — leaving the box checked until the next load. Happy to open that separately if you think it is worth chasing; it is not what this thread is about.

PR #885, opened 2026-08-18
Generate a trigger service that calls an agent

dan-niles:
Adds the language-server support for generating a Ballerina service that calls an agent when a trigger fires — the code-generation side of connecting an agent to a channel like WhatsApp, Telegram, or GitHub.

- New `AgentTriggerServiceBuilder`, used when a service-creation request carries an agent.
- Chat channels (WhatsApp Business, Telegram, Google Chat, the built-in chat endpoint) each own their reply client and reply logic inside the generated service, so deleting the service removes everything it needs.
- Event-source connectors (GitHub, Kafka, CDC, and the rest of the gallery) share one generic builder driven entirely by the connector's own schema — no per-connector code required.
- The generated handler always offloads the agent call with `start`, so a slow run doesn't cause webhook retries or duplicate replies.
- 44 new tests in `AgentTriggerGenerationTest`.


packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/model/context/GetServiceInitModelContext.java
```diff
@@ -34,9 +34,11 @@
...
  * @param isLocalRepository whether the connector was picked from a Ballerina local-repository search
  *                          result rather than Central (see {@code ServiceModelRequest.isLocalRepository})
+ * @param agentName         the agent variable this trigger is created for, or {@code null}
+ * @param agentOrgName      the publishing org of that agent, deciding {@code .run} vs {@code ->run}
  * @since 1.3.0
  */
 public record GetServiceInitModelContext(String orgName, String packageName, String moduleName, String version,
                                          Project project, SemanticModel semanticModel, Document document,
-                                         boolean isLocalRepository) {
+                                         boolean isLocalRepository, String agentName, String agentOrgName) {
```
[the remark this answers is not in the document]
dan-niles: Not fixing. This is an internal builder context with one production call site (`ServiceBuilderRouter:166`) and one test, both in this module and both updated; accessors are unchanged, so readers stay compatible. Its canonical constructor has been widened four times before (`a91a92a`, `b3ce8ff`, `32f49f6`, `4c087e1`) with no compat overload — that is the convention here. The two siblings you cite are not counter-examples: `ServiceModelRequest` is the Gson-deserialized RPC payload and already had those ctors, and `TriggerBasicInfo`'s exists so six in-module call sites didn't have to change.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/SchemaDrivenSourceGenerator.java
```diff
@@ -142,29 +143,33 @@ public static Map<String, List<TextEdit>> buildAddServiceEditsForTrigger(Service
...
+            if (slash <= 0 || slash == moduleRef.length() - 1) {
+                continue;
+            }
+            String org = moduleRef.substring(0, slash);
+            String module = moduleRef.substring(slash + 1);
+            if (!Utils.importExists(rootNode, org, module)) {
+                imports.append(Utils.getImportStmt(org, module));
             }
         }
         return imports.toString();
```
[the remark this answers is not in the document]
dan-niles: Not fixing. I checked `importStatements` across all 26 bundled trigger models: none lists its own module, `ballerina/log` or `ballerina/http`, and none repeats an entry — the only non-empty ones are the four CDC models. `channel.imports()` is `ballerina/log`, or `ballerina/http` for AiChat whose self-module is `ballerina/ai`. So the collision isn't reachable on shipped data, and the un-deduped loop predates this PR. Separately, there is a real duplicate-import bug here you didn't flag: for `ballerinax/mysql.cdc.driver as _` the module parses as `mysql.cdc.driver as _`, so `Utils.importExists` never matches and a second CDC service in the same file re-emits it. Pre-existing; tracking separately.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/agent/EventAgentTriggerChannel.java
```diff
@@ -0,0 +1,324 @@
...
+        }
+        String chosen = binding.get().handlerName();
+        List<String> members = new ArrayList<>();
+        for (TriggerUISchemaModel.FunctionModel handler : handlers(context)) {
+            if (handler.name().equals(chosen)) {
+                members.add(binding.get().handler());
+            } else if (isPresent(handler)) {
+                members.add(render(handler, context, ""));
+            }
+        }
```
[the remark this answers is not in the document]
dan-niles: Not fixing. `trigger-metadata.schema.json` marks `name` required for handlers and no bundled model has an unnamed function, so this needs malformed schema data. `byName`'s guard is there because its `name` *argument* is routinely null (`formValue(HANDLER)` and `PREFERRED_HANDLER.get(...)`), not because handler names are. The suggested inversion also isn't null-safe: `chosen` is `primary.name()` and `primary` can be `handlers.getFirst()`, so `chosen.equals(...)` NPEs in exactly the same case where the current code returns false. Only `Objects.equals` would work, and with a null name `buildHandlerSource` emits `remote function null(...)` anyway.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/AgentTriggerServiceBuilder.java
```diff
@@ -0,0 +1,252 @@
...
+        SchemaDrivenSourceGenerator.ResolvedListener listener = channel.listener(rootNode, emitAlias)
+                .orElseGet(() -> SchemaDrivenSourceGenerator.resolveListener(filledModel, emitAlias));
+        AgentTriggerContext channelContext = new AgentTriggerContext(emitAlias, listener.varName(),
+                formValues.get(AGENT_NAME_PROPERTY), formValues.getOrDefault(AGENT_ORG_PROPERTY, BALLERINA_ORG),
+                formValues, filledModel, triggerModel);
+
+        Optional<ServiceDeclarationNode> existing = listener.declaration() != null ? Optional.empty()
+                : findService(rootNode, listener.varName(), channelContext.serviceDescriptor());
+        Optional<AgentTriggerChannel.HandlerBinding> binding = channel.handlerBinding(channelContext);
+        if (existing.isPresent() && binding.isPresent()) {
```
LakshanWeerasinghe, line 153: **The duplicate-service guard can never fire for the chat channels.**

The guard is `existing.isPresent() && binding.isPresent()`, but `TelegramChannel`, `WhatsAppBusinessChannel` and `GoogleChatChannel` never override `handlerBinding`, so `binding` is always empty for them.

Repro: wire agent A to Telegram (creates a new listener → `service telegram:TelegramService on telegramListener`), then wire agent B to Telegram picking the *existing* listener (`telegram.json` has an `existingListener` branch). `listener.declaration()` is null and `findService` *does* match — the descriptor resolves to `telegram:TelegramService` — but `binding` is empty, so we fall through to the append path below and emit a second identical `service telegram:TelegramService on telegramListener { … }`. Two services of the same type on one listener fail to register.

`EventAgentTriggerChannel` is protected against exactly this case (`testASecondEventIsWiredIntoTheServiceThatAlreadyExists`); the chat channels aren't.

LakshanWeerasinghe (commented):
## Review: agent trigger service generation

Went through the new `AgentTriggerServiceBuilder` + `builder/service/agent/*` channels, the `SchemaDrivenSourceGenerator` visibility/refactor hunks, `ServiceBuilderRouter` routing, the `TriggerBasicInfo` / `ServiceModelRequest` widening, and `trigger_properties.json`. The schema-driven channel is a nice shape — most of what's below is about the bespoke channels not carrying the same protections as the event one.

Highest-impact three:

1. The duplicate-service guard in `AgentTriggerServiceBuilder` is unreachable for the chat channels (they never populate `handlerBinding`), so wiring a second agent to an existing Telegram/WhatsApp/Google Chat listener appends a duplicate service.
2. `TelegramChannel`'s `{{token}}` resolves to empty for two of the three `authentication` branches, emitting `check new ({token: });`.
3. The offload idempotence check is a raw substring match, so a user edit to the generated handler parameter makes re-wiring emit a duplicate method.

Rest inline. Also flagged a couple of listing/builder disagreements (Central search, `trigger_properties.json`) that fail silently rather than loudly.

Things I specifically checked and found correct, for what it's worth: the text-block indentation arithmetic in `REPLY_METHOD` / `render` / `mergeIntoService`; `escapeTemplate`'s `${`-before-backtick ordering; that the new FORM children grafted under `SERVICE_TYPE_DESCRIPTOR` aren't traversed by `collectListenerArgs` / `findServiceType` / `findBasePath` and so don't leak into listener args or annotations; descriptor defaults matching `serviceTypes[0].name` in every shipped model; the `buildImports` / `rewritePrefix` / `composeParameters` refactors being behaviour-preserving; and the compatibility constructors covering all existing call sites.

<sub>🤖 Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/agent/TelegramChannel.java
```diff
@@ -0,0 +1,77 @@
...
+            """;
+
+    @Override
+    public AgentTriggerKind kind() {
+        return AgentTriggerKind.CHAT;
+    }
+
+    @Override
+    public String serviceBlock(AgentTriggerContext context) {
+        return context.fill(SERVICE_BLOCK)
```
LakshanWeerasinghe, line 73: **`{{token}}` can resolve to empty and emit a syntax error.**

`context.formValue("token")` returns `""` when the key is absent from the flattened form. In `trigger-models/telegram.json`, `token` lives inside the *first* branch of the `authentication` CHOICE ("provide exactly one of `secretToken` or `token`"); the `secretToken` branch and the `existingListener` branch both carry no `token`.

Either selection generates:

```ballerina
self.telegramClient = check new ({token: });
```

which doesn't parse. There's no fallback, no validation, and no channel-level required-property entry for it — compare `WhatsAppBusinessChannel`, which declares `accessToken` as required. The new Telegram tests never assert on the token, so this passes CI.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/AgentTriggerServiceBuilder.java
```diff
@@ -0,0 +1,252 @@
...
+                .filter(FunctionBodyBlockNode.class::isInstance)
+                .map(FunctionBodyBlockNode.class::cast)
+                .findFirst();
+        LineRange lastMember = members.isEmpty() ? service.openBraceToken().lineRange()
+                : members.get(members.size() - 1).lineRange();
+        if (body.isEmpty()) {
+            return List.of(new TextEdit(Utils.toRange(lastMember.endLine()),
+                    TWO_NEW_LINES + binding.handler() + TWO_NEW_LINES + binding.replyMethod()));
+        }
+        if (body.get().toSourceCode().contains(binding.offload())) {
```
LakshanWeerasinghe, line 194: **Substring-based idempotence check re-emits the handler after an edit.**

The guard is a raw `body.toSourceCode().contains(binding.offload())`, and when it misses we re-emit *both* the offload and `binding.replyMethod()` with no duplicate check on the method itself.

Repro: wire an agent to `onOpened` → generates `_ = start self.runAgentOnOpened(payload);` plus `function runAgentOnOpened(github:IssuesEvent payload)`. Rename the parameter `payload` → `event` in the editor. Re-wire the same event: the offload is regenerated from the schema as `…(payload)`, `contains` misses, and the merge inserts a *second* `runAgentOnOpened` definition plus a call referencing the now-nonexistent `payload` — duplicate-symbol and undefined-symbol errors.

Matching on the method name (as the `existing` handler lookup does) rather than the full statement text would be more robust here.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/agent/AgentTriggerChannels.java
```diff
@@ -0,0 +1,83 @@
...
+    private AgentTriggerChannels() {
+    }
+
+    public static Optional<AgentTriggerChannel> forModule(String orgName, String moduleName) {
+        return forModule(orgName, moduleName, null, false);
+    }
+
+    public static Optional<AgentTriggerChannel> forModule(String orgName, String moduleName, String version,
+                                                          boolean isLocalRepository) {
+        AgentTriggerChannel bespoke = bespoke(orgName, moduleName);
```
LakshanWeerasinghe, line 53: **`BESPOKE.get(moduleName)` ignores `orgName`.**

`orgName` is accepted and then dropped, so any module named `ai`, `telegram`, `whatsapp.business` or `googleapis.chat` from *any* org resolves to the bespoke channel. A local-repository fork surfaced by `searchLocalRepository` (e.g. `myorg/telegram`) gets `TelegramChannel`'s hardcoded `telegram:TelegramService` / `telegram:Client` / `message.chat.id` template and generates code against types the selected package doesn't have.

The schema-driven fallback directly below this correctly keys on org + module — worth making `BESPOKE` do the same. (Same applies at line 69.)

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/ServiceModelGeneratorService.java
```diff
@@ -465,9 +467,13 @@ public CompletableFuture<TriggerListResponse> searchTriggers(TriggerListRequest
                     .collect(Collectors.toSet());
             String query = request == null ? null : request.query();
             List<TriggerBasicInfo> centralTriggers = TriggerSearchUtil.searchCentral(
-                    RemoteCentral.getInstance(), query, null, localKeys);
+                    RemoteCentral.getInstance(), query, null, localKeys).stream()
+                    .map(AgentTriggerChannels::withAgentKind)
```
LakshanWeerasinghe, line 471: **Central search advertises agent support that the builder then refuses.**

`withAgentKind` is applied to `TriggerSearchUtil.searchCentral` results, but `toTriggerBasicInfo` there hardcodes `EVENT_TYPE` for every keyword-matched package, so `kindOf` returns `EVENT` unconditionally. The builder side (`AgentTriggerChannels.forModule`) additionally requires that `getSchemaDrivenTriggerModel` actually resolve *and* have `kind == "event"`.

So a Central trigger package with no `trigger-authoring.json` / `trigger-metadata.json` is advertised to the client as agent-capable; when the user picks it, `AgentTriggerServiceBuilder.handles` returns false and the trigger is created with no agent wiring at all — silently, no error.

`AgentTriggerGenerationTest` asserts that "the listing and the builder must agree on what is offerable", which holds for the curated `trigger_properties.json` list but not for search results.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/agent/EventAgentTriggerChannel.java
```diff
@@ -0,0 +1,324 @@
...
+        Optional<HandlerBinding> binding = handlerBinding(context);
+        if (binding.isEmpty()) {
+            return "";
+        }
+        String chosen = binding.get().handlerName();
+        List<String> members = new ArrayList<>();
+        for (TriggerUISchemaModel.FunctionModel handler : handlers(context)) {
+            if (handler.name().equals(chosen)) {
+                members.add(binding.get().handler());
+            } else if (isPresent(handler)) {
```
LakshanWeerasinghe, line 230: **A contract-required handler can be dropped when a different one is selected.**

`isPresent(handler)` requires `enabled() && !optional`, but every `schemaFunctions`-based connector ships `enabled: false`, so in practice only the *chosen* handler is emitted.

For `asb`, `mqtt`, `solace` and `solace.jms`, `onMessage` is `optional: false` — i.e. contract-required. The `agentEventHandler` dropdown built at line ~148 offers *every* handler, `onError` included. Selecting `onError` yields a service containing only `onError` and no `onMessage`, which doesn't satisfy `solace:Service` / `mqtt:Service`.

Either filter non-payload handlers out of the dropdown, or always emit the `optional: false` ones alongside the selection.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/agent/EventAgentTriggerChannel.java
```diff
@@ -0,0 +1,324 @@
...
+    @Override
+    public Optional<HandlerBinding> handlerBinding(AgentTriggerContext context) {
+        List<TriggerUISchemaModel.FunctionModel> handlers = handlers(context);
+        if (handlers.isEmpty()) {
+            return Optional.empty();
+        }
+        TriggerUISchemaModel.ServiceTypeModel serviceType = context.serviceType();
+        TriggerUISchemaModel.FunctionModel primary = selectHandler(handlers, context.formValue(HANDLER),
+                serviceType == null ? "" : serviceType.name());
+        String replyMethodName = REPLY_METHOD_PREFIX + capitalize(primary.name());
```
LakshanWeerasinghe, line 195: **`handlerName` diverges from the name actually generated for variant-bearing handlers** (latent).

`HandlerBinding.handlerName` is `primary.name()`, but the function `SchemaDrivenSourceGenerator.buildFunctionSource` emits is named by `effectiveFunctionName(function)` — which returns the selected variant's `codedata.originalName()` when a parameter is a `VARIATION_SELECTOR`. `TriggerFunctionAdapter.toFunction` handles this divergence explicitly; this path doesn't.

Consequence: for a variant-bearing handler the merge lookup `fn.functionName().text().equals(binding.handlerName())` misses the function it just generated, so `mergeIntoService` takes the `body.isEmpty()` branch and appends a whole second handler with the same effective name — duplicate method definition.

No shipped `trigger-models/*.json` uses variants today, so this isn't reachable yet, but library-supplied manifests can trigger it.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/AgentTriggerServiceBuilder.java
```diff
@@ -0,0 +1,252 @@
...
+            return;
+        }
+        for (Map.Entry<String, Value> entry : properties.entrySet()) {
+            Value field = entry.getValue();
+            if (field == null) {
+                continue;
+            }
+            String value = field.getValue();
+            if (value != null && !value.isBlank()) {
+                flat.putIfAbsent(entry.getKey(), value);
```
LakshanWeerasinghe, line 238: **Depth-first flattening lets a nested field shadow the channel-owned keys.**

`collect` recurses depth-first with `flat.putIfAbsent`, so a nested leaf wins over a same-named field encountered later at the top level. `agentName`, `agentOrg` and `instructions` are appended *last* by `getServiceInitModel` (`ServiceInitModel.properties` is a `LinkedHashMap`), so any connector whose init form nests a field keyed `agentName` / `agentOrg` / `instructions` shadows them, and the generated code calls a connector-config value as if it were the agent variable.

Reading those three keys directly off the top-level property map, rather than through the flattened view, would remove the collision class entirely.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger_properties.json
```diff
@@ -344,5 +344,18 @@
     "version": "0.9.0",
     "icon": "https://bcentral-packageicons.azureedge.net/images/ballerinax_azure.storage.files_0.9.0.png",
     "kind": "file"
+  },
+  "26": {
```
LakshanWeerasinghe, line 348: **The new `ai` / "Agent Chat" entry pins `1.0.0` and has no `icon`.**

Every other entry ships an `"icon"`; this one doesn't, and it pins `"version": "1.0.0"`. Because `version` and `kind` are both present, `getTriggerBasicInfoByName(TriggerProperty)` takes the fast path, so the client is told the version is 1.0.0 and the icon is `CommonUtils.generateIcon("ballerina", "ai", "1.0.0")` → `ballerina_ai_1.0.0.png`, which 404s. The PR's own test projects depend on `ballerina/ai` 1.13.0.

Separately: nothing in `filterTriggers` excludes this entry, so "Agent Chat" now shows up in the ordinary trigger picker too. Picked without an agent, `handles()` is false and it routes to `AiChatServiceBuilder`, whose `getServiceInitModel` comes from the sqlite index rather than `AiChatChannel` — probably not the intended UX.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/AgentTriggerServiceBuilder.java
dan-niles: Confirmed and fixed in [`cf1551e0`](https://github.com/ballerina-platform/ballerina-vscode/pull/885/commits/cf1551e08b). `findService` does match for the chat channels, and the empty `binding` dropped it to the append path. `buildEdits` now throws when a service already sits on the listener and the channel has no binding to merge into, rather than emitting the twin.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/agent/TelegramChannel.java
dan-niles: Confirmed and fixed in [`cf1551e0`](https://github.com/ballerina-platform/ballerina-vscode/pull/885/commits/cf1551e08b). `token` only exists under `listener.choices[0] -> listenerConfig -> authentication.choices[0]`, so the `secretToken` and existing-listener branches both emitted `check new ({token: });`. The reply client now takes its own required `botToken`, matching WhatsApp's `accessToken`. Note this means the create-new + token branch asks for the bot token twice — I preferred that over reusing the key `token`, where the nested listener value would silently shadow whatever the user typed.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/AgentTriggerServiceBuilder.java
dan-niles: Confirmed and fixed in [`cf1551e0`](https://github.com/ballerina-platform/ballerina-vscode/pull/885/commits/cf1551e08b). The reply-method half was the worse one — it did not even need a rename, deleting just the offload line and re-running the wizard duplicated the method. The offload is now matched on `start self.<replyMethod>(` and the reply method is only appended when no method of that name exists; `HandlerBinding` carries `replyMethodName` for both.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/agent/EventAgentTriggerChannel.java
dan-niles: Confirmed and fixed in [`cf1551e0`](https://github.com/ballerina-platform/ballerina-vscode/pull/885/commits/cf1551e08b). `isPresent` is now `!optional || enabled`, so contract-required handlers survive a different selection. Checked the blast radius across all shipped models: it changes exactly `asb`, `mqtt`, `solace` and `solace.jms` `Service.onMessage`. github is all `enabled:true`, and kafka/rabbitmq/cdc are all-optional, so their output is unchanged.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger_properties.json
dan-niles: Fixed the data in [`cf1551e0`](https://github.com/ballerina-platform/ballerina-vscode/pull/885/commits/cf1551e08b): version `1.0.0` -> `1.13.0` and an explicit icon. On the picker point — the agent popup dodges the broken icon via `channelIcon`'s codicon special-case, so it only showed in the ordinary picker. Leaving the visibility question open; that one is a UX call, not a bug fix.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/agent/AgentTriggerChannels.java
dan-niles: Agreed — fixed in [`3ccd5e37`](https://github.com/ballerina-platform/ballerina-vscode/pull/885/commits/3ccd5e3740), before this review landed. `BESPOKE` is keyed on `org/module` and `kindOf` takes `orgName`. One narrowing: `searchTriggers` results never reach the agent popup (it reads `triggers.local`), so what was actually wrong was the `agentTriggerKind` label on those rows, not the builder routing.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/agent/EventAgentTriggerChannel.java
dan-niles: Fixed in [`4a864103`](https://github.com/ballerina-platform/ballerina-vscode/pull/885/commits/4a86410318). `effectiveFunctionName` is now public, and both the binding's handler name and the reply-method name derive from it; `serviceBlock`'s loop compares effective names on both sides so the two stay consistent. No-op on current data, as you noted — no shipped model uses a VARIATION_SELECTOR.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/AgentTriggerServiceBuilder.java
dan-niles: Fixed in [`4a864103`](https://github.com/ballerina-platform/ballerina-vscode/pull/885/commits/4a86410318). Added `reclaimChannelKeys`, which re-asserts the channel-owned keys over the flattened view after `flattenFormValues`. Scoped to `channel.additionalProperties().keySet()` plus `agentName`/`agentOrg` rather than "top level always wins", so connector-field precedence is unchanged.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/ServiceModelGeneratorService.java
dan-niles: Traced this one and I do not think it holds as an agent bug. `synthesizeTriggerModel` hardcodes `"event"` as the kind for every synthesized model, so any Central or local-repository package that resolves passes the `kind == "event"` filter in `forModule`. The kind check only ever rejects bundled entries (the four `file` ones and `mcp`) — it is a no-op for search results, so nothing from Central is advertised agent-capable and then refused on kind grounds.

What is left is narrower: `forModule` still returns empty when the package ships no trigger metadata at all. But such a package cannot produce any trigger — the ordinary picker falls through to `DefaultServiceBuilder` for it too. Same root cause, pre-existing, not agent-specific. The useful follow-up is excluding metadata-less packages from trigger search generally, which fixes both surfaces; happy to raise that separately.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/builder/service/agent/AgentTriggerChannels.java
dan-niles: Correction to my earlier reply on this thread: I said search results never reach the agent popup, and that is wrong. `AddAgentTriggerPopup` renders `CentralSearchPanel` (behind `ballerina.enableAdditionalTriggerSearch`, default off), so a local-repository `myorg/telegram` really could reach `forModule`. Your original framing was right — the org-key fix corrected actual routing, not just a label.

LakshanWeerasinghe (approved):

PR #886, opened 2026-08-18
Add and delete an agent's triggers from its focus diagram

dan-niles:
Wires real data into the caller rail and adds the ability to add or delete a trigger directly from the agent's focus diagram.

image

- Caller rows are populated from the design model, grouped by service, capped at 5, and open the actual call site rather than always the entry point.
- A `+` tile opens a channel picker (Chat Channels / Event Sources, searchable, with live Ballerina Central results) that lands in the shared service-creation wizard with the agent pre-filled.
- A `⋮` on a row deletes the trigger's service and, if nothing else uses it, its listener — re-resolving the listener's position from a fresh design model so a stale range can't remove the wrong text.
- Fixed `PullingStatus.ERROR` never being set, which left a failed fetch stuck on a dead Retry button.
- New `agentUsages.test.ts` coverage.


packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/index.tsx
```diff
@@ -162,6 +175,9 @@ export function BIFocusFlowDiagram(props: BIFocusFlowDiagramProps) {
         });
         rpcClient.onParentPopupSubmitted((parent: ParentPopupData) => {
             console.log(">>> on parent popup submitted", parent);
+            if (!parent?.recentIdentifier && !parent?.artifactType) {
```
kaumini, line 180: **[low-med] Guard may drop legitimate reloads**

`ParentPopupData` is `{ recentIdentifier, artifactType, dataMapperMetadata? }` and the extension's `notifyChange` always emits all three, so this early-return skips the reload for a `dataMapperMetadata`-only submit even though previously *every* parent-popup submit reloaded the focus model. Only skip a submit that is genuinely empty of payload:

```suggestion
            if (!parent?.recentIdentifier && !parent?.artifactType && !parent?.dataMapperMetadata) {
```

packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/index.tsx
```diff
@@ -259,6 +275,105 @@ export function BIFocusFlowDiagram(props: BIFocusFlowDiagramProps) {
...
+                    return;
+                }
+                const usages = findAgentUsages(response.designModel, {
+                    filePath,
+                    startLine: pos.startLine,
+                    symbol: typeof renderNode.properties?.variable?.value === "string"
+                        ? renderNode.properties.variable.value.trim()
+                        : undefined,
+                }, triggerProtocols);
+                usagesDirtyRef.current = false;
```
kaumini, line 310: **[medium] Usage rail can get stuck stale**

Clearing the dirty flag here is guarded only by `requestId`, but a content change can land *during* an in-flight fetch without bumping `requestId` — the reload it schedules is still sitting in the `debouncedGetAgentModel` debounce window. The stale fetch then resolves, passes the `requestId` guard, clears `usagesDirtyRef`, and caches a result computed from the pre-change design model. When the debounced reload finally runs, `loadAgentUsages` hits `if (!usagesDirtyRef.current && cached) return;` and never refetches — so the rail keeps showing outdated callers until some unrelated content update comes along.

Suggested fix — track a content "version" that bumps on every content update, snapshot it at fetch start, and bail if it changed mid-fetch (mirrors the existing `requestId` guard):

```ts
// with the other refs
const usagesContentRef = useRef(0);

// in onProjectContentUpdated
usagesDirtyRef.current = true;
usagesContentRef.current++;

// in loadAgentUsages' setTimeout callback
const contentAtStart = usagesContentRef.current;          // snapshot before the awaits
...
if (requestId !== usageRequestIdRef.current || usagesContentRef.current !== contentAtStart) {
    return;                                                // leave dirty set; the pending reload refetches
}
...
usagesDirtyRef.current = false;                           // now only reached for an up-to-date fetch
```

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AddAgentTriggerPopup.tsx
```diff
@@ -0,0 +1,230 @@
...
+            })
+            .catch((error: unknown) => console.error(">>> Error fetching trigger models", error))
+            .finally(() => {
+                if (!cancelled) {
+                    setIsLoading(false);
+                }
+            });
+        return () => {
+            cancelled = true;
+        };
```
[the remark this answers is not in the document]
dan-niles: Not taking this — it would fork the error UX for one popup.

The `.catch(err => console.error(">>> Error fetching trigger models", err))` here is the established pattern at the other two `getTriggerModels` call sites (`ComponentListView/index.tsx:95-111`, `CreateIntegrationWizard/index.tsx:307-310`); this popup is a near-copy of the former. If we want a failure state it should land at all three, not here alone.

Also not fully dead on failure: typing a query still renders `CentralSearchPanel`.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/ServiceCreationView.tsx
```diff
@@ -169,42 +200,44 @@ export function ServiceCreationView(props: ServiceCreationViewProps) {
...
+            if (!isMountedRef.current) {
+                return;
+            }
+
+            setHeaderInfo({ title: initModel.displayName, moduleName: initModel.moduleName });
+            setServiceInitModel(initModel);
+            setFormFields(mapPropertiesToFormFields(initModel.properties));
+            setFilePath(target.filePath);
+            setTargetLineRange({ startLine: endOfFile, endLine: endOfFile });
+            setPullingStatus(undefined);
```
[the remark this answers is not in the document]
dan-niles: Not taking this — the state is unreachable, and it is not a regression.

`joinProjectPath` only returns an empty `filePath` when `StateMachine.context().projectPath` is unset, which is the workspace-overview state (`stateMachine.ts:680`). Both mounts of this view — the MainPanel route and `PopupPanel.tsx:106` (`BIAddAgentTrigger`, opened from an agent focus diagram) — require an open project, which sets it.

Pre-PR the render guard was the same `filePath && targetLineRange`, so an empty path produced the same dead form before this change.

Worth revisiting if workspace-level trigger creation ever ships.

packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/index.tsx
dan-niles: Fixed in ef308d88a1. Added a content version that bumps on every content update; the fetch snapshots it at start and drops its result if it moved, so the dirty flag survives for the pending reload.

packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/index.tsx
dan-niles: Leaving this as is. The guard can only ever drop a redundant reload: any real source change also fires `onProjectContentUpdated`, and that handler is not gated, so the rail refreshes anyway.

The data mapper case cannot reach here either — `startAddAgentTrigger` is the only popup this view opens, and `InlineDataMapper` is only opened from `FlowDiagram` (never mounted at the same time). The empty payload the guard drops is the trigger popup closing: it is opened without an `artifactType` and closed with no `parent`, so every close emits all-undefined.

kaumini (approved):

PR #888, opened 2026-08-18
Add Metadata JSON related details for service types in to the WI Copilot context

SasinduDilshara:
Add Metadata JSON related details for service types in to the WI Copilot context

Fixes https://github.com/wso2/product-integrator/issues/2125

## Summary

- Added schema v2 support for trigger and service metadata.
- Enabled WI Copilot to resolve service types, listeners, handlers, parameters, annotations, constraints, imports, and platform dependencies.
- Replaced intermediate JSON processing with typed model objects while preserving wire output.
- Improved service selection, type-reference scanning, library filtering, and passthrough handling.
- Added service rendering, annotation, schema, wire-contract, corpus, and integration tests.
- Updated dependency metadata, pinned corpus versions, and related build configuration.
- Added safeguards for recursive type expansion and metadata version validation.


packages/ballerina-extension/src/features/ai/utils/libs/to-syntax-string.ts
```diff
@@ -418,35 +1282,1036 @@ function deriveListenerAlias(listenerName: string): string | null {
...
+    lines.push(...renderServiceAnnotationLines(service.annotations, listenerAlias));
+
+    // Spec §3 and §2 `deprecated`, as doc sections rather than notes -- see `renderDeprecationSection`.
+    // Last among the `#` lines and immediately before the annotations, because Ballerina metadata puts
+    // every `#` line ahead of every annotation and `@deprecated` is the annotation this pairs with.
+    //
+    // The listener's own deprecation is stated here too: a service is written `on new <listener>(...)`, so
+    // a superseded listener is a fact about the declaration a reader is about to write, and stating it
+    // anywhere else would be stating it nowhere they would look.
+    lines.push(...renderDeprecationSection(service.deprecated, ""));
```
RNViththagan, line 2076: This emits a `#` documentation line after an `@` annotation, which Ballerina metadata rejects (`bal build`: "missing close bracket token").

`renderServiceAnnotationLines` (line 2272) returns `[...notes, ...attachments]`, so it already pushes the `@alias:ServiceConfig {...}` attachment. The `renderDeprecationSection` calls here then push `# # Deprecated ...` lines *after* that attachment, so a service that both carries a §8 annotation obligation and is deprecated renders `@... {...}` followed by `#` — uncompilable.

The comment just above says the deprecation block is "immediately before the annotations", but with the flattened `renderServiceAnnotationLines` it lands after them. `renderHandlers` (line ~2109) handles the identical case correctly by using `splitAnnotationRequirementLines` and interleaving `depBlock` before `obligationAttachments`. Suggest doing the same here so the deprecation `#` lines always precede the attachment.

Latent today (no corpus document is both deprecated and §8-obligated, as the surrounding comment notes), but broken by construction the moment one is.

packages/ballerina-extension/src/features/ai/utils/libs/to-syntax-string.ts
```diff
@@ -378,11 +622,617 @@ function renderStandaloneFunction(func: RemoteFunction): string {
...
+
+    // The type the reader would actually WRITE for this shape — which is what has to be compared against
+    // what is already on the page. Comparing the bare bound instead suppressed nothing for `array` and
+    // `stream`: ftp's csv slot is declared `string[][]` while its bound is `string[]`, so the two never
+    // matched and all four of its shapes restated types the signature had already given.
+    const completion = shape.completionType
+        ? qualifyDeclaredType(shape.completionType, listenerAlias)
+        : "";
+    const written = shape.form === "stream"
+        ? (completion ? `stream<${bound}, ${completion}>` : `stream<${bound}>`)
```
RNViththagan, line 840: Minor / possibly unreachable — worth a sanity check: `written` for the stream form has a space after the comma (`stream<${bound}, ${completion}>`), but the `visible` set is built from `qualifyDeclaredType(param.type)`. If that renders streams without the space (`stream<X,error?>`), `visible.has(written)` never matches and the note restates a type the signature already shows — the array case (`${bound}[]`) matches but the stream case wouldn't. Likely latent since parameter-side stream binding shapes don't appear in the corpus today (stream bindings are return-side). If a param-side stream shape can occur, normalizing the spacing on both sides would keep suppression consistent with the array case.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/service/ServiceLoader.java
```diff
@@ -53,49 +77,137 @@ private ServiceLoader() {
...
+        Set<String> absorbed = new HashSet<>();
+        List<Service> services = new ArrayList<>();
+        for (Service svc : fixedServices) {
+            String name = svc.getName();
+            if (name != null && genericNames.contains(name)) {
+                if (!schemaDerived) {
+                    continue;
+                }
+                absorbed.add(name);
+                InstructionLoader.loadServiceInstruction(libraryName)
```
RNViththagan, line 188: `loadServiceInstruction(libraryName)` is called inside the `for (Service svc : fixedServices)` loop with a loop-invariant argument, so it re-reads the same `service.md` classpath resource once per colliding service. `CopilotLibraryManager.augmentServicesWithInstructions` hoists this exact read above its loop (once per library) — suggest matching that here: load once before the loop and reuse. Low impact today (only http/graphql collide), consistency only.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReader.java
```diff
@@ -62,8 +62,19 @@ public final class LibraryMetadataReader {
...
+    /**
+     * Sized for the Copilot, which walks every library in one request. At the designer's bound of 2 the
+     * bundled documents evicted each other and nearly every request re-parsed the same JSON.
+     *
+     * <p>Held comfortably above the corpus rather than at it. The 2026-08-19 corpus is 32 documents, and a
+     * bound of 20 would have reintroduced exactly the eviction this cache exists to remove — silently, as a
+     * slow request rather than a wrong answer. The headroom is what stops the next document from doing so
+     * again; the entries are small parsed models, so the cost of the slack is negligible.
+     */
+    private static final int PACKAGED_METADATA_CACHE_SIZE = 64;
```
LakshanWeerasinghe:
```suggestion
    private static final int PACKAGED_METADATA_CACHE_SIZE = 3;
```

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReader.java
```diff
@@ -84,11 +95,45 @@ public static LibraryMetadataReader getInstance() {
...
+     * the sample project picks, neither of which suits a caller walking every library against a release it
+     * already compiled.
+     *
+     * <p>Empty means "no readable document here" and nothing more; where to look next is the caller's
+     * policy, composable with {@link Optional#or}.
+     *
+     * @param pkg the already-resolved package (may be {@code null})
+     * @return the parsed document, or empty when the package ships none that can be read
+     */
+    public Optional<TriggerMetadataModel> getTriggerMetadataModel(Package pkg) {
```
LakshanWeerasinghe, line 118: remove this and use getTriggerMetadataModel(ModuleInfo). To create a ModuleInfo object use `public static ModuleInfo from(PackageDescriptor packageDescriptor)`

submodules/wso2-vscode-extensions
```diff
@@ -1 +1 @@
-Subproject commit d8ef67e08f1b71ed2cbd26e42816bb562af98530
+Subproject commit 09b37c86c93420e45f7d5b5ffb2b3c200c4c70d3
```
RNViththagan: This submodule bump goes **backward**: `09b37c86` (Aug 5) is an ancestor of the base's `d8ef67e0` (Aug 19) — `git merge-base --is-ancestor` confirms it, 106 commits behind. Among the dropped commits are the ones adding `bi-smb.svg` and `bi-oracledb.svg` to font-wso2-vscode — and this PR itself adds `ballerinax/smb` and `ballerinax/oracledb` to the corpus, so their icons would be missing. Looks like a leftover from the conflict resolution (the head commit is an "after-conflict" merge). Could you re-point the submodule at `d8ef67e0` or newer?

packages/ballerina-language-server/gradle.properties
```diff
@@ -99,6 +99,35 @@ persistCoreVersion=1.9.1
...
+stdlibAiVersion=1.11.2
+stdlibAvroVersion=1.2.1
+stdlibCacheVersion=3.10.0
+ballerinaCloudVersion=4.0.0
+stdlibDataCsvVersion=0.8.2
+stdlibEmailVersion=2.13.0
+stdlibFileVersion=1.13.0
+stdlibJwtVersion=2.15.1
+stdlibMathVectorVersion=1.2.0
+stdlibMcpVersion=1.0.3
```
RNViththagan: Two things in this new block:

1. `stdlibMcpVersion=1.0.3` — but this same PR bumps mcp to **1.2.0** in `Ballerina.toml`/`Dependencies.toml`, and the block's own header says "Versions must match build-config/ballerina_dependencies/Dependencies.toml". Same skew for `stdlibAiVersion=1.11.2` (lock has 1.13.0) and `stdlibEmailVersion=2.13.0` (lock has 2.14.0), so the offline test distribution stages jars the lock doesn't ask for.
2. 13 of these keys (`stdlibMcpVersion`, `stdlibAiVersion`, `stdlibEmailVersion`, …) already exist higher up in this file — Java properties are last-wins, so these later duplicates silently override the earlier definitions, and anyone bumping the earlier `stdlibMcpVersion` will see no effect. Only `stdlibProtobufVersion` and `stdlibWebsubVersion` are genuinely new; suggest updating the existing keys in place instead of re-declaring them.

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
```diff
@@ -593,6 +560,105 @@ function getOwnTypeDefsForLib(
...
+                add(annotation?.typeConstraint);
+            }
+            // Spec §9: every type a binding note can name. The envelope matters most — the renderer tells
+            // the reader to write `*kafka:AnydataConsumerRecord;`, which is unusable unless that record is
+            // defined in the same prompt.
+            //
+            // Walks `typedescs[]`, the shape §9 now takes. It walked the removed `modes[]` until this was
+            // fixed, which silently emptied the whole branch and dropped every envelope, bound and excluded
+            // type out of the closure.
+            for (const variant of param.binding?.typedescs ?? []) {
```
RNViththagan, line 696: This binding walk exists only for **method parameters**, but the renderer emits type names from three more binding positions that this closure never visits:

- `template.parameters[].binding` → `renderBindingNotes(template, …)` (to-syntax-string.ts:1983)
- `template.return.binding` → `renderReturnBindingNotes(template, …)` (:1988)
- `method.return.binding` → `renderReturnBindingNotes(method, …)` (:2091)

The template loop above (line 601) adds `param.type`/`alternatives`/`annotationRefs` but not `param.binding`, and both `return` positions only get `add(….return?.type)`. Since both scanners (`getOwnRecordRefs` at 679 and `getExternalTypeDefRefs` at 866) share this function, a binding envelope/completionType from those positions is missed by **both** the same-library typeDefs closure and the external fetch list — so the prompt can say `may bind to a record that includes *mod:SomeEnvelope;` while defining `SomeEnvelope` nowhere. The return path is the sharpest case because `renderReturnBindingNotes` deliberately suppresses nothing ("Nothing is suppressed here", :828).

Suggest factoring this typedescs walk into a helper and calling it from the template param loop and both return positions.

packages/ballerina-extension/src/features/ai/utils/libs/to-syntax-string.ts
```diff
@@ -92,6 +153,59 @@ function escapeRegExp(str: string): string {
...
+ *    foreign prefix (`http:Headers`), and neither takes an alias.
+ *
+ * Deduplicated by record name, so a union naming the same type twice cannot be prefixed twice.
+ */
+function qualifyDeclaredType(type: Type | undefined, listenerAlias: string | null): string {
+    if (!type) {
+        return "";
+    }
+    const externalLinks = collectExternalLinks(type);
+    if (externalLinks.length > 0) {
```
RNViththagan, line 174: Two questions about the external branch here:

1. It early-returns after `applyPrefixToTypeName`, so a type whose `links` carry **both** an external and an internal entry would get its home-module members left bare (`Session|http:Request` instead of `mcp:Session|http:Request`). The schema path (`TypeResolver`) never mixes categories on one type, but the index JSON and the symbol-path builders can — is a mixed-links type impossible on every path that reaches this function? If not, applying the internal pass to the external result (instead of returning) would make it order-independent.
2. `applyPrefixToTypeName` still matches with `\b…\b` — the internal branch below was explicitly moved to lookarounds because a recordName can end in `]` (`AnydataConsumerRecord[]`) and `\b` after `]` never matches. If an external link can carry a `[]`-suffixed recordName the same way internal ones provably do, the external branch has the exact bug the internal one fixed. Worth mirroring the lookaround fix there too.

packages/ballerina-extension/src/features/ai/utils/libs/to-syntax-string.ts
```diff
@@ -418,35 +1282,1036 @@ function deriveListenerAlias(listenerName: string): string | null {
...
+function renderHandlers(service: FixedService, listenerAlias: string | null,
+                        terminator: string): string[] {
+    const lines: string[] = [];
+    for (const method of service.methods ?? []) {
+        // `.trimEnd()` is load-bearing, not tidiness. An index-served description arrives with a trailing
+        // newline, and appending another produced a BLANK LINE between the description and the `# + param`
+        // lines below it. A blank line terminates a Ballerina doc comment, so the description became a
+        // dangling comment and only the parameter docs attached to the handler. Four of salesforce's
+        // handlers are affected.
+        const desc = method.description ? `    # ${method.description.trimEnd()}\n` : "";
```
RNViththagan, line 1842: `desc` (and `paramDocs`/`paramDeprecations` below) emit the text on a single `# `-prefixed line without splitting on `\n`. A description with an **internal** newline (the `.trimEnd()` comment shows corpus descriptions already arrive with trailing ones) puts the continuation at column 0 with no `#` — a bare line inside a service body, which both detaches the doc comment and breaks the syntax. `renderDescription` (:282) already handles this with `.split("\n").map(l => "# " + l)` — worth routing these through the same treatment.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/service/HandlerCatalogAspect.java
```diff
@@ -0,0 +1,203 @@
...
+            return;
+        }
+        Set<String> usedNames = new HashSet<>();
+        for (TriggerMetadataModel.ServiceType.Param param : params) {
+            if (param != null && param.name() != null) {
+                usedNames.add(param.name());
+            }
+        }
+        for (int i = 0; i < params.size(); i++) {
+            ParamScope paramScope = new ParamScope(handlerScope, params.get(i), null, i, usedNames);
```
RNViththagan, line 183: The pre-seed loop just above guards `param != null && param.name() != null`, and the sibling `buildFromOptions` does `if (option == null || option.name() == null) continue;` — but this loop passes `params.get(i)` into `ParamScope` unguarded. A null slot in a document's `params[]` NPEs inside name resolution (`ParamTypeResolver.resolveName` → `param.name()`), and because `TriggerSchemaServiceLoader.load` catches `RuntimeException` and returns `empty(documentResolved=true)`, `ServiceLoader.loadAllServices` then withholds the index fallback — so one malformed slot silently drops the **whole library's** services instead of one parameter. A `continue` matching the pre-seed guard keeps the failure per-entry like everywhere else in this pipeline.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/CopilotLibraryManager.java
```diff
@@ -166,30 +180,28 @@ public List<Library> loadFilteredLibraries(String[] libraryNames) {
...
             List<Symbol> moduleSymbols = semanticModel.moduleSymbols();
-            CopilotDeprecationEnricher.enrich(servicesJson, moduleSymbols);
-            CopilotListenerNameEnricher.enrich(servicesJson, moduleSymbols);
-            List<Service> services = new ArrayList<>();
-            for (JsonElement serviceElement : servicesJson) {
-                Service service = GSON.fromJson(serviceElement, Service.class);
-                services.add(service);
-            }
+            CopilotDeprecationEnricher.enrich(services, moduleSymbols);
+            CopilotListenerNameEnricher.enrich(services, moduleSymbols);
```
RNViththagan, line 194: Now that the enrichers also run over the schema-derived services: `CopilotListenerNameEnricher` resolves a **single** override class (the first class type-including `Listener` when no canonical `Listener` exists) and rewrites the class part of *every* service's `listener.name`. The schema path has already resolved listeners per-service via `ListenerPairingResolver`, and they can legitimately differ per service. For the current corpus this is a no-op or idempotent (mcp has a canonical `Listener`; cdc-style packages ship one listener), but a schema-driven package with two non-canonical listener classes would have all its services collapsed onto whichever class comes first in `moduleSymbols`. Should the enricher skip services whose listener the schema path already resolved?

packages/ballerina-extension/src/features/ai/utils/libs/to-syntax-string.ts
RNViththagan, line 2076: Two more details on top of my earlier comment here, from a deeper pass:

1. `service.listener.deprecated` (line 2282) triggers the same `#`-after-`@` ordering, not only `service.deprecated` — a deprecated *listener* on an annotated service breaks identically.
2. The §8 ordering test ("documentation precedes every annotation, including @deprecated") can't catch this: it uses `isDeprecated: true`, which emits only the bare `@deprecated` and never the prose `# # Deprecated` section, and no test combines prose deprecation with service annotations. So a regression test for the fix needs `deprecated: "..."` prose plus an annotation obligation on the same service.

PR #893, opened 2026-08-19
Add support for `ProductMode` flag for the Integrator and Agent Builder surfaces

dan-niles:
Introduces `ProductMode`, replacing the ad-hoc `agentBuilder` flag, to decide whether the extension shows the full WSO2 Integrator surface or the focused Agent Builder one. Integrator's own behaviour is unchanged.

- `getProductMode()` reads `WSO2_PRODUCT_MODE` once at startup; exposed via `StateMachine.productMode()` and `useProductMode()`.
- `productTerms` / `useProductTerms()` centralises mode-specific wording instead of an `isAgentBuilder ? … : …` ternary at every call site.
- Copilot presents as "WSO2 Agent Builder Intelligence" in agent mode, with matching theming and a run-state orb.
- New `agent-builder` Copilot skill that pins the exact source shapes the language server recognises, so agent code Copilot writes renders correctly in the diagram.
- `PackageOverview` picks up the new orb and a shared `PageHeader`.

## Summary

- Added centralized `ProductMode` support for Integrator and Agent Builder environments.
- Exposed product mode through core state helpers, RPC APIs, and visualizer hooks.
- Replaced fixed assistant names with mode-aware names and terminology.
- Added Agent Builder-specific orb styling, welcome content, templates, and `agent-builder` skill documentation.
- Added shared product terms and `PageHeader` components.
- Improved package overview and integration creation flows.
- Added Central icon and entry-type glyph resolution helpers.
- Updated AI prompts, authentication messages, status displays, and error messages to use dynamic product names.


packages/ballerina-visualizer/src/components/AgentStatusOrb/CopilotHeroBox.tsx
```diff
@@ -147,7 +151,7 @@ export function CopilotHeroBox({ placeholder }: { placeholder: string }) {
     const state = status?.state ?? "idle";
     const active = state !== "idle";
     const colors = ORB_COLORS[state];
```
gigara, line 153: This component was left out of the Agent Builder theming: it still uses `ORB_COLORS[state]` and renders the WebGL `ShaderOrb`, and its `AmbientFrame` below gets neither `$agentBuilder` nor `$colors`. Every other orb surface (floating orb, welcome orb, composer frame, footer loading orb) switches to the accent palette / CSS sphere in agent-builder mode — but the hero box in the package-overview empty state keeps the Integrator purple/orange orb and glow.

Note the agent-builder palette contains `var(...)`/`color-mix(...)` strings the shader can't consume, so the fix needs the same `cssSphere` gate used in `AgentStatusOrb/index.tsx`.

gigara (commented):
Reviewed the full diff with the branch checked out; the PR-touched packages type-check clean (`ballerina-core`, `ballerina-rpc-client`, `ballerina-visualizer`, `ballerina-extension`).

Overall this is well built: the `ProductMode` plumbing is consistent end-to-end (env var → `StateMachine.productMode()` → capability handshake → `agentBuilderModeEnabled` RPC → `useProductMode`), the new early return in `PackageOverview` sits safely after all hooks, and the `agent-builder` SKILL.md is genuinely well written. Inline comments below cover the specific findings; two general notes that don't anchor to a single line:

**Formatting churn.** `WaitingForLoginSection`, `MigrationPanel`, and `orchestrator.ts` contain whitespace-only reindents from a different formatter, and the `AmbientFrame` `:focus-within` block in `shared.ts` ended up *misindented* (the arrow-function body was de-indented below its property). The `{showTopBar && (` wrapper in `CreateIntegrationWizard/index.tsx` also leaves the inner `<WizardTopBar>` at its old indent level. Worth one cleanup pass so the diff reflects real changes.

**Unused API surface.** Several additions have no caller in this repo: `nameOnly`/`nameLabel` on `CreateIntegrationWizard`, `bordered` on `TopNavigationBar`, `HERO_GLOW` in `AgentStatusOrb/shared.ts`, the icon helpers in `ballerina-core` (see inline comment), and `PageHeader`'s `hideDivider`/`hasTopNavigationBar`/`isBallerinaWorkspace`/`hideUndoRedo` props. If these are consumed by a follow-up PR in the `agent-triggers` stack that's fine — otherwise they should land together with their consumers.

packages/ballerina-visualizer/src/components/AgentStatusOrb/CopilotHeroBox.tsx
```diff
@@ -124,6 +125,9 @@ const SendButton = styled.button`
 `;
 
 export function CopilotHeroBox({ placeholder }: { placeholder: string }) {
+    const assistantName = useAssistantName();
+    const productMode = useProductMode();
+    const shortName = shortAssistantName(productMode);
```
gigara, line 130: Nit: `useShortAssistantName()` exists in the same hook module, but this (and `MiniChat`, and `PackageOverview` via `shortAssistantName(productMode)`) recompute it by hand. Worth picking one pattern.

packages/ballerina-extension/src/features/ai/agent/prompts.ts
```diff
@@ -66,7 +67,7 @@ import { WEB_SEARCH_TOOL_NAME, WEB_FETCH_TOOL_NAME } from "./tools/web-tools";
  * Generates the system prompt for the design agent
  */
 export function getSystemPrompt(projects: ProjectSource[], op: OperationType, userSkills: ProjectSkillMeta[], disabledSkills?: Set<string>, disabledSkillMetas?: Array<{ name: string; trigger: string }>): string {
-    return `You are WSO2 Integration Intelligence (WII) — you can also go by "Wii" — an expert assistant specialized in Ballerina help with relavant integration usecases. You will be helping with designing a solution for user query in a step-by-step manner.
+    return `You are ${aiAssistantName()}, an expert assistant specialized in Ballerina help with relavant integration usecases. You will be helping with designing a solution for user query in a step-by-step manner.
```
gigara: Two things on this line:

1. The `(WII) — you can also go by "Wii"` alias was dropped from the system prompt, but the Integrator welcome screen (`AIChat/Welcome/index.tsx`) still tells users *"You can call me Wii."* — the UI now invites a name the model is no longer told about.
2. Since the line was edited anyway: `relavant` → `relevant`, `usecases` → `use cases`.

packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/index.tsx
```diff
@@ -536,6 +538,10 @@ const AIChat: React.FC = () => {
                                     }
                                     activeScaffoldKeyRef.current = key;
                                 }
+                                if (defaultPrompt.newThread) {
```
gigara, line 541: `newThread` has no producer anywhere in this repo, and it is only honored inside the `autoSubmit` branch — a prompt with `newThread: true` but no `autoSubmit` silently ignores it. If a follow-up PR produces it, could you at least document the autoSubmit-only constraint on the field in `AIPanelPrompt`? Otherwise this reads as dead code.

packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
```diff
@@ -59,6 +59,64 @@ export const ORB_COLORS: Record<AgentRunState, [string, string, string]> = {
...
+const STATE_BASE = {
+    "running": "var(--vscode-progressBar-background)",
+    "awaiting-input": "var(--vscode-editorWarning-foreground)",
+    "completed": "var(--vscode-editorGutter-addedBackground)",
+    "error": "var(--vscode-statusBarItem-errorBackground)",
+} as const;
+
+/** Agent Builder's own palette — theme-variable-based; Integrator keeps ORB_COLORS above as-is. */
+export const AGENT_BUILDER_ORB_COLORS: Record<AgentRunState, [string, string, string]> = {
+    "idle": ["#6b5ce8", BRAND_ORANGE, "#ffb199"],
```
gigara: This `idle` entry is unreachable — `orbColors()` returns `ACCENT_SPHERE` for agent-builder idle — and it's also misleading: hardcoded Integrator colors inside the "Agent Builder's own palette" table. Suggest removing the entry (or wiring it through `orbColors`) so the table matches reality.

packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
```diff
@@ -215,6 +301,25 @@ export const Sphere = styled.div<SphereProps>`
...
+    inset: -3px;
+    border-radius: 50%;
+    border: 2px solid transparent;
+    border-top-color: ${(props: { color: string }) => props.color};
+    animation: ${spin} 1.1s linear infinite;
+    pointer-events: none;
+
+    @media (prefers-reduced-motion: reduce) {
+        animation: none;
+        opacity: 0.6;
```
gigara, line 319: Behavior change for Integrator too: the old local `SpinArc` used `display: none` under `prefers-reduced-motion`; this shared one shows a static arc at `opacity: 0.6` in both modes. If intentional (arguably better — still signals "running"), fine; if not, it's an accidental regression for reduced-motion users.

packages/ballerina-visualizer/src/hooks/useProductMode.ts
```diff
@@ -0,0 +1,67 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { useEffect, useState } from "react";
+import { useRpcContext } from "@wso2/ballerina-rpc-client";
+import { ProductMode, assistantName, shortAssistantName } from "@wso2/ballerina-core";
+
+/** One fetch per webview; the setting needs a reload to change. */
+let cached: ProductMode | undefined;
```
gigara: Two consequences of this design worth considering:

1. The hook defaults to `INTEGRATOR` until the RPC resolves, so an Agent Builder webview briefly paints "WSO2 Integration Intelligence" branding on first load.
2. A transient RPC failure caches `INTEGRATOR` permanently for the webview's lifetime.

Since the extension host generates the webview HTML and already knows the mode, injecting it into the webview bootstrap state (or using it as the seed here) would remove both.

packages/ballerina-extension/src/rpc-managers/common/rpc-manager.ts
```diff
@@ -284,6 +284,11 @@ export class CommonRpcManager implements CommonRPCAPI {
         return extension.ballerinaExtInstance.enabledAdditionalTriggerSearch();
     }
 
+    async agentBuilderModeEnabled(): Promise<boolean> {
+        const { ProductMode } = await import("../../utils/config");
```
gigara: The dynamic import is unnecessary — this PR moved `ProductMode` to `@wso2/ballerina-core`, which this file already imports statically at the top. Adding it to that import (it's a runtime enum, no cycle risk) lets this method drop the `async` indirection.

packages/ballerina-core/src/icons/brand-icon-registry.ts
```diff
@@ -93,3 +93,19 @@ export const KIND_DEFAULT_ICON: Record<string, BrandIcon> = {
 export function resolveKindDefaultIcon(kind: string | undefined | null): BrandIcon {
     return (kind && KIND_DEFAULT_ICON[kind]) || KIND_DEFAULT_ICON.http;
 }
+
+const CENTRAL_ICON_MODULE = /\/[^/]*?_(.+)_\d[^/]*\.(?:png|svg)$/i;
+
+export function resolveBrandIconFromUrl(url: string | undefined | null): BrandIcon | undefined {
```
gigara, line 99: `resolveBrandIconFromUrl`, `ENTRY_TYPE_GLYPH`, and `resolveEntryTypeGlyph` are exported but have no caller in this repo. If they belong to a follow-up PR in the stack, fine — otherwise they should land with their consumers.

packages/ballerina-extension/src/features/ai/agent/skills/registry.ts
```diff
@@ -83,7 +84,22 @@ export const skillCreatorSkill: Skill = {
     default: false,
 };
 
+// agent-builder skill
+const agentBuilder = parseSkillMd(agentBuilderMd);
+if (!agentBuilder.name || !agentBuilder.description) {
+    throw new Error(`[agent-builder] SKILL.md is missing required frontmatter fields (name="${agentBuilder.name}", description="${agentBuilder.description}")`);
+}
+
+export const agentBuilderSkill: Skill = {
```
gigara, line 93: The load/parse/validate/build block is now repeated three times, and `skill-creator` is the only one skipping the frontmatter validation that `data-map` and `agent-builder` both do. A small `buildSkill(md, name, flags)` helper would remove the duplication and the drift.

packages/ballerina-extension/src/utils/config.ts
```diff
@@ -268,6 +258,14 @@ export function getProductMode(): ProductMode {
         : ProductMode.INTEGRATOR;
 }
 
+export function aiAssistantName(): string {
```
gigara, line 263: The docstring on `getProductMode` above still says callers should read `StateMachine.productMode()` rather than calling it directly — but these two helpers call it directly, and the branded constants in `features/ai/constants.ts` now call it at module-load time. Worth updating the comment to the actual contract (env-derived, safe to call anywhere).

packages/ballerina-extension/src/features/ai/constants.ts
```diff
@@ -15,20 +15,21 @@
...
+import { aiAssistantName } from "../../utils/config";
 
 export const WI_EXTENSION_ID = 'wso2.wso2-integrator';
 
 export const CONFIG_FILE_NAME = "Config.toml";
 export const CONFIGURE_DEFAULT_MODEL_COMMAND = "ballerina.configureWso2DefaultModelProvider";
 
 export const CLOSE_AI_PANEL_COMMAND = "ballerina.close.ai.panel";
-export const SIGN_IN_BI_COPILOT = "Sign in to WSO2 Integration Intelligence";
+export const SIGN_IN_BI_COPILOT = `Sign in to ${aiAssistantName()}`;
```
gigara, line 26: These template literals are evaluated at import time, freezing the product name at module load. Safe today (the env var is fixed before the host starts), but it's a trap if the mode ever becomes a setting — and it pulls `utils/config` (→ vscode, core) into previously dependency-free constants modules. Consider making these functions, matching the call-site interpolation used elsewhere in this PR. Same applies to `natural-programming/constants.ts`.

packages/ballerina-visualizer/src/views/BI/components/PageHeader.tsx
```diff
@@ -0,0 +1,142 @@
...
+                    <EditableTitle title={title} onCommit={onTitleEdit} validate={validateTitle}>
+                        <ProjectTitle>{title}</ProjectTitle>
+                    </EditableTitle>
+                ) : (
+                    <ProjectTitle>{title}</ProjectTitle>
+                )}
+                {subtitle && <ProjectSubtitle>{subtitle}</ProjectSubtitle>}
+            </TitleContainer>
+            <HeaderControls>
+                {!hideUndoRedo && <UndoRedoGroup key={Date.now()} />}
```
gigara, line 137: `key={Date.now()}` remounts `UndoRedoGroup` on every render of the host page. This was inherited from the old inline header, but extracting it into a shared component was the moment to fix it (or at least TODO it).

packages/ballerina-extension/src/views/ai-panel/errorCodes.ts
```diff
@@ -35,7 +36,7 @@ export const PARSING_ERROR: ErrorCode = {
 
 export const UNKNOWN_ERROR: ErrorCode = {
     code: 4,
-    message: "An unknown error occurred while generating code. Try login again to WSO2 Integration Intelligence"
+    message: "An unknown error occurred while generating code. Try login again to copilot"
```
gigara: Grammar + this is now the only message that lost the product name entirely. Suggest:

```suggestion
    message: `An unknown error occurred while generating code. Try logging in again to ${aiAssistantName()}.`
```

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
```diff
@@ -1209,18 +1185,27 @@ export function PackageOverview(props: PackageOverviewProps) {
...
+                                                {/* Disabled rather than hidden mid-turn: reverting the
+                                                    generation restores a pre-turn checkpoint and drops
+                                                    whatever is not in it, including an artifact added now. */}
+                                                <Button
+                                                    appearance="primary"
+                                                    onClick={() => setShowAddIntegration(true)}
+                                                    disabled={agentWorking}
+                                                    tooltip={agentWorking ? `Available once ${shortName} finishes` : undefined}
+                                                >
+                                                    <Codicon name="add" sx={{ marginRight: 8 }} /> Add Integration
```
gigara: The "Add Integration" button, the `AddIntegrationPanel` title, and the empty-state guidance strings here are hardcoded, so Agent Builder users see "Add Integration" while the rest of that surface says "agentic integration". `productTerms` exists precisely for this — `terms.integrationLabel` etc.

packages/ballerina-visualizer/src/views/AIPanel/components/SkillsManager/index.tsx
```diff
@@ -298,7 +300,7 @@ const SkillsManager: React.FC<SkillsManagerProps> = ({ onClose, onSkillsChange }
                     <Loader label="Loading skills…" />
                 ) : (
                     <>
-                        {renderSection(SkillTier.BUILTIN, "Built-in", "Skills shipped with the Copilot.", builtinSkills, "No built-in skills.")}
+                        {renderSection(SkillTier.BUILTIN, "Built-in", `Skills shipped with the ${shortName}.`, builtinSkills, "No built-in skills.")}
```
gigara: Minor wording: "Skills shipped with the Integration Intelligence." reads awkwardly with the article — `Skills shipped with ${shortName}.` scans better in both modes.

packages/ballerina-extension/src/views/visualizer/webview.ts
```diff
@@ -233,13 +237,12 @@ export class VisualizerWebview {
...
         const isDevantEditor = vscode.commands.executeCommand('getContext', 'devant.editor') !== undefined;
 
-        const biExtension = isInWI() || vscode.extensions.getExtension('wso2.ballerina-integrator');
         // After a wizard submit this HTML is the first frame post-reload, so it continues
         // the wizard's "Creating <name>" screen (also handed to the React app below).
         const startupProgress = getStartupIntegrationProgress(
             StateMachine.context().workspacePath || StateMachine.context().projectPath
         );
-        const productTitle = biExtension ? VisualizerWebview.biTitle : VisualizerWebview.ballerinaTitle;
+        const productTitle = VisualizerWebview.webviewTitle;
```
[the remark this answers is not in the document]
dan-niles: Pre-existing — this PR's only change to this file is the tab title; it adds no subscriptions. The undisposed listeners date back to b4e24ac20f (2025-06-30) and are the case the `.coderabbit.yaml` rule 4 note already carves out. Tracking as a separate fix.

packages/ballerina-extension/src/rpc-managers/common/rpc-handler.ts
```diff
@@ -76,6 +77,7 @@ export function registerCommonRpcHandlers(messenger: Messenger) {
     messenger.onRequest(selectFileOrFolderPath, () => rpcManger.selectFileOrFolderPath());
     messenger.onRequest(experimentalEnabled, () => rpcManger.experimentalEnabled());
     messenger.onRequest(additionalTriggerSearchEnabled, () => rpcManger.additionalTriggerSearchEnabled());
+    messenger.onRequest(agentBuilderModeEnabled, () => rpcManger.agentBuilderModeEnabled());
```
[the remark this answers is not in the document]
dan-niles: Skipping — `rpc-handler.ts` is generated (`THIS FILE INCLUDES AUTO GENERATED CODE`), and per `docs/TEST_GUIDE.md` the RPC contract isn't its own tier; the product-mode behaviour is covered at L1 in `singleIntegrationLanding.test.ts`.

packages/ballerina-visualizer/src/hooks/useProductMode.ts
```diff
@@ -0,0 +1,67 @@
...
+    return mode;
+}
+
+export function useAssistantName(): string {
+    return assistantName(useProductMode());
+}
+
+export function useShortAssistantName(): string {
+    return shortAssistantName(useProductMode());
+}
```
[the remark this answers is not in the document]
dan-niles: Skipping for now — `useProductMode`'s module-level cache is likely to be replaced by seeding the mode from the webview bootstrap, so a test written against it now would be thrown away.

packages/ballerina-visualizer/src/components/AgentStatusOrb/CopilotHeroBox.tsx
dan-niles: Skipping — once the stack lands, `MainPanel` swaps `PackageOverview` for `AgentBuilderOverview` in agent-builder mode, so this component only ever renders in Integrator (the Agent Builder empty state builds its own hero and reuses just `openCopilotPanel`/`submitPromptToCopilot` from here).

packages/ballerina-extension/src/features/ai/agent/prompts.ts
dan-niles: Fixed in 377b203262 — the alias is back for Integrator only (Agent Builder has none), and both typos are gone.

packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
dan-niles: Fixed in 377b203262 — `idle` now holds `ACCENT_SPHERE` and `orbColors` just indexes the table.

packages/ballerina-extension/src/rpc-managers/common/rpc-manager.ts
dan-niles: Fixed in 377b203262 — `ProductMode` now comes from the existing static core import.

packages/ballerina-extension/src/utils/config.ts
dan-niles: Fixed in 377b203262 — the docstring now says it is env-derived and safe to call anywhere.

packages/ballerina-extension/src/views/ai-panel/errorCodes.ts
dan-niles: Fixed in 377b203262 — took your suggestion verbatim.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.tsx
dan-niles: Fixed in 377b203262 — both the button and the panel title now use `terms.integrationLabel`.

packages/ballerina-visualizer/src/views/AIPanel/components/SkillsManager/index.tsx
dan-niles: Fixed in 377b203262 — dropped the article.

packages/ballerina-visualizer/src/views/BI/components/PageHeader.tsx
dan-niles: Leaving as-is — the remount is what refreshes the buttons after an edit (the effect only depends on `undoing`/`redoing`), so dropping the key would leave undo/redo stale; worth fixing properly with a real refresh trigger across all four call sites.

packages/ballerina-extension/src/features/ai/constants.ts
dan-niles: Leaving as-is — no cycle results (the chain terminates at `migrationEvents`), and the env is fixed before the host starts, so converting six constants to functions would churn every call site for no behaviour change.

packages/ballerina-extension/src/features/ai/agent/skills/registry.ts
dan-niles: Leaving the helper — data-map's block carries keywords and command templates so it would not collapse cleanly, though you are right that `skill-creator` is the odd one out on validation.

packages/ballerina-visualizer/src/components/AgentStatusOrb/CopilotHeroBox.tsx
dan-niles: Leaving as-is — all three sites need `productMode` for other reasons anyway, and `CopilotHeroBox` turns out to be Integrator-only once `AgentBuilderOverview` lands.

packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/index.tsx
dan-niles: Left undocumented for now — there is still no producer, so the constraint is better written down by whichever PR starts emitting it.

packages/ballerina-core/src/icons/brand-icon-registry.ts
dan-niles: Removed in 0a939017d6 — confirmed no consumer in any branch of the stack, so they can land again with whatever needs them.

dan-niles: Correction — reverted in 832af825a4: my earlier check was wrong (a bad pathspec), and these do have consumers higher in the stack (`bi-diagram` ConnectorIcon and AgentNodeWidget, and AgentBuilderWorkspaceOverview), so removing them broke the build.

packages/ballerina-visualizer/src/hooks/useProductMode.ts
dan-niles: Fixed in b737976211 — the host now writes `window.productMode` in `getLibraryWebViewContent`, so the hook seeds itself on the first render, and a failed RPC clears the in-flight promise instead of caching Integrator.

packages/ballerina-visualizer/src/components/AgentStatusOrb/shared.ts
dan-niles: Intentional, keeping it — `Halo` in the same file and the same running state already freezes rather than hides (`animation: none; opacity: 0.4`), so the shared `SpinArc` now matches it and the old `display: none` was the outlier; reduced motion should drop the motion, not the cue.

PR #894, opened 2026-08-19
Fix agent canvas moving issues, and open the call site from a rail row

dan-niles:
Fixes the agent focus canvas sliding and jumping around — after a Copilot edit, when the terminal opens, in a narrow window, or on window blur — caused by the node's rendered width being unstable rather than any RPC latency.

- Transient `readOnly` state no longer changes what renders, which was the actual root cause.
- The fit now measures painted content instead of the reserved layout box, and re-centres when the node's footprint changes.
- Lowered the minimum zoom floor and added a bail-out for a collapsed canvas.
- The canvas now clips its content, and the pathfinding router no longer measures a canvas it doesn't own.
- A caller rail row now opens where the agent is actually called, not just the entry point.
- Deletions re-resolve positions before cascading, so a stale range can't remove the wrong text.

## Summary

- Stabilized agent focus canvas fitting, positioning, resizing, visibility, and animation.
- Prevented canvas sliding and jumping after edits, terminal actions, window resizing, and focus changes.
- Added painted-content measurement, bounded zoom, centering, overflow clipping, and safer pathfinding behavior.
- Replayed trace animations through a dedicated `useTraceAnimationBridge` hook.
- Updated caller rail navigation to open agents at their actual call sites.
- Re-resolved listener positions before cascading trigger deletions.
- Improved toolkit class resolution for MCP tool deletion.
- Added assistant-name support across diagram context, tooltips, and agent-focus views.
- Improved brand icon resolution and agent-node read-only controls.
- Added tests for listener and helper-function positioning.


packages/bi-diagram/src/components/nodes/AgentWidget/agentFocusFit.ts
```diff
@@ -0,0 +1,170 @@
...
+            model.setZoomLevel(targetZoomPct);
+            model.setOffset(targetOffsetX, targetOffsetY);
+            diagramEngine.repaintCanvas();
+        });
+        canvas.style.transition = "none";
+        canvas.style.transform = "translate(0px, 0px) scale(1)";
+    };
+    canvas.addEventListener("transitionend", finalize, { once: true });
+    const fallback = setTimeout(finalize, AGENT_FOCUS_FIT_ANIMATION_MS + 100);
+}
```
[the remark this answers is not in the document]
dan-niles: Fixed in 5754586209, with one change to the proposed approach.

The stale `finalize` was real, but cancelling only in the animate branch leaves the second half of the bug: the re-entrant `computeAgentFocusFit` reads `nodeElement.getBoundingClientRect()` through the half-finished transform while the model still reports its pre-animation zoom, so the new target is wrong on its own.

So `animateAgentFocusFit` now returns a cancel function that drops the transform *without* committing the target, and `fitToContainer` calls it **before it measures** rather than inside the animate branch. The DOM and the model then agree, a double-click simply restarts the same animation, and the unmount cleanup calls it too so no `flushSync` runs against a torn-down engine.

packages/ballerina-visualizer/src/hooks/useTraceAnimationBridge.ts
```diff
@@ -0,0 +1,49 @@
...
+
+let wired = false;
+
+export function useTraceAnimationBridge(): void {
+    const { rpcClient } = useRpcContext();
+
+    useEffect(() => {
+        if (wired || !rpcClient) {
+            return;
+        }
```
kaumini, line 32: `useTraceAnimationBridge` is defined here but never imported or called anywhere. Since this PR also removed the `rpcClient.onTraceAnimationChanged(...)` registration that used to live in `BIFlowDiagram`, nothing wires up trace animations anymore — `setTraceAnimationActive` / `setTraceAnimationInactive` will never fire.

**Suggested fix** — call the hook from the component that previously hosted the registration (`views/BI/FlowDiagram/index.tsx`, and/or `FocusFlowDiagram`):

```ts
import { useTraceAnimationBridge } from "../../../hooks/useTraceAnimationBridge";

// inside BIFlowDiagram, alongside the other hooks:
useTraceAnimationBridge();
```

Minor: the module-level `wired` flag is never reset and the effect returns no cleanup, so if `rpcClient` is recreated the listener won't re-register. Consider deregistering in an effect cleanup instead.

dan-niles: `useTraceAnimationBridge` is wired from MainPanel in https://github.com/ballerina-platform/ballerina-vscode/pull/907 , which is stacked directly on this branch, so the two land together; on the cleanup point, onTraceAnimationChanged returns void rather than a disposable and rpcClient is created once per webview, so the wired guard is deliberate.

kaumini (approved):

PR #905, opened 2026-08-19
Update the Azure Files trigger model for the 1.0.0 data binding contract

YasanPunch:
PRAM https://github.com/wso2/product-integrator/issues/2234

## Purpose

Aligns the bundled Azure Files trigger model with the connector's 1.0.0 data binding contract. The 1.0.0 compiler plugin narrowed what the typed content handlers accept, so the handlers generated by the current model no longer compile. This also adds the stream toggle that the other file listeners already have.

## Changes

- `onFileJson`: the generated content parameter is now `json` instead of `map<json>`. The 1.0.0 plugin accepts only `json` or a record type, and `map<json>` is rejected, so an Add Handler action produced code that failed to build.
- `onFileCsv`: the generated content parameter is now `record {}[]` instead of `string[][]`. CSV binding is records only in 1.0.0. The row schema button still applies the `{{type}}[]` template, so choosing a record `R` yields `R[]`.
- Added the `Stream (Large Files)` toggle to `onFile` and `onFileCsv`, matching the `ftp` and `smb` models. Azure Files was the only file listener without it, although the connector accepts both stream forms. With the toggle on, the composed parameter is `stream<byte[], error?>` for `onFile` and `stream<record {}, error?>` for `onFileCsv`, which are the forms the 1.0.0 `ContentFunctionValidator` accepts. The remaining handlers are unchanged, since the connector takes no stream for them.
- Version references moved from 0.9.0 to 1.0.0 (three in the trigger model, two in the Azure Files entry of `trigger_properties.json`, including the version stamped icon URL).
- Help text corrections, all describing existing 1.0.0 behaviour:
  - The `On Error` post-processing action no longer claims to cover content binding failures unconditionally. With an `onError` handler declared, the binding failure is disposed of by `onError`'s own annotation instead.
  - The `onError` description no longer implies that errors returned by a handler are delivered to it. They are not.
  - The `err` parameter description now covers failed polls as well as content binding failures.

## Merge gate

Ballerina Central currently publishes `ballerinax/azure.storage.files` 0.9.0 only. Because the model now declares 1.0.0, this should merge together with or after the 1.0.0 release, otherwise the wizard cannot resolve the package.

## Validation

- `TriggerModelReaderTest` (12/12) and `PayloadComposerTest` (5/5) pass.
- Type changes verified against the 1.0.0 `ContentFunctionValidator`, and the help text against the connector spec sections 5.4 and 5.5.
- The stream node mirrors the `ftp` and `smb` shape exactly: a `PAYLOAD_MODIFIER` sibling of `payload` inside the parameter's `COMPLEX_PAYLOAD` container.

## Note for reviewers

The `onError` node has no File Handling Options group, so its `@files:FunctionConfig` cannot be set from the editor. In 1.0.0 that annotation is what disposes of a file after a content binding failure when `onError` is declared, so a service built in the editor leaves such a file in place and it is picked up again on the next poll. Happy to add the group in this PR or a follow-up if you would like it exposed.

## Dependencies

The latest commit points the model at `ballerinax/azure.storage.files` **1.0.1**, which restores the string CSV row forms (`string[][]` and `stream<string[], error?>`) in the connector's compiler plugin. Depends on [module-ballerinax-azure.storage.files#9](https://github.com/ballerina-platform/module-ballerinax-azure.storage.files/pull/9) merging and the 1.0.1 patch reaching Ballerina Central; please merge this PR only after that release.

### DIFF
--- a/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger-models/azure.storage.files.json
+++ b/packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger-models/azure.storage.files.json
@@ -7,9 +7,9 @@
   "orgName": "ballerinax",
   "packageName": "azure.storage.files",
   "moduleName": "azure.storage.files",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "type": "files",
-  "icon": "https://bcentral-packageicons.azureedge.net/images/ballerinax_azure.storage.files_1.0.0.png",
+  "icon": "https://bcentral-packageicons.azureedge.net/images/ballerinax_azure.storage.files_1.0.1.png",
   "kind": "file",
   "listenerKind": "SINGLE_SELECT_LISTENER",
   "initProperties": {
@@ -1213,7 +1213,7 @@
               "typeMembers": [
                 {
                   "type": "ServiceConfiguration",
-                  "packageInfo": "ballerinax:azure.storage.files:1.0.0",
+                  "packageInfo": "ballerinax:azure.storage.files:1.0.1",
                   "packageName": "azure.storage.files",
                   "kind": "RECORD_TYPE",
                   "selected": true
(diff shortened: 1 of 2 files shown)


packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger-models/azure.storage.files.json
```diff
@@ -1294,7 +1293,7 @@
                       "label": "Row Schema",
                       "description": "Create a record type describing one row of the file. Each row arrives as one record."
                     },
-                    "value": "string[][]",
+                    "value": "record {}[]",
```
niveathika, line 1296: default is kept at string[][]

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger-models/azure.storage.files.json
```diff
@@ -1628,7 +1627,7 @@
                 "afterError": {
                   "metadata": {
                     "label": "On Error",
-                    "description": "Action when the handler fails or the file content does not match the expected format."
+                    "description": "Action when the handler returns an error. Also applies when the file content does not match the expected format, unless an onError handler is declared."
```
niveathika, line 1630: If the function return an error we log it and go.

niveathika (changes requested):

niveathika (approved):

LakshanWeerasinghe (approved):

PR #912, opened 2026-08-20
Escape reserved keywords in module name derivation

### LINKED ISSUE #1849: Incorrect Ballerina Code Generated for HubSpot "Import" Connection
### Description

When creating a HubSpot CRM "Import" connection using low-code, the generated Ballerina code in `connections.bal` is malformed. This may be because the "Import" identifier conflicts with Ballerina's reserved keyword import, which does not appear to be properly escaped.

Generated code 

Image

Expected code

Image

### Steps to Reproduce

- Add a new HubSpot Import connector connection.
- leave the configuration details as default 
- Click 'Save Connection'.

Image

### Version

5.0.0.1

_No response_

thuva9872:
## Purpose
Fixes: https://github.com/wso2/product-integrator/issues/1849

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary

- Escape reserved-keyword module names when generating Ballerina imports, qualifiers, and type signatures.
- Preserve raw module names during derivation and normalize escaped names during comparisons.
- Update flow-model generators and language-server services to use the new emission helpers.
- Add comprehensive tests for escaping, unescaping, round trips, and real type signatures.
- Fix malformed code generation for HubSpot CRM `Import` connections.


packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/CommonUtils.java
```diff
@@ -854,15 +860,28 @@ public static String getImportStatement(String orgName, String packageName, Stri
             importStatement.append(orgName).append("/");
         }
         if (moduleName != null && moduleName.startsWith(packageName + ".")) {
-            importStatement.append(moduleName);
+            importStatement.append(escapeModuleName(moduleName));
```
pasindufernando1: **Blocking: the escaped string leaks into module identifiers.**

This return value isn't only import text. `getImportStatements()` joins it into `Property.imports` (Property.java:553-577), and consumers parse it back with `ModuleInfo.from` (ModuleInfo.java:50-57), which takes the path segment verbatim as `packageName`. So `SourceBuilder.getExpressionBodyText` (:413) ends up calling `centralApi.latestPackageVersion("ballerinax", "hubspot.crm.'import")` and `PackageName.from("hubspot.crm.'import")` — neither resolves, the method returns empty, and `DataMapperDefinitionBuilder:160` throws `IllegalStateException("Failed to produce the data mapper output")`. Pre-PR this resolved fine and only the emitted import line was invalid, so this is a regression. Same leak at `ExpressionEditorService.java:297`.

Second half of the same issue at L252: the now-escaped `'import:Rec` no longer matches the `packageMap` key, which is built from the real package name via `getDefaultModulePrefix`, so `BallerinaCompilerApi.getType` misses the module even once the pull is fixed.

Suggest keeping the shared signature unescaped and escaping only at render time (`SourceBuilder.TokenBuilder` / `generateImportsTextEdits`).

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/CommonUtils.java
```diff
@@ -991,7 +1010,8 @@ public static boolean isNaturalExpressionBodiedFunction(SyntaxTree syntaxTree, F
     }
 
     public static String getClassType(String packageName, String clientName) {
-        String importPrefix = packageName.substring(packageName.lastIndexOf('.') + 1);
+        String importPrefix =
```
pasindufernando1: **Blocking: the other prefix derivations are still unescaped.**

This one is fixed, but the same `substring(lastIndexOf('.') + 1)` one-liner is duplicated in `Codedata.java:76`, `FunctionCall.java:77`, `ModelProviderBuilder.java:94`, `EmbeddingProviderBuilder.java:95`, `ActivityCallBuilder.java:557`, `TypeResolver.java:118` and `ServiceIndexLoader.java:282`.

So a function-call node now writes a valid `import ballerinax/hubspot.crm.'import;` and `import:createBatch(...)` beside it, which fails with `ERROR invalid token 'import'` (verified). The goal isn't met for function calls, model/embedding providers or activity calls — the error just moves from the import line to the call site. These should all route through one helper.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/CommonUtils.java
```diff
@@ -854,15 +860,28 @@ public static String getImportStatement(String orgName, String packageName, Stri
...
     }
 
+    /**
+     * Escapes each dot-separated segment of a module name against Ballerina reserved keywords.
+     * e.g. "hubspot.crm.import" -> "hubspot.crm.'import"
+     *
+     * @param moduleName the dot-separated module name
+     * @return the module name with each reserved-keyword segment escaped
+     */
+    private static String escapeModuleName(String moduleName) {
```
pasindufernando1: **Blocking: escaping the writer without the readers breaks import dedupe — reproduced as a duplicate unescaped import plus a retry loop.**

Adding a variable whose type comes from a keyword-named module emits a second, unescaped import next to the escaped one and then retries indefinitely.

`Codedata.getImportSignature()` (Codedata.java:67) is raw `org + "/" + module` and never calls `getImportStatement`, so this escaping doesn't run on the expression-editor path (`ExpressionEditorService.java:209` → `applyModuleImport`). That unescaped string is then the needle for the existence check at `ExpressionEditorContext.java:136-139` (`importText.startsWith("import " + importStatement)`), which can't match the escaped import already in the file — so it never converges, re-emitting the edit and re-firing `PackageUtil.pullModuleAndNotify` each request.

`CommonUtils.importExists` has the mirrored bug: it joins `IdentifierToken::text()`, which carries the leading quote (cf. `QNameRefCompletionUtil.java:94`, which strips it for this reason), and compares against an unescaped name — affecting `TypesManager:1039,1046`, `DataMapManager:3428,4373`, `ConfigEditorV2Service:416`, `XSDTypeGenerator:336`. `TypesManager.getImportStmt` (:1057) also formats `import %s/%s;` raw.

Fix needs an `unescapeReservedKeyword` next to `CommonUtil.escapeReservedKeyword:468` and normalization on both sides of those comparisons.

pasindufernando1 (commented):
Approach is right (verified on bal 2201.13.4: `import testpkg.crm.'import;` compiles and `'import:greet()` / `'import:Rec` resolve), but the escape is applied to two writers and no readers, and those readers use the same string as a package identifier and as a dedupe key. Three blocking issues inline.

Non-blocking: `escapeModuleName` duplicates `ModuleUtil.escapeModuleName` (ModuleUtil.java:196) minus its `PRE_DECLARED_LANG_LIBS` guard; and the new tests assert `SyntaxInfo.isKeyword`, the same predicate `escapeReservedKeyword` uses, so they can't catch a wrong output shape and don't cover the new `isPredefinedLangLib` branch at L252.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/CommonUtils.java
```diff
@@ -991,7 +1010,8 @@ public static boolean isNaturalExpressionBodiedFunction(SyntaxTree syntaxTree, F
     }
 
     public static String getClassType(String packageName, String clientName) {
```
dulajdilshan, line 1179: `getClassType` now escapes the derived prefix unconditionally, but it is missing the predefined-lang-lib exemption this same PR deliberately adds to `getTypeSignature` (lines 249-253).

`getClassType` takes the last dot segment of the module, so for the lang libraries that segment *is* a keyword: `lang.string` → `'string`, `lang.map` → `'map`, `lang.error` → `'error`, `lang.int`, `lang.xml`, `lang.table`, `lang.stream`, `lang.typedesc`, `lang.function`, `lang.object`, `lang.transaction`, `lang.future`, `lang.boolean`, `lang.decimal`, `lang.float`. Those prefixes are legal unescaped — exactly the reason the guard was added in the `getTypeSignature` hunk.

`CallBuilder.setConcreteTemplateData` (`CallBuilder.java:141`) calls `CommonUtils.getClassType(codedata.module(), codedata.object())` for every node kind other than FUNCTION_CALL/AGENT/CLASS_INIT/ACTIVITY_CALL/TYPED_AGENT — i.e. including `METHOD_CALL` — and `CallBuilder.isLangLibFunction` (line 442) confirms `codedata.module()` legitimately holds `lang.*` values on that path. So a lang-lib method-call node's connection property type flips from `string:…` to `'string:…`.

Suggest reusing the `isPredefinedLangLib(orgName, packageName)` guard here too, which means either passing the org into `getClassType` or pre-filtering at the callers.

packages/ballerina-language-server/model-generator-commons/src/test/java/io/ballerina/modelgenerator/commons/CommonUtilsTest.java
```diff
@@ -0,0 +1,87 @@
...
+                {"ballerinax", "x.function.y", "x.function.y"},            // keyword middle segment
+                {"ballerinax", "type", "import"},                          // keyword pkg + keyword module
+                {"ballerinax", "foo", "foo.import"},                       // startsWith branch, keyword tail
+                {"", "hubspot.crm.import", "hubspot.crm.import"},          // no org
+        };
+    }
+
+    @Test(dataProvider = "importStatements")
+    public void testGetImportStatementEscapesKeywordSegments(String orgName, String packageName, String moduleName) {
+        String importStatement = CommonUtils.getImportStatement(orgName, packageName, moduleName);
```
dulajdilshan: Both new tests assert only the invariant "every segment is escaped iff it is a keyword" and never assert the actual produced string, so a regression that keeps every segment individually well-formed but produces the wrong path still passes.

For example `getImportStatement("ballerinax", "hubspot.crm.import", "import")` yields `ballerinax/hubspot.crm.'import.'import` (duplicated tail — a shape the current `getImportStatement` branch structure can produce when a caller passes the bare module prefix as `moduleName`), and this test would report success.

Adding expected-value assertions for the seven data rows would make the suite actually pin the fix.

dulajdilshan (commented):
**`ClassMemberManager.resolveImport` emits an unescaped import for an escaped prefix** (not in this diff, but broken by it)

`flow-model-generator-core/.../ClassMemberManager.java:445` derives `requestedPrefix` from the type string (`type.substring(0, type.indexOf(':'))`), which is now produced by the escaping `getClassType`, so it becomes `'import`. The import text emitted on the very next lines is still built from raw codedata:

```java
String moduleId = codedata.org() + "/" + codedata.module();
... "import " + moduleId + ...
```

For a keyword-named module this writes `import ballerinax/hubspot.crm.import;` — unescaped, a syntax error — while the member's type refers to prefix `'import`, which that declaration never defines.

Additionally, the alias-collision loop compares `'import` against `usedPrefixes` collected from existing (unescaped) import declarations, so an already-present import of the same module is never matched and a duplicate declaration is added.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/CommonUtils.java
thuva9872: addressed in [6a5458e](https://github.com/ballerina-platform/ballerina-vscode/pull/912/commits/6a5458ec454153a58426f61a4b46fb729c1dc4dc)

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/CommonUtils.java
thuva9872: addressed in [6a5458e](https://github.com/ballerina-platform/ballerina-vscode/pull/912/commits/6a5458ec454153a58426f61a4b46fb729c1dc4dc)

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/CommonUtils.java
thuva9872: addressed in [6a5458e](https://github.com/ballerina-platform/ballerina-vscode/pull/912/commits/6a5458ec454153a58426f61a4b46fb729c1dc4dc)

packages/ballerina-language-server/model-generator-commons/src/test/java/io/ballerina/modelgenerator/commons/CommonUtilsTest.java
thuva9872: addressed in [6a5458e](https://github.com/ballerina-platform/ballerina-vscode/pull/912/commits/6a5458ec454153a58426f61a4b46fb729c1dc4dc)

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/CommonUtils.java
thuva9872: addressed in [6a5458e](https://github.com/ballerina-platform/ballerina-vscode/pull/912/commits/6a5458ec454153a58426f61a4b46fb729c1dc4dc)

pasindufernando1 (dismissed):
Blockers from my earlier comment are resolved — thanks for the quick turnaround.

- **`ImportPrefixReader`** — `moduleNameOf` unescapes for lookup while `importPrefixOf` keeps the escaped token for emission; that split is the right one. Good call covering the `as hsImport` case in the new test.
- **`ClassMemberManager`** — `importModuleId` unescapes so the existing import is recognised, and the emitted text goes through `escapeImportStatement`. Nice catch on the `removeIf` at :170 as well.
- **`ActivityGenerator.needsModuleImport`** — unescapes before matching, so the connector import is emitted again.

Approving. Two follow-ups, neither a regression from this PR, so I'd be fine with them as separate issues:

1. The activity signature itself is still raw — `resolveConnectionType()` (:156) and `returnType` (:189 / :211) go through `TokenBuilder.name(String)`, which is a bare append. With `needsModuleImport` fixed we now write `import ballerinax/x.'import;` alongside `import:Client` in the same generated function, so this path still produces an invalid token. `SourceBuilder:1138` (the types.bal import edit) and the two data-mapper builders have the same shape.
2. Prefix-collision sets are mixed-domain: `importedPrefixes` returns `'import` while `ModuleAliasResolver.allocate` compares the raw `selfPrefix(module)`. If a file imports `ballerinax/a.import` and the generator needs `ballerinax/b.import`, the collision isn't detected and the escaped prefix is emitted twice. Narrow, but worth a comment on `importedPrefixes` at minimum.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/CommonUtils.java
```diff
@@ -1146,6 +1211,49 @@ public static String getDefaultModulePrefix(String packageName) {
...
+        if (signature == null || signature.indexOf(':') < 0) {
+            return signature;
+        }
+        Matcher matcher = TYPE_MODULE_PREFIX_PATTERN.matcher(signature);
+        StringBuilder result = new StringBuilder();
+        while (matcher.find()) {
+            String prefix = matcher.group(1);
+            String replacement = CommonUtil.PRE_DECLARED_LANG_LIBS.contains(LANG_LIB_PREFIX + prefix)
+                    ? prefix : CommonUtil.escapeReservedKeyword(prefix);
+            matcher.appendReplacement(result, Matcher.quoteReplacement(replacement) + ":");
```
rosensilva, line 1251: Will this also change text inside string literals? For example, "import:Rec" becomes "'import:Rec", changing the allowed value. Can we skip string literals?

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/main/java/io/ballerina/flowmodelgenerator/extension/ExpressionEditorService.java
```diff
@@ -349,12 +361,13 @@ private String boundPrefix(String filePathString, Codedata codedata, boolean wil
             // The owning organization is passed so an org-less import in the file -- which can only name a
             // module of its own package -- is not taken as a match for a foreign module of the same name.
             String currentOrg = document.map(doc -> doc.module().descriptor().org().value()).orElse(null);
-            return willWriteImport
+            String sourcePrefix = willWriteImport
                     ? ImportPrefixReader.resolve(rootNode, codedata.org(), codedata.module(), null, currentOrg)
                     : ImportPrefixReader.boundPrefix(rootNode, codedata.org(), codedata.module(), currentOrg);
+            return CommonUtils.unescapeModuleName(sourcePrefix);
```
rosensilva, line 367: Will this remove the required quote from prefixes like 'import, producing invalid calls like import:println(). The expression editor uses this text directly

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/ClassMemberManager.java
```diff
@@ -475,8 +482,11 @@ private static TextEdit importEdit(ModulePartNode modulePart, String importText)
 
     private static String importModuleId(ImportDeclarationNode declaration) {
         String org = declaration.orgName().map(value -> value.orgName().text().trim()).orElse("");
-        String module = declaration.moduleName().stream().map(token -> token.text().trim())
-                .reduce((left, right) -> left + "." + right).orElse("");
+        // Unescape so an import read from source (reserved-keyword segments carry the leading quote) compares
+        // equal to the raw module id the model stores.
+        String module = CommonUtils.unescapeModuleName(declaration.moduleName().stream()
+                .map(token -> token.text().trim())
+                .reduce((left, right) -> left + "." + right).orElse(""));
```
rosensilva, line 489: Will this finds an existing keyword import, replacePrefix() adds another quote to the already escaped type, producing ''import:Client. Can we prevent this and test adding a second connection?

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/CommonUtils.java
thuva9872: No. This is only called whenever we encounter a type descriptor.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/main/java/io/ballerina/flowmodelgenerator/extension/ExpressionEditorService.java
thuva9872: The value that is inserted in the expression editor is unescaped one. We escape it only when we generate the source.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/ClassMemberManager.java
thuva9872: yes. We are unescaping all the imports on read time and escaping on source emission. So, it will work fine when finding existing import.

rosensilva (approved):

PR #921, opened 2026-08-20
Serialize shared sample-project package resolution and metadata queries

IsuruMaduranga:
Serializes all access to the singleton sample project's non-thread-safe `PackageResolver` behind `SAMPLE_RESOLUTION_LOCK` (offline resolutions now route through the memoized cache; `cachedVersion`/`isModuleUnresolved` are synchronized), preventing concurrent HashMap corruption in the compiler's `EnvironmentPackageCache` during parallel Language Server requests.

Fixes wso2/product-integrator#2193

- Serialized shared sample-project `PackageResolver` access with `SAMPLE_RESOLUTION_LOCK` to improve stability during parallel Language Server requests.
- Updated offline resolution and unresolved-module checks to use synchronized resolver operations and memoized version data.
- Simplified `AiUtils` by using `PackageUtil.cachedVersion` directly.

### DIFF
--- a/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
+++ b/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
@@ -84,16 +84,8 @@ import io.ballerina.modelgenerator.commons.PackageUtil;
 import io.ballerina.projects.DependenciesToml;
 import io.ballerina.projects.Document;
 import io.ballerina.projects.Package;
-import io.ballerina.projects.PackageDescriptor;
-import io.ballerina.projects.PackageName;
-import io.ballerina.projects.PackageOrg;
 import io.ballerina.projects.Project;
 import io.ballerina.projects.TomlDocument;
-import io.ballerina.projects.environment.PackageMetadataResponse;
-import io.ballerina.projects.environment.PackageResolver;
-import io.ballerina.projects.environment.ResolutionOptions;
-import io.ballerina.projects.environment.ResolutionRequest;
-import io.ballerina.projects.environment.ResolutionResponse;
 import io.ballerina.tools.diagnostics.Location;
 import org.ballerinalang.langserver.common.utils.CommonUtil;
 import org.ballerinalang.langserver.commons.BallerinaCompilerApi;
@@ -105,7 +97,6 @@ import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -633,21 +624,7 @@ public class AiUtils {
     }
 
     public static Optional<String> resolvePackageVersion(String org, String packageName) {
-        try {
-            PackageResolver resolver = PackageUtil.getSampleProject()
-                    .projectEnvironmentContext().getService(PackageResolver.class);
-            ResolutionRequest resolutionRequest = ResolutionRequest.from(
-                    PackageDescriptor.from(PackageOrg.from(org), PackageName.from(packageName)));
-            Collection<PackageMetadataResponse> metadataResponses = resolver.resolvePackageMetadata(
-                    Collections.singletonList(resolutionRequest),
-                    ResolutionOptions.builder().setOffline(true).build());
-            return metadataResponses.stream().findFirst()
-                    .filter(meta -> meta.resolutionStatus() != ResolutionResponse.ResolutionStatus.UNRESOLVED)
-                    .map(PackageMetadataResponse::resolvedDescriptor)
-                    .map(descriptor -> descriptor.version().value().toString());
-        } catch (RuntimeException e) {
-            return Optional.empty();
-        }
+        return Optional.ofNullable(PackageUtil.cachedVersion(org, packageName));
     }
 
     private static synchronized void ensureDependentModulesResolved() {
--- a/packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
+++ b/packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
@@ -117,8 +117,9 @@ public class PackageUtil {
             new ConcurrentHashMap<>();
 
     /**
-     * Serializes cache-miss resolutions: before caching, every call built its own sample project
-     * (and resolver), so the shared resolver was never used concurrently. Keep that property.
+     * Serializes access to the shared sample project's non-thread-safe resolver — cache-miss
+     * resolutions, offline resolutions, and metadata queries — so concurrent Language Server
+     * requests cannot corrupt it.
      */
     private static final Object SAMPLE_RESOLUTION_LOCK = new Object();
 
@@ -164,21 +165,24 @@ public class PackageUtil {
      * i.e. the version the build has provisioned. Returns null if not cached.
      */
     public static String cachedVersion(String org, String name) {
-        try {
-            PackageResolver resolver = SAMPLE_PROJECT.projectEnvironmentContext().getService(PackageResolver.class);
-            Collection<PackageMetadataResponse> responses = resolver.resolvePackageMetadata(
-                    Collections.singletonList(ResolutionRequest.from(
-                            PackageDescriptor.from(PackageOrg.from(org), PackageName.from(name)))),
-                    ResolutionOptions.builder().setOffline(true).build());
-            Optional<PackageMetadataResponse> first = responses.stream().findFirst();
-            if (first.isPresent()
-                    && first.get().resolutionStatus() != ResolutionResponse.ResolutionStatus.UNRESOLVED) {
-                return first.get().resolvedDescriptor().version().toString();
+        synchronized (SAMPLE_RESOLUTION_LOCK) {
+            try {
+                PackageResolver resolver = SAMPLE_PROJECT.projectEnvironmentContext()
+                        .getService(PackageResolver.class);
+                Collection<PackageMetadataResponse> responses = resolver.resolvePackageMetadata(
+                        Collections.singletonList(ResolutionRequest.from(
+                                PackageDescriptor.from(PackageOrg.from(org), PackageName.from(name)))),
+                        ResolutionOptions.builder().setOffline(true).build());
+                Optional<PackageMetadataResponse> first = responses.stream().findFirst();
+                if (first.isPresent()
+                        && first.get().resolutionStatus() != ResolutionResponse.ResolutionStatus.UNRESOLVED) {
+                    return first.get().resolvedDescriptor().version().toString();
+                }
+            } catch (RuntimeException ignored) {
+                // fall through to null
             }
-        } catch (RuntimeException ignored) {
-            // fall through to null
+            return null;
         }
-        return null;
     }
 
     /**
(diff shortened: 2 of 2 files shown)


packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
```diff
@@ -398,42 +397,49 @@ private static Optional<Path> resolveLatestModuleBala(BuildProject buildProject,
...
+                        PackageDescriptor.from(PackageOrg.from(org), PackageName.from(name),
+                                PackageVersion.from(version)));
+                PackageResolver packageResolver = SAMPLE_PROJECT.projectEnvironmentContext()
+                        .getService(PackageResolver.class);
+                return packageResolver.resolvePackageMetadata(Collections.singletonList(resolutionRequest),
+                                ResolutionOptions.builder().setOffline(true).build()).stream()
+                        .findFirst()
+                        .map(response -> response.resolutionStatus() == ResolutionResponse.ResolutionStatus.UNRESOLVED)
+                        .orElse(false);
+            } catch (RuntimeException ignored) {
```
RNViththagan: Small correctness worry with this catch. Returning `false` means "resolved", and there's one call site where that's load-bearing — `service-model-generator/.../extension/util/Utils.java:1659` uses this as the sole guard around the pull:

```java
if (PackageUtil.isModuleUnresolved(orgName, packageName, targetVersion)) {
    notifyClient(MessageType.Info, PULLING_THE_MODULE_MESSAGE, moduleInfo, lsClientLogger);
    Optional<SemanticModel> semanticModel = PackageUtil.getSemanticModel(moduleInfo);
    ...
}
```

Unlike `pullModuleAndNotify`, there's no fall-through `getModulePackage` after the `if`, so a swallowed exception there doesn't just drop the notification — it skips the pull entirely and returns as though the module were already present. `targetVersion` on that path comes from `centralApi.latestPackageVersion(...)`, and `PackageVersion.from` throws `ProjectException` (a `RuntimeException`) on a null or malformed version, so it's reachable rather than theoretical.

Could we at least log it, and ideally separate "resolved" from "couldn't tell"? Anything that stops a failure from looking like a success works for me.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
```diff
@@ -117,8 +117,8 @@ private static BuildOptions balaBuildOptions() {
             new ConcurrentHashMap<>();
 
     /**
-     * Serializes cache-miss resolutions: before caching, every call built its own sample project
-     * (and resolver), so the shared resolver was never used concurrently. Keep that property.
+     * Serializes access to the shared sample project's resolver (cache misses, offline resolutions,
+     * and metadata queries) to ensure thread-safety across concurrent Language Server requests.
```
RNViththagan: On the widened scope described here — flagging the trade-off rather than objecting to it. `memoizedSampleBala` holds this monitor for the whole of `resolution.get()`, which on a miss can be a Central pull. Now that `cachedVersion` (L176) and `isModuleUnresolved` (L427) take the same monitor, two cheap offline metadata reads on hot paths — `fetchVersionIfNotExists`, and `FunctionDataBuilder:402` — can end up blocked for the length of an unrelated download.

I don't think a finer-grained lock is available: `EnvironmentPackageCache` keeps plain `java.util.Map`s (`projectsById`, `projectsByOrgNameVersion`), so reads race with writes too and a single mutex is the honest answer.

One cheap mitigation if you like it: memoize `cachedVersion` the same way. Its javadoc already says it returns "the version the build has provisioned", so it's stable for the session, and caching it would take the hot path off the lock entirely after the first call per coordinate.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
```diff
@@ -117,8 +117,9 @@ private static BuildOptions balaBuildOptions() {
             new ConcurrentHashMap<>();
 
     /**
-     * Serializes cache-miss resolutions: before caching, every call built its own sample project
-     * (and resolver), so the shared resolver was never used concurrently. Keep that property.
+     * Serializes access to the shared sample project's non-thread-safe resolver — cache-miss
+     * resolutions, offline resolutions, and metadata queries — so concurrent Language Server
+     * requests cannot corrupt it.
      */
     private static final Object SAMPLE_RESOLUTION_LOCK = new Object();
```
LakshanWeerasinghe, line 124: This monitor is held across network I/O, and the doc update doesn't mention it. `memoizedSampleBala` (line 135) runs its supplier inside `synchronized (SAMPLE_RESOLUTION_LOCK)`, and both suppliers can hit Central: `resolveVersionedModuleBala` retries with `offline=false`, and `resolveLatestModuleBala` calls `RemoteCentral.getInstance().latestPackageVersion(...)` and then an online `resolvePackages` — which is the actual module pull behind `pullModuleAndNotify`.

Before this PR, `cachedVersion` / `isModuleUnresolved` / `getModulePackageOffline` were offline-only and never blocked. Now they serialize behind that pull: while one thread downloads a connector from Central (seconds, or the HTTP timeout on a bad network), every concurrent `getFlowModel` (`FunctionDataBuilder:402` → `isModuleUnresolved`), service-model request (`Utils:1659`), trigger-model request (`getModulePackageOffline`) and AI request (`AiUtils.resolveAiModuleVersion` → `cachedVersion`) stalls. A bare `synchronized` also makes those waits uninterruptible, so LSP cancellation can't release them and the request pool can be drained into an apparent LS freeze.

The PR description claims "offline resolutions now route through the memoized cache", but none of these three are memoized — each call still does a fresh resolver round-trip under the global monitor. Memoizing them (both answer immutable questions) would match the description and stop cheap queries from repeatedly queueing behind the pull; alternatively use a `ReentrantLock` with `tryLock(timeout)` for the read-only queries and degrade instead of blocking.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
```diff
@@ -398,42 +402,48 @@ private static Optional<Path> resolveLatestModuleBala(BuildProject buildProject,
...
-        PackageResolver packageResolver = buildProject.projectEnvironmentContext().getService(PackageResolver.class);
-        Collection<PackageMetadataResponse> packageMetadataResponses = packageResolver.resolvePackageMetadata(
-                Collections.singletonList(resolutionRequest),
-                ResolutionOptions.builder().setOffline(true).build());
-        Optional<PackageMetadataResponse> pkgMetadata = packageMetadataResponses.stream().findFirst();
-        if (pkgMetadata.isEmpty() ||
-                pkgMetadata.get().resolutionStatus() == ResolutionResponse.ResolutionStatus.UNRESOLVED) {
-            return Optional.empty();
-        }
+        synchronized (SAMPLE_RESOLUTION_LOCK) {
```
LakshanWeerasinghe, line 405: Two issues with wrapping the whole method body:

1. **The critical section is wider than it needs to be.** Only the two `packageResolver` calls touch shared state. `BalaProject.loadProject(defaultBuilder, balaPath)` on line 430 uses `ProjectEnvironmentBuilder.getDefaultBuilder()`, i.e. a *fresh* `Environment` per call — it shares nothing and needs no lock, yet it is the expensive part (bala read + project/env setup). Holding the global monitor across it makes every trigger-model/metadata lookup strictly serial (`TriggerModelReader:403`, `TriggerModelReader:465`, `LibraryMetadataReader:249`), e.g. cross-module introspection over N referenced packages now runs one-at-a-time with a full bala load in each slot. Please hoist the resolve out and leave the load outside the lock.

2. **The lock doesn't match the parameter.** The method takes an arbitrary `buildProject`, but locks the sample-project-specific monitor. Today all callers pass `PackageUtil.getSampleProject()`, so it happens to be correct — but if anyone passes a user project, the lock protects nothing relevant (that project's resolver is still reachable unlocked via `getModulePackage`/`resolveVersionedModuleBala`) while still serializing unrelated work. Either drop the parameter (or assert `buildProject == SAMPLE_PROJECT`), or lock per project the way `PROJECT_LOCKS` does.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
```diff
@@ -398,42 +402,48 @@ private static Optional<Path> resolveLatestModuleBala(BuildProject buildProject,
...
-                Collections.singletonList(ResolutionRequest.from(pkgMetadata.get().resolvedDescriptor())),
-                ResolutionOptions.builder().setOffline(true).build());
-        Optional<ResolutionResponse> resolutionResponse = resolutionResponses.stream().findFirst();
-        if (resolutionResponse.isEmpty()) {
-            return Optional.empty();
-        }
+            Collection<ResolutionResponse> resolutionResponses = packageResolver.resolvePackages(
+                    Collections.singletonList(ResolutionRequest.from(pkgMetadata.get().resolvedDescriptor())),
+                    ResolutionOptions.builder().setOffline(true).build());
+            Optional<ResolutionResponse> resolutionResponse = resolutionResponses.stream().findFirst();
```
LakshanWeerasinghe, line 422: Pre-existing, but this hunk rewrites the exact lines, so worth fixing while here: `findFirst()` only checks emptiness, not resolution status. `resolveResponse`'s own javadoc in this file documents this precise trap — `resolvePackages` can return a non-empty collection whose single entry is `UNRESOLVED` with a `null` `resolvedPackage()`, so line 429 `resolutionResponse.get().resolvedPackage().project()` NPEs. Reachable whenever the metadata query resolves a descriptor but the bala itself isn't in the local cache (e.g. metadata present from a stale index, bala pruned/partially copied). `TriggerModelReader` and `LibraryMetadataReader` catch `Throwable`, so it silently degrades to "module not found" instead of surfacing.

Suggest `.filter(r -> r.resolutionStatus() == ResolutionResponse.ResolutionStatus.RESOLVED)` (or `r.resolvedPackage() != null`, matching `resolveLatestModuleBala`) before `findFirst()`. Separately, this path calls `BalaProject.loadProject(defaultBuilder, balaPath)` without `balaBuildOptions()`, unlike the `loadBalaPackage` helper right above — worth reusing `loadBalaPackage(balaPath)` here so the two stay in sync.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
```diff
@@ -633,21 +624,7 @@ private static String resolveAiModuleVersion(Project project) {
...
-                    Collections.singletonList(resolutionRequest),
-                    ResolutionOptions.builder().setOffline(true).build());
-            return metadataResponses.stream().findFirst()
-                    .filter(meta -> meta.resolutionStatus() != ResolutionResponse.ResolutionStatus.UNRESOLVED)
-                    .map(PackageMetadataResponse::resolvedDescriptor)
-                    .map(descriptor -> descriptor.version().value().toString());
-        } catch (RuntimeException e) {
-            return Optional.empty();
-        }
+        return Optional.ofNullable(PackageUtil.cachedVersion(org, packageName));
```
LakshanWeerasinghe, line 627: Behaviour is equivalent (`PackageVersion.toString()` delegates to the same `SemanticVersion`, and a `null` `resolvedDescriptor()` NPEs into `cachedVersion`'s own `catch (RuntimeException)` → `null` → `Optional.empty()`, matching the old `Optional.map` behaviour), so the dedup is good.

The one thing to be aware of: this call is now a *blocking* one. `resolveAiModuleVersion` runs on the AI/agent request path and only reaches here when `Dependencies.toml` has no pinned `ballerina/ai`, and `cachedVersion` now takes `SAMPLE_RESOLUTION_LOCK` — which `memoizedSampleBala` can hold across a Central pull. Covered in my comment on the lock declaration; no change needed here if that one is addressed.

LakshanWeerasinghe (commented):
### Review: correctness of the new `SAMPLE_RESOLUTION_LOCK` coverage

I verified the mechanical parts first, and they're fine:

* The imports dropped from `AiUtils` (`PackageDescriptor`, `PackageOrg`, `PackageMetadataResponse`, `PackageResolver`, `Resolution*`, `Collections`) are genuinely unused afterwards — `Collection` is still needed and still imported, so this compiles.
* `resolvePackageVersion` is behaviour-preserving: `PackageVersion.toString()` is literally `version.toString()` on the same `SemanticVersion` (verified in `ballerina-lang` bytecode), so `descriptor.version().value().toString()` == `descriptor.version().toString()`. A `null` `resolvedDescriptor()` on a non-`UNRESOLVED` response used to yield `Optional.empty()` and now NPEs inside `cachedVersion`, which its own `catch (RuntimeException)` converts to `null` → `Optional.empty()`. Same result.
* Each `Project` gets its own `Environment` (`EnvironmentBuilder.buildDefault()` allocates a fresh one per call), so the corruption really is scoped to whoever shares `SAMPLE_PROJECT`'s environment. `loadBalaPackage()` uses `getDefaultBuilder()` (fresh env) and is correctly left unlocked.

Two things I'd like addressed before merge — one behavioural regression from the widened lock, and one hole the fix doesn't close (see inline comments for the rest).

#### 1. The lock this reuses is held across Ballerina Central network I/O

`memoizedSampleBala` holds `SAMPLE_RESOLUTION_LOCK` while running its supplier, and both suppliers can go to the network: `resolveVersionedModuleBala` falls back to `resolveResponse(..., offline=false)`, and `resolveLatestModuleBala` calls `RemoteCentral.getInstance().latestPackageVersion(...)` plus an online `resolvePackages` — i.e. the actual package **pull** in `pullModuleAndNotify` happens inside the critical section.

That was tolerable while only cache-miss resolutions contended for it. After this PR, `cachedVersion`, `isModuleUnresolved` and `getModulePackageOffline` — all strictly offline, previously never blocking — queue behind it. Concrete scenario: a user adds a connector, `pullModuleAndNotify` starts a multi-second Central download while holding the monitor; every concurrent `getFlowModel` (`FunctionDataBuilder` → `isModuleUnresolved`), service-model request (`Utils#1659`), trigger-model request (`TriggerModelReader` → `getModulePackageOffline`) and AI request (`AiUtils.resolveAiModuleVersion` → `cachedVersion`) blocks for the full duration of that download, or the HTTP timeout if the network is flaky. Because it's a bare `synchronized` monitor, those threads are also uninterruptible — LSP cancellation can't free them, and the LS request pool can be drained into an apparent freeze.

The PR description says "offline resolutions now route through the memoized cache", but none of the three newly-locked methods are memoized — each call still does a fresh resolver round-trip under the global monitor. Actually memoizing them (they answer immutable questions: `org:name` → provisioned version, `org:name:version` → resolvable) would both match the description and keep these cheap queries from repeatedly queueing behind the pull. Failing that, please consider a `ReentrantLock` with `tryLock(timeout)` for the read-only queries (degrade to "unknown" instead of blocking), or a separate lock/scope for the network-bound pull.

#### 2. Still-unguarded sharers of `SAMPLE_PROJECT`'s environment (this is the "affects other places" part)

`LibraryMetadataReader` builds its local repository handle straight off the sample project's environment:

```java
// LibraryMetadataReader.java:180-181
private static final PackageRepository INSTANCE = BallerinaUserHome.from(
        PackageUtil.getSampleProject().projectEnvironmentContext().environment()).localPackageRepository();
```

`localRepository().getPackage(...)` (`getCompiledPackageFromLocalRepository`, reached from `getTriggerMetadataModelFromLocalRepository` / `getTriggerUISchemaModelFromLocalRepository` / `localPackageRoot`) lands in `FileSystemRepository.getPackage`, which does `ProjectEnvironmentBuilder.getBuilder(environment)` + `BalaProject.loadProject` — i.e. it loads into **the same `Environment`**, whose `ProjectPackageCache.cache(...)` delegates to the shared `EnvironmentPackageCache` (plain `HashMap`s). Worse, `TriggerModelReader#synthesizeTriggerModel` then calls `PackageUtil.getCompilation(pkg)` on that package, so `PackageResolution` drives the **shared sample resolver** during compilation — entirely outside `SAMPLE_RESOLUTION_LOCK`, and `PROJECT_LOCKS` doesn't help because it keys on the bala source root, not on the resolver.

So two concurrent trigger-model requests — one resolving from the local repository, one going through `getModulePackageOffline`/`cachedVersion` — can still corrupt exactly the `EnvironmentPackageCache` this PR is trying to protect, and #2193 would reproduce. If the local-repository path is meant to be covered, it needs the same lock (or the local repository should be built on its own `EnvironmentBuilder.buildDefault()` environment, which would decouple it entirely — probably the cleaner fix).

Adjacent, same class of bug, outside this PR's stated scope but worth a follow-up: `AgentsManagerService.java:423-433` calls `resolvePackageMetadata` on the **user project's** resolver obtained from `workspaceManager.loadProject(...)`. That environment is shared with the workspace manager and the compiler, so if the premise "`PackageResolver` is not thread-safe" holds, that call races with concurrent compilations of the same project.

PR #922, opened 2026-08-20
Lower the Copilot orb below side panels and modals

RNViththagan:
Related to https://github.com/wso2/product-integrator/issues/2155

The floating orb rendered at z-index 10000, above every side panel and modal (1900–2100), so it covered the save/cancel buttons at the bottom of side-panel forms.

- Lower the orb from z-index 10000 to 1800 and the mini chat from 10001 to 1801, so an open side panel, popup or modal renders over them and its controls stay reachable — panels take priority when open
- Widen the data mapper import panel to 400 (matching every other panel) so the orb's idle invite box no longer pokes past its edge

- Lowered the Copilot floating orb z-index to `1800` and the mini chat z-index to `1801`.
- Allows side panels, popups, and modals with higher z-index values to remain accessible.
- Increased the data mapper import panel width to `400` so the orb invite box stays within the panel boundary.

### DIFF
--- a/packages/ballerina-visualizer/src/components/AgentStatusOrb/MiniChat.tsx
+++ b/packages/ballerina-visualizer/src/components/AgentStatusOrb/MiniChat.tsx
@@ -355,7 +355,8 @@ const spin = keyframes`
 
 const Panel = styled.div`
     position: fixed;
-    z-index: 10001;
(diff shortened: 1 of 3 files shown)


SasinduDilshara (approved):

samithkavishke (approved):
LGTM

PR #923, opened 2026-08-20
Show the watched path on SMB and Azure Files entry points

niveathika:
Fixes https://github.com/wso2/product-integrator/issues/2233

## Purpose
In the project explorer, the three file-based triggers rendered their entry points inconsistently:

| Connector | Before | After |
|---|---|---|
| ftp | `FTP Integration - /` | `FTP Integration - /` |
| smb | `SMB Integration` | `SMB Integration - /incoming` |
| azure.storage.files | `files:Service` | `Azure Files Integration - /invoices` |

Three independent causes:

1. **smb** — `Artifact.moduleAnnotationFields` had an `ftp` entry but no `smb` one, so the builder never looked for the annotation field. smb does emit `@smb:ServiceConfig { path: "/incoming" }` (its `initProperties.path` codedata is `SERVICE_ANNOTATION`, same as ftp), so the value was always in the source, just never read.
2. **Azure Files, wrong map key** — `entryPointMap` was keyed `"files"`, the import prefix, but `Artifact.Builder#icon` sets `module` from `ModuleID#moduleName()`, which is `"azure.storage.files"`. Nothing matched, so `resolveServiceName` fell through and returned the type descriptor verbatim. Same class of dead key that #35faff8990 worked around for SAP JCo (`"jco"` vs `sap.jco`) by adding `serviceTypeMap`.
3. **Azure Files, path is the attach point** — Azure Files keeps its watched path in the service attach point, not an annotation. `ServiceConfiguration` in `ballerinax/azure.storage.files:0.9.0` holds only `recursive`, `fileNamePattern` and `minFileAgeSeconds`; its doc states *"The watched path itself is the service's attach point (for example `service /invoices on lsn`)."* But `ModuleNodeTransformer#transform` resolved `typeDescriptor()` first and never fell through to `absoluteResourcePath()`.

While tracing (3) I found a second, related bug: the Azure Files `readOnlyMetadata` declared `kind: SERVICE_ANNOTATION, path: "ServiceConfig.path"`. Since that field does not exist, `AnnotationExtractor` resolves nothing and the **"Monitored Path" chip on the service card is permanently blank**.

Resolves N/A — found during a review of the file-trigger models, no tracking issue filed.

## Goals
Every file-based trigger shows its watched path in the project explorer, resolved from wherever that connector actually stores it, and the Azure Files service card shows its monitored path.

## Approach
`Artifact.java`
- Added `"smb" -> {"path"}` to `moduleAnnotationFields`.
- Re-keyed the Azure Files `entryPointMap` entry from `"files"` to `"azure.storage.files"`, and aligned the label with the trigger palette's `displayName` ("Azure Files Integration", was "Azure Storage Files Integration").
- Added `attachPointNamedModules` (Azure Files and SMB) plus a static `usesAttachPointAsName(String module)` beside `resolveServiceName`, guarded on `entryPointMap` so a member with no entry-point label cannot emit a bare path. SMB's `@smb:ServiceConfig { path }` still wins when present, so the attach point is only its fallback.
- `unquote` strips the quotes a string-literal attach point arrives with.

`ModuleNodeTransformer.java`
- The attach-point branch is tried before the type-descriptor branch, but only for modules in that set.

Scoping note: the general rule — append the attach point to *any* entry point that has one — would also change MCP (`MCP Service - /mcp`) and Salesforce (`Salesforce Event Integration - /data/AccountChangeEvent`), and would arguably make them consistent with how HTTP and GraphQL already read. That is deliberately out of scope here; this PR changes nothing outside the file triggers. Happy to follow up if reviewers prefer the general rule.

`trigger-models/azure.storage.files.json`
- `readOnlyMetadata.path` is now `kind: "SERVICE_BASE_PATH"`, matching how mcp and salesforce already declare a base-path chip.

### Behaviour changes worth calling out
These fell out of the fix rather than being its subject, so they are easy to miss in the diff:

- **`unquote` is not scoped to the attach-point set.** Any connector whose service uses a string-literal attach point now loses the quotes from its explorer label, so asb/solace/mqtt-style services are affected too. The rabbitmq expectation was the only one carrying quotes, so nothing else moves.
- **Azure Files `multipleListenersAllowed`: `true` -> `false`.** Azure Files, unlike ftp and smb, does not support multiple listener attachment. `ServiceRules.resolveCardinality` reads this flag as permissive-unless-forbidden, so `false` makes the Copilot contract reject `service files:Service /invoices on lsnA, lsnB` - the intended effect, not a side effect. ftp and smb keep `true`.
- **Azure Files `identifier.presence`: `optional` -> `required`.** `TriggerModelSynthesizer` maps presence onto the synthesized property's `optional`, which disagreed with the trigger model marking `path` as `optional: false`. Azure Files has no path annotation, so the attach point is the only carrier of the watched path and optional would permit a service watching nothing.
- **smb's `identifier` and `$pathSource` rule are removed**, aligning smb with ftp, which already carried neither on `staging/5.14.0`. Note the consequence: with `smb:SmbServiceConfig.path` declared `string path?`, nothing in either manifest now requires an SMB service to state its watched path. Pre-existing for ftp; worth a separate fix for both.

## UI Component Development
N/A — no UI components changed; the labels are produced by the language server.

## User stories
As an integration developer with several file triggers in one project, I can tell them apart in the project explorer by the path each one watches, without opening each service.

## Release note
The project explorer now shows the monitored path on SMB and Azure Files service entry points, and Azure Files entry points are labelled "Azure Files Integration" instead of `files:Service`. Fixed the blank "Monitored Path" field on the Azure Files service card.

## Documentation
N/A — no documented behaviour changes; these are display labels in the project explorer and the service card.

## Training
N/A

## Certification
N/A — display-label fix with no impact on exam content.

## Marketing
N/A

## Automation tests
- Unit tests
  - New `ServiceGeneratorTests#testFileTriggerServiceDisplayNames` locks the module-name keying for all three file triggers, so the Azure Files re-key cannot silently regress.
  - New `ArtifactNamingTest` covers the two genuinely new paths directly: `usesAttachPointAsName` (true for `azure.storage.files` and `smb`, false for ftp/http/null/unknown) and `unquote`'s quote stripping.
  - Adjusted a stale comment in `TriggerModelReaderTest#testReadAzureStorageFiles`; the assertion it sits above (`SERVICE_BASE_PATH`) was already correct, the comment claimed the opposite.
  - `ServiceGeneratorTests` 2/2, `TriggerModelReaderTest` 12/12, `TriggerSearchUtilTest` 9/9. `compileJava`, `compileTestJava`, `checkstyleMain`, `checkstyleTest` all clean.
  - The composed end-to-end label still has no fixture: `ballerinax/azure.storage.files` is not in `build-config/ballerina_dependencies`, so a RabbitMQ-style `azure.storage.files.bal` fixture cannot resolve until the library is provisioned. `usesAttachPointAsName` and `unquote` are unit-tested directly in the meantime; the composed label was verified by hand against a generated Azure Files service.
- Integration tests
  - None added.

## Security checks
- Ran FindSecurityBugs plugin and verified report? no — display-label change with no new input handling or IO

## Samples
N/A

## Related PRs
None.

## Migrations (if applicable)
N/A — no persisted state; labels are recomputed from source on every artifact generation.

## Test environment
- JDK 21 (Temurin), Gradle 8.11.1
- Ubuntu on WSL2 (kernel 6.6.87.2)
- Read against `ballerina/ftp:2.20.2`, `ballerina/smb:2.0.1`, `ballerinax/azure.storage.files:0.9.0`

## Learning
The two naming layers are easy to conflate and were the root of this bug: `codedata.moduleName` in the trigger models carries the **import prefix** (`files`), while `Artifact.module` carries the **semantic module name** from `ModuleID#moduleName()` (`azure.storage.files`). `entryPointMap` is keyed by the latter, so any entry written with a prefix silently never matches.

Worth flagging for a follow-up: `entryPointMap` still holds several such dead keys — `jco`, `business`, `chat`, `github`, `hubspot`, `shopify`, `twilio` — where the real module names are `sap.jco`, `whatsapp.business`, `googleapis.chat`, `trigger.github`, `trigger.hubspot`, `trigger.shopify`, `trigger.twilio` (the `trigger-models/` filenames are the module names). I left them alone to keep this PR scoped.

Separately, a review of the three connectors' `trigger-metadata-models` turned up divergences that need regenerating the bundled models, so they are not in this PR: smb marks `$functionConfig` as `presence: required` where ftp and azure mark it optional (none of `FunctionConfiguration`'s fields are required); smb marks `$serviceConfig` required, which contradicts its own `$pathSource` rule; only smb carries a `$pathSource` rule at all; and handler shapes drift across the three (`contents` vs `content`, `ftpError` vs `err`, differing parameter order, and `onError` carrying `$functionConfig` only on ftp).

- Fix file-trigger entry-point labels for FTP, SMB, and Azure Files.
- Resolve SMB paths from `@smb:ServiceConfig`.
- Map Azure Files by `azure.storage.files` and resolve paths from the service attach point.
- Use `SERVICE_BASE_PATH` for Azure Files `readOnlyMetadata`.
- Refactor attach-point naming and quote handling.
- Align parameter names and ordering in Copilot schema tests.
- Add coverage for naming, display labels, and Azure Files metadata.


LakshanWeerasinghe (dismissed):

LakshanWeerasinghe (dismissed):

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/artifactsgenerator/Artifact.java
```diff
@@ -287,14 +295,23 @@ public Builder child(Artifact child) {
...
             } else {
-                this.name = entryPointMap.get(module) + " - " + path;
+                this.name = entryPointMap.get(module) + " - " + attachPoint;
             }
             return this;
         }
 
+        // Azure Files-style triggers watch the path in `service files:Service "/invoices" on lsn`, so the attach
+        // point is the only place the watched path appears.
+        public boolean usesAttachPointAsName() {
```
dulajdilshan: Nit: a boolean query on a `Builder` reads a little oddly, and the caller in `ModuleNodeTransformer` still has to repeat the `!resourcePaths.isEmpty()` check. A static helper taking the module name would keep the branch condition in one place.

Also, this is now the fourth module-keyed table in this class (`entryPointMap`, `serviceTypeMap`, `moduleAnnotationFields`, `attachPointNamedModules`). The dead-key cleanup you flagged as a follow-up would be a good moment to fold them into one per-module record.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/test/java/io/ballerina/architecturemodelgenerator/core/ServiceGeneratorTests.java
```diff
@@ -48,6 +48,15 @@ public void testSapJcoServiceDisplayNames() {
                 "SAP JCo RFC Service");
     }
 
+    @Test(description = "Resolve file trigger entry points by their semantic module name")
+    public void testFileTriggerServiceDisplayNames() {
```
dulajdilshan, line 52: This test only covers the static map lookup, which the PR didn't really change. Neither of the two genuinely new code paths — the attach-point branch and `unquote` — is covered here. Right now `unquote`'s only coverage is incidental, via the updated RabbitMQ expected file.

I know the description says a compiled fixture would be needed, but `architecture-model-generator-ls-extension/src/test/resources/artifacts/config/rabbitmq.json` shows that style of test already works. An equivalent `azure.storage.files` fixture would lock in `Azure Files Integration - /invoices` and stop the branch from silently regressing.

packages/ballerina-language-server/model-generator-commons/src/main/resources/trigger-metadata-models/azure.storage.files/trigger-metadata.json
```diff
@@ -17,7 +17,7 @@
       "doc": "Receives files detected on the watched Azure Files share, typed by extension, with an untyped fallback and an error handler.",
       "type": { "name": "Service" },
       "concrete": false,
-      "multipleListenersAllowed": true,
+      "multipleListenersAllowed": false,
```
dulajdilshan, line 20: This flip from `true` to `false` isn't mentioned anywhere in the description and doesn't seem related to the path fix. Is it intentional? If so, could you add the reason to the description — it changes what users can build, unlike the rest of the PR.

packages/ballerina-language-server/model-generator-commons/src/main/resources/trigger-metadata-models/azure.storage.files/trigger-metadata.json
```diff
@@ -17,7 +17,7 @@
...
       "annotations": ["$serviceConfig"],
       "identifier": { "presence": "optional", "form": ["basePath"] },
```
dulajdilshan: This says the identifier is `optional`, but `trigger-models/azure.storage.files.json` marks the `path` property `"optional": false`. If the attach point is where Azure Files keeps its watched path, shouldn't this be `required` so the two agree?

packages/ballerina-language-server/model-generator-commons/src/main/resources/trigger-metadata-models/ftp/trigger-metadata.json
```diff
@@ -241,8 +241,8 @@
             "annotations": ["$functionConfig"],
             "params": [
               {
-                "id": "$service.onError.ftpError",
-                "name": "ftpError",
+                "id": "$service.onError.err",
+                "name": "err",
```
dulajdilshan, line 245: Renaming this to `err` matches what `trigger-models/ftp.json` already ships and what smb/azure use, so it's the right direction.

One leftover though: `service-model-index-generator/src/main/resources/service_artifacts.json` still has `ftpError` for this same parameter. Nothing breaks today (the schema path wins for FTP, per `testSchemaDrivenLibraryIsNeverSubstitutedByTheIndex`), but the PR leaves two sources of truth disagreeing. Worth updating here or filing a follow-up.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger-models/azure.storage.files.json
```diff
@@ -1091,7 +1091,7 @@
           "selected": false
         }
       ],
-      "value": "\"/\"",
+      "value": "/",
```
dulajdilshan, line 1094: This matches how `mcp.json` and `salesforce.json` declare their base-path defaults, so it's consistent.

One thing worth confirming: `SchemaDrivenSourceGenerator.findBasePath` returns this value and it's written straight into the source between the descriptor and `on`, with no quoting or identifier escaping. With the default now showing the unquoted form, a user is likely to type `/invoices-2026` — and hyphens, dots and spaces aren't legal in an unescaped Ballerina path segment, so that source wouldn't compile.

mcp and salesforce get away with this because their paths are fixed strings; an Azure Files directory name is user input. Does the form escape or quote such values somewhere? If not, keeping the quoted string-literal default (and letting the new `unquote` handle the display side) would be the safer pairing.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger-models/azure.storage.files.json
```diff
@@ -4029,10 +4029,9 @@
   ],
   "readOnlyMetadata": [
     {
-      "key": "path",
+      "key": "basePath",
```
dulajdilshan: Confirming for other reviewers that renaming the key was necessary, not just tidy-up: `TriggerReadOnlyMetadataAdapter` routes `SERVICE_BASE_PATH` to `ServiceDescriptionExtractor`, which only recognises `basepath`, `attachpoint`, `servicetype` and `servicename`. With the old `path` key the chip would have stayed blank even after fixing the `kind`.

packages/ballerina-language-server/model-generator-commons/src/main/resources/trigger-metadata-models/azure.storage.files/trigger-metadata.json
NipunaRanasinghe, line 20: This flip isn't mentioned anywhere in the PR description or the commit messages, and ftp/smb keep `true` in the same PR. `ServiceRules.resolveCardinality` reads this flag as permissive-unless-forbidden, so `false` now actively rejects `service files:Service /invoices on lsnA, lsnB` in the Copilot contract, which was previously allowed. If Azure Files does support multiple listeners, this makes Copilot refuse valid source. Was it intentional?

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/artifactsgenerator/Artifact.java
```diff
@@ -136,9 +137,16 @@ public record Artifact(String id, LineRange location, String type, String name,
...
             "ftp", new String[]{"path"},
+            "smb", new String[]{"path"},
             "rabbitmq", new String[]{"queueName"}
     );
 
+    /**
+     * Modules that carry the service's watched path in its attach point instead of a service annotation, so the
+     * entry-point label appends the attach point even though the service also has a type descriptor.
+     */
+    private static final Set<String> attachPointNamedModules = Set.of("azure.storage.files");
```
NipunaRanasinghe: smb looks like it belongs in this set too. For `service smb:Service "/incoming" on lsn` without `@smb:ServiceConfig { path }`, `trySetNameFromAnnotation` returns false, `usesAttachPointAsName()` is false, and the type-descriptor branch wins, so the label is just `SMB Integration` and the path is dropped. Two such services are then indistinguishable in the explorer, which is the same bug this PR fixes for Azure Files. It also reads oddly next to the smb manifest changes here, which delete smb's `identifier` and the `$pathSource` rule that said the identifier is a valid source for the watched path.

packages/ballerina-language-server/model-generator-commons/src/main/resources/trigger-metadata-models/smb/trigger-metadata.json
```diff
@@ -301,23 +301,13 @@
             { "kind": "handler", "id": "$service.onFile" },
             { "kind": "handler", "id": "$service.onFileDelete" }
           ]
-        },
-        {
-          "id": "$pathSource",
```
NipunaRanasinghe, line 306: With this rule gone and `identifier` removed above, nothing in the manifest requires an SMB service to state its watched path any more. `$serviceConfig` with `presence: required` only requires the annotation to be present, not its `path` field, and this rule's `prefer: "fromAnnotation"` was what pointed a consumer at that field. Copilot can now emit `@smb:ServiceConfig {}` with no path and the service quietly watches whatever the listener defaults to. The service-designer UI is fine either way since `trigger-models/smb.json` still has `path` as `optional: false`.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/artifactsgenerator/Artifact.java
```diff
@@ -287,14 +295,23 @@ public Builder child(Artifact child) {
         }
 
         public Builder serviceNameWithPath(String path) {
+            // A string-literal attach point (`service files:Service "/invoices" on lsn`) reaches here with its
+            // quotes, unlike the identifier form (`service /invoices on lsn`).
+            String attachPoint = unquote(path);
```
NipunaRanasinghe, line 301: `unquote` here applies to every module, not just Azure Files, so any connector whose service uses a string-literal attach point gets a different explorer label. The rabbitmq fixture was updated for this; I grepped and it's the only expectation carrying the quotes, so nothing breaks. Just worth calling out in the description, since it's a display change for asb/solace/mqtt-style services that the PR says are out of scope.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/artifactsgenerator/Artifact.java
```diff
@@ -287,14 +295,23 @@ public Builder child(Artifact child) {
...
-                this.name = entryPointMap.get(module) + " - " + path;
+                this.name = entryPointMap.get(module) + " - " + attachPoint;
             }
             return this;
         }
 
+        // Azure Files-style triggers watch the path in `service files:Service "/invoices" on lsn`, so the attach
+        // point is the only place the watched path appears.
+        public boolean usesAttachPointAsName() {
+            return module != null && attachPointNamedModules.contains(module);
```
NipunaRanasinghe: Minor, for future safety: if someone adds a module here without a matching `entryPointMap` entry, `serviceNameWithPath` falls into `this.name = attachPoint` and the label becomes a bare `/invoices` with no connector name, which is worse than the `serviceName` branch this now pre-empts. Adding `&& entryPointMap.containsKey(module)` would keep the two collections consistent.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/test/java/io/ballerina/architecturemodelgenerator/core/ServiceGeneratorTests.java
```diff
@@ -48,6 +48,15 @@ public void testSapJcoServiceDisplayNames() {
                 "SAP JCo RFC Service");
     }
 
+    @Test(description = "Resolve file trigger entry points by their semantic module name")
```
NipunaRanasinghe, line 51: This locks the map key, which is useful, but it tests `resolveServiceName`, which is the path Azure Files now routes around (it goes through `usesAttachPointAsName` into `serviceNameWithPath`). So `attachPointNamedModules`, the new `ModuleNodeTransformer` branch, and `unquote` are all still uncovered. `unquote` at least is a pure static and can be tested directly without a compiled fixture.

packages/ballerina-language-server/model-generator-commons/src/main/resources/trigger-metadata-models/ftp/trigger-metadata.json
NipunaRanasinghe, line 245: `service_artifacts.json:711` still has this param as `ftpError`, so after this rename the SQLite index and the schema disagree for ftp. It stays latent because the schema path wins today (see `testSchemaDrivenLibraryIsNeverSubstitutedByTheIndex`), but anything falling back to the index will emit the old name. Worth updating in the same PR.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/artifactsgenerator/Artifact.java
```diff
@@ -287,14 +295,23 @@ public Builder child(Artifact child) {
...
-                this.name = path;
+                this.name = attachPoint;
             } else {
-                this.name = entryPointMap.get(module) + " - " + path;
+                this.name = entryPointMap.get(module) + " - " + attachPoint;
             }
             return this;
         }
 
+        // Azure Files-style triggers watch the path in `service files:Service "/invoices" on lsn`, so the attach
```
NipunaRanasinghe: Nit: the example here uses the quoted form, but this PR changes `trigger-models/azure.storage.files.json` to generate the unquoted attach point, and the description quotes the library doc as `service /invoices on lsn`. Probably clearer to show the unquoted form.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger-models/azure.storage.files.json
niveathika: We only support /a/B on the service definition. Only resource paths are supported with azure files and using quoted should throw an error

packages/ballerina-language-server/model-generator-commons/src/main/resources/trigger-metadata-models/azure.storage.files/trigger-metadata.json
niveathika: This is intentional. Azure files unlike ftp and smb does not support multiple listener attachment

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/artifactsgenerator/ModuleNodeTransformer.java
```diff
@@ -159,7 +159,9 @@ public Optional<Artifact> transform(ServiceDeclarationNode serviceDeclarationNod
         Optional<TypeDescriptorNode> typeDescriptorNode = serviceDeclarationNode.typeDescriptor();
         NodeList<Node> resourcePaths = serviceDeclarationNode.absoluteResourcePath();
         if (!serviceBuilder.trySetNameFromAnnotation(serviceDeclarationNode)) {
-            if (typeDescriptorNode.isPresent()) {
+            if (!resourcePaths.isEmpty() && serviceBuilder.usesAttachPointAsName()) {
```
NipunaRanasinghe: Nit: this repeats the `serviceNameWithPath(getPathString(resourcePaths))` call from the last branch. The two could fold into one condition:

```java
if (!resourcePaths.isEmpty() && (serviceBuilder.usesAttachPointAsName() || typeDescriptorNode.isEmpty())) {
    serviceBuilder.serviceNameWithPath(getPathString(resourcePaths));
} else if (typeDescriptorNode.isPresent()) {
    ...
```

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/artifactsgenerator/Artifact.java
```diff
@@ -136,9 +137,16 @@ public record Artifact(String id, LineRange location, String type, String name,
             "postgresql", new String[]{"tables"},
             "mysql", new String[]{"tables"},
             "ftp", new String[]{"path"},
+            "smb", new String[]{"path"},
             "rabbitmq", new String[]{"queueName"}
     );
 
+    /**
+     * Modules that carry the service's watched path in its attach point instead of a service annotation, so the
```
NipunaRanasinghe: Nit: "Modules ... in its attach point" — should be "their attach point".

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/test/java/io/ballerina/architecturemodelgenerator/core/ServiceGeneratorTests.java
```diff
@@ -48,6 +48,15 @@ public void testSapJcoServiceDisplayNames() {
...
+    public void testFileTriggerServiceDisplayNames() {
+        Assert.assertEquals(Artifact.resolveServiceName("ftp", "ftp:Service"), "FTP Integration");
```
NipunaRanasinghe: Nit: the second argument is ignored for any module found in `entryPointMap`, so these three assertions would pass with any string there. They pin the map keys, which is the point, but the `ftp:Service` / `files:Service` values read as if they matter.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/artifactsgenerator/Artifact.java
niveathika: Made it a static `Artifact.usesAttachPointAsName(String module)` beside `resolveServiceName`, so the module-keyed tables are only ever consulted by statics; the transformer now calls it with `serviceBuilder.module()`.

Being straight about the second half of your first point: the `!resourcePaths.isEmpty()` check can't collapse into the helper — the transformer owns `resourcePaths` and the third branch needs the same check regardless. So the Builder trades the domain predicate for a plain `module()` accessor rather than losing a query method entirely.

Leaving the four-table fold for the dead-key cleanup as you suggested — it touches ~30 entries and would swamp a label fix.

25543d4

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/test/java/io/ballerina/architecturemodelgenerator/core/ServiceGeneratorTests.java
niveathika: Added `ArtifactNamingTest` covering both new paths directly: `usesAttachPointAsName` (true for `azure.storage.files`, false for ftp/smb/http/null) and `unquote`'s quote stripping. It lives in `io.ballerina.artifactsgenerator` so `unquote` stays package-private instead of being widened for a test; registered in `testng.xml`.

On the fixture — it's blocked rather than declined: `ballerinax/azure.storage.files` isn't in `build-config/ballerina_dependencies`, so an `azure.storage.files.bal` fixture wouldn't resolve. RabbitMQ works because `ballerinax/rabbitmq` *is* provisioned; `CopilotSchemaServicesTest` names azure.storage.files among the seven unprovisioned libraries for the same reason. Once it's provisioned the fixture is the better test, and I'd rather add it then than pull dependency provisioning into this PR.

25543d4

packages/ballerina-language-server/model-generator-commons/src/main/resources/trigger-metadata-models/azure.storage.files/trigger-metadata.json
niveathika: Good catch — changed to `"presence": "required"`. Azure Files has no `path` annotation the way ftp and smb do, so the attach point is the only carrier of the watched path; optional would permit a service watching nothing. `TriggerModelSynthesizer` maps presence straight onto the synthesized property's `optional`, which is exactly what disagreed with the model's `optional: false`. `TriggerModelReaderTest` and `TriggerModelSynthesizerTest` still pass.

25543d4

packages/ballerina-language-server/model-generator-commons/src/main/resources/trigger-metadata-models/ftp/trigger-metadata.json
niveathika: Leaving this one: that path isn't in use now, as you note the schema-driven model wins for FTP, and Lakshan is cleaning up the legacy index entries separately.

Worth flagging it's not only the JSON — the checked-in `service-index.sqlite` carries `ftpError` too, so a JSON-only rename would desync the index from the file that seeds it. Better done as one regeneration in that cleanup.

packages/ballerina-language-server/model-generator-commons/src/main/resources/trigger-metadata-models/azure.storage.files/trigger-metadata.json
niveathika: Confirmed intentional — Azure Files, unlike ftp and smb, doesn't support multiple listener attachment, so the Copilot contract rejecting `service files:Service /invoices on lsnA, lsnB` is the effect I wanted rather than a side effect. Thanks for pinning down `resolveCardinality`'s permissive-unless-forbidden reading; that's the precise mechanism and it's what makes the flip meaningful.

You and @dulajdilshan both landed on this, so I've added a "Behaviour changes worth calling out" section to the description covering it.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/artifactsgenerator/Artifact.java
niveathika: Agreed, and fixed in 54de0c5 — `smb` now joins `attachPointNamedModules`.

Since `trySetNameFromAnnotation` runs first, `@smb:ServiceConfig { path }` still wins whenever it's present, so this only changes the annotation-less case you described: `service smb:Service "/incoming" on lsn` now reads `SMB Integration - /incoming` instead of dropping the path.

On the manifest oddity you noticed alongside it — see the `$pathSource` thread; that removal was aligning smb to ftp, which already carried neither.

packages/ballerina-language-server/model-generator-commons/src/main/resources/trigger-metadata-models/smb/trigger-metadata.json
niveathika: You're right, and it isn't compiler-caught either: `smb:SmbServiceConfig` declares `string path?`, so `@smb:ServiceConfig {}` compiles and nothing rejects a pathless service.

The removal was deliberate alignment rather than an oversight — ftp on `staging/5.14.0` already had no `identifier` and no `$pathSource`, so smb was the lone outlier carrying both. But that makes the gap pre-existing for ftp and now shared by smb, which is a worse position than I'd framed it, so thanks for pushing on it.

I'd rather fix it for both together than re-add the rule for smb alone here: with `identifier` gone, `prefer: "fromAnnotation"` has nothing left to choose against, so the rule would need reshaping into a plain "annotation field required" constraint that ftp should get too. Noted in the description; leaving this thread open for that follow-up.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/artifactsgenerator/Artifact.java
niveathika: Correct, and thanks for grepping it rather than just flagging it. Now documented in the description: the broader effect is intended, and since rabbitmq was the only expectation carrying quotes, no other connector's label actually moves.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/artifactsgenerator/Artifact.java
niveathika: Applied in 54de0c5 — `usesAttachPointAsName` now also requires `entryPointMap.containsKey(module)`, so a member with no entry-point label falls through to the `serviceName` branch instead of emitting a bare path. `ArtifactNamingTest` covers the unknown-module case.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/test/java/io/ballerina/architecturemodelgenerator/core/ServiceGeneratorTests.java
niveathika: Addressed in 25543d4 — `ArtifactNamingTest` now covers `usesAttachPointAsName` and `unquote` directly, taking your point that the pure static needs no compiled fixture. It lives in `io.ballerina.artifactsgenerator` so `unquote` stays package-private instead of being widened for a test. 54de0c5 extends it for the smb case.

The composed end-to-end label is still uncovered: `ballerinax/azure.storage.files` isn't in `build-config/ballerina_dependencies`, so a RabbitMQ-style `.bal` fixture can't resolve until the library is provisioned.

packages/ballerina-language-server/model-generator-commons/src/main/resources/trigger-metadata-models/ftp/trigger-metadata.json
niveathika: Leaving this one: that path isn't in use now, and Lakshan is cleaning up the legacy index entries separately.

Worth flagging it's not only the JSON — the checked-in `service-index.sqlite` carries `ftpError` too, so a JSON-only rename would desync the index from the file that seeds it. Better done as one regeneration in that cleanup.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/artifactsgenerator/Artifact.java
niveathika: Switched to the unquoted form in 54de0c5, matching the library doc and the description.

I kept the quoted example in `serviceNameWithPath`'s inline comment, since that comment exists specifically to explain what `unquote` is there to strip.

LakshanWeerasinghe (approved):

NipunaRanasinghe (approved):
LGTM

PR #937, opened 2026-08-21
Show `Set up AI` modal from the Agent Builder overview when not logged in

dan-niles:
### Purpose

Sending a prompt from the Agent Builder overview silently did nothing when the AI panel was not authenticated. This PR lets the user complete AI setup from the overview itself, then sends the held prompt.

https://github.com/user-attachments/assets/1900cf77-147e-4fe2-9b99-3ea1ded455e6

### Goals

- The composer checks the AI machine state before it dispatches a prompt.
- If setup is incomplete, the login flow opens in a `PopupModal` and the prompt is held. When the state reaches `Authenticated`, the prompt is dispatched.
- A footer link opens the same modal on demand: **Set up AI to generate** when unauthenticated, **Finish setting up AI** while a provider flow is open.
- A notice appears if the run never starts, so the prompt is not lost.

### Approach

**Overview (`AgentBuilderOverview/EmptyState.tsx`)**
- Reads `getAIMachineSnapshot()` and refreshes on `onAIPanelStateChanged`, mapping the machine state to `unknown | authenticated | unauthenticated | connecting | disabled`.
- Renders `LoginPanel`, `WaitingForLogin`, or `DisabledWindow` inside the modal, per that state.

**Shared auth surfaces**
- `LoginPanel` and `WaitingForLogin` take an `embedded` prop, so they drop the full-viewport wrapper when they render inside the modal.
- `AlertBox` now exports `AuthPanel` / `AuthTitle` / `AuthSubTitle` / `AuthActions`. `WaitingForLogin` uses them instead of its own duplicated `AlertContainer`, and its cancel button becomes a leading **Back** action.
- `LoginPanel` shows the orb in its header and lays the provider cards out in a grid on wide viewports.

**Supporting changes**
- `OrbHolder` moves from `EmptyState` into `AgentStatusOrb/shared` as `OrbAura`, so the login header can reuse it.
- `PopupModal` gains `autoHeight` and `maxWidth` for the shorter auth forms.
- `assistantTagline` / `useAssistantTagline` replace the hardcoded Integrator tagline in the login header.
- `RPCLayer` forwards AI state transitions to the visualizer webview. Before, only the AI panel webview received them.

## Summary

- Enable AI setup from the Agent Builder overview.
- Hold prompts during authentication and resume them after setup.
- Add setup actions, state-based labels, and run-start failure feedback.
- Reuse embedded authentication components and support configurable modal sizing.
- Add product-specific assistant taglines and shared orb styling.
- Forward AI state transitions to the visualizer webview.

### DIFF
--- a/packages/ballerina-core/src/state-machine-types.ts
+++ b/packages/ballerina-core/src/state-machine-types.ts
@@ -52,6 +52,15 @@ export function shortAssistantName(mode: ProductMode): string {
     return SHORT_ASSISTANT_NAMES[mode];
 }
 
+const ASSISTANT_TAGLINES: Record<ProductMode, string> = {
+    [ProductMode.INTEGRATOR]: 'Your AI pair programmer for integration development',
+    [ProductMode.AGENT_BUILDER]: 'Your AI partner for building agents'
+};
+
+export function assistantTagline(mode: ProductMode): string {
+    return ASSISTANT_TAGLINES[mode];
+}
+
 export type MachineStateValue =
     | 'initialize'
     | 'lsError'
--- a/packages/ballerina-extension/src/RPCLayer.ts
+++ b/packages/ballerina-extension/src/RPCLayer.ts
@@ -60,6 +60,7 @@ const DISRUPTIVE_TRANSITION_EVENTS = new Set(['VIEW_UPDATE', 'UPDATE_PROJECT_STR
 
 export class RPCLayer {
     static _messenger: Messenger = new Messenger({ ignoreHiddenViews: false });
+    private static _aiForwardingBound = false;
 
     constructor(webViewPanel: WebviewPanel | WebviewView) {
         if (isWebviewPanel(webViewPanel)) {
@@ -77,9 +78,13 @@ export class RPCLayer {
             window.onDidChangeActiveColorTheme((theme) => {
                 RPCLayer._messenger.sendNotification(currentThemeChanged, { type: 'webview', webviewType: VisualizerWebview.viewType }, theme.kind);
             });
-            AIStateMachine.service().onTransition((state) => {
-                RPCLayer._messenger.sendNotification(aiStateChanged, { type: 'webview', webviewType: VisualizerWebview.viewType }, state.value);
-            });
+            // Register once: onTransition has no unsubscribe, and each delivery costs the visualizer an RPC.
+            if (!RPCLayer._aiForwardingBound) {
+                RPCLayer._aiForwardingBound = true;
+                AIStateMachine.service().onTransition((state) => {
+                    RPCLayer._messenger.sendNotification(aiStateChanged, { type: 'webview', webviewType: VisualizerWebview.viewType }, state.value);
+                });
+            }
         } else if (isMigrationPanel(webViewPanel)) {
             // Migration panel is a WebviewPanel but does not use the AI state machine.
             RPCLayer._messenger.registerWebviewPanel(webViewPanel as WebviewPanel);
--- a/packages/ballerina-extension/src/RPCLayer.ts
+++ b/packages/ballerina-extension/src/RPCLayer.ts
@@ -77,6 +77,9 @@ export class RPCLayer {
             window.onDidChangeActiveColorTheme((theme) => {
                 RPCLayer._messenger.sendNotification(currentThemeChanged, { type: 'webview', webviewType: VisualizerWebview.viewType }, theme.kind);
             });
+            AIStateMachine.service().onTransition((state) => {
+                RPCLayer._messenger.sendNotification(aiStateChanged, { type: 'webview', webviewType: VisualizerWebview.viewType }, state.value);
+            });
         } else if (isMigrationPanel(webViewPanel)) {
             // Migration panel is a WebviewPanel but does not use the AI state machine.
             RPCLayer._messenger.registerWebviewPanel(webViewPanel as WebviewPanel);
--- a/packages/ballerina-visualizer/src/components/PopupModal/index.tsx
+++ b/packages/ballerina-visualizer/src/components/PopupModal/index.tsx
@@ -49,11 +49,15 @@ const Backdrop = styled(PopupOverlay)`
     }
 `;
 
-const Box = styled(PopupContainer) <{ $expanded?: boolean }>`
-    width: ${(props: { $expanded?: boolean }) => props.$expanded ? "90%" : "80%"};
-    max-width: ${(props: { $expanded?: boolean }) => props.$expanded ? "1000px" : "800px"};
-    height: ${(props: { $expanded?: boolean }) => props.$expanded ? "90vh" : "80vh"};
-    max-height: ${(props: { $expanded?: boolean }) => props.$expanded ? "none" : "800px"};
+type BoxProps = { $expanded?: boolean; $autoHeight?: boolean; $maxWidth?: number };
+
+const Box = styled(PopupContainer) <BoxProps>`
+    width: ${(props: BoxProps) => props.$expanded ? "90%" : "80%"};
+    max-width: ${(props: BoxProps) => props.$maxWidth ? `${props.$maxWidth}px` : props.$expanded ? "1000px" : "800px"};
+    transition: max-height 180ms ease, max-width 180ms ease;
+    height: ${(props: BoxProps) => props.$autoHeight ? "auto" : props.$expanded ? "90vh" : "80vh"};
+    max-height: ${(props: BoxProps) => props.$autoHeight ? "80vh" : props.$expanded ? "none" : "800px"};
+    min-height: ${(props: BoxProps) => props.$autoHeight ? "0" : "480px"};
     animation: ${popIn} ${ENTER_MS}ms cubic-bezier(0.16, 1, 0.3, 1) both;
     &.closing {
         animation-duration: ${EXIT_MS}ms;
@@ -98,13 +102,15 @@ export interface PopupModalProps {
     dismissOnBackdropClick?: boolean;
     dismissOnEscape?: boolean;
     expanded?: boolean;
+    autoHeight?: boolean;
+    maxWidth?: number;
     zIndexBase?: number;
     ariaLabelledBy?: string;
     children: (close: () => void) => ReactNode;
 }
 
 export function PopupModal(props: PopupModalProps) {
-    const { onClose, dismissOnBackdropClick, dismissOnEscape, expanded, zIndexBase, ariaLabelledBy, children } = props;
+    const { onClose, dismissOnBackdropClick, dismissOnEscape, expanded, autoHeight, maxWidth, zIndexBase, ariaLabelledBy, children } = props;
     const [closing, setClosing] = useState(false);
     const exitTimer = useRef<ReturnType<typeof setTimeout>>(null);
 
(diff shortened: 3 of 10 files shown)


packages/ballerina-core/src/state-machine-types.ts
```diff
@@ -52,6 +52,15 @@ export function shortAssistantName(mode: ProductMode): string {
...
 }
 
+const ASSISTANT_TAGLINES: Record<ProductMode, string> = {
+    [ProductMode.INTEGRATOR]: 'Your AI pair programmer for integration development',
+    [ProductMode.AGENT_BUILDER]: 'Your AI partner for building agents'
+};
+
+export function assistantTagline(mode: ProductMode): string {
+    return ASSISTANT_TAGLINES[mode];
+}
```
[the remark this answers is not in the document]
dan-niles: `assistantTagline` is a total `Record<ProductMode, string>` lookup, so TypeScript already fails the build on an unmapped mode — the same reason its siblings `assistantName` and `shortAssistantName` carry no test.

packages/ballerina-extension/src/RPCLayer.ts
```diff
@@ -77,6 +77,9 @@ export class RPCLayer {
             window.onDidChangeActiveColorTheme((theme) => {
                 RPCLayer._messenger.sendNotification(currentThemeChanged, { type: 'webview', webviewType: VisualizerWebview.viewType }, theme.kind);
             });
+            AIStateMachine.service().onTransition((state) => {
+                RPCLayer._messenger.sendNotification(aiStateChanged, { type: 'webview', webviewType: VisualizerWebview.viewType }, state.value);
+            });
```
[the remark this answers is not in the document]
dan-niles: On the test: `packages/ballerina-extension/jest.config.js` scopes `roots` to `src/test-support`, so a spec for `src/RPCLayer.ts` would not be collected at all.

On disposal: correct but pre-existing rather than introduced here — the four sibling listeners in this same block are undisposed too and the `else` branch already had this exact `AIStateMachine.onTransition` call before this PR, so I'll give `RPCLayer` one disposal path for all five in a follow-up instead of making the block inconsistent.

packages/ballerina-visualizer/src/views/BI/AgentBuilderOverview/EmptyState.tsx
```diff
@@ -536,6 +567,22 @@ export function EmptyState({ onCreateFromScratch, isLibrary }: EmptyStateProps)
...
+            .catch((): void => undefined);
+    }, [rpcClient]);
+
+    useEffect(() => {
+        if (!rpcClient) {
+            return;
+        }
+        refreshAiSnapshot();
+        rpcClient.onAIPanelStateChanged(refreshAiSnapshot);
+    }, [rpcClient, refreshAiSnapshot]);
```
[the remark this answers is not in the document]
dan-niles: `onAIPanelStateChanged` wraps `vscode-messenger-webview`'s `onNotification`, which is `handlerRegistry.set(type.method, handler)` returning the messenger — so there is no disposer to return, and re-registering replaces the handler rather than accumulating one.

kaumini (commented):
A few follow-ups from reviewing the AI-setup-from-overview flow. Four inline comments below (a modal-reopen bug, a shared `PopupModal` regression, and two subscription-cleanup leaks), each with a suggested fix.

packages/ballerina-visualizer/src/views/BI/AgentBuilderOverview/EmptyState.tsx
```diff
@@ -597,25 +648,115 @@ export function EmptyState({ onCreateFromScratch, isLibrary }: EmptyStateProps)
     const runDetail =
         state === "completed" ? undefined : status?.label ?? (running ? "Working on it…" : undefined);
     const showOpenCopilot = !aiPanelOpen;
+    const authState = toAuthState(aiSnapshot?.state);
+    const substate = authenticatingSubstate(aiSnapshot?.state);
+    const showSetup =
```
kaumini, line 653: `setupRequested` is set by `openSetup` and only ever cleared by `dismissSetup`. When setup is opened from the footer link and login succeeds, the modal closes here (once `authState === "authenticated"`, `showSetup` is false) but `setupRequested` stays `true`. If `authState` later leaves `Authenticated` (token expiry, or sign-out from the AI panel), this condition flips back to true and the modal reopens on the overview with no user action.

Suggest clearing the flag once auth completes, regardless of how setup was opened:

```tsx
// Reset the setup surface once auth completes, however it was opened.
useEffect(() => {
    if (authState === "authenticated") {
        setSetupRequested(false);
    }
}, [authState]);
```

packages/ballerina-visualizer/src/components/PopupModal/index.tsx
```diff
@@ -49,11 +49,15 @@ const Backdrop = styled(PopupOverlay)`
...
-    max-height: ${(props: { $expanded?: boolean }) => props.$expanded ? "none" : "800px"};
+type BoxProps = { $expanded?: boolean; $autoHeight?: boolean; $maxWidth?: number };
+
+const Box = styled(PopupContainer) <BoxProps>`
+    width: ${(props: BoxProps) => props.$expanded ? "90%" : "80%"};
+    max-width: ${(props: BoxProps) => props.$maxWidth ? `${props.$maxWidth}px` : props.$expanded ? "1000px" : "800px"};
+    transition: max-height 180ms ease, max-width 180ms ease;
+    height: ${(props: BoxProps) => props.$autoHeight ? "auto" : props.$expanded ? "90vh" : "80vh"};
+    max-height: ${(props: BoxProps) => props.$autoHeight ? "80vh" : props.$expanded ? "none" : "800px"};
+    min-height: ${(props: BoxProps) => props.$autoHeight ? "0" : "480px"};
```
kaumini, line 60: This adds a `480px` floor to **every** non-`autoHeight` modal. The previous Box had no `min-height`, so all existing consumers (`AddAgentPopup`, `TemplateModal`, `AddLibraryArtifactPopup`, `AIAgentSidePanel`, …) now inherit it. On a short editor viewport (under ~600px tall) the box is forced past its `80vh` height and overflows/clips the viewport — a regression for modals that don't opt into `autoHeight`.

Capping the floor at `80vh` keeps the intent on tall viewports without the regression:

```suggestion
    min-height: ${(props: BoxProps) => props.$autoHeight ? "0" : "min(480px, 80vh)"};
```

packages/ballerina-extension/src/RPCLayer.ts
```diff
@@ -77,6 +77,9 @@ export class RPCLayer {
             window.onDidChangeActiveColorTheme((theme) => {
                 RPCLayer._messenger.sendNotification(currentThemeChanged, { type: 'webview', webviewType: VisualizerWebview.viewType }, theme.kind);
             });
+            AIStateMachine.service().onTransition((state) => {
```
kaumini: This `onTransition` subscriber is registered every time the RPCLayer constructor runs — i.e. every time a visualizer / evaluation / agent-chat `WebviewPanel` is created — and is never removed when the panel is disposed. In xstate v4 `onTransition` returns the interpreter (not an unsubscribe), so nothing cleans it up: reopening a panel accumulates subscribers, and each one re-sends `aiStateChanged` to the visualizer on every AI transition (N duplicate notifications + N `getAIMachineSnapshot` round-trips per transition).

The forwarding doesn't depend on the specific panel — it always targets `VisualizerWebview.viewType` via the static `_messenger` — so registering it once removes the leak entirely:

```ts
private static _aiForwardingBound = false;

// …inside the isWebviewPanel branch of the constructor:
if (!RPCLayer._aiForwardingBound) {
    RPCLayer._aiForwardingBound = true;
    AIStateMachine.service().onTransition((state) => {
        RPCLayer._messenger.sendNotification(aiStateChanged, { type: 'webview', webviewType: VisualizerWebview.viewType }, state.value);
    });
}
```

(The sibling `StateMachine` / `StateMachinePopup` / theme subscriptions above have the same latent issue. The alternative, if you'd rather fix them together, is to capture each subscription and remove it via `service.off(listener)` / `disposable.dispose()` in the panel's `onDidDispose`.)

packages/ballerina-visualizer/src/views/BI/AgentBuilderOverview/EmptyState.tsx
```diff
@@ -536,6 +567,22 @@ export function EmptyState({ onCreateFromScratch, isLibrary }: EmptyStateProps)
...
+            .then(setAiSnapshot)
+            .catch((): void => undefined);
+    }, [rpcClient]);
+
+    useEffect(() => {
+        if (!rpcClient) {
+            return;
+        }
+        refreshAiSnapshot();
+        rpcClient.onAIPanelStateChanged(refreshAiSnapshot);
```
kaumini, line 583: This subscription is never cleaned up — the effect returns no teardown. After `EmptyState` unmounts, the stale `refreshAiSnapshot` keeps running a `getAIMachineSnapshot` RPC and calling `setAiSnapshot` on every AI transition for the life of the webview. `onAIPanelStateChanged` returns `void` today (the underlying messenger stores one handler per method in a Map with no removal API), so the effect can't truly unsubscribe.

Two options:

- **Local mitigation** — guard against post-unmount updates:

```tsx
useEffect(() => {
    if (!rpcClient) {
        return;
    }
    let active = true;
    refreshAiSnapshot();
    rpcClient.onAIPanelStateChanged(() => {
        if (active) {
            refreshAiSnapshot();
        }
    });
    return () => {
        active = false;
    };
}, [rpcClient, refreshAiSnapshot]);
```

- **Proper fix** — have `onAIPanelStateChanged` return a disposable and return it from the effect, mirroring `onProjectContentUpdated` (which already returns an unsubscribe fn). This needs the messenger layer to support handler removal.

packages/ballerina-visualizer/src/views/BI/AgentBuilderOverview/EmptyState.tsx
dan-niles: Fixed in 82f6854 — an effect now clears `setupRequested` once `authState` reaches `authenticated`.

packages/ballerina-visualizer/src/components/PopupModal/index.tsx
dan-niles: `Box` extends `PopupContainer`, which has carried `min-height: 480px` since d851625 (Jan 14), so this line keeps the existing floor rather than adding one — it only relaxes it to `0` for `autoHeight`.

packages/ballerina-extension/src/RPCLayer.ts
dan-niles: Fixed in 82f6854 with the static flag; left the three siblings alone since their handlers absorb duplicates (debounced fetch, idempotent `setState`) whereas this one costs an RPC per delivery.

packages/ballerina-visualizer/src/views/BI/AgentBuilderOverview/EmptyState.tsx
dan-niles: The messenger keeps one handler per method in a `Map`, so the next mount overwrites the stale one rather than stacking — leaving this as-is to match `onStateChanged` / `onPromptUpdated` elsewhere.

kaumini (approved):

PR #940, opened 2026-08-23
Order and group trigger handler form inputs from the UI schema

LakshanWeerasinghe:
Adds an **optional** `layout` block to a trigger UI schema's handler definitions, so a connector author can dictate the order of — and group — the input fields in the handler form.

Until now the handler form's section order was hard-coded in JSX (`TriggerHandlerForm/index.tsx`), and order within a section was whatever the JSON happened to declare. There was no way for a connector to say "put these inputs together under a heading, and show it first".

### Schema shape

```json
"schemaFunctions": [{
  "name": "onMessage",
  "layout": [
    { "fields": ["$variant", "$description", "content", "stream", "*rest"] },
    { "id": "adv", "label": "Advanced", "description": "Rarely needed.",
      "advanced": true, "fields": ["caller", "$headers"] }
  ]
}]
```

- **Opt-in at every level.** No `layout` → the form renders exactly as before; a partial layout only has to name what it wants to move. Every connector that shipped before this keeps its current form.
- An author's own identifiers are used bare (a parameter name, a `properties` key, a payload `bindingGroup`). The form's built-in units use reserved `$`-prefixed ids — `$variant`, `$description`, `$name`, `$documentation`, `$parameters`, `$returnType`, `$headers`. This is load-bearing, not cosmetic: real connectors ship parameters literally named `headers` (mcp) and `parameters` (sap.jco), so bare ids would address the wrong unit.
- `*rest` is the placement directive for everything unclaimed. `*`-prefixed rather than `$` because it names no unit, and because a `properties` key is an arbitrary schema-authored string — a field called `$rest` is possible, whereas nothing can be called `*rest`.
- A labeled section renders as a divider + heading; an unlabeled one is an ordered run that keeps each unit's default chrome. `advanced: true` moves a labeled group inside the collapsed "Advanced Configurations" box.
- An id matching no unit is skipped with a dev-console warning — a handler variant may legitimately lack a field its siblings have.

Layout is **presentation only**. The generated handler signature still follows `parameters` order.

### What this fixes in the shipped models

**`ftp`/`smb` `onFileCsv`** — the Stream flag rendered *above* the Row Schema it modifies, so the user was asked to tick "Stream (Large Files)" before there was a schema to stream. (The flag isn't a handler property in the schema at all: it lives inside the payload parameter's `COMPLEX_PAYLOAD` tree and the LS lifts it out ahead of the payload.) The layout swaps the two.

**`mcp` `StreamableHttpService/newTool`** — the transport-level opt-ins are now grouped:

```
Advanced Configurations
  ☐ Meta
  Transport Parameters
  Access transport-level request data.
  HTTP Headers          + Header
  Request Access
  ☐ Request   ☐ Headers
```

### Changes by layer

| | |
|---|---|
| Schema model | `TriggerUISchemaModel.LayoutSection` + `FunctionModel.layout` |
| Wire model | new `LayoutSection`, `Function.layout`, carried in `TriggerFunctionAdapter.toFunction` so every fanned-out variant inherits it |
| Core types | `HandlerLayoutSection` + `FunctionModel.layout` |
| Frontend | new pure `handlerLayout.ts` (`handlerUnitsOf` / `resolveHandlerLayout`); `TriggerHandlerForm` renders from resolved sections instead of hard-coded JSX |

Synthesis from `trigger-metadata.json` deliberately emits no layout — layout is presentation, and that model is presentation-free by design.

Two hardcoded strings in the form also changed, and so apply to every schema-driven connector: the advanced box is now titled **"Advanced Configurations"** (was "Advanced Parameters") and the add-header link reads **"+ Header"** (was "+ Add Header").

### Testing

- **41 new L1 tests** (`handlerLayout.test.ts`), including a zero-regression invariant that a handler with no `layout` resolves to the form's historical order, and guards for the two real id collisions above.
- **New `TriggerLayoutTest`** — a corpus guard over all 25 bundled models: every `layout` id must resolve to a real unit (against the *wire* model, so lifted composition siblings count), `*rest` at most once, no unknown `*` directive, and no field name colliding with a reserved id.
- 6 new `TriggerFunctionAdapterTest` cases: layout survives the schema → wire hop, every variant inherits it, absent/empty normalises to null.
- Full suites green: 510 service-model-generator + 78 model-generator-commons + 423 ballerina-visualizer. `tsc` and the test-typecheck gate clean. Verified end-to-end by driving the real shipped models through `TriggerFunctionAdapter` into the frontend resolver.

One coverage gap worth naming: the "suppress a unit's own blurb inside a labeled group" rule is JSX-only, and `TriggerHandlerForm` won't instantiate under jsdom (CodeMirror), so it is verified manually rather than by test. Pinning it needs an L4 assertion.

---

**Stacked PR.** Based on `sap-jco-driver-deps` (the SAP JCo driver-dependency work), which still needs its own PR against `staging/5.14.0`. Review that one first; this becomes mergeable once it lands.

## Summary

- Added optional `layout` sections to trigger UI schema models.
- Propagated layouts through wire models and `TriggerFunctionAdapter`.
- Added frontend layout resolution for ordering, grouping, reserved form units, and `*rest`.
- Preserved existing behavior and handler signatures when layouts are omitted.
- Updated FTP/SMB and MCP layouts, advanced section text, and header text.
- Added adapter, layout resolution, bundled-model validation, and UI rendering tests.
- Added a wait for the GraphQL add button in the end-to-end test flow.


packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger-models/ftp.json
```diff
@@ -1130,6 +1130,17 @@
               }
             }
           ],
+          "layout": [
```
NipunaRanasinghe, line 1133: `onFile` (Raw Bytes) has the identical defect and doesn't get the fix.

`onFileCsv` and `onFile` are sibling variants of the **same `onCreate` group** — the user picks between them in the Format dropdown — and both carry the same shape:

```
onFileCsv  param content → subprops { stream: PAYLOAD_MODIFIER, payload: PAYLOAD_TYPE }
onFile     param content → subprops { stream: PAYLOAD_MODIFIER, payload: PAYLOAD_TYPE }
```

`handleVariantChange` re-resolves the layout from the newly selected variant's model, so switching Format from CSV to Raw Bytes flips the Stream checkbox back above the schema it modifies — the exact bug this PR fixes, still live in the same form.

Same in `smb.json`. Suggest adding the same `layout` block to `onFile` in both models. Worth a look at the remaining `onCreate` variants (`onFileJson`/`onFileXml`/`onFileText`) too, so the field order doesn't shift as the user cycles the dropdown — they have no `stream`, so it's ordering consistency rather than a bug.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/test/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerLayoutTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+        for (String id : addressable) {
+            Assert.assertFalse(RESERVED_IDS.contains(id) || REST_DIRECTIVE.equals(id),
+                    where + ": the field \"" + id + "\" collides with a reserved layout id, so a layout "
+                            + "could never address it. Rename the field.");
+        }
+    }
+
+    /** Every bare id an author may write for this handler, unioned over its wire variants. */
+    private Set<String> addressableIds(TriggerUISchemaModel.FunctionModel handler) {
```
NipunaRanasinghe, line 163: This set is a superset of what the form can actually address, so the guard passes ids that still get silently dropped.

`addressableIds` collects *every* parameter name, binding group, and property key. `handlerUnitsOf` only builds units for:

- grouped payload params (`groupedPayloadParametersOf`)
- params with `advanced === true`
- properties whose `codedata.type` is `METADATA_FLAG` / `PAYLOAD_MODIFIER` / `COMPLEX_FUNCTION_ANNOTATION` / `ANNOTATION_ATTACHMENT`
- the `$`-reserved ids and `schema.header`

Plain non-advanced, non-payload params are common in the corpus — `asb/onMessage` `message` and `caller`, `file/onCreate` `event`, `googleapis.chat` `event`/`caller`, `ftp/onError` `ftpError`. An author who writes `"fields": ["message", "*rest"]` for `asb/onMessage` passes this test and gets a `console.warn` plus a dropped field at runtime, which is what the class javadoc says this test exists to prevent.

Suggest narrowing `addressableIds` to the same predicate `handlerUnitsOf` uses, so the build guard and the resolver agree on what's addressable.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
```diff
@@ -803,264 +1173,66 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
...
-                                                marginTop: index === 0 ? 0 : 8,
-                                                description: param.metadata?.description,
-                                            }}
-                                        />
-                                    </CheckBoxGroup>
-                                ))}
-                                {headerTemplate && (
-                                    <>
+                                {looseAdvancedUnits.length > 0 && renderAdvancedBody(looseAdvancedUnits)}
+                                {advancedSections.map((section) => (
```
NipunaRanasinghe: Missing the emptiness guard that `bodySections` has.

`bodySections.map` computes `renderSectionUnits(section)` first and returns `null` when the body is empty. This loop emits `<Typography>{section.label}</Typography>` and the description unconditionally, then calls `renderSectionUnits`. A labeled advanced section whose units all render `null` — a `$description` on a handler with no description, a non-bindable payload, a property that isn't on the model — leaves a bare heading inside "Advanced Configurations".

`resolveHandlerLayout` already drops sections with zero *resolved* units, so this only bites when units resolve but render nothing; none of the three shipped layouts hit it today. Still, it's the same guard for one line.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
```diff
@@ -783,6 +800,359 @@ export function TriggerHandlerForm(props: TriggerHandlerFormProps) {
     const handlerDescription = functionModel.metadata?.description?.trim();
     const showAnnotationsDivider = hasVariants || metadataFlags.length > 0 || modifierFlags.length > 0;
 
+    const sections = resolveHandlerLayout(functionModel, artifactFields.map((field) => field.key));
```
NipunaRanasinghe: This runs in the render body, so it re-derives every unit and re-emits every diagnostic on each render — an authoring mistake `console.warn`s once per keystroke, not once. `buildOrderedArtifactFields` was deliberately kept out of render for the same reason.

`useMemo` on `[functionModel, artifactFields]` covers both.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/handlerLayout.test.ts
```diff
@@ -0,0 +1,572 @@
...
+    it("treats a blank description as absent", () => {
+        const [section] = resolveHandlerLayout(fn({
+            ...mcpNewTool,
+            layout: [{ id: "g", label: "G", description: "   ", fields: [LAYOUT_ID_HEADERS] }],
+        }), MCP_ARTIFACT_KEYS);
+        expect(section.description).toBeUndefined();
+    });
+});
+
+describe("the layouts actually shipped in trigger-models", () => {
```
NipunaRanasinghe: These don't read the shipped models — both tests re-declare hand-built fixtures that mirror `ftp.json` / `mcp.json` (the comment below says as much). Editing either JSON's `layout` won't fail them, despite the describe name.

`TriggerLayoutTest` does read the real models, but it only checks that ids resolve, never order — so "ftp/smb onFileCsv defines the row schema before offering the stream flag" isn't pinned anywhere. Either load the JSON here, or rename the block so it doesn't read as a guard over the shipped models.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/resources/trigger-models/ftp.json
LakshanWeerasinghe: Good catch — added the identical `layout` block to `onFile` in both `ftp.json` and `smb.json` (content before stream), in 05e24122ea.

Left `onFileJson`/`onFileXml`/`onFileText` alone though: they have no `stream` modifier property, so their default (layout-less) resolution order already places `content` before the remaining fields — same result an explicit layout would give. Since there's no live bug there today (as you noted, it'd be consistency rather than a fix), I didn't want to add a purely speculative layout block for variants nothing is currently wrong with.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/test/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerLayoutTest.java
LakshanWeerasinghe: You're right, `addressableIds` was a strict superset of what `handlerUnitsOf` actually renders. Narrowed it to the same predicate — a payload parameter (mirroring `isPayloadParameter` from payloadComposer.ts) or `advanced === true`, plus properties with one of the four addressable codedata types — so the build guard and the resolver agree on what's addressable. `TriggerLayoutTest` still passes 25/25 since none of the currently authored layouts happen to reference a plain non-advanced/non-payload param like `message`/`caller` yet, but it'll now catch it if one does. Fixed in 05e24122ea.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
LakshanWeerasinghe: Fixed — `advancedSections.map` now computes `renderSectionUnits(section)` first and returns `null` when it's empty, same guard `bodySections` already had. In 05e24122ea.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/index.tsx
LakshanWeerasinghe: Wrapped it in `useMemo` keyed on `[functionModel, artifactFields]`, in 05e24122ea. Had to move it above the `if (!functionModel) return null` early return rather than leave it where the old line sat — a hook called after a conditional return would violate the Rules of Hooks (inconsistent hook count between renders). `resolveHandlerLayout` itself is guarded to handle a null `functionModel` inside the memo callback.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/handlerLayout.test.ts
LakshanWeerasinghe: Went with the rename option — loading the real JSON here isn't really practical since the visualizer only ever sees the wire model after the Java-side adapter transforms the authored `layout`, so a raw JSON import wouldn't be equivalent to what actually renders.

Renamed the block to 'layout resolution against shapes mirroring the ftp/mcp trigger models' and expanded the comment to spell out that these are hand-built fixtures, not the shipped files, and to point at `TriggerLayoutTest.java` as the real guard over the bundled models (which checks id-resolution, not order). In 05e24122ea.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/test/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerLayoutTest.java
```diff
@@ -0,0 +1,203 @@
...
+
+    /** Whether a parameter is a payload (data-binding) parameter; mirrors {@code isPayloadParameter} in
+     * payloadComposer.ts. */
+    private boolean isPayloadParameter(Parameter parameter) {
+        if (Constants.DATA_BINDING.equals(parameter.getKind())) {
+            return true;
+        }
+        Codedata codedata = parameter.getType() != null ? parameter.getType().getCodedata() : null;
+        String type = codedata != null ? codedata.getType() : null;
+        return Constants.CD_TYPE_PAYLOAD_TYPE.equals(type) || Constants.CD_TYPE_PAYLOAD_TYPE_INCLUDED_RECORD.equals(type);
```
NipunaRanasinghe: This line is 122 characters and will fail `checkstyleTest`.

`build-config/checkstyle/build/checkstyle.xml` sets `LineLength` to `max=120, severity=error`; `gradle/javaProject.gradle` wires up `checkstyleTest` alongside `checkstyleMain` and excludes only `**/module-info.java`. The `Javadoc.*` test-source suppression doesn't help here, and `SuppressionFilter` is commented out in the config regardless.

Wrapping after the `||` puts it back under the limit:

```java
        return Constants.CD_TYPE_PAYLOAD_TYPE.equals(type)
                || Constants.CD_TYPE_PAYLOAD_TYPE_INCLUDED_RECORD.equals(type);
```

It's the only line added in `05e24122ea` over 120 — everything else tops out at 113.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/test/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerLayoutTest.java
```diff
@@ -0,0 +1,203 @@
...
+        for (String id : addressable) {
+            Assert.assertFalse(RESERVED_IDS.contains(id) || REST_DIRECTIVE.equals(id),
+                    where + ": the field \"" + id + "\" collides with a reserved layout id, so a layout "
+                            + "could never address it. Rename the field.");
+        }
+    }
+
+    /** Property {@code codedata.type} values {@code handlerUnitsOf} in handlerLayout.ts renders as a unit. */
+    private static final Set<String> ADDRESSABLE_PROPERTY_TYPES = Set.of(
+            "METADATA_FLAG", Constants.CD_TYPE_PAYLOAD_MODIFIER, Constants.CD_TYPE_COMPLEX_FUNCTION_ANNOTATION,
```
NipunaRanasinghe: Nit: `"METADATA_FLAG"` is a bare literal sitting between three `Constants.CD_TYPE_*` references. There's genuinely no `CD_TYPE_METADATA_FLAG` in `Constants` today, so adding one next to `CD_TYPE_PAYLOAD_MODIFIER` (line 136) would make this set uniform and keep the string in one place.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/handlerLayout.ts
```diff
@@ -0,0 +1,322 @@
...
+    });
+
+    const remainder = units.filter((unit) => !claimed.has(unit));
+    let remainderKey: string | undefined;
+    if (remainder.length > 0) {
+        if (restAt === -1) {
+            remainderKey = LAYOUT_ID_REST;
+            sections.push({ key: remainderKey, units: remainder });
+        } else {
+            remainderKey = sections[restAt].key;
```
NipunaRanasinghe: Minor: `*rest`'s position *within* a section's `fields` is ignored — the remainder is always appended to the end of the hosting section, whatever index the directive sat at. So `fields: ["*rest", "stream"]` resolves to `[...remainder, stream]`, not `[...remainder]` then `stream`.

Harmless for the three shipped layouts (all put it last), and the section-level placement is covered by the "places the remainder where `*rest` says" test. But the docs on `LAYOUT_ID_REST` and in `service.ts` call it "the placement directive", which reads as intra-section. Worth a sentence saying placement is section-granular, or honouring the index here.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/handlerLayout.ts
```diff
@@ -0,0 +1,322 @@
...
+            sections[restAt] = { ...sections[restAt], units: [...sections[restAt].units, ...remainder] };
+        }
+    }
+
+    const declaredKeys = new Set(
```
NipunaRanasinghe: Minor: excluding `remainderKey` also excludes a section that explicitly named artifact fields, when that same section happens to host `*rest`.

```js
layout: [
  { id: "g1", fields: ["$name", "*rest"] },
  { id: "g2", label: "G2", fields: ["$returnType"] },
]
```

`remainderKey` becomes `"g1"`, so `declaredKeys` is `{"g2"}`, `split` is `["g2"]` — length 1, no warning — and `consolidateArtifactUnits` renders the whole block in `g2`, even though the author named `$name` first in `g1`. The comment says `declaredKeys` "excludes the remainder, so the split warning fires only on a real mistake", but this *is* a real split and it goes unreported.

Tracking whether a section declared any field explicitly (separately from whether it hosts the remainder) would fix both the warning and the choice of `first`.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/handlerLayout.ts
```diff
@@ -0,0 +1,322 @@
...
+
+    if (fn.schema?.header) {
+        units.push({ id: LAYOUT_ID_HEADERS, kind: "HEADERS" });
+    }
+
+    return units.filter((unit) => !!unit.id);
+}
+
+/** Index every unit by its primary id, then by its alt ids. First registration wins. */
+function indexUnits(units: HandlerUnit[]): Map<string, HandlerUnit> {
```
NipunaRanasinghe, line 198: Minor: first-registration-wins silently makes the loser unaddressable.

Two units can share a primary id — a parameter that is both `advanced: true` and a payload parameter yields both a `PAYLOAD` and an `ADVANCED_PARAM` unit under the same name, and a `properties` key equal to a parameter name collides the same way. The shadowed unit can never be named, so it always falls into the remainder, and nothing warns: not here, and not in `TriggerLayoutTest` (whose id set is flat, so the collision is invisible there too).

Also unvalidated: two sections may share an `id`, which becomes a duplicate React key in `bodySections.map` / `advancedSections.map`.

Neither is reachable in the current corpus. A `warn` on a duplicate primary id here, and a duplicate-`id` assertion in `TriggerLayoutTest`, would keep it that way.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/test/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerLayoutTest.java
LakshanWeerasinghe: Already fixed in d914dcc5cc (wrapped after the `||`) — this was pushed a little before your comment landed. Confirmed `checkstyleTest` passes clean now.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/test/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerLayoutTest.java
LakshanWeerasinghe: Added `Constants.CD_TYPE_METADATA_FLAG` next to `CD_TYPE_PAYLOAD_MODIFIER` and swapped in the reference. In d789ae9bce.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/handlerLayout.ts
LakshanWeerasinghe: Went with documenting it rather than honouring the index — changing actual placement behavior felt like a bigger, less-tested change than the bug warranted for something none of the three shipped layouts hit. Added a sentence to both `LAYOUT_ID_REST`'s doc comment here and `HandlerLayoutSection`'s in service.ts spelling out that placement is section-granular, with the `["*rest", "stream"]` example. In d789ae9bce.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/handlerLayout.ts
LakshanWeerasinghe: Good catch, confirmed with your exact repro. Fixed by capturing `explicitlyDeclaredKeys` right after the `layout.map` pass — before the remainder gets merged into whichever section hosts `*rest` — so a section is judged by whether it picked its own fields, not by whether it happens to be the remainder's landing spot. `g1` now correctly counts as a real split. In d789ae9bce.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/handlerLayout.ts
LakshanWeerasinghe: Added both: `indexUnits` now warns when a later unit's primary id shadows an earlier one, and `resolveHandlerLayout` warns (and disambiguates the React key) when a layout section id is reused. Added the matching assertion in `TriggerLayoutTest` for duplicate section ids too. In d789ae9bce.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/handlerLayout.ts
NipunaRanasinghe, line 1: Non-blocking, for a follow-up: the three resolver changes in `d789ae9bce` landed without touching `handlerLayout.test.ts`, so they're the only untested paths left in a module that otherwise has 41 L1 tests.

- **`explicitlyDeclaredKeys` split detection.** The repro from the earlier thread is the test: `[{ id: "g1", fields: ["$name", "*rest"] }, { id: "g2", label: "G2", fields: ["$returnType"] }]` should now warn and consolidate into `g1`. The two existing cohesion tests both pass under the old `declaredKeys` too, so nothing currently distinguishes the fix from the bug.
- **Duplicate primary id warning in `indexUnits`.** Needs a fixture where a unit id collides — e.g. a parameter that is both `advanced: true` and a payload parameter.
- **Section-id disambiguation.** `key = ${key}-${index}` changes the rendered React keys, and no test pins which key a reused-id section ends up with.

All three are behaviour, not just diagnostics, so they're worth pinning before the next edit to this file quietly changes one.

NipunaRanasinghe (dismissed):
LGTM.

One non-blocking follow-up inline: the three resolver changes in `d789ae9bce` have no test coverage.

Note I haven't run the suites locally — the reported results are taken as given.

packages/ballerina-visualizer/src/views/BI/ServiceDesigner/Forms/TriggerHandlerForm/handlerLayout.ts
LakshanWeerasinghe: Good call, and sorry for the noise on my earlier replies — turns out those three resolver fixes (from `d789ae9bce`) had been made in a worktree that got cleaned up before merging back into this branch, so the branch was actually still on the pre-fix resolver despite what I said. Recovered them in 6d4a887c.

Added tests pinning all three behaviors in d7273cf9, using your exact repros:
- `explicitlyDeclaredKeys` split detection: `[{ id: "g1", fields: ["$name", "*rest"] }, { id: "g2", label: "G2", fields: ["$returnType"] }]` now warns and consolidates into `g1`.
- Duplicate primary id warning in `indexUnits`, via a parameter that's both `advanced: true` and a bindable payload (same bare id, two units).
- Section-id disambiguation (`key = ${key}-${index}`) for two layout sections sharing one id.

Verified all three fail against the pre-fix resolver (swapped it back in locally, confirmed 3/3 failures) before restoring the fix, so they actually catch a regression rather than passing incidentally either way — same gap you flagged in the two existing cohesion tests.

NipunaRanasinghe (dismissed):
LGTM

packages/ballerina-extension/playwright-report/index.html
NipunaRanasinghe, line 1: Do we need to commit this file? If not lets add to gitignore

LakshanWeerasinghe: removed

NipunaRanasinghe (approved):

PR #942, opened 2026-08-24
Update issue reporting URL

### LINKED ISSUE #1808: Issue reporting URL in the VS Code plugin points to an invalid GitHub repo
### Description

Image

### Steps to Reproduce

See description

### Version

Ballerina plugin: 5.12.2, WI Plugin: 1.0.1

_No response_

pasindufernando1:
## Purpose
Fixes : https://github.com/wso2/product-integrator/issues/1808

### Summary

- Updated issue-reporting links to use the Product Integrator GitHub issues URL.
- Centralized the URL in `PRODUCT_INTEGRATOR_ISSUES_URL`.
- Replaced local URL constants and removed unused imports.


rosensilva (dismissed):

dulajdilshan (commented):
Can't we put this URL into a common place? (e.g. `/common`)

packages/ballerina-core/src/interfaces/constants.ts
```diff
@@ -16,6 +16,8 @@
  * under the License.
  */
 
+export const BALLERINA_INTEGRATOR_ISSUES_URL = "https://github.com/wso2/product-integrator/issues";
```
samithkavishke: We should use the Product Integrator URL only when the user is using Product Integrator. If the user is using only the Ballerina Extension, we should navigate them to https://github.com/ballerina-platform/ballerina-lang/issues

dulajdilshan:
```suggestion
export const PRODUCT_INTEGRATOR_ISSUES_URL = "https://github.com/wso2/product-integrator/issues";
```

pasindufernando1: Ideally the issue should be reported in the product integrator. Even with the ballerina extension, the user can use the visualize option and move to the integrator aspects. Hence if it is a lang issue, we should be moving them to the lang side, rather than creating an issue in the lang repo directly

dulajdilshan: It does make sense for users who use pro-code to report issues in ballerina-lang since there is a possibility of them seeing these issues as language issues. But for low-code users, we expect them to report in the product-integrator.  Overall, we should stick with redirecting to one URL ( = https://github.com/wso2/product-integrator/issues)

pasindufernando1: Fixed

dulajdilshan (dismissed):
LGTM!

packages/ballerina-core/src/interfaces/constants.ts
```diff
@@ -16,6 +16,8 @@
  * under the License.
  */
 
+export const PRODUCT_INTEGRATOR_ISSUES_URL = "https://github.com/wso2/product-integrator/issues";
```
NipunaRanasinghe, line 19: This extension is used by the open source Ballerina community as well, and those users have not opted into WSO2 Integrator. Pointing every "create an issue" link at `wso2/product-integrator` means a plain Ballerina user gets sent to the Integrator issue tracker to report what is, to them, a Ballerina bug. This might be confusing for the reporter.

Could we pick a better alternative for the Ballerina users before this goes in? Worth agreeing the Ballerina-side target with the Ballerina team (this repo vs `ballerina-lang`) rather than defaulting everyone to the Integrator tracker.

rosensilva: Earlier, we decided to keep all issues related to the product-integrator (including standard libraries, etc.) in one place. This way, we can ensure that no issues are overlooked, even if they're reported in multiple spots. Plus, it shouldn't be a big concern because we guide users through the bug reporting link, which is what they see when they report an issue. I believe they might take the URL quite seriously. WDYT?

dulajdilshan: Also, if the URL does matter, we can improve the issue template by adding a redirection to ballerina-lang for language issues (But still hard to manage, as Rosen said). 

Additional thing: These notifications are not coming because of direct language bugs. These are, most of the time, limitations of the lang-server. So IMO, we need to track them via `product-integrator`

NipunaRanasinghe: @rosensilva @dulajdilshan your points make sense. +1 to keep the existing behaviour.

NipunaRanasinghe (approved):

PR #945, opened 2026-08-24
Filter out KBs from connections + Refactoring KB filter

damsith0:
> Solves the issue https://github.com/wso2/product-integrator/issues/2221

- Added shared logic to identify Marketplace items tagged `"knowledge-base-as-a-service"`.
- Filtered Knowledge Base services from the WSO2 cloud Connections list.
- Updated the Knowledge Base page to use the shared filtering logic.
- Over-fetched Marketplace results before filtering to preserve the requested page size.
- Added tests for filtering, ordering, pagination, and internal service handling.

### DIFF
--- a/packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/CloudKnowledgeBasePage.tsx
+++ b/packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/CloudKnowledgeBasePage.tsx
@@ -24,11 +24,7 @@ import ButtonCard from "../../../../components/ButtonCard";
 import { BodyTinyInfo } from "../../../styles";
 import { usePlatformExtContext } from "../../../../providers/platform-ext-ctx-provider";
 import { ConnectorsGrid, Section, SectionHeader, SectionTitle } from "../AddConnectionPopup/styles";
-import { ProgressWrap } from "./utils";
-
-// Devant tags knowledge base services with this tag.
-const KB_TAG = "knowledge-base-as-service";
-const isKnowledgeBase = (item: MarketplaceItem) => item.tags?.includes(KB_TAG) ?? false;
+import { isKnowledgeBaseService, ProgressWrap } from "./utils";
 
 interface CloudKnowledgeBasePageProps {
     // Opens a blank CloudKnowledgeBase create form (manual entry, no Devant service pre-selected).
@@ -68,7 +64,7 @@ export function CloudKnowledgeBasePage(props: CloudKnowledgeBasePageProps) {
                 request: getMarketPlaceParams,
             }),
         enabled: isSignedIn,
-        select: (data) => ({ ...data, data: (data?.data || []).filter(isKnowledgeBase) }),
+        select: (data) => ({ ...data, data: (data?.data || []).filter(isKnowledgeBaseService) }),
     });
 
     const items: MarketplaceItem[] = knowledgeBases?.data || [];
--- a/packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/DevantConnectorList.tsx
+++ b/packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/DevantConnectorList.tsx
@@ -34,7 +34,7 @@ import {
     SectionHeader,
     SectionTitle,
 } from "../AddConnectionPopup/styles";
-import { DevantConnectionFlow, getKnownAvailableNode, isKnowledgeBaseService, ProgressWrap } from "./utils";
+import { DevantConnectionFlow, filterConnectionMarketplaceItems, getKnownAvailableNode, ProgressWrap } from "./utils";
 
 // Target number of connections to show. Knowledge base services can only be filtered out client-side
 // (the marketplace tag filter is include-only), so over-fetch and trim to this size so filtered-out
@@ -165,18 +165,12 @@ export function DevantConnectorList(props: DevantConnectorListProps) {
             filterType !== "databases" && platformExtState.isLoggedIn && !!platformExtState?.selectedContext?.project,
         select: (data) => ({
             ...data,
-            data: data.data
-                .filter((item) => {
-                    // Knowledge base services are listed under "Add Knowledge Base", not here.
-                    if (isKnowledgeBaseService(item)) {
-                        return false;
-                    }
-                    if (filterType === "internal-services") {
-                        return item.component?.componentId !== platformExtState?.selectedComponent?.metadata?.id;
-                    }
-                    return true;
-                })
-                .slice(0, CONNECTIONS_PAGE_SIZE),
+            data: filterConnectionMarketplaceItems(
+                data.data,
+                filterType,
+                platformExtState?.selectedComponent?.metadata?.id,
+                CONNECTIONS_PAGE_SIZE,
+            ),
         }),
     });
 
--- a/packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/DevantConnectorList.tsx
+++ b/packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/DevantConnectorList.tsx
@@ -34,7 +34,7 @@ import {
     SectionHeader,
     SectionTitle,
 } from "../AddConnectionPopup/styles";
-import { DevantConnectionFlow, getKnownAvailableNode, ProgressWrap } from "./utils";
+import { DevantConnectionFlow, getKnownAvailableNode, isKnowledgeBaseService, ProgressWrap } from "./utils";
 
 interface DevantConnectorListProps {
     onItemSelect: (
@@ -160,6 +160,10 @@ export function DevantConnectorList(props: DevantConnectorListProps) {
         select: (data) => ({
             ...data,
             data: data.data.filter((item) => {
+                // Knowledge base services are listed under "Add Knowledge Base", not here.
+                if (isKnowledgeBaseService(item)) {
+                    return false;
+                }
                 if (filterType === "internal-services") {
                     return item.component?.componentId !== platformExtState?.selectedComponent?.metadata?.id;
                 }
--- a/packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/DevantConnectorList.tsx
+++ b/packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/DevantConnectorList.tsx
@@ -36,6 +36,12 @@ import {
 } from "../AddConnectionPopup/styles";
 import { DevantConnectionFlow, getKnownAvailableNode, isKnowledgeBaseService, ProgressWrap } from "./utils";
 
+// Target number of connections to show. Knowledge base services can only be filtered out client-side
+// (the marketplace tag filter is include-only), so over-fetch and trim to this size so filtered-out
+// KBs don't shrink the visible connections list (or search results).
+const CONNECTIONS_PAGE_SIZE = 24;
+const MARKETPLACE_OVERFETCH_LIMIT = 100;
+
 interface DevantConnectorListProps {
     onItemSelect: (
         flow: DevantConnectionFlow | null,
@@ -131,7 +137,7 @@ export function DevantConnectorList(props: DevantConnectorListProps) {
     };
 
     const getMarketPlaceParams: GetMarketplaceItemsParams = {
-        limit: 24,
+        limit: MARKETPLACE_OVERFETCH_LIMIT,
         offset: 0,
         networkVisibilityFilter: "all",
         networkVisibilityprojectId: platformExtState?.selectedContext?.project?.id,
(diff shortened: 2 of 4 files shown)


packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/DevantConnectorList.tsx
```diff
@@ -160,6 +160,10 @@ export function DevantConnectorList(props: DevantConnectorListProps) {
         select: (data) => ({
             ...data,
             data: data.data.filter((item) => {
+                // Knowledge base services are listed under "Add Knowledge Base", not here.
+                if (isKnowledgeBaseService(item)) {
```
pasindufernando1: KB services are filtered client-side, so they still consume slots from the single fixed page we fetch (`limit: 24, offset: 0`, no pagination UI). An org with enough KB items in that page gets a shorter connections list — worst case "No services running" while connectable services exist past the page. Same truncation applies to search results.

`GetMarketplaceItemsParams` has a `tags` field — does the API support excluding by tag? If so this could move server-side; otherwise over-fetching would keep the visible count independent of how many KBs land in the page. Non-blocking if KB counts are expected to stay small relative to 24.

damaya0: GetMarketplaceItemsParams API only has a poistive tags filter, since no other service type has tags other than knowledge bases, we wont be able to filter by using that. 

As an alternate fix we can overfetch marketplace items, and then filter out KBs, and trim to match the page limit of 24.

pasindufernando1: Agreed on the API constraint — positive-only `tags` can't express the exclusion, so client-side is the only option. Over-fetch + trim sounds right; `rpc-manager.ts:1004` already uses `limit: 100` for a similar call, so there's precedent for the bound.

Two things when you push it:

1. `CloudKnowledgeBasePage.tsx:47` needs the same treatment, and it's the more likely failure. It filters *for* KBs out of the same fixed 24 sorted by `createdTime` — an org with 24 newer non-KB services sees an empty Knowledge Bases page even though KBs exist.
2. The PR description already lists "Over-fetched Marketplace results before filtering, then limited the list to the requested page size" — that's not on the branch yet (`4545b326` still has `limit: 24`). Worth fixing the description or landing the change, so this doesn't get approved on a summary that's ahead of the code.

damaya0: Okay will change the connection page to fetch 100 and then trim to 24 after filtering out KBs.

Just realized that we can use the tag filtering in the API itself for the knowledge base filtering in CloudKnowledgeBasePage.tsx. Will send a PR for that improvement :-)

packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/DevantConnectorList.tsx
```diff
@@ -34,7 +34,13 @@ import {
...
     SectionTitle,
 } from "../AddConnectionPopup/styles";
-import { DevantConnectionFlow, getKnownAvailableNode, ProgressWrap } from "./utils";
+import { DevantConnectionFlow, filterConnectionMarketplaceItems, getKnownAvailableNode, ProgressWrap } from "./utils";
+
+// Target number of connections to show. Knowledge base services can only be filtered out client-side
+// (the marketplace tag filter is include-only), so over-fetch and trim to this size so filtered-out
+// KBs don't shrink the visible connections list (or search results).
+const CONNECTIONS_PAGE_SIZE = 24;
+const MARKETPLACE_OVERFETCH_LIMIT = 100;
```
lnash94: Does this keep fetching until 24 non-KB connections are found, or does it only filter the first 100 results?

damaya0: Only the first 100 results akke.

lnash94: Can there be more than 100 marketplace items, and if so, how are items after the first 100 shown? If filtering leaves fewer than 24 connections, how will connections beyond the first 100 be displayed?

damaya0: The fetching is per page. If those 100 does not contain more than 24 non KB connections, then that page will have less than 24 items.

damaya0: Correction - WI only renders one UI component with 24 components and there is no pagination. The api has default limit of 24. Created an issue for this - https://github.com/wso2/product-integrator/issues/2267. UNtil this issue is solved, we can use this method to list the connections.

lnash94: Since this issue https://github.com/wso2/product-integrator/issues/2267 already existed and was not introduced by this fix, let's address it separately.

lnash94 (dismissed):

packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/utils.ts
```diff
@@ -27,6 +27,36 @@ export const ProgressWrap = styled.div`
...
+    filterType: string,
+    selectedComponentId: string | undefined,
+    pageSize: number,
+): MarketplaceItem[] {
+    return items
+        .filter((item) => {
+            if (isKnowledgeBaseService(item)) {
+                return false;
+            }
+            if (filterType === "internal-services") {
```
SasinduDilshara: What is this internal services contains? is it already had name?

Btw lets add a meaningful constant for that

packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/utils.ts
```diff
@@ -27,6 +27,36 @@ export const ProgressWrap = styled.div`
     padding: 40px;
 `;
 
+// Devant tags knowledge base services with this tag.
+export const KB_SERVICE_TAG = "knowledge-base-as-service";
+export const isKnowledgeBaseService = (item: MarketplaceItem) => item.tags?.includes(KB_SERVICE_TAG) ?? false;
```
SasinduDilshara, line 33: include or ==?

packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/utils.ts
damaya0: Internal services are the services running inside the user's own WSO2 Cloud org.

That filtering was there previously, (In code diff - deleted line 163 in DevantConnectorList.tsx file. Filter type definition at line 56) The whole filtering was moved to the utils from this PR.

packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/utils.ts
damaya0: .include was chosen since tags is an array and in the future multiple tags can be added to one service.

SasinduDilshara: so `item` is an array?, Im got confussed because of the name

damaya0: item.tags is the array, thats where we check for KB_SERVICE_TAG

SasinduDilshara (dismissed):

lnash94 (dismissed):

packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/CloudKnowledgeBasePage.tsx
```diff
@@ -68,7 +64,7 @@ export function CloudKnowledgeBasePage(props: CloudKnowledgeBasePageProps) {
                 request: getMarketPlaceParams,
             }),
         enabled: isSignedIn,
-        select: (data) => ({ ...data, data: (data?.data || []).filter(isKnowledgeBase) }),
+        select: (data) => ({ ...data, data: (data?.data || []).filter(isKnowledgeBaseService) }),
```
madushajg, line 67: The over-fetch rationale added for the connections list applies with more force here, and this page didn't get it. This query still uses limit: 24, offset: 0 sorted by `createdTime` and then keeps only KB-tagged items client-side, so in an org with more than 24 marketplace services, "Existing WSO2 Cloud Knowledge Bases" will render short or empty even when KBs exist.
Unlike the exclusion case, this one can be done server-side: `GetMarketplaceItemsParams` has `tags?: string | null` ("Optionally filter services based on tags. Multiple tags can be provided as a comma separated list."). Setting tags: `KB_SERVICE_TAG` on `getMarketPlaceParams` makes the fetch itself KB-only, removes the truncation, and makes the client-side `.filter(isKnowledgeBaseService)` a cheap guard rather than the actual filter.

packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/DevantConnectorList.tsx
madushajg: The 100/24 relationship is an unstated assumption: the list silently renders fewer than `CONNECTIONS_PAGE_SIZE` if more than 76 of the first 100 items are KBs (or, in `internal-services` mode, the selected component itself), and there's no "load more" to recover. Worth stating that explicitly in the comment, e.g. `MARKETPLACE_OVERFETCH_LIMIT` defined as a multiple of `CONNECTIONS_PAGE_SIZE` with a note that the page shrinks past that ratio, so a future change to the page size doesn't silently change the safety margin. Also note this makes every debounced search request pull 100 items instead of 24.

packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/DevantConnectorList.tsx
```diff
@@ -159,12 +165,12 @@ export function DevantConnectorList(props: DevantConnectorListProps) {
...
-                    return item.component?.componentId !== platformExtState?.selectedComponent?.metadata?.id;
-                }
-                return true;
-            }),
+            data: filterConnectionMarketplaceItems(
+                data.data,
+                filterType,
+                platformExtState?.selectedComponent?.metadata?.id,
+                CONNECTIONS_PAGE_SIZE,
+            ),
```
madushajg, line 184: `filterType` is a well-typed union here but the util takes filterType: string, so the type is thrown away at the call boundary and a typo would fail silently. Export the union from utils.ts (e.g. export type `ConnectionFilterType = "all" | "internal-services" | "third-party-services" | "databases"`) and use it both for the useState and the parameter.
Related: the new `INTERNAL_SERVICES` constant is only referenced inside `utils.ts`, this file still hardcodes "internal-services" at lines 149, 208, 231 and 232. Either use the constant here too or drop it; as-is it's a second source of truth for the same literal.

packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/utils.ts
```diff
@@ -27,6 +27,37 @@ export const ProgressWrap = styled.div`
...
+    selectedComponentId: string | undefined,
+    pageSize: number,
+): MarketplaceItem[] {
+    return items
+        .filter((item) => {
+            if (isKnowledgeBaseService(item)) {
+                return false;
+            }
+            if (filterType === INTERNAL_SERVICES) {
+                return item.component?.componentId !== selectedComponentId;
```
madushajg, line 57: When `selectedComponentId` is undefined, `item.component?.componentId` !== `selectedComponentId` evaluates to `undefined` !== `undefined` → `false` for every item that has no `component`, so they all get dropped from `internal-services`. The old inline code had the same flaw, but the extracted signature now explicitly admits undefined, so it's worth guarding:

```ballerina
if (filterType === INTERNAL_SERVICES && selectedComponentId) {
    return item.component?.componentId !== selectedComponentId;
}
```
None of the new tests cover `selectedComponentId` === `undefined`, so this would pass unnoticed either way.

packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/CloudKnowledgeBasePage.tsx
damaya0: We will open a separate PR for this change since this is an improvement and is not related to the issue this PR is solving. Hope it's okay

packages/ballerina-visualizer/src/views/BI/Connection/DevantConnections/utils.ts
damaya0: Did the suggested changes. and added a test too.

lnash94 (approved):

SasinduDilshara (approved):

madushajg (approved):

PR #946, opened 2026-08-24
Fix branch rendering when a comment is the only child

pasindufernando1:
Fixes : https://github.com/wso2/product-integrator/issues/2223
### Problem

A branch whose only child is a comment rendered as nothing — no widget, no links — so the diagram drew a connector straight past it. Repro:

```ballerina
do {
    if true {
        // Hi
    }
} on fail error e {
    log:printError("Error occurred", 'error = e);
    return e;
}
```

The `Then` branch disappears and the dashed line runs from the `If` diamond to the Error Handler.

### Cause

Comments render as note chips on another node rather than as widgets, so a comment-only branch has no renderable content — but it was not treated as empty:

- `InitVisitor` — empty-branch checks used `children.length === 0`, so no `EMPTY` placeholder was inserted.
- `NodeFactoryVisitor` — the branch placeholder widget is created only behind checks a `COMMENT` sibling breaks (`children.length === 1`, `children.at(0)`), so even with a placeholder present nothing was created.

Both layers had to change; fixing only the first is a no-op.

### Fix

Adds `getRenderableChildren()` to skip `COMMENT` children in those checks, and uses it for the last-child lookups so a trailing comment is no longer mistaken for a branch's last node. Covers If/Match, While/Foreach/Lock and the ErrorHandler body branch.

### Verification

Ran the visitor pipeline over the reported flow model and compared against the unfixed baseline:

```
before:  36031 -> 36031-Else -> 36031-endif          (Else only)
after:   36031 -> 36031-Then -> 36031-endif          (Then now wired)
         36031 -> 36031-Else -> 36031-endif
```

The `before` run reproduces the reported diagram. Also checked while-loop with comment-only body, if with a trailing comment, a normal if, and a genuinely empty if — all correct. `diff.test.ts` and `WorkflowNodeCompatibility.test.ts` pass (36 tests).

### Not covered

The note chip still attaches to the `If` node rather than inside the branch, since comments flush to the last real node and the placeholder is an `EMPTY` node with no chip support. Existing behaviour, left as-is.

## Summary

- Fixed rendering for branches whose only child is a comment.
- Excluded `COMMENT` nodes from branch emptiness checks and final-node lookup.
- Added `EMPTY` placeholders for comment-only branches in conditional, loop, container, and error-handler branches.
- Added snapshot tests for layout nodes, links, and rendering flags.
- Verified normal, empty, comment-only, loop, and trailing-comment cases.

### DIFF
--- a/packages/bi-diagram/src/visitors/InitVisitor.ts
+++ b/packages/bi-diagram/src/visitors/InitVisitor.ts
@@ -51,6 +51,12 @@ export class InitVisitor implements BaseVisitor {
         return true;
     }
 
+    // Comments render as note chips on another node, not as widgets of their own, so a branch
+    // whose children are all comments has nothing to link to and must be treated as empty too.
+    private isBranchEmpty(branch: Branch): boolean {
+        return !branch.children || branch.children.every((child) => child.codedata.node === "COMMENT");
+    }
+
     beginVisitNode(node: FlowNode, parent?: FlowNode): void {
         if (!this.validateNode(node)) return;
         node.viewState = this.getDefaultViewState();
@@ -109,8 +115,8 @@ export class InitVisitor implements BaseVisitor {
                 }
             }
 
-            // if branch is empty add empty node
-            if (!branch.children || branch.children.length === 0) {
+            // if branch is empty (or only holds comments) add empty node
+            if (this.isBranchEmpty(branch)) {
                 // empty branch
                 // add empty node as `add new node` button
                 const emptyNode: FlowNode = {
@@ -180,8 +186,8 @@ export class InitVisitor implements BaseVisitor {
                 }
             }
 
-            // if branch is empty add empty node
-            if (!branch.children || branch.children.length === 0) {
+            // if branch is empty (or only holds comments) add empty node
+            if (this.isBranchEmpty(branch)) {
                 // empty branch
                 // add empty node as `add new node` button
                 const emptyNode: FlowNode = {
@@ -282,8 +288,8 @@ export class InitVisitor implements BaseVisitor {
             }
         }
 
-        // add empty node if the branch is empty
-        if (!branch.children || branch.children.length === 0) {
+        // add empty node if the branch is empty (or only holds comments)
+        if (this.isBranchEmpty(branch)) {
             // empty branch
             // add empty node as `add new node` button
             const emptyNode: FlowNode = {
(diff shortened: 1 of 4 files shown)


packages/bi-diagram/src/visitors/InitVisitor.ts
```diff
@@ -51,6 +51,12 @@ export class InitVisitor implements BaseVisitor {
         return true;
     }
 
+    // Comments render as note chips on another node, not as widgets of their own, so a branch
+    // whose children are all comments has nothing to link to and must be treated as empty too.
+    private isBranchEmpty(branch: Branch): boolean {
+        return !branch.children || branch.children.every((child) => child.codedata.node === "COMMENT");
```
gigara:
```suggestion
        return !branch.children || branch.children.length === 0 || branch.children.every((child) => child.codedata.node === "COMMENT");
```

packages/bi-diagram/src/visitors/InitVisitor.ts
```diff
@@ -51,6 +51,12 @@ export class InitVisitor implements BaseVisitor {
         return true;
     }
 
+    // Comments render as note chips on another node, not as widgets of their own, so a branch
+    // whose children are all comments has nothing to link to and must be treated as empty too.
+    private isBranchEmpty(branch: Branch): boolean {
```
gigara: isEmptyBranchOrOnlyComments

packages/bi-diagram/src/visitors/NodeFactoryVisitor.ts
```diff
@@ -181,6 +181,11 @@ export class NodeFactoryVisitor implements BaseVisitor {
         }
     }
 
+    // Comments render as note chips on another node, so they never count as branch content.
+    private getRenderableChildren(branch: Branch): FlowNode[] {
```
gigara: getRenderableBranchChildren

packages/bi-diagram/src/visitors/InitVisitor.ts
pasindufernando1: Addressed

packages/bi-diagram/src/visitors/InitVisitor.ts
pasindufernando1: Addressed

packages/bi-diagram/src/visitors/NodeFactoryVisitor.ts
pasindufernando1: Addressed

gigara (dismissed):

gigara (approved):

NipunaRanasinghe (approved):

PR #947, opened 2026-08-24
Re-land review-mode empty-diff fix (#855) and fix PackageUtil sample-project concurrency

IsuruMaduranga:
## What this does

This PR re-lands the review-mode empty-diff fixes and performance optimizations (originally #855, reverted in #928) and resolves the concurrency hazard in `PackageUtil`'s sample-project package resolution (fixes [wso2/product-integrator#2193](https://github.com/wso2/product-integrator/issues/2193)).

---

#### Root Cause (Upstream Compiler Internals)
* In `ballerina-lang` (`io.ballerina.projects.internal.environment.EnvironmentPackageCache`), the package cache maintains two plain, unsynchronized `java.util.HashMap` fields (`projectsById` and `projectsByOrgNameVersion`).
* Calling `PackageResolver.resolvePackages()` or `resolvePackageMetadata()` invokes `EnvironmentPackageCache.cache(Package pkg)`, which directly mutates these maps via `computeIfAbsent`.
* When `SAMPLE_PROJECT` was previously held as a global singleton, concurrent LS requests (flow-model generation, trigger models, connector resolution, library metadata) concurrently mutated these maps, causing `HashMap` corruption (lost updates, tree-bin corruption, infinite loops on resize).

> **Note for Future Upstream Fix in `ballerina-platform/ballerina-lang`:**  
> The permanent compiler-level fix belongs in:  
> `compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/environment/EnvironmentPackageCache.java`  
> Synchronizing `EnvironmentPackageCache` (e.g., using `ConcurrentHashMap` or synchronizing `cache()` / `getPackage()`) will make `BuildProject` environments safe for concurrent shared resolution upstream.

#### Solution in this PR (Application-Level Isolation)
Instead of coarse locking (`SAMPLE_RESOLUTION_LOCK`, which serialized unrelated LS requests), this PR decouples the immutable cache from the mutable compiler resolver:

1. **Per-Thread Resolver (`ThreadLocal<BuildProject>`)**:
   - `SAMPLE_PROJECT` is now a `ThreadLocal<BuildProject>`. Each LS worker thread resolves through its own isolated `Environment`, `DefaultPackageResolver`, and `EnvironmentPackageCache`.
   - Eliminates cross-thread mutable state with zero locking on hot LS paths.
2. **Lock-Free Immutable Cache (`SAMPLE_RESOLUTION_CACHE`)**:
   - Retains the performance optimization by caching resolved `.bala` file paths (`ConcurrentHashMap<String, Optional<Path>>`).
   - Cache hits never touch the resolver; cache misses resolve on the calling thread's isolated environment and populate via `putIfAbsent`.
3. **Closed Secondary Access Paths (`localPackageRepository()`)**:
   - `LibraryMetadataReader` and `AiUtils` now use `PackageUtil.localPackageRepository()` (also `ThreadLocal`), preventing secondary un-synchronized environment access via `BallerinaUserHome`.
4. **Memory & Disk Hygiene**:
   - Per-thread projects load in-memory from a single shared, static directory (`SAMPLE_PROJECT_DIR`) with `deleteOnExit`. No per-thread temporary directories are created on disk.
   - Thread-local instances are held in the worker thread's `ThreadLocalMap` and garbage-collected when the thread terminates.
5. **Forward Compatibility**:
   - If/when `ballerina-lang` synchronizes `EnvironmentPackageCache`, `PackageUtil` can collapse `ThreadLocal` back to a shared instance with zero API breakage to callers.

---

#### A. Language Server (`SemanticDiffComputer` & Services)
* **Comprehensive Module & Construct Coverage**:
  * **All Modules & Tests**: `SemanticDiffComputer` now iterates over all packages and submodules (`pkg.modules()`) including test documents, with URIs resolved via exact document paths.
  * **New Construct Diff Support**: Emits semantic diffs for module variables/configurables (`MODULE_VAR`), constants (`CONSTANT`), listeners (`LISTENER`), class definitions (`CLASS_DEFINITION`), enum definitions (`ENUM_DEFINITION`), and imports (`IMPORT`).
  * **Class Diffing**: Diffed method-by-method as `FUNCTION` diffs; class headers (metadata, visibility, `isolated` / class-type qualifiers, class keyword, and name) and non-method members are diffed as `CLASS_DEFINITION`.
  * **Trivia-Insensitive Import Diffing**: Imports are compared strictly by semantic identity (`org/module:alias`) rather than raw source code trivia. Prevents untouched imports from showing as "modified" when leading comments re-attach due to reordering/insertions. Added/deleted imports render in canonical `import <key>;` form.
  * **Type Deletion & External Function Support**: Explicitly tracks type definition deletions and external function body modifications (previously omitted).
* **`copilotAgentService/ensureAiBaseline` RPC**:
  * Replaced fire-and-forget `didOpen` LSP notifications with an awaitable request that explicitly evicts, reloads, and applies baseline snapshot contents before tool execution starts. Kills the race where the LS indexed mid-edit files from disk.
* **`copilotAgentService/prewarmDependencies` RPC**:
  * Resolves and pre-compiles a package's direct dependencies asynchronously in the background at turn start, removing ~3.5s of dependency resolution off the critical review click path.
* **Virtual Document Event Optimization**:
  * `ProjectUpdateEventPublisher` skips `PROJECT_UPDATE` subscriber fan-out for `ai://` virtual documents, preventing unnecessary package recompilations for frozen baseline files.

#### B. VS Code Extension Host (`AgentExecutor`, `ApprovalViewManager`, Checkpoints)
* **Robust Baseline Seeding & Mid-Run Package Seeding**:
  * `AgentExecutor` awaits `ensureAiBaseline` prior to running the agent loop.
  * When a new package is created mid-run (`Ballerina.toml` write), its baseline is initialized with empty files so subsequent writes review cleanly as additions rather than diffing against itself.
* **Package Mapping & Error Isolation**:
  * Normalizes file path casing, trailing slashes, and Windows drive letters when matching modified files to packages.
  * Unions `Ballerina.toml` and `ProjectSource` package lists so empty package arrays don't orphan modified files.
  * If `getSemanticDiff` fails for one package, sibling packages still render their diffs, and an explicit `semanticDiffError` is recorded rather than masquerading as a no-op turn.
* **Review Reseed Invalidation**:
  * Finalizes pending reviews on all threads when a new generation starts (since the single `ai://` LS baseline slot is reseeded for the new run).
* **Safe Revert & Checkpoint Verification**:
  * `restoreWorkspaceSnapshot` returns an explicit boolean status; `revertGeneration` verifies that the workspace restore succeeded before marking the generation reverted in state.
* **Concurrent Execution Bounds**:
  * Parallelizes `getSemanticDiff` calls (concurrency cap: 3).
  * Parallelizes baseline seeding (concurrency cap: 4).
  * Chunks workspace checkpoint file reads in 32-way concurrent batches.

#### C. Visualizer Webview UI (`ReviewMode`, `ReviewBar`, Diagram Engine)
* **`ReadonlySourceDiff` Component**:
  * Renders non-diagram constructs (module variables, constants, listeners, classes, enums, imports) as syntax-highlighted old/new source diff blocks with VS Code theme border variables.
* **Unified Labels & Declarations Grouping**:
  * Centralized `NodeKindEnum` in `ballerina-core` with `nodeKindLabels.ts` shared between `ReviewMode` and `ReviewBar`.
  * Groups non-service constructs under a collapsible `declarations` group in the `ReviewBar`. Class methods label as "method" rather than "resource".
* **Review Session Model Cache (`reviewModelCache.ts`)**:
  * Implemented a review-session promise cache shared across diagram components. Prevents redundant LS round-trips during prev/next navigation or `Old`/`New`/`Diff` toggle switches and collapses concurrent requests.
  * **Outward Prefetching**: Automatically warms models for all items in the review payload starting outward from the active item using 2 background workers.
* **Diagram Engine Render Optimization**:
  * Initialized `DiagramEngine` lazily with `useState(() => new DiagramEngine())` in `bi-diagram`, eliminating the repeated creation/destruction of ~20 node/link factories on every re-render.
* **Non-`.bal` Changes & Failure States**:
  * Displays modified non-`.bal` files (e.g., `Config.toml`, `Ballerina.toml`) rather than reporting "No changes to review".
  * Replaced permanent spinners with warning banners and explicit failure states when old-side flow-model compilations fail.

---

### 3. Why Parallel Execution is Critical for Review Mode

Review Mode transforms raw code diffs into interactive visual flow, type, and component diff diagrams. Sequential execution previously caused substantial latency and UI stalls during review generation and navigation. We introduced bounded parallel execution across four key paths:

1. **Simultaneous Old vs. New Model Generation**:
   - Rendering a unified diff diagram requires fetching both the **original model** (`file://`) and the **modified model** (`ai://`) from the Language Server.
   - Sequential fetches doubled user wait time (`T_old + T_new` ~4s). Fetching both concurrently halves the latency (`max(T_old, T_new)` ~2s) per diagram.

2. **Multi-Package Diff Collection**:
   - In multi-package workspaces or migration workflows, Copilot frequently modifies multiple packages in a single turn.
   - The extension collects `getSemanticDiff` across affected packages concurrently (bounded at 3 concurrent workers) so multi-core LS workers process compilations in parallel rather than compounding sequential compilation times.

3. **Background Prefetching for Zero-Lag Navigation**:
   - An AI generation often touches 5–10 functions/types. Without prefetching, navigating between items (clicking "Next" or clicking review chips in chat) incurred a 1–2s loading spinner on every single item.
   - Two background workers prefetch and warm models in the review-session cache outward from the active item. By the time the user reviews the first change, subsequent diagrams are already cached in memory and render instantly.

4. **Turn-Start Baseline Seeding & Checkpoint Reads**:
   - Pre-seeding baseline packages into the LS (`ensureAiBaseline`) runs with bounded concurrency (cap: 4), and workspace file snapshots are read in 32-way concurrent chunks rather than one file at a time.

> **Direct Connection to the Concurrency Fix**:  
> Because Review Mode intentionally dispatches these concurrent requests to maximize UI responsiveness, the Language Server's resolution paths (e.g., `PackageUtil`, `LibraryMetadataReader`, `TriggerModelReader`) frequently receive overlapping requests on separate worker threads. This makes the thread-isolation fix in `PackageUtil` essential to prevent concurrent compiler cache corruption.

---

### 4. Testing & Verification

* **Concurrency Stability Test**:
  * Added `PackageUtilConcurrencyTest` in `flow-model-generator-core`.
  * Runs 16 threads synchronized via `CyclicBarrier` across 25 iterations exercising `cachedVersion`, `isModuleUnresolved`, `getModulePackageOffline`, and `getModulePackage`.
  * Asserts execution completes without exceptions, hangs, or divergences from a single-threaded baseline.
* **Unit & Integration Suites**:
  * Language Server test suites pass (`./gradlew test`).
  * Extension unit tests (`pnpm run test:unit`) pass.
  * Visualizer unit tests pass.

## Summary

- Re-lands review-mode empty-diff fixes and performance improvements.
- Adds broader semantic diff support for declarations, source-only changes, and trivia-only changes.
- Adds reliable baseline seeding, dependency prewarming, bounded concurrency, caching, prefetching, and improved failure handling.
- Improves review UI rendering for source diffs, diagram fallbacks, compilation errors, and partial results.
- Isolates `PackageUtil` sample-project resolution per thread and adds concurrency coverage.
- Improves workspace restore validation and diagnostic error reporting.

### DIFF
--- a/packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
@@ -23,6 +23,7 @@ import { ModelMessage, stepCountIs, streamText, TextStreamPart } from 'ai';
 import { getAnthropicClient, getProviderCacheControl, getProviderModelOptions, addCacheControlToMessages, ANTHROPIC_SONNET } from '../utils/ai-client';
 import { populateHistoryForAgent, getErrorMessage, getErrorCode, buildChatError } from '../utils/ai-utils';
 import { seedAiBaselines } from '../utils/project/ls-schema-notifications';
+import { mapWithConcurrency } from '../utils/concurrency';
 import { getSystemPrompt, getUserPrompt } from './prompts';
 import { FollowupSituation, startFollowupSuggestions } from './followups';
 import { prepareAgentsMdForTurn } from './agents-md';
@@ -307,6 +308,18 @@ export class AgentExecutor extends AICommandExecutor<GenerateAgentCodeRequest> {
                 await seedAiBaselines(tempProjectPath, projects);
             } else {
                 console.log(`[AgentExecutor] Skipping ai:// baseline seed (skipFreshProjectSetup)`);
+            }
+
+            // Fire-and-forget: warm the LS module-package caches for each package's
+            // dependencies while the generation streams, so the first review-diff diagram
+            // does not pay the one-time dependency resolution/compilation cost (seconds)
+            // interactively. Failures are irrelevant — the fetch path pays the cost lazily.
+            for (const project of projects) {
+                const pkgRoot = project.packagePath
+                    ? path.join(tempProjectPath, project.packagePath)
+                    : tempProjectPath;
+                StateMachine.langClient().prewarmDependencies({ projectPath: pkgRoot })
+                    .catch(() => { /* best-effort warm-up only */ });
             }
 
             const workspaceId = this.config.executionContext.workspacePath || this.config.executionContext.projectPath;
@@ -1073,9 +1086,15 @@ Generation stopped by user. The last in-progress task was not saved. Any complet
             const isWorkspace = StateMachine.context().projectInfo?.projectKind === PROJECT_KIND.WORKSPACE_PROJECT;
             let semanticDiffError: string | undefined;
             let diffedPackageCount = 0;
-            for (const pkg of affectedPackages) {
+            // Each package's diff is an independent LS request against its own project root,
+            // so fetch them concurrently; results are folded back in affectedPackages order
+            // below to keep diffPackageMap/semanticDiffs alignment and error-message order
+            // deterministic. Bounded, because each request can trigger two full package
+            // compilations and migration turns can touch many packages at once.
+            const diffPackages = affectedPackages.filter(
                 // Skip workspace root — it only contains Ballerina.toml, not a real package
-                if (isWorkspace && pkg === workingProjectPath) { continue; }
+                pkg => !(isWorkspace && pkg === workingProjectPath));
+            const packageResults = await mapWithConcurrency(diffPackages, 3, async pkg => {
                 const pkgName = path.basename(pkg);
                 try {
                     const res = await langClient.getSemanticDiff({ projectPath: pkg });
@@ -1085,23 +1104,29 @@ Generation stopped by user. The last in-progress task was not saved. Any complet
                     if (res?.errorMsg) {
                         throw new Error(res.errorMsg);
                     }
-                    if (res) {
-                        diffedPackageCount++;
-                        diffPackageMap.push(...Array(res.semanticDiffs.length).fill(pkgName));
-                        semanticDiffs.push(...res.semanticDiffs);
-                        loadDesignDiagrams = loadDesignDiagrams || res.loadDesignDiagrams;
-                        // Partial success: diffs are valid but the package failed to compile,
-                        // so flow diagrams will likely be unavailable. Keep the diffs and
-                        // surface the reason as a warning.
-                        semanticDiffError = semanticDiffError ?? res.compilationError ?? undefined;
-                    }
+                    return { pkgName, res, error: undefined as string | undefined };
                 } catch (err) {
-                    // One package failing must not discard the diffs already collected for
-                    // its siblings — keep going and report the failure alongside them.
+                    // One package failing must not discard the diffs collected for its
+                    // siblings — record the failure and report it alongside them.
                     console.error(`[AgentExecutor] getSemanticDiff failed for package ${pkg}; keeping other packages' diffs`, err);
                     const message = err instanceof Error ? err.message : String(err);
-                    const packageError = isWorkspace ? `${pkgName}: ${message}` : message;
-                    semanticDiffError = semanticDiffError ? `${semanticDiffError}\n${packageError}` : packageError;
+                    return { pkgName, res: undefined, error: isWorkspace ? `${pkgName}: ${message}` : message };
+                }
+            });
+            for (const { pkgName, res, error } of packageResults) {
+                if (error !== undefined) {
+                    semanticDiffError = semanticDiffError ? `${semanticDiffError}\n${error}` : error;
+                    continue;
+                }
+                if (res) {
+                    diffedPackageCount++;
+                    diffPackageMap.push(...Array(res.semanticDiffs.length).fill(pkgName));
+                    semanticDiffs.push(...res.semanticDiffs);
+                    loadDesignDiagrams = loadDesignDiagrams || res.loadDesignDiagrams;
+                    // Partial success: diffs are valid but the package failed to compile,
+                    // so flow diagrams will likely be unavailable. Keep the diffs and
+                    // surface the reason as a warning.
+                    semanticDiffError = semanticDiffError ?? res.compilationError ?? undefined;
                 }
             }
 
--- a/packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
@@ -1085,6 +1085,9 @@ Generation stopped by user. The last in-progress task was not saved. Any complet
                 : await determineAffectedPackages(accumulatedModifiedFiles, context.projects, context.ctx, workingProjectPath);
             const isWorkspace = StateMachine.context().projectInfo?.projectKind === PROJECT_KIND.WORKSPACE_PROJECT;
             let semanticDiffError: string | undefined;
+            const appendDiffError = (msg: string) => {
+                semanticDiffError = semanticDiffError ? `${semanticDiffError}\n${msg}` : msg;
+            };
             let diffedPackageCount = 0;
             // Each package's diff is an independent LS request against its own project root,
             // so fetch them concurrently; results are folded back in affectedPackages order
@@ -1115,7 +1118,7 @@ Generation stopped by user. The last in-progress task was not saved. Any complet
             });
             for (const { pkgName, res, error } of packageResults) {
                 if (error !== undefined) {
-                    semanticDiffError = semanticDiffError ? `${semanticDiffError}\n${error}` : error;
+                    appendDiffError(error);
                     continue;
                 }
                 if (res) {
@@ -1125,8 +1128,11 @@ Generation stopped by user. The last in-progress task was not saved. Any complet
                     loadDesignDiagrams = loadDesignDiagrams || res.loadDesignDiagrams;
                     // Partial success: diffs are valid but the package failed to compile,
                     // so flow diagrams will likely be unavailable. Keep the diffs and
-                    // surface the reason as a warning.
-                    semanticDiffError = semanticDiffError ?? res.compilationError ?? undefined;
+                    // surface the reason as a warning — appended so no package's reason
+                    // is dropped.
+                    if (res.compilationError) {
+                        appendDiffError(res.compilationError);
+                    }
                 }
             }
 
--- a/packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
@@ -1040,22 +1040,34 @@ Generation stopped by user. The last in-progress task was not saved. Any complet
                 ? cachedAffectedPackages
                 : await determineAffectedPackages(accumulatedModifiedFiles, context.projects, context.ctx, workingProjectPath);
             const isWorkspace = StateMachine.context().projectInfo?.projectKind === PROJECT_KIND.WORKSPACE_PROJECT;
+            let semanticDiffError: string | undefined;
             for (const pkg of affectedPackages) {
                 // Skip workspace root — it only contains Ballerina.toml, not a real package
                 if (isWorkspace && pkg === workingProjectPath) { continue; }
                 const pkgName = path.basename(pkg);
                 try {
                     const res = await langClient.getSemanticDiff({ projectPath: pkg });
+                    // errorMsg means the LS could not compute diffs at all (e.g. the package
+                    // fails to compile) — semanticDiffs is absent, so treat it as a failure
+                    // instead of reading past it.
+                    if (res?.errorMsg) {
+                        throw new Error(res.errorMsg);
+                    }
                     if (res) {
                         diffPackageMap.push(...Array(res.semanticDiffs.length).fill(pkgName));
                         semanticDiffs.push(...res.semanticDiffs);
                         loadDesignDiagrams = loadDesignDiagrams || res.loadDesignDiagrams;
+                        // Partial success: diffs are valid but the package failed to compile,
+                        // so flow diagrams will likely be unavailable. Keep the diffs and
+                        // surface the reason as a warning.
+                        semanticDiffError = semanticDiffError ?? res.compilationError ?? undefined;
                     }
                 } catch (err) {
                     console.error(`[AgentExecutor] getSemanticDiff failed for package ${pkg}, falling back to plain modifiedFiles`, err);
                     semanticDiffs.length = 0;
                     diffPackageMap.length = 0;
                     loadDesignDiagrams = false;
+                    semanticDiffError = err instanceof Error ? err.message : String(err);
                     break;
                 }
             }
@@ -1070,6 +1082,7 @@ Generation stopped by user. The last in-progress task was not saved. Any complet
                 modifiedFiles: accumulatedModifiedFiles,
                 tempProjectPath: workingProjectPath,
                 isWorkspace,
+                semanticDiffError,
             };
 
             approvalViewManager.openReviewMode(context.messageId, reviewData, false);
@@ -1082,7 +1095,7 @@ Generation stopped by user. The last in-progress task was not saved. Any complet
                 tempProjectPath: workingProjectPath,
                 modifiedFiles: accumulatedModifiedFiles,
                 affectedPackagePaths: affectedPackages,
-                reviewView: { semanticDiffs, loadDesignDiagrams, isWorkspace },
+                reviewView: { semanticDiffs, loadDesignDiagrams, isWorkspace, semanticDiffError },
             });
 
             context.eventHandler({
--- a/packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
@@ -22,7 +22,7 @@ import { StateMachine } from '../../../stateMachine';
 import { ModelMessage, stepCountIs, streamText, TextStreamPart } from 'ai';
 import { getAnthropicClient, getProviderCacheControl, getProviderModelOptions, addCacheControlToMessages, ANTHROPIC_SONNET } from '../utils/ai-client';
 import { populateHistoryForAgent, getErrorMessage, getErrorCode, buildChatError } from '../utils/ai-utils';
-import { sendAgentDidOpenForFreshProjects } from '../utils/project/ls-schema-notifications';
+import { seedAiBaselines } from '../utils/project/ls-schema-notifications';
 import { getSystemPrompt, getUserPrompt } from './prompts';
 import { FollowupSituation, startFollowupSuggestions } from './followups';
 import { prepareAgentsMdForTurn } from './agents-md';
@@ -147,6 +147,18 @@ function computeTokenBreakdown(
 }
 
 /**
+ * Normalizes a relative path for package-membership comparison: trims, collapses `.`
+ * segments, converts backslashes, and strips leading `./` and trailing slashes. The two
+ * sides being compared come from different authors (the workspace toml's `packages` list
+ * vs the LLM's verbatim tool `file_path` args), so raw string comparison misses trivially
+ * equivalent forms like `./orders` vs `orders/main.bal`.
+ */
+export function normalizeRelativePath(p: string): string {
+    const normalized = path.normalize(p.trim()).replace(/\\/g, '/').replace(/\/+$/, '');
+    return normalized === '.' ? '' : normalized;
+}
+
+/**
  * Determines which packages have been affected by analyzing modified files
  * Returns temp directory package paths for use with Language Server semantic diff API
  * @param modifiedFiles Array of relative file paths that were modified
@@ -173,18 +185,23 @@ async function determineAffectedPackages(
         return Array.from(affectedPackages);
     }
 
-    // Re-read workspace Ballerina.toml from temp to get the current package list
-    // (the agent may have added new packages during the session)
+    // Union of the current workspace Ballerina.toml package list (re-read so packages the
+    // agent added mid-run are present) and the generation-start ProjectSource packages.
+    // An empty toml `packages` array must fall back too — `??` alone doesn't cover it.
     const workspaceToml = await getWorkspaceTomlValues(tempProjectPath);
-    const packagePaths: string[] = workspaceToml?.workspace?.packages ?? projects.map(p => p.packagePath).filter(p => p !== "");
+    const tomlPackages = (workspaceToml?.workspace?.packages ?? []).map(normalizeRelativePath);
+    const sourcePackages = projects.map(p => normalizeRelativePath(p.packagePath ?? ''));
+    const packagePaths: string[] = Array.from(new Set([...tomlPackages, ...sourcePackages])).filter(p => p !== '');
 
     // For workspace scenario with multiple packages
     // We need to map modified files to their temp package paths
+    const unmatchedFiles: string[] = [];
     for (const modifiedFile of modifiedFiles) {
+        const normalizedFile = normalizeRelativePath(modifiedFile);
         let matched = false;
 
         for (const pkgPath of packagePaths) {
-            if (modifiedFile.startsWith(pkgPath + '/') || modifiedFile === pkgPath) {
+            if (normalizedFile.startsWith(pkgPath + '/') || normalizedFile === pkgPath) {
                 const tempPackagePath = path.join(tempProjectPath, pkgPath);
                 affectedPackages.add(tempPackagePath);
                 matched = true;
@@ -196,8 +213,19 @@ async function determineAffectedPackages(
         if (!matched) {
             // File at workspace root (e.g. root Ballerina.toml)
             affectedPackages.add(tempProjectPath);
+            unmatchedFiles.push(modifiedFile);
             console.log(`[determineAffectedPackages] File '${modifiedFile}' is at workspace root (temp): ${tempProjectPath}`);
         }
+    }
+
+    // Unmatched .bal files are a red flag: the workspace root holds no sources, so their
+    // package's diff would silently never be computed. Loud log so this is diagnosable.
+    const unmatchedBalFiles = unmatchedFiles.filter(f => f.endsWith('.bal'));
+    if (unmatchedBalFiles.length > 0) {
+        console.error(
+            `[determineAffectedPackages] ${unmatchedBalFiles.length} modified .bal file(s) matched no workspace package — their diffs will be missing.`,
+            { unmatchedBalFiles, packagePaths }
+        );
     }
 
     const result = Array.from(affectedPackages);
@@ -271,9 +299,12 @@ export class AgentExecutor extends AICommandExecutor<GenerateAgentCodeRequest> {
             );
 
             // 2. Seed the ai:// baseline, unless the caller explicitly asked to skip it
-            // (migration reusing the same directory across sequential stages).
+            // (migration reusing the same directory across sequential stages). Awaited so
+            // the baseline is guaranteed in place before the first edit can touch disk —
+            // the LS-side ensureAiBaseline request applies the pre-edit contents before
+            // responding, unlike the old fire-and-forget didOpen seed.
             if (!this.config.lifecycle?.skipFreshProjectSetup) {
-                sendAgentDidOpenForFreshProjects(tempProjectPath, projects);
+                await seedAiBaselines(tempProjectPath, projects);
             } else {
                 console.log(`[AgentExecutor] Skipping ai:// baseline seed (skipFreshProjectSetup)`);
             }
@@ -1041,6 +1072,7 @@ Generation stopped by user. The last in-progress task was not saved. Any complet
                 : await determineAffectedPackages(accumulatedModifiedFiles, context.projects, context.ctx, workingProjectPath);
             const isWorkspace = StateMachine.context().projectInfo?.projectKind === PROJECT_KIND.WORKSPACE_PROJECT;
             let semanticDiffError: string | undefined;
+            let diffedPackageCount = 0;
             for (const pkg of affectedPackages) {
                 // Skip workspace root — it only contains Ballerina.toml, not a real package
                 if (isWorkspace && pkg === workingProjectPath) { continue; }
@@ -1054,6 +1086,7 @@ Generation stopped by user. The last in-progress task was not saved. Any complet
                         throw new Error(res.errorMsg);
                     }
                     if (res) {
+                        diffedPackageCount++;
                         diffPackageMap.push(...Array(res.semanticDiffs.length).fill(pkgName));
                         semanticDiffs.push(...res.semanticDiffs);
                         loadDesignDiagrams = loadDesignDiagrams || res.loadDesignDiagrams;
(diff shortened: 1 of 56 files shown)


packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
```diff
@@ -118,8 +227,49 @@ public static String cachedVersion(String org, String name) {
...
+    public static PackageRepository localPackageRepository() {
+        return LOCAL_REPOSITORY.get();
+    }
+
+    /**
+     * Whether {@code buildProject} is the calling thread's sample project (i.e. its resolution is
+     * memoizable). Every caller passes {@link #getSampleProject()} on the same thread, so an
+     * identity compare against this thread's {@code ThreadLocal} value is exact and needs no registry.
+     */
+    private static boolean isSampleProject(BuildProject buildProject) {
```
dulajdilshan: **The `BuildProject` parameter on `getModulePackage` is now dead weight — and reading it has a side effect.**

Every caller of `getModulePackage` in the repo passes `PackageUtil.getSampleProject()` — `CopilotLibraryManager` (2), `EvalTemplateSearchCommand`, `servicemodelgenerator/util/Utils`, `pullModuleAndNotify`, plus the tests. There is not one real-project caller. So `isSampleProject` never returns `false`, and the uncached "caller's real project" branch in both overloads is unreachable.

The bigger problem is this identity compare itself: `SAMPLE_PROJECT.get()` *instantiates* the calling thread's sample project as a side effect. Any thread that ever reaches `getModulePackage` — including one that only meant to pass its own project — builds a full resolver environment just to answer a boolean.

This PR already did the right thing for `getModulePackageOffline`: dropped the parameter, resolved internally. Doing the same for both `getModulePackage` overloads deletes `isSampleProject` entirely, removes the accidental instantiation, and turns the thread-confinement contract from something documented into something a caller cannot state wrongly.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
```diff
@@ -87,7 +90,48 @@ private static BuildOptions balaBuildOptions() {
...
+     * own sample project (hence its own environment, resolver and cache) removes the shared mutable
+     * state entirely, so these hot, shared LS paths need no locking. The expensive part — avoiding
+     * a Central round trip per lookup — is preserved by {@link #SAMPLE_RESOLUTION_CACHE} below,
+     * which stays shared because it holds only immutable data (resolved bala paths).
+     *
+     * <p>This is the app-side fix for the shared-resolver hazard. If the compiler later makes the
+     * resolver / package cache concurrency-safe, this can collapse back to a single shared instance
+     * with no change to callers.
+     *
+     * <p>Cost/lifecycle: the per-thread project is a lightweight in-memory load of the single shared
```
dulajdilshan: **ThreadLocal footprint on a pooled LS is larger than this phrasing suggests.**

"reclaimed by GC when the thread dies ... live memory tracks the number of threads currently resolving" is technically correct but reads as a transient cost. The LS runs long-lived thread pools, so in practice live memory is (pool width x one lang-lib resolver environment) for the whole process lifetime — those threads don't die. Combined with the accidental instantiation from `isSampleProject` (see my comment there), the resident set could be wider than the set of threads that actually do sample resolution.

The design choice is right — removing shared mutable state beats locking it. But could we get one heap measurement (LS steady state, before/after) into the PR before merge? That's the number the language-server team will want when deciding whether this is the permanent shape or a bridge until the compiler-side cache is made thread-safe.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
```diff
@@ -96,13 +140,78 @@ private static BuildOptions balaBuildOptions() {
...
+     */
+    private static Optional<Path> memoizedSampleBala(String key, Supplier<Optional<Path>> resolution) {
+        Optional<Path> cached = SAMPLE_RESOLUTION_CACHE.get(key);
+        if (cached != null) {
+            return cached;
+        }
+        Optional<Path> resolved;
+        try {
+            resolved = resolution.get();
+        } catch (RuntimeException e) {
```
dulajdilshan: **Silent `RuntimeException` swallow.**

```java
} catch (RuntimeException e) {
    return Optional.empty();
}
```

No log, no `LSClientLogger`. A resolver failure — a corrupted cache, a malformed descriptor, an IO error surfaced as unchecked — becomes indistinguishable from "package legitimately absent", and the caller has no way to tell them apart.

That is the same silent-empty failure mode that payload 1 of this very PR exists to fix on the review side (`surface package-compilation failures ... instead of silently producing an empty diff`). Worth applying the same principle here: log at debug/warn with the coordinate key before returning empty.

packages/ballerina-language-server/spotbugs-exclude.xml
```diff
@@ -265,4 +265,11 @@
         <Class name="io.ballerina.projectservice.core.baltool.BalToolsUtil" />
         <Bug pattern="DP_CREATE_CLASSLOADER_INSIDE_DO_PRIVILEGED"/>
     </Match>
+    <Match>
+        <!-- Returning the shared sample project is the point: it is a process-wide memoized
+             instance whose per-call re-creation defeated resolution caching (see PackageUtil). -->
+        <Class name="io.ballerina.modelgenerator.commons.PackageUtil" />
+        <Method name="getSampleProject" />
+        <Bug pattern="MS_EXPOSE_REP"/>
```
dulajdilshan: **This justification now contradicts the fix.**

> Returning the shared sample project is the point: it is a **process-wide memoized instance** whose per-call re-creation defeated resolution caching

That was true when this exclusion was written (as part of the #855 caching work). This PR's whole point is that it is *no longer* a process-wide instance — it is per-thread. The comment now tells the next reader the opposite of what the code does.

The suppression may still be warranted, but the reason has to be rewritten to the new one: the returned project is thread-confined and the contract is documented on the getter. And if the `getModulePackage` parameter is dropped (see my comment on `isSampleProject`), `getSampleProject` loses its last production caller and this exclusion can probably go away entirely.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/test/java/io/ballerina/flowmodelgenerator/core/PackageUtilConcurrencyTest.java
```diff
@@ -0,0 +1,230 @@
...
+                    try {
+                        startLine.await(); // release all threads at once to maximise contention
+                        for (int i = 0; i < ITERATIONS && !stop.get(); i++) {
+                            // Exercise the load path (getModulePackage*, each of which builds a fresh
+                            // lang-lib environment) only on the first pass: that is where every thread
+                            // first-misses at once and hits the resolver. Doing it every pass would
+                            // allocate thousands of environments and exhaust the test heap without
+                            // adding coverage (later passes are memoized path-cache hits anyway). The
+                            // cheap metadata methods run every pass to sustain resolver contention.
+                            boolean exerciseLoads = i == 0;
```
dulajdilshan: **The load paths are not actually exercised concurrently — the cache is already warm.**

`baselineFor()` runs single-threaded over every coordinate before the pool starts, and it calls `getModulePackageOffline` (and `cachedVersion`/`isModuleUnresolved`) for each one. That populates `SAMPLE_RESOLUTION_CACHE` for the whole `COORDS` list. By the time the 16 threads are released, every `getModulePackageOffline` / `getModulePackage` call is a memoized path-cache hit that never reaches a resolver.

So this comment is inaccurate:

> that is where every thread first-misses at once and hits the resolver

There is no first-miss — the baseline consumed it. Only `cachedVersion` and `isModuleUnresolved` (both unmemoized) reach the thread-local resolver under contention; the `exerciseLoads` branch is measuring the `ConcurrentHashMap`.

Two options: clear `SAMPLE_RESOLUTION_CACHE` between the baseline and the concurrent phase (needs a test hook), or keep it as-is and correct the comment.

Either way — the class javadoc is admirably honest that this "does not by itself prove the race is present or absent", and I'd ask that the same honesty reach the PR description, which currently presents `PackageUtilConcurrencyTest passes` as the verification for this fix. With a warm cache it is a smoke test for the metadata paths, not a guard on the resolution path the bug lives in.

packages/ballerina-extension/src/views/ai-panel/checkpoint/checkpointUtils.ts
```diff
@@ -54,27 +54,39 @@ export async function captureWorkspaceSnapshot(messageId: string): Promise<Check
...
-                    // later-scanned files unrevertible on restore.
-                    console.warn(`[Checkpoint] Snapshot size exceeded max limit (${totalSize} bytes) — aborting capture, no partial checkpoint will be saved`);
-                    vscode.window.showWarningMessage(
-                        `Workspace is too large to checkpoint (exceeds ${Math.round(config.maxSnapshotSize / 1024 / 1024)}MB). This generation will not be revertible.`
-                    );
+        // Read in bounded-concurrency chunks: one-at-a-time reads dominate run-start
+        // latency on large workspaces, while unbounded Promise.all can exhaust file
+        // handles. Chunks keep fileList in findFiles order and let the size cap abort
+        // between chunks.
+        const READ_CONCURRENCY = 32;
```
dulajdilshan: **Third bounded-concurrency idiom in a PR that introduces a shared helper for exactly this.**

`AgentExecutor.ts` and `ls-schema-notifications.ts` both use the new `mapWithConcurrency`; this site hand-rolls chunked `Promise.all`. I can see why — the size cap needs to abort between batches, which `mapWithConcurrency` has no way to express — but that reasoning should be in the comment, or the helper should grow an abort hook, otherwise the next person reasonably copies whichever one they find first.

One real behaviour change worth noting: the cap is now checked once per chunk rather than per file, so capture reads and buffers up to 31 files past `maxSnapshotSize` before deciding to abort. Fine for a 32-file chunk of source, but it is a genuine (small) regression in the fail-closed guarantee the comment above describes.

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
```diff
@@ -1040,26 +1084,69 @@ Generation stopped by user. The last in-progress task was not saved. Any complet
...
+            // Each package's diff is an independent LS request against its own project root,
+            // so fetch them concurrently; results are folded back in affectedPackages order
+            // below to keep diffPackageMap/semanticDiffs alignment and error-message order
+            // deterministic. Bounded, because each request can trigger two full package
+            // compilations and migration turns can touch many packages at once.
+            const diffPackages = affectedPackages.filter(
                 // Skip workspace root — it only contains Ballerina.toml, not a real package
-                if (isWorkspace && pkg === workingProjectPath) { continue; }
+                pkg => !(isWorkspace && pkg === workingProjectPath));
+            const packageResults = await mapWithConcurrency(diffPackages, 3, async pkg => {
```
dulajdilshan: Nit: the concurrency limits are bare literals here (`3`) and in `ls-schema-notifications.ts` (`4`), both tuning constants for the same underlying resource (LS compilation fan-out). Worth naming them — `REVIEW_DIFF_CONCURRENCY` and friends — so the relationship between the two is visible and they can be tuned together.

Separately, on the fold-back loop below: `semanticDiffError = semanticDiffError ?? res.compilationError ?? undefined` keeps only the first message and blends two different severities into one string — a hard "this package could not be diffed at all" failure and a soft "diffs are valid but flow diagrams may be unavailable" warning end up in the same field. The UI then can't tell the user which one they are looking at. Not blocking, but a `{severity, message}` shape would carry what the comment already knows.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/copilotagent/core/SemanticDiffComputer.java
```diff
@@ -90,8 +98,64 @@ public SemanticDiffComputer(Project originalProject,
...
+        Throwable cause = throwable;
+        while (cause instanceof CompletionException
+                && cause.getCause() != null && cause.getCause() != cause) {
+            cause = cause.getCause();
+        }
+        String message = cause.getMessage();
+        return message == null || message.isBlank() ? cause.getClass().getName() : message;
+    }
+
+    private static Map<String, Module> collectModules(Project project) {
```
dulajdilshan: Architectural note, not blocking this PR: `SemanticDiffComputer` is now **1,118 lines across 41 methods**, and this is the second PR to grow it.

The per-construct diffing — imports, classes, functions, types, constants, enums, plus `nonMethodMembersSource`, the metadata builders and the addition/deletion emitters — is a natural strategy set keyed by `NodeKind`, which would also make `STNodeRefMap`'s parallel `Map<String, X>`-per-kind fields fall out as one registry instead of six hand-maintained ones. Adding a construct kind currently means touching both classes in matching places.

Please don't fold that into this PR — it is already a re-land plus an unrelated fix. But it should happen before a third feature lands here.

packages/ballerina-extension/src/rpc-managers/ai-panel/rpc-manager.ts
```diff
@@ -551,12 +551,25 @@ export class AiPanelRpcManager implements AIPanelAPI {
...
+            if (!checkpoint) {
+                const reason = "No checkpoint exists for this generation (checkpoints may be disabled, or the workspace exceeded the snapshot size limit), so the changes cannot be reverted automatically. Use source control to undo them if needed.";
+                console.error(`[Review Actions] Revert refused for ${doneGeneration.id}: ${reason}`);
+                window.showErrorMessage(`Could not revert the Copilot changes: ${reason}`);
+                throw new Error(reason);
+            }
+            // restoreWorkspaceSnapshot reports its own failures to the user but does not throw, so a
+            // failed applyEdit must not fall through to marking the generation reverted and telling
+            // the model the files were restored when they weren't — the same refusal as no checkpoint.
+            const restored = await restoreWorkspaceSnapshot(checkpoint, true);
```
dulajdilshan, line 568: **The same bug is still at line 799.** `revertToCheckpoint` ignores the flag you just added:

```ts
await restoreWorkspaceSnapshot(checkpoint);          // returns false on failure — discarded
const restored = chatStateStorage.restoreThreadToCheckpoint(...);   // runs anyway
```

If the restore fails, the chat history is truncated while the files stay unchanged. That is worse than the case fixed here, because truncating history can't be undone.

Same fix:

```ts
const restored = await restoreWorkspaceSnapshot(checkpoint);
if (!restored) {
    throw new Error("Restoring the workspace failed; the conversation was not rewound.");
}
```

The third caller (`activator.ts:185`) is a test command — fine to leave.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/copilotagent/core/SemanticDiffComputer.java
```diff
@@ -124,100 +188,188 @@ public Result computeSemanticDiffs() {
...
+
+    /**
+     * Signature used to decide whether a class needs a {@code CLASS_DEFINITION} diff once its methods
+     * have been diffed separately. It combines the class header — metadata, visibility, class-type
+     * qualifiers, the {@code class} keyword and the name — with the non-method members, so a
+     * header-only change (for example adding {@code isolated} or an annotation) is not lost when the
+     * members are unchanged. Qualifier/keyword/name tokens use {@code text()} to stay trivia-insensitive,
+     * matching the trivia-only handling elsewhere in this computer.
+     */
+    private static String classHeaderAndMembersSource(ClassDefinitionNode classNode) {
```
dulajdilshan, line 365: **`metadata()` uses `toSourceCode()`, so this is trivia-sensitive.** Every other part uses `text()`:

```java
classNode.metadata().ifPresent(m -> header.append(m.toSourceCode().trim()));      // includes trivia
classNode.visibilityQualifier().ifPresent(q -> header.append(q.text().trim()));   // no trivia
```

Result: editing a comment inside a class's doc comment now produces a `CLASS_DEFINITION` diff with no real change behind it. That's the same false positive `4c128bbcb2` removes for imports.

Fix by reading metadata's tokens with `text()`, or narrow the javadoc claim above (it currently says this is trivia-insensitive).

Minor: there's no separator between metadata and the qualifier — `@Foo` + `public` becomes `@Foopublic class A`. Harmless for a comparison key, but a delimiter would be cleaner.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
```diff
@@ -118,8 +234,49 @@ public static String cachedVersion(String org, String name) {
...
+    public static PackageRepository localPackageRepository() {
+        return LOCAL_REPOSITORY.get();
+    }
+
+    /**
+     * Whether {@code buildProject} is the calling thread's sample project (i.e. its resolution is
+     * memoizable). Every caller passes {@link #getSampleProject()} on the same thread, so an
+     * identity compare against this thread's {@code ThreadLocal} value is exact and needs no registry.
+     */
+    private static boolean isSampleProject(BuildProject buildProject) {
```
dulajdilshan, line 264: **The `BuildProject` parameter can be dropped, and reading it has a side effect.**

I checked all 7 callers of `getModulePackage` — every one passes `getSampleProject()`. So `isSampleProject` always returns `true`, and the "real project" branch in both overloads is dead code.

The side effect: `SAMPLE_PROJECT.get()` *creates* the thread's sample project. So a thread that only wanted to pass its own project builds an entire resolver environment just to evaluate a boolean.

You already solved this for `getModulePackageOffline` — dropped the parameter, resolve internally. Doing the same for both `getModulePackage` overloads:

- deletes `isSampleProject` and the accidental instantiation
- makes the thread-confinement contract impossible to get wrong, instead of documented
- retires the `MS_EXPOSE_REP` suppression, since `getSampleProject` loses its last production caller

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
```diff
@@ -87,7 +94,48 @@ private static BuildOptions balaBuildOptions() {
...
+     * a Central round trip per lookup — is preserved by {@link #SAMPLE_RESOLUTION_CACHE} below,
+     * which stays shared because it holds only immutable data (resolved bala paths).
+     *
+     * <p>This is the app-side fix for the shared-resolver hazard. If the compiler later makes the
+     * resolver / package cache concurrency-safe, this can collapse back to a single shared instance
+     * with no change to callers.
+     *
+     * <p>Cost/lifecycle: the per-thread project is a lightweight in-memory load of the single shared
+     * {@link #SAMPLE_PROJECT_DIR} (no per-thread temp directory), but its resolver environment does
+     * retain that thread's lang-lib packages. A thread's instance is reclaimed by GC when the thread
```
dulajdilshan, line 119: **"reclaimed by GC when the thread dies" understates the cost on a pooled LS.**

LS threads come from long-lived pools — they don't die. So the real footprint is *pool width x one resolver environment*, held for the life of the process. With the accidental instantiation from `isSampleProject` (see my other comment), that can cover more threads than the ones actually doing sample resolution.

The design is right — removing shared state beats locking it. Could you add one heap measurement (LS at steady state, before/after) to the PR? That's the number the LS team will want when deciding if this stays or gets reverted once the compiler-side cache is fixed.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
```diff
@@ -87,7 +94,48 @@ private static BuildOptions balaBuildOptions() {
...
+     * which is a plain, unsynchronized {@code HashMap}; resolving a package writes into it. While
+     * this was a single shared {@code BuildProject}, concurrent LS requests — flow-model, trigger,
+     * connector and library-metadata generation all resolve here — mutated that one {@code HashMap}
+     * from multiple threads and corrupted it (wso2/product-integrator#2193). Giving each thread its
+     * own sample project (hence its own environment, resolver and cache) removes the shared mutable
+     * state entirely, so these hot, shared LS paths need no locking. The expensive part — avoiding
+     * a Central round trip per lookup — is preserved by {@link #SAMPLE_RESOLUTION_CACHE} below,
+     * which stays shared because it holds only immutable data (resolved bala paths).
+     *
+     * <p>This is the app-side fix for the shared-resolver hazard. If the compiler later makes the
```
dulajdilshan, line 113: **This fixes the sample project, but #2193 asks a broader question that's left open.**

First — the fix itself is complete. I checked all seven call sites in the issue and each one now goes through the ThreadLocal, and `SAMPLE_RESOLUTION_LOCK` is gone.

The issue asked us to decide one of two things:

- `PackageResolver` is unsafe to share -> serialise the seven callers
- it is safe to share -> drop the lock

This PR drops the lock *and* restructures to avoid sharing, so it acts on both answers without recording either. The unsafe-ness of `EnvironmentPackageCache` stays undocumented.

That matters because **the race belongs to the package cache, not to the sample project** — and it's still reachable elsewhere:

```java
// AgentsManagerService.java:424
Project project = this.workspaceManager.loadProject(projectPath);   // cached per path -> shared
PackageResolver resolver = project.projectEnvironmentContext().getService(PackageResolver.class);
resolver.resolvePackageMetadata(...);                               // no lock
```

`BallerinaWorkspaceManager` caches projects per path, so two concurrent requests on one project share an environment and its unsynchronized cache. Same race, real project, and a per-thread sample project doesn't help. (I checked the other two doors: `BallerinaConnectorService:165` builds a fresh environment each call — safe; `LSPackageLoader:176` is one-time init — fine.)

Suggested follow-ups, none of which grow this PR much:

1. File the ballerina-lang issue to make `EnvironmentPackageCache` thread-safe (likely just a `ConcurrentHashMap` + `computeIfAbsent`) and link it here. That turns the "if the compiler later makes the resolver safe..." note into a tracked dependency.
2. Note in #2193 which of the two options this represents, so it isn't closed as fully answered.
3. Decide whether `AgentsManagerService:424` is in scope now or gets its own issue — it shouldn't resurface later as a new bug.

packages/ballerina-extension/src/views/ai-panel/checkpoint/checkpointUtils.ts
dulajdilshan, line 62: **Third way of doing bounded concurrency in one PR.** `AgentExecutor.ts` and `ls-schema-notifications.ts` use the new `mapWithConcurrency`; this one hand-rolls chunked `Promise.all`.

The reason is good — the size cap has to abort between batches, which `mapWithConcurrency` can't express — but it isn't written down, so the next person copies whichever they find first. Either add that to the comment or give the helper an abort hook.

Also worth noting: the cap is now checked per chunk instead of per file, so this reads and buffers up to 31 files past `maxSnapshotSize` before aborting. Small, but it does weaken the fail-closed guarantee described below.

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
dulajdilshan, line 1100: Nit: the limits are bare numbers here (`3`) and in `ls-schema-notifications.ts` (`4`). Both bound the same resource (LS compilation fan-out), so naming them would make that link visible and let them be tuned together.

Separately, in the loop below: `semanticDiffError = semanticDiffError ?? res.compilationError ?? undefined` keeps only the first message and mixes two severities in one field — "this package couldn't be diffed at all" and "diffs are fine but flow diagrams may be missing" end up indistinguishable, so the UI can't tell the user which one happened. A `{severity, message}` shape would carry what the comment already knows.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/copilotagent/core/SemanticDiffComputer.java
dulajdilshan, line 143: Not blocking — an architectural note for later. `SemanticDiffComputer` is now **1,136 lines / 41 methods**, and this is the second PR to grow it.

The per-construct diffing (imports, classes, functions, types, constants, enums, plus the source/metadata helpers) is a natural strategy set keyed by `NodeKind`. That would also collapse `STNodeRefMap`'s six parallel `Map<String, X>` fields into one registry. Today, adding a construct kind means editing both classes in matching places.

Please don't do it in this PR — it's already a re-land plus an unrelated fix. But it's worth doing before a third feature lands here.

packages/ballerina-extension/src/rpc-managers/ai-panel/rpc-manager.ts
IsuruMaduranga: Fixed in eb8c5d07a5 — revertToCheckpoint now checks the restore result and throws before truncating the thread, same guard as the revert-generation path.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/copilotagent/core/SemanticDiffComputer.java
IsuruMaduranga: Fixed in eb8c5d07a5. The comparison key is now built from token text only (a small triviaFreeSource walk), so metadata trivia can't produce a false CLASS_DEFINITION diff, and the space-joined tokens also fix the missing separator. Same fix applied to computeSourceDiffs (listeners, module vars, constants, enums), which had the identical false positive. Displayed source stays toSourceCode().

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
IsuruMaduranga: Agreed it's worth doing — deferring to a follow-up rather than this re-land: it changes a public LS API signature (7 call sites across 4 modules), which also adds upstream-merge friction. Worth noting the accidental instantiation is latent today: every caller already calls getSampleProject() to pass it in, so the thread-local is initialized either way.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
IsuruMaduranga: Fair — pool width × resolver environment is the real steady-state number. Will capture a before/after heap measurement at LS steady state and add it to the PR description before merge.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
IsuruMaduranga: This PR takes the first option: the resolver/package cache is unsafe to share, so sharing is removed rather than serialized. Will file the ballerina-lang issue for EnvironmentPackageCache thread-safety and link it here + note the option taken in #2193. AgentsManagerService:424 should get its own issue rather than growing this PR. One more for the same family, found while reviewing: BallerinaWorkspaceManager.sourceRootToProject is a plain HashMap guarded only by per-project locks, so concurrent loadProject calls for two different roots can race the map — pre-existing LS-wide, but this PR's 3-way getSemanticDiff concurrency raises exposure. Will include it in the issue.

packages/ballerina-extension/src/views/ai-panel/checkpoint/checkpointUtils.ts
IsuruMaduranga: Documented in 97ca32be5b — the comment now says why this is hand-rolled (the size cap must abort between batches, which mapWithConcurrency can't express). On the per-chunk cap: accepted as a trade-off; it reads at most 31 files past the limit before aborting, and the capture still fails closed.

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
IsuruMaduranga: The aggregation half is fixed in eb8c5d07a5 — every package's compilationError is now appended (with a unit test covering two failing packages), not just the first. Naming the two limits and splitting semanticDiffError into {severity, message} are deferred; the severity split touches the UI contract, so it fits better in a follow-up.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/copilotagent/core/SemanticDiffComputer.java
IsuruMaduranga: Agreed — a NodeKind-keyed strategy set is the right shape, and it would collapse STNodeRefMap's parallel maps too. Will keep it out of this PR and do it before a third feature lands in this class.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/main/java/io/ballerina/copilotagent/extension/CopilotAgentService.java
```diff
@@ -56,6 +73,115 @@ public Class<?> getRemoteInterface() {
...
+                            .ifPresent(PackageUtil::getCompilation);
+                    warmed++;
+                }
+                response.setWarmedDependencyCount(warmed);
+            } catch (Exception e) {
+                response.setError(e);
+            }
+            return response;
+        });
+    }
```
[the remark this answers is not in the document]
rosensilva: @IsuruMaduranga shall we address this, please?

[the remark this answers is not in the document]
rosensilva: @IsuruMaduranga please address this as well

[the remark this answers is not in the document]
IsuruMaduranga: This is fixed

dulajdilshan (approved):
LGTM!

rosensilva (approved):

PR #950, opened 2026-08-25
Auto-refresh the WSO2 default provider token before run, debug, and test

dan-niles:
## Purpose

Fixes https://github.com/wso2/product-integrator/issues/2151

Auto-refreshes `Config.toml`'s WSO2 default model/embedding provider access token current across every run/debug/test entry point, so users don't have to manually re-run "Configure default WSO2 model provider" whenever it expires.

## What changed

- `refreshDefaultProviderToken` (`features/ai/utils.ts`) refreshes the `accessToken` in `Config.toml` (and `tests/Config.toml`) before a run, if a fresher token is available.
- Wired into every run/debug/test entry point: `prepareAndGenerateConfig` (covers Run, BI Run, and Debug via `DebugConfigProvider`), the Config.toml "Run" command, the legacy `ballerina.project.test` command, and both Test Explorer implementations (BI and non-BI).
- If the project's source no longer calls `ai:getDefaultModelProvider()` / `ai:getDefaultEmbeddingProvider()`, the stale `[ballerina.ai.wso2ProviderConfig]` entry is removed instead of refreshed, avoiding a later "unused configuration" error from `bal run`/`bal test`.
- Doesn't auto-create the entry: a project that has never run "Configure default WSO2 model provider" is left untouched.

## Summary

- Refreshes WSO2 default provider tokens before run, debug, configuration, and test operations.
- Updates both `Config.toml` files when a newer token is available.
- Removes stale provider settings when projects no longer use the default providers.
- Prompts users to configure the provider when source code uses it without configuration.
- Reuses sign-in and retry handling for token refresh failures.
- Skips affected test runs and reports configuration errors when refresh fails.


packages/ballerina-extension/src/features/ai/utils.ts
```diff
@@ -326,6 +336,192 @@ export async function addConfigFile(
...
+        if (credentials?.loginMethod !== LoginMethod.BI_INTEL) {
+            return true;
+        }
+
+        writeDefaultModelConfigToProject(projectPath, credentials.secrets.accessToken);
+        return true;
+    } catch (error) {
+        console.error('Failed to refresh the WSO2 default model provider token:', error);
+        vscode.window.showWarningMessage(DEFAULT_PROVIDER_TOKEN_REFRESH_FAILED);
+        return true;
```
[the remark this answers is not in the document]
dan-niles: Intentional fail-open: this catch wraps the whole body, so returning false would block runs on transient fs errors or an offline STS refresh (which rejects on a 5-minute pre-expiry buffer, while the token in Config.toml is often still valid) — pre-PR those runs always proceeded.

packages/ballerina-extension/src/features/test-explorer/runner.ts
```diff
@@ -239,10 +240,15 @@ export async function runHandler(request: TestRunRequest, token: CancellationTok
...
     const langClient = extension.ballerinaExtInstance.langClient;
+    const unconfiguredProjectPaths = new Set<string>();
     for (const projectPath of projectPaths) {
         if (token.isCancellationRequested) {
             break;
         }
+        if (!(await refreshDefaultProviderToken(projectPath))) {
+            unconfiguredProjectPaths.add(projectPath);
+            continue;
+        }
```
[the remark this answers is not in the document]
dan-niles: Debug is already covered: startDebugging routes through DebugConfigProvider.resolveDebugConfiguration, which calls prepareAndGenerateConfig at config-provider.ts:125 (outside the debugTests guard) and so already refreshes — adding it here would double-refresh and could prompt twice.

SasinduDilshara (dismissed):

packages/ballerina-extension/src/features/ai/utils.ts
```diff
@@ -326,6 +336,199 @@ export async function addConfigFile(
...
+            return true;
+        }
+
+        const isReferenced = await isDefaultProviderReferencedInSource(projectPath);
+
+        if (!hasConfiguredProviderToken(projectPath)) {
+            return !isReferenced || await promptToConfigureDefaultProvider(projectPath);
+        }
+
+        if (!isReferenced) {
```
madushajg, line 524: `isDefaultProviderReferencedInSource` relies on `getProjectSource`, which currently only scans root `*.bal` files and `modules/*/`. It doesn't include the `generated/` module directory, even though we treat it as a valid module root elsewhere (`ai-utils.ts:122`, `rpc-manager.ts:373`, `temp-project.ts:396`).

As a result, a single reference in an unscanned file can make `isReferenced` return `false`, causing us to silently remove the entire `[ballerina.ai.wso2ProviderConfig]` table—including any additional keys and comments—from both `Config.toml` copies.

Suggested Fix:

* Include `generated/` in the source scan.
* Consider gating this destructive branch behind an explicit signal (or confirmation) rather than relying on a text-based heuristic.

A false positive here only leaves a stale entry, whereas a false negative can result in user configuration being lost.

packages/ballerina-extension/src/features/ai/utils.ts
```diff
@@ -326,6 +336,199 @@ export async function addConfigFile(
...
+    try {
+        // Single-file projects report the .bal file as their path; nothing to scan or refresh.
+        if (!fs.existsSync(projectPath) || !fs.statSync(projectPath).isDirectory()) {
+            return true;
+        }
+
+        const isReferenced = await isDefaultProviderReferencedInSource(projectPath);
+
+        if (!hasConfiguredProviderToken(projectPath)) {
+            return !isReferenced || await promptToConfigureDefaultProvider(projectPath);
```
madushajg, line 521: `hasConfiguredProviderToken` only checks the project's `Config.toml`, so configurations provided via `BAL_CONFIG_FILES`, `BAL_CONFIG_VAR_*`, or CI injection are treated as missing and prompt on every run.

Also, dismissing the notification returns `false` and silently aborts the run, making the **Run** button appear to do nothing.

Suggested Fix:
Proceed when the prompt is dismissed and letting `bal run` report the actual error, or showing an explicit **"Run cancelled"** message with an option to stop prompting.

packages/ballerina-extension/src/features/ai/utils.ts
```diff
@@ -326,6 +336,199 @@ export async function addConfigFile(
...
+        let timeoutHandle: ReturnType<typeof setTimeout> | null = null;
+        const subscription = AIStateMachine.service().subscribe((state) => {
+            if (state.value !== 'Authenticated') {
+                return;
+            }
+            if (timeoutHandle !== null) {
+                clearTimeout(timeoutHandle);
+            }
+            pendingAuthRetries = null;
+            subscription.unsubscribe();
```
madushajg: `xstate` v4 calls the subscription listener synchronously with the current state, so if the machine is already `Authenticated`, `subscription.unsubscribe()` runs before `subscription` is initialized, causing a TDZ `ReferenceError`.

This leaves `pendingAuthRetries` non-null and the timeout is never armed, so subsequent sign-in prompts silently push into a dead array.

Please declare `let subscription` before subscribing (or use a handled flag), and clear `pendingAuthRetries` in a `catch`/`finally`.

packages/ballerina-extension/src/features/ai/utils.ts
dan-niles: `getProjectSource` only walked `modules/*`, never `generated/`, even though `generated/<module>` is a valid module root elsewhere (e.g. `getModuleDirectory` in rpc-manager.ts). Fixed by scanning both dirs in fd0687bd02 (deduped into a shared `readModulesDir` helper).

packages/ballerina-extension/src/features/ai/utils.ts
dan-niles: Fixed in fd0687bd02: `hasConfiguredProviderToken` now short-circuits true when `BAL_CONFIG_FILES` or any `BAL_CONFIG_VAR_*` env var is set (presence check, not a validity one — we don't parse the external file/vars, we just back off and let `bal run` report the real error). Went with your first suggested option there rather than trying to reconstruct Ballerina's env-var-to-config-key mapping, since a false-negative there would be worse than occasionally deferring to `bal run`.

Also added an explicit "Run cancelled: the WSO2 default AI provider isn't configured" warning when the prompt is dismissed, so the Run button no longer appears to silently do nothing.

packages/ballerina-extension/src/features/ai/utils.ts
dan-niles: Fixed in fd0687bd02: `subscription` is now declared with `let` before calling `.subscribe()`, with optional-chained `.unsubscribe()` calls, so an already-`Authenticated` synchronous callback no longer hits the TDZ ReferenceError.

madushajg (approved):

SasinduDilshara (approved):

PR #951, opened 2026-08-25
Classify Try It services by module prefix

RadCod3:
## Purpose
Try It classified services by substring-matching the design model's type name, checking `http` before `mcp`. A streamable-HTTP MCP service is reported as `mcp:StreamableHttpService`, which contains `Http`, so it was classified as HTTP and routed through `getOpenAPIDefinition`. The OpenAPI tool correctly returns no content for an MCP service, so the empty-match guard threw "No service matches the base path" and the MCP Inspector was never opened.

The design model always reports module-qualified type names, in both generator paths: the explicit service type descriptor, and the fallback that builds `<listener module>:Service`. This PR updates the logic to match on that module prefix instead. Verified against the bundled language server jar that `http:Service`, `graphql:Service`, `ai:Service`, `mcp:Service` and `mcp:StreamableHttpService` all resolve as expected, with only the last changing behaviour.

Manually tested "Try It" for the 4 service types as well.

Fixes https://github.com/wso2/product-integrator/issues/2177

## Summary

- Updated Try It service classification to use the module prefix from the service type.
- Added shared `ServiceType` and `resolveServiceType` logic.
- Added unit tests for supported, unsupported, and comment-prefixed service types.
- Correctly routes `mcp:StreamableHttpService` to the MCP Inspector.

### DIFF
--- a/packages/ballerina-extension/src/features/tryit/activator.ts
+++ b/packages/ballerina-extension/src/features/tryit/activator.ts
@@ -35,6 +35,7 @@ import { getCurrentProjectRoot } from "../../utils/project-utils";
 import { requiresPackageSelection, selectIntegrationOrPrompt } from "../../utils/command-utils";
 import { VisualizerWebview } from "../../views/visualizer/webview";
 import { TracerMachine } from "../tracing";
+import { resolveServiceType, ServiceType } from "./service-type";
 
 // File constants
 const FILE_NAMES = {
@@ -416,29 +417,6 @@ async function findServiceForResource(services: ServiceInfo[], resourceMetadata:
     }
 }
 
-// The design model reports service types as module-qualified names (e.g. 'mcp:StreamableHttpService'),
-// so the module prefix is what identifies the kind. A substring match on the type name would classify
-// MCP services as HTTP, since 'StreamableHttpService' contains 'Http'.
-function resolveServiceType(type: string): ServiceType | undefined {
-    // The design model omits the type for services whose listener could not be resolved.
-    if (!type) {
-        return undefined;
-    }
-
-    switch (type.split(':')[0].trim().toLowerCase()) {
-        case 'http':
-            return ServiceType.HTTP;
-        case 'graphql':
-            return ServiceType.GRAPHQL;
-        case 'mcp':
-            return ServiceType.MCP;
-        case 'ai':
-            return ServiceType.AGENT;
-        default:
-            return undefined;
-    }
-}
-
 async function getAvailableServices(projectDir: string): Promise<ServiceInfo[] | null> {
     try {
         const langClient = StateMachine.langClient();
@@ -455,7 +433,7 @@ async function getAvailableServices(projectDir: string): Promise<ServiceInfo[] |
                 const trimmedPath = absolutePath.trim();
                 const name = displayName || (trimmedPath.startsWith('/') ? trimmedPath.substring(1) : trimmedPath);
 
-                const serviceType = resolveServiceType(type);
+                const serviceType = resolveServiceType(type)!;
 
                 const listener = {
                     name: attachedListeners
@@ -1086,13 +1064,6 @@ function disposeErrorWatcher() {
 }
 
 // Service information interface
-enum ServiceType {
-    HTTP = 'HTTP',
-    AGENT = 'AI Agent',
-    GRAPHQL = 'GraphQL',
-    MCP = 'MCP'
-}
-
 interface ServiceInfo {
     name?: string;
     basePath: string;
--- a/packages/ballerina-extension/src/features/tryit/activator.ts
+++ b/packages/ballerina-extension/src/features/tryit/activator.ts
@@ -416,6 +416,29 @@ async function findServiceForResource(services: ServiceInfo[], resourceMetadata:
     }
 }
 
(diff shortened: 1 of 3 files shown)


packages/ballerina-extension/src/features/tryit/activator.ts
```diff
@@ -416,6 +416,29 @@ async function findServiceForResource(services: ServiceInfo[], resourceMetadata:
     }
 }
 
+// The design model reports service types as module-qualified names (e.g. 'mcp:StreamableHttpService'),
+// so the module prefix is what identifies the kind. A substring match on the type name would classify
+// MCP services as HTTP, since 'StreamableHttpService' contains 'Http'.
+function resolveServiceType(type: string): ServiceType | undefined {
```
NipunaRanasinghe: Nit: this is a pure function, so it would be easy to unit test, but it isn't exported so there's no way to reach it from a test. Not asking you to change that here. If you want the MCP case pinned against future regressions, an MCP spec next to `e2e-playwright-tests/tryit/http-try-it-existing.spec.ts` would do it. Optional, given you tested all four types by hand.

packages/ballerina-extension/src/features/tryit/activator.ts
```diff
@@ -416,6 +416,29 @@ async function findServiceForResource(services: ServiceInfo[], resourceMetadata:
...
+    // The design model omits the type for services whose listener could not be resolved.
+    if (!type) {
+        return undefined;
+    }
+
+    switch (type.split(':')[0].trim().toLowerCase()) {
```
NipunaRanasinghe: The `trim()` is needed because the anon-listener path in the design model builds the type from `typeDescriptor().toSourceCode()` (`CodeAnalyzer.java:215`), which carries the leading trivia. Worth noting it only covers whitespace, so something like `on new // comment\n    http:Listener(8080)` would still fall through to `default`. The old substring match happened to survive that. Contrived enough that I'd leave it alone, just flagging so it isn't a surprise later.

packages/ballerina-extension/src/features/tryit/activator.ts
```diff
@@ -427,25 +450,12 @@ async function getAvailableServices(projectDir: string): Promise<ServiceInfo[] |
         });
 
         const services = response.designModel.services
-            .filter(({ type }) => {
-                const lowerType = type.toLowerCase();
-                return lowerType.includes('http') || lowerType.includes('ai') || lowerType.includes('graphql') || lowerType.includes('mcp');
-            })
+            .filter(({ type }) => resolveServiceType(type) !== undefined)
```
NipunaRanasinghe, line 431: Nit: `resolveServiceType` ends up running twice per service, once here and once in the map. The service list is tiny so the cost is nothing, and I think this reads better than fusing the two into one pass. Fine as is.

packages/ballerina-extension/src/features/tryit/activator.ts
```diff
@@ -427,25 +450,12 @@ async function getAvailableServices(projectDir: string): Promise<ServiceInfo[] |
...
-                if (lowerType.includes('http')) {
-                    serviceType = ServiceType.HTTP;
-                } else if (lowerType.includes('graphql')) {
-                    serviceType = ServiceType.GRAPHQL;
-                } else if (lowerType.includes('mcp')) {
-                    serviceType = ServiceType.MCP;
-                } else {
-                    serviceType = ServiceType.AGENT;
-                }
+                const serviceType = resolveServiceType(type);
```
NipunaRanasinghe: This is `ServiceType | undefined` now, and it goes into `ServiceInfo.type`, which is a plain `ServiceType`. It only compiles because `strictNullChecks` is off (`tsconfig.json:17`). The value is genuinely never undefined here since the filter above already ruled that out, so this is safe. It just means the types no longer say so, and it would start failing if that flag is ever turned on.

packages/ballerina-extension/src/features/tryit/activator.ts
RadCod3: Added unit tests for now. Will add an e2e test separately.

packages/ballerina-extension/src/features/tryit/activator.ts
RadCod3: Addressed in 77f0c7011ba6

packages/ballerina-extension/src/features/tryit/activator.ts
RadCod3: Added a regex replace to handle the example with a comment and a unit test to cover that in 64039cff9fc1. Note that it seems if the comment has a colon the LS reports its type incorrectly. 

```ballerina
// Comment on its OWN line, between `new` and the type descriptor.
service /ownline on new
        // a standalone comment
        http:Listener(9201) {
    resource function get a() returns string => "a";
}

// Same, but the comment itself contains a colon.
service /colon on new
        // see: the docs
        http:Listener(9202) {
    resource function get a() returns string => "a";
}
```

Below is how the LS reports the above 2 services in order,
- `        // a standalone comment\n        http:Service` 
- `        // see:Service`

The first case is handled fine by the regex but the latter is not. Weird edge case tho :/

NipunaRanasinghe (approved):
LGTM

MohamedSabthar (approved):

PR #954, opened 2026-08-28
Fix Central agent search not working

dan-niles:
## Purpose

Fixes https://github.com/wso2/product-integrator/issues/2201

https://github.com/user-attachments/assets/5100df7e-4541-45a0-af39-fd40e05480ce

The `Add Agent` popup only ever listed agents from the local index and the current workspace, so agents published to Ballerina Central were never found.

## What changed

- **Search Central directly.** `AgentSearchCommand` now queries Central with the `Type/Agent` keyword filter instead of relying on the local index. The `Organization` filter scopes results to the user's own packages (or the current org when not logged in).
- **Pick a definition from a package.** Central returns packages, not classes. Selecting one pulls the package and lists the agent definitions inside it in a new `PackageAgentsView`. If a package has only one definition, it goes straight to the config form.
- **Resolve the class before building the form.** `TypedAgentBuilder` fills in the agent class name when a Central result arrives without one.

## Notes

- The keyword filter must lead the Central query (`keywords:"Type/Agent" AND <text>`); the reverse order mis-associates on multi-word text and returns nothing.
- `org` and `user-packages` are ANDed by Central, so only one is sent.

## Summary

- Fixes Add Agent search for agent definitions published to Ballerina Central.
- Searches Central with the `Type/Agent` filter and scopes results to the user or organization.
- Resolves Central packages to their agent definitions.
- Adds package selection when multiple agent definitions exist.
- Opens configuration directly when a package contains one definition.
- Resolves missing agent class names for Central results.
- Preserves workspace agents before Central results in the All tab.
- Adds tests for Central search, package resolution, and agent-class handling.


RadCod3 (dismissed):

MohamedSabthar (dismissed):

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/AgentSearchCommand.java
```diff
@@ -151,24 +160,87 @@ private List<AvailableNode> getLandingAgents() {
...
+            centralQueryMap.put("user-packages", "true");
+            return true;
+        }
+        String currentOrg = project.currentPackage().packageOrg().value();
+        if (currentOrg == null || currentOrg.isEmpty()) {
+            return false;
+        }
+        centralQueryMap.put("org", currentOrg);
+        return true;
+    }
```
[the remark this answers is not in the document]
dan-niles: Added TestNG coverage via `RemoteCentral.setTestInstance(...)` with a recording `CentralAPI` — asserts the keyword-first `q` and both the authorized (`user-packages`) and unauthorized (`org`) scope params. Placed in `agentsmanager/AgentSearchCommandTest` rather than `flow-model-generator-core`, since `SearchCommand`'s constructor needs `search-index.sqlite`, which only exists on `flow-model-generator-ls-extension`'s classpath.

MohamedSabthar (dismissed):

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AddAgentPopup/AddAgentPopupContent.tsx
```diff
@@ -291,13 +333,45 @@ export function AddAgentPopupContent(props: AddAgentPopupContentProps) {
...
+    };
+
     const handleSelectAgent = (agent: AvailableNode) => {
         if (dependencyMode) {
             onAgentSelectedForDependency?.(agent);
             return;
         }
-        onPendingAgentChange(agent);
-        onViewChange("configure");
+        if (!agent.codedata.object) {
```
pasindufernando1, line 386: Dependency mode returns above this branch, so a Central package node (no `codedata.object`) is never expanded. `agentDependencyDraftFromNode` then falls back to `agent.metadata?.label` — the *package* name — so selecting `myorg/chatagents` via "Use Agent" generates the type `chatagents:chatagents` instead of `chatagents:ChatAgent`, which does not compile. Before this PR every gallery node carried `object`, so this path is newly reachable.

Suggest expanding (or resolving the agent class) before the `dependencyMode` early return.

dan-niles: Fixed in e8735e3292. Moved the dependency-mode check after the `codedata.object` check so `expandPackage` always runs first; `onAgentSelectedForDependency` now only fires once the package is resolved to an actual definition (including through the multi-definition picker, which had the same gap).

pasindufernando1: Confirmed addressed in e8735e3292 — `selectResolvedAgent` guarantees `codedata.object` is set before the dependency callback fires, and good catch extending it to the single-definition shortcut. Thanks!

pasindufernando1 (dismissed):
LGTM

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AddAgentPopup/AddAgentPopupContent.tsx
```diff
@@ -291,13 +333,50 @@ export function AddAgentPopupContent(props: AddAgentPopupContentProps) {
...
+                selectResolvedAgent(found[0]);
+                return;
+            }
+            setPackageAgents(found);
+            onPendingAgentChange(agent);
+            onViewChange("package");
+        } finally {
+            setIsExpanding(false);
+        }
+    };
```
madushajg, line 383: The `try` here has only a finally. If the `search` RPC rejects, this produces an unhandled promise rejection and the popup silently stays on the gallery with no feedback. Since this call triggers a Central pull plus a package compile on the LS side, failure is realistic; please add a `catch` that surfaces an error state (e.g. reuse `setLoadError`) rather than swallowing it.

dan-niles: Fixed in 33d8179c6a — added a `catch` that logs and surfaces `showErrorMessage` so the failure isn't swallowed.

MohamedSabthar (dismissed):

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/AgentSearchCommand.java
```diff
@@ -151,24 +160,87 @@ private List<AvailableNode> getLandingAgents() {
...
+        centralQueryMap.put("limit", String.valueOf(limit));
+        centralQueryMap.put("offset", String.valueOf(offset));
+
+        if (orgScoped && !addOrgScope(centralQueryMap, centralClient)) {
+            return null;
+        }
+        return centralClient.searchPackages(centralQueryMap);
+    }
+
+    // Central ANDs `org` with `user-packages`, so sending both drops everything the user owns under another org.
```
dulajdilshan: **[Major] Stated invariant contradicts shipped code.** The comment asserts Central *ANDs* `org` with `user-packages` "so sending both drops everything the user owns under another org" — but `CentralSearchUtil.searchConnectorsByOrganization:146-152` deliberately sends **both**, and its javadoc claims the opposite: *"When the client has authorized access, user-owned packages are also included in the results."*

I read both sites; they can't both be right about the same API, so one of them returns the wrong package set to users today — either connector search silently hides teammates' packages, or agent search needlessly narrows. Nothing pins the real behaviour: `AgentSearchCommandTest:82-95` asserts the map this code itself built.

Could you confirm the semantics against Central and make both sites agree? Whichever way it lands, the loser is a live bug — and the next Central-backed search will copy whichever it finds first.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/AgentSearchCommand.java
```diff
@@ -151,24 +160,87 @@ private List<AvailableNode> getLandingAgents() {
...
+    private boolean addOrgScope(Map<String, String> centralQueryMap, CentralAPI centralClient) {
+        if (centralClient.hasAuthorizedAccess()) {
+            centralQueryMap.put("user-packages", "true");
+            return true;
```
dulajdilshan: **[Spec · Major] Partial implementation — signing in narrows the Organization tab.** When authorized this sends only `user-packages=true` (your own packages, any org); the anonymous branch at `:216-221` sends `org=<currentOrg>` and returns the whole org. So a logged-in user sees *fewer* agents than a logged-out one, and the tab is labelled `"Organization"` (`AddAgentPopupContent.tsx:58`).

The issue's bar is *"An Agent definition that was created and published to central does not show up when trying to add an agent from a separate project"* — a teammate's published definition stays invisible in the browse path, i.e. the reported symptom survives for the team case.

The comment above shows you hit a real Central constraint here, so I'm asking rather than asserting: should the authorized branch also send `org`, or union the two queries client-side, so the tab matches its label?

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/AgentSearchCommand.java
```diff
@@ -151,24 +160,87 @@ private List<AvailableNode> getLandingAgents() {
...
+    }
+
+    private PackageResponse getPackageResponse(String searchQuery, boolean orgScoped) {
+        CentralAPI centralClient = RemoteCentral.getInstance();
+        Map<String, String> centralQueryMap = new HashMap<>();
+        // Keyword must lead: `<text> AND keywords:"..."` mis-associates on multi-word text and returns nothing.
+        String q = searchQuery == null || searchQuery.isEmpty()
+                ? AGENT_KEYWORD_FILTER
+                : AGENT_KEYWORD_FILTER + " AND " + searchQuery;
+        centralQueryMap.put("q", q);
```
dulajdilshan: **[Medium] Missing guard — raw user text goes into the Central `q`.** The search term is concatenated after ` AND ` with no escaping, so a trailing operator (`chat AND`) yields a dangling clause and a quote yields an unbalanced literal; the resulting failure is swallowed at `:190` and the user just sees an empty list.

I looked for escaping in `CentralSearchUtil:158`, `RestClient.searchPackages:119` and `TriggerSearchUtil:88` — there is none anywhere, so nothing downstream saves it. The blank and multi-word cases the description calls out are already correct; it's the hostile ones that aren't.

Stripping bare boolean operators before appending would be the minimal fix — quoting the whole term would work too but turns every search into a phrase search, so which behaviour do you want? (Both axes flagged this independently.)

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/AgentSearchCommand.java
```diff
@@ -151,24 +160,87 @@ private List<AvailableNode> getLandingAgents() {
...
         addCategory(LOCAL_AGENTS_CATEGORY, filterAgents(getWorkspaceAgents(), searchQuery));
         return rootBuilder.build().items();
     }
 
     private List<Item> getOrganizationAgents(String searchQuery) {
-        String currentOrg = project.currentPackage().packageOrg().value();
-        addCategory(LOCAL_AGENTS_CATEGORY, filterAgents(getWorkspaceAgents(), searchQuery).stream()
-                .filter(agent -> agent.codedata().org().equalsIgnoreCase(currentOrg))
-                .toList());
+        addCategory(CENTRAL_AGENTS_CATEGORY, fetchAgentsFromCentral(searchQuery, true));
```
dulajdilshan, line 184: **[Spec · Medium] The Organization tab lost its workspace agents.** `getOrganizationAgents` now adds only `CENTRAL_AGENTS_CATEGORY`; the diff removes the `addCategory(LOCAL_AGENTS_CATEGORY, …)` call that filtered workspace agents to the current org. They survive in All (`:164`) and Project (`:173`), so nothing is lost outright.

The issue notes *"It works as intended when trying to access the definition from the same workspace its defined in"* — that path is narrower than before in this one tab, and a narrowing nobody asked for is the kind that gets reported as a new bug.

Was this deliberate — is the tab now Central-only by design — or should the org-filtered local agents come back?

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AddAgentPopup/AddAgentPopupContent.tsx
```diff
@@ -180,6 +191,33 @@ export function AddAgentPopupContent(props: AddAgentPopupContentProps) {
...
+                const orgAgents = toAgents(model);
+                setAgents((current) => {
+                    const seen = new Set(current.map(moduleId));
+                    return [...current, ...orgAgents.filter((agent) => !seen.has(moduleId(agent)))];
+                });
+            })
+            .finally(() => {
+                if (request === searchRequestRef.current) {
+                    setIsLoadingOrgAgents(false);
+                }
```
dulajdilshan, line 223: **[Medium] Unhandled rejection — same class as the `expandPackage` fix, different function.** `loadOrganizationAgents` chains `.then(...).finally(...)` with no `.catch`, so a rejected `search` RPC is an unhandled rejection and the user gets no feedback — exactly what @madushajg raised for `expandPackage`, in code this PR added (`0a6496e05d`) rather than the code that was fixed.

Both neighbours already handle it: `:126` uses `.catch(() => {})` and `expandPackage:370` now does `catch` + `showErrorMessage`. This call is a Central round trip, so failure is realistic.

Would a `.catch` matching `expandPackage`'s shape work here — log, surface, and let the `finally` clear `isLoadingOrgAgents`?

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AddAgentPopup/AddAgentPopupContent.tsx
```diff
@@ -196,7 +234,10 @@ export function AddAgentPopupContent(props: AddAgentPopupContentProps) {
             })
             .then((model) => {
                 if (request === searchRequestRef.current) {
-                    setAgents((model.categories ?? []).flatMap((category) => (category.items ?? []) as AvailableNode[]));
+                    setAgents(toAgents(model));
+                    if (!text && filter === "All") {
+                        loadOrganizationAgents(request);
+                    }
```
dulajdilshan, line 246: **[Spec · Medium] Cross-org agents are discoverable only by typing.** The lazy Central round trip fires only when `!text && filter === "All"` and requests the org-scoped source; the *unscoped* Central search (`AgentSearchCommand.java:168`, `orgScoped=false`) runs only once the user types something.

So an agent published under a different organization never appears by browsing. Demoted from the sub-agent's Major once I traced it — typing does reach it, so this is a discoverability gap, not the bug persisting.

Is search-only cross-org discovery intentional? If so, would the empty/landing state be the place to say "type to search Ballerina Central"?

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/TypedAgentBuilder.java
```diff
@@ -61,6 +66,39 @@ public void setConcreteTemplateData(TemplateContext context) {
...
+    }
+
+    static TemplateContext withAgentClass(TemplateContext context, String className) {
+        Codedata resolved = new Codedata.Builder<>(null).from(context.codedata()).object(className).build();
+        return new TemplateContext(context.workspaceManager(), context.filePath(), context.position(), resolved,
+                context.lsClientLogger());
+    }
+
+    static Optional<String> findAgentClass(Package agentPackage) {
+        return AiUtils.findAgentClasses(agentPackage).stream().findFirst().flatMap(ClassSymbol::getName);
```
dulajdilshan: **[Medium] Speculative generality — the resolving arm may be unreachable.** `resolveAgentClass` only fires for TYPED_AGENT codedata with no `object`, and every producer was checked: `generateCentralAgentNode:225` is the only object-less one, `buildLocalAgentNode:302` always sets it, and `agents_landing.json` is `[]`. The only consumer, `handleSelectAgent:379`, routes object-less nodes to `expandPackage` — never to `getNodeTemplate`.

That leaves four new methods and a 128-line test for a branch the webview can't reach, and its `findFirst()` would pick an arbitrary class if it ever did (YAGNI). Both review axes landed here independently.

Is there a non-UI caller I've missed? If not, would deleting it — leaving `getPackageAgents` as the single resolution point — be cleaner than keeping a net nothing throws into? If you'd rather keep it, saying "compatibility net for older webviews" in the comment would stop the next reader re-deriving this.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/AgentSearchCommand.java
dan-niles: Confirmed by testing the live Central API directly: `org=ballerinax` alone returns 727 results, `user-packages=true` alone returns 7, both together return 0. Central ANDs the two, so the original comment here was right and `CentralSearchUtil`'s javadoc is the one that's wrong (likely a latent bug there too, out of scope for this PR).

Fixed by querying `org` and `user-packages` as two separate Central requests and merging/deduping the results client-side (`fetchOrganizationScopedAgents`), so an authorized user now sees both their own packages and the org's, instead of only one.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/AgentSearchCommand.java
dan-niles: Same fix as the comment on `addOrgScope` above — the authorized branch now also queries `org` (as a separate request, merged with `user-packages`), so a signed-in user sees the whole org again, matching what the anonymous branch already returned.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/AgentSearchCommand.java
dan-niles: Good catch. Went with stripping bare boolean operators (and stray quotes) rather than quoting the whole term, matching the precedent already in `CentralLibrarySearchAccessor.SOLR_BOOLEAN_OPERATORS` for the same reason. `chat AND` now sanitizes to `chat` before it's appended.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/AgentSearchCommand.java
dan-niles: Deliberate — restored: `getOrganizationAgents` now adds the org-filtered `LOCAL_AGENTS_CATEGORY` back alongside the Central category, matching what it did before this diff.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AddAgentPopup/AddAgentPopupContent.tsx
dan-niles: Fixed — added the same `.catch` shape as `expandPackage`: logs the error and shows `showErrorMessage`, before `finally` clears `isLoadingOrgAgents`.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AddAgentPopup/AddAgentPopupContent.tsx
dan-niles: Went with the cheap partial fix rather than making the unscoped search eager by default (that would mean an unbounded cross-org Central query on every empty-state "All" tab open — a real perf/product tradeoff I didn't want to make unilaterally). Added a hint in the empty state: "No agents found. Type to search for agents from other organizations."

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/TypedAgentBuilder.java
dan-niles: Confirmed unreachable, traced independently rather than taking this on faith: every path that reaches `getNodeTemplate` for a TYPED_AGENT already has `object` set — the popup resolves it via `expandPackage` before selection, the nested-sub-agent flow (`useCreateNode`/`CreateAgentForm`) gets it from an already-resolved `ClassSymbol` via `AiUtils.markAgentParams`, and editing an existing agent gets it straight from the compiler. Removed `resolveAgentClass`, `needsAgentClass`, `withAgentClass`, `findAgentClass`, and the test class covering them.

dulajdilshan (approved):
LGTM!

SasinduDilshara (approved):

PR #955, opened 2026-08-28
Fix agent definition form layout

dan-niles:
## Purpose

Fixes https://github.com/wso2/product-integrator/issues/2226

https://github.com/user-attachments/assets/3f6b11fa-e231-414d-8ef7-b675caf03ee4

Simplifies the agent definition creation form by moving the library name field under advanced configurations. (Package settings dropdown drawer has been converted to the standard "Advanced Configurations" drawer used in other forms).

- Updated the Agent Definition form to use an “Advanced Configurations” section for library package fields.
- Added separate layout behavior for modal and full-page forms.
- Fixed codicon sizing in button cards.
- Improved agent tool label and menu alignment with a flex layout.


packages/ballerina-visualizer/src/views/BI/AIChatAgent/AgentDefinitionForm.tsx
```diff
@@ -93,33 +88,19 @@ const LibraryDetails = styled.div`
     gap: 16px;
 `;
 
-const PackageSettings = styled.div`
-    border-top: 1px solid var(--vscode-panel-border);
-    padding-top: 12px;
-`;
-
-const PackageSettingsButton = styled.button`
+const AdvancedRow = styled.div`
```
samithkavishke: Do we need to define separate styling for this component? can't we reuse the already defined styles in the connection popup

MohamedSabthar (dismissed):

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AgentDefinitionForm.tsx
dan-niles: Replaced it with `FormStyles.Row` from `Forms/styles`, the same style the Devant connector popup already uses for its advanced-configurations row.

samithkavishke (dismissed):
LGTM

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AgentDefinitionForm.tsx
```diff
@@ -152,7 +127,7 @@ function LibraryDetailsForm({
     touchedFields,
     hasAttemptedSubmit,
 }: LibraryDetailsFormProps) {
-    const [showPackageSettings, setShowPackageSettings] = useState(false);
+    const [showAdvancedOptions, setShowAdvancedOptions] = useState(false);
```
madushajg, line 130: Moving **Library Name** into a default-collapsed section creates a silent dead-end: `handleCreate` still returns early on `!isLibraryFormValid`, but the error is now hidden, so the **Submit** button appears to do nothing.

This is reachable with names like `AI` (derived library name < 3 chars) or names containing `/`, and can happen permanently if `getDefaultOrgName()` rejects and `orgName` remains empty.

Please auto-expand the section when it contains an error, similar to `ProjectForm/AddComponentFields.tsx:252-258`, or lift `showAdvancedOptions` and force it open on failed submit.

samithkavishke: **HIGH**: `showAdvancedOptions` defaults to `false`, and the Library Name/Package Name/Organization fields only render when it's `true` (see the `{showAdvancedOptions && (...)}` block below). `handleCreate` (further down, outside this diff) guards on `!isLibraryFormValid` and just sets `hasAttemptedSubmit(true)` before returning — it never sets `showAdvancedOptions(true)`.

Repro: with `destination === "library"`, if any of those fields fails validation (e.g. a library name under 3 chars, or `getDefaultOrgName()` failing/returning empty), clicking "Create" silently no-ops: the section stays collapsed, so the fields carrying the errors are never mounted and the user gets no feedback at all.

Suggest expanding the section (and/or surfacing an error indicator on the collapsed header, similar to `CollapsibleSection`'s `hasError` prop in `../ProjectForm/components/CollapsibleSection.tsx`) whenever `hasAttemptedSubmit` is true and the library form is invalid.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AgentDefinitionForm.tsx
```diff
@@ -470,16 +457,26 @@ export function AgentDefinitionForm({ projectPath, submitText = "Create Agent De
...
-                    buttonSx={{ width: "100%", height: "35px" }}
-                >
-                    {creating ? <Typography variant="progress">Creating...</Typography> : submitText}
-                </SubmitButton>
+            <Actions
+                style={inModal
+                    ? { justifyContent: "center", padding: "16px 36px", borderTop: "1px solid var(--vscode-panel-border)" }
+                    : { justifyContent: "flex-end", marginTop: 20 }}
+            >
+                {inModal ? (
```
samithkavishke: **LOW**: `inModal` is threaded as several separate special-cased style branches in this shared form (FormFields padding, Actions justify/padding/border, and this duplicated `SubmitButton` vs. `Button` JSX branch) instead of letting each caller supply its own layout chrome. Consider stripping the modal-specific styling (padding, footer divider, full-width button) out of `AgentDefinitionForm` and having the modal wrapper apply it around a layout-neutral form, similar to how `AddAgentDefinition.tsx` already owns its own full-page layout.

packages/ballerina-visualizer/src/components/ButtonCard/index.tsx
```diff
@@ -104,6 +104,9 @@ const IconContainer = styled.div`
         height: 24px;
         font-size: 24px;
     }
+    > div:first-child > i {
+        font-size: inherit;
```
samithkavishke: **LOW**: `> div:first-child > i { font-size: inherit; }` reaches two levels into the shared `Codicon` component's internal DOM structure. If `Codicon`'s markup changes (e.g. drops the wrapper `div`, or switches to an `svg`), this selector silently stops matching with no compile-time signal and icon sizing regresses here. The font-size inheritance fix likely belongs in `Codicon` itself rather than in this consumer.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/AgentDefinitionForm.tsx
dan-niles: Fixed in https://github.com/ballerina-platform/ballerina-vscode/commit/bd9242ce6b

madushajg (approved):

MohamedSabthar (approved):

PR #959, opened 2026-08-28
Improve title and submit button text in project creation form

madushajg:
## Purpose
Addresses: https://github.com/wso2/product-integrator/issues/2081

- Updated the project creation form title to “Create Integration Project”.
- Updated the project and library creation submit buttons to use “Create”.


gigara (dismissed):

rosensilva (dismissed):

gigara (dismissed):

packages/ballerina-visualizer/src/views/BI/ProjectForm/embedded/integrator-form/CreateProjectChooser.tsx
```diff
@@ -89,7 +89,7 @@ export function CreateProjectChooser({
 
     return (
         <CreateFlowShell
-            title="Create"
+            title="Create Integration Project"
```
NipunaRanasinghe: "Integration Project" merges two concepts we otherwise keep separate. The subtitle right below says a project organizes your integrations and libraries, so "integration project" reads like a third, distinct thing. This flow creates a project that can then contain either an integration or a library, so `Create Project` seems more accurate here.

madushajg: Yeah that makes sense. Fixed with 37d9b77d8918a8d8429dc705256dc71871376cfb

NipunaRanasinghe (approved):
LGTM

PR #960, opened 2026-08-28
Generate GraphQL object types as distinct service classes

NipunaRanasinghe:
Fixes : https://github.com/wso2/product-integrator/issues/1822

### Problem
GraphQL union members must be `distinct` service classes. `SourceCodeGenerator.generateGraphqlClassType` — the generator behind `typesManager/createGraphqlClassType`, which every "Object" creation in the GraphQL type editor goes through — emitted a bare `service class`. A union built out of editor-created objects therefore failed to compile:

```
ERROR [main.bal:(58:5,63:6)] invalid GraphQL union type member 'Teacher' found. Only distinct service types are allowed
```

### Fix
Emit `distinct service class`. Unconditional rather than only for union members, because the class is created before it is known whether it will be used in a union or as an interface implementation, and a distinct service class with no inclusions is still a plain object type.

Checked against `bal` 2201.13.4 that `distinct` only restricts assignment *into* the class type and never removes assignability out of it. Identical compile results with and without it for: direct instantiation, assignment to a structural `service object` type, type inclusion in both directions (a distinct class carries its own type ID *plus* its inclusions', so `http:Service`-style library interfaces and listener attach are unaffected), and object-constructor expressions with an expected type. `readonly distinct service class` — what the readonly branch now produces — also compiles.

The one behaviour change: a separately declared, structurally identical class is no longer assignable to the generated type, and `is` / `ensureType` against that type returns false for such a value. Both require hand-written code that mirrors an editor-generated class.

Note this also covers the non-GraphQL "Service Class" option in the type diagram, which routes through the same endpoint (the `TODO: for TypeDiagram we need to give a generic class creation` in `TypeEditor.tsx`).

### Known gap
Unions built from *pre-existing* non-distinct service classes still fail. Fixing that means rewriting other type definitions when they are pulled into a union, which is out of scope here.

### Tests
Updated the 7 expected-output configs under `types_manager/create_graphql_class_type/config/`. `CreateGraphqlClassTypeTest` was not run locally (the LS build needs `packageUser`/`packagePAT`), so it is covered by CI on this PR.

- Generate GraphQL service types as `distinct service class` so editor-created union members compile correctly.
- Preserve compatibility with direct instantiation, structural assignment, type inclusions, listeners, and object-constructor expressions.
- Update expected outputs for GraphQL and Service Class type diagram configurations.
- Existing non-distinct service classes in unions remain unsupported.


pasindufernando1 (approved):
LGTM

PR #963, opened 2026-08-31
Display memory store in agent node widget

dan-niles:
## Purpose

Fixes https://github.com/wso2/product-integrator/issues/2212

https://github.com/user-attachments/assets/e39c8e67-03a4-40d8-8f28-057533ea277b

Summary:

- Agent node now shows which memory store an agent's memory is using, with a clickable icon to open the store's settings
- Memory store field moved out of "Advanced Configurations" into the main form, since it's the main thing users set
- Fixed memory not showing up at all for some agents (nilable memory type wasn't recognized)
- In-memory store's size field is now visible outside advanced settings
- Fixed hover highlights on the agent node overlapping each other (box, card, buttons all lighting up together)
- Removed a scrollbar in the memory form that didn't actually scroll anything

## Summary

- Display the agent’s configured memory store in the high-level diagram.
- Add a clickable memory store icon that opens its settings.
- Support memory store creation and editing through the agent editor.
- Resolve memory stores from variables and scoped agent declarations.
- Recognize optional AI types such as `ai:Memory?` and `ai:ModelProvider?`.
- Improve hover behavior to prevent overlapping highlights.
- Expose in-memory store size and remove the ineffective scrollbar.


MohamedSabthar (dismissed):

samithkavishke (changes requested):
@dan-niles Please add test cases for this and improve the test coverage. Which reported as 48% here

packages/ballerina-visualizer/src/views/BI/AIChatAgent/MemoryManagerConfig.tsx
```diff
@@ -435,6 +437,36 @@ export function MemoryManagerConfig(props: MemoryConfigProps): JSX.Element {
...
+        const overlayId = openOverlay({
+            title: "Configure Memory Store",
+            content: (
+                <LoaderContainer>
+                    <RelativeLoader />
+                </LoaderContainer>
+            ),
+            onBack: closeTopOverlay,
+        });
+        const storeNode = await findFlowNodeByModuleVarName(storeName, rpcClient);
```
samithkavishke: Since `AiUtils.resolveComponent` on the backend now resolves memory stores declared as local variables or class fields (not just module-level connections), this lookup should probably match. `findFlowNodeByModuleVarName` only searches `rpcClient.getModuleNodes()`, so if a user declares the memory store as a local variable or class field, this will resolve to `null`, log "Memory store declaration not found", and silently close the overlay — no error shown to the user.

Meanwhile the other new edit entry point for the same feature — clicking the memory-store icon on the agent node itself (`selectMemoryStore` in `useAgentEditorController.ts`) — uses a position-scoped lookup (`findFlowNode`/`findAgentScopedNode`) that does resolve local variables and class fields correctly.

Could this use the same position-scoped lookup as `selectMemoryStore`, so both edit paths support the store declarations this PR adds backend support for?

dan-niles: Fixed via https://github.com/ballerina-platform/ballerina-vscode/pull/963/commits/24e3a4ec9d3d13a01b22ab5a6c4480192660ff47
Extracted the position-scoped lookup into a shared `findAgentScopedNode` util in utils.ts, so `selectMemoryStore` and `handleEditStore` now use the same code path. Local vars and class fields resolve for both edit entry points now.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/test/java/io/ballerina/flowmodelgenerator/core/AiUtilsComponentResolutionTest.java
```diff
@@ -0,0 +1,184 @@
...
+                newExpressionArgumentsOf("memoryWithNamedStore")), "memoryStoreVar");
+    }
+
+    @Test(description = "A memory store held in a class field is not resolved: CommonUtils#isAiMemoryStore "
+            + "filters by symbol kind before resolveComponent ever sees the field, and its getClassSymbol "
+            + "helper does not unwrap a ClassFieldSymbol (unlike the VariableSymbol/ParameterSymbol it does "
+            + "handle) - a pre-existing gap in a shared helper outside this change, not a regression here")
+    public void testMemoryStoreFromClassFieldIsNotYetResolved() {
+        Assert.assertNull(
+                AiUtils.getMemoryStoreData(semanticModel, newExpressionArgumentsOf("memoryWithFieldStore")));
```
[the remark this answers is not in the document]
dan-niles: Deferring this. The gap here is deeper than unwrapping `ClassFieldSymbol`: `fieldStore` is declared with the abstract `ai:ShortTermMemoryStore` interface type, not a concrete class, and `hasAiTypeInclusion` only checks a `ClassSymbol`'s own `typeInclusions()` — it never matches when the declared type *is* the marker interface itself. Fixing it properly means widening the shared `isAi*` type-check logic (used by 10 predicates: model provider, agent, knowledge base, vector store, memory, memory store, etc.), which is a bigger change than fits this PR. Filing as a follow-up.

samithkavishke (commented):
## Review follow-up

Confirmed both existing thread concerns are addressed: the reported memory-store edit lookup bug is fixed via the shared `findAgentScopedNode` lookup, and CodeRabbit's class-field marker-interface gap is a disclosed, scoped-out follow-up (agreed by CodeRabbit).

Found one more issue while re-checking the migration to `findAgentScopedNode`:

### 🔴 HIGH: `deleteMemory` wasn't migrated to `findAgentScopedNode`

**`packages/ballerina-visualizer/src/views/BI/AIChatAgent/useAgentEditorController.ts:171`**

`selectMemory` and `selectMemoryStore` were updated to resolve declarations via the new position-scoped `findAgentScopedNode`, but `deleteMemory` still resolves via the old module-only `findFlowNodeByModuleVarName`. For memory declared as a local variable or class field (exactly the case this PR adds support for), that lookup returns `null`, the delete is silently skipped, yet the code still clears the agent's `memory` property and reports success — orphaning the original declaration in source with no user-visible warning.

Suggest applying the same fix here as in `selectMemory`/`selectMemoryStore`.

### 🟡 MEDIUM: class-field memory stores via the abstract marker interface aren't resolved

**`packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/CommonUtils.java` (`getClassSymbol`)**

Already flagged by CodeRabbit and acknowledged as an out-of-scope follow-up (`getClassSymbol` doesn't unwrap `ClassFieldSymbol`, so a store field typed by the abstract `ai:ShortTermMemoryStore` interface isn't recognized). Documented in `AiUtilsComponentResolutionTest.testMemoryStoreFromClassFieldIsNotYetResolved`. Flagging again here just to keep it visible alongside the HIGH item above — not blocking, since it's already tracked.

samithkavishke (dismissed):
LGTM

packages/ballerina-visualizer/src/components/ConnectionSelector/ConnectionConfig.tsx
```diff
@@ -197,16 +196,6 @@ export function ConnectionConfig(props: ConnectionConfigProps): JSX.Element {
     const handleOnChange = useCallback((fieldKey: string, value: any) => {
         if (fieldKey !== "connection" || value === selectedConnectionValue) return;
 
-        const isKnownConnection = value && connectionNodesMap.current.has(value);
```
madushajg, line 200: Removing the guard means `updateFieldsForConnection(value)` now runs for unknown values and calls `setSelectedConnectionValue(value)`, which is the key on `<ArtifactForm>`.

Since `handleOnChange` fires on every keystroke, switching the connection field to `fx` and typing a free expression changes the key on every character, remounting the entire form. This drops focus mid-typing and causes the "Loading form data..." indicator to flash on each remount.

Please key the form on connection identity rather than raw text, e.g.:

```tsx
key={
    connectionNodesMap.current.has(selectedConnectionValue)
        ? selectedConnectionValue
        : "__expression__"
}

File: packages/ballerina-visualizer/src/views/BI/AIChatAgent/MemoryStoreConfig.tsx [hunk 4]
```diff
@@ -0,0 +1,86 @@
...
+
+    const [template, setTemplate] = useState<FlowNode>();
+    const [isSaving, setIsSaving] = useState(false);
+    const filePath = useRef<string>("");
+
+    useEffect(() => {
+        void (async () => {
+            const lineRange = storeNode.codedata.lineRange;
+            filePath.current = (await rpcClient.getVisualizerRpcClient()
+                .joinProjectPath({ segments: [lineRange.fileName] })).filePath;
```
madushajg: This calls `joinProjectPath` unconditionally, but `MemoryManagerConfig.resolveFilePath` exists precisely because `codedata.lineRange.fileName` may already be absolute. Joining then doubles the project prefix. Since the store-edit path goes through that code, please reuse `resolveFilePath` instead of `joinProjectPath` here.

packages/ballerina-visualizer/src/components/ConnectionSelector/ConnectionConfig.tsx
dan-niles: The guard removal meant `updateFieldsForConnection` ran on every keystroke while typing an expression, changing the `ArtifactForm` key each time. Fixed by keying on connection identity (known connection value, the default-provider sentinel, or a stable `__expression__` placeholder for free text) in 87417f42a5.

packages/ballerina-visualizer/src/views/BI/AIChatAgent/MemoryStoreConfig.tsx
dan-niles: Agreed. Extracted MemoryManagerConfig's resolveFilePath into a shared util (AIChatAgent/utils.ts) and switched MemoryStoreConfig to use it instead of the raw joinProjectPath call, so it's consistent with the memory node's own file-path resolution. Also extended the absolute-path check to recognize Windows UNC paths (\\server\share) while I was in there. Fixed in 87417f42a5.

MohamedSabthar (approved):

dulajdilshan (approved):

PR #965, opened 2026-08-31
[Agent Builder] Add try-it button for agent triggers

dan-niles:
This PR adds a try-it button in the 3-dot menu of the supported triggers for the agent node.

https://github.com/user-attachments/assets/096124a2-235b-423b-9cdf-5d412eeeac16

Changes include:

- Added a "Try It" button so you can quickly test a trigger (HTTP, chat, GraphQL, MCP) right from the diagram
- Fixed hover and click areas on trigger rows so empty space next to a label also responds
- Fixed the trigger menu overlapping long tool names
- Fixed the trace animation not playing when an agent is called over HTTP


packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/agentUsages.ts
```diff
@@ -296,3 +317,47 @@ export async function getAgentTriggerScopes(rpcClient: BallerinaRpcClient): Prom
...
+    const unresolved = [...new Set(present)].filter((protocol) => !known.has(protocol) && !centralScopes.has(protocol));
+
+    await Promise.all(unresolved.map(async (protocol) => {
+        try {
+            const models = await rpcClient
+                .getServiceDesignerRpcClient()
+                .searchTriggers({ query: protocol, includeLocalRepository: true });
+            const results = [...(models?.local ?? []), ...(models?.localRepositoryResults ?? [])];
+            const match = results.some((trigger) => trigger.listenerProtocol === protocol);
+            centralScopes.set(protocol, match ? "SERVICE" : undefined);
```
kaumini: Hard-coding `"SERVICE"` here discards data the language server already sends, and this path drives source deletion.

`searchTriggers` results go through `AgentTriggerChannels::withAgentKind` on the LS side (`ServiceModelGeneratorService.java:472-476`), exactly like `getTriggerModels` does — so every result already carries `agentTriggerKind` and `deletionScope`. `getAgentTriggerScopes` four lines up reads both:

```js
.filter((trigger) => trigger.agentTriggerKind && trigger.listenerProtocol && trigger.deletionScope)
.map((trigger) => [trigger.listenerProtocol, trigger.deletionScope])
```

Two reachable failures:

1. A Central trigger declaring `ENTRY_POINT` gets classified `SERVICE`, so the row only ever offers whole-service deletion instead of endpoint-only. (`AgentTriggerGenerationTest` asserts `ballerina/http` → `ENTRY_POINT`, so this shape exists.)
2. A trigger whose `deletionScope` is null gets a working Delete menu. `TriggerBasicInfo`'s javadoc says null means the trigger cannot be deleted — `ftp` is the asserted example.

Dropping the `agentTriggerKind` filter also lets a non-agent Central trigger on a matching protocol enable the menu.

Mitigating factor: `confirmTriggerDeletion` shows an accurate modal for whatever it is about to do, so nothing is removed without a warning — it is the wrong option set rather than silent data loss.

Suggested fix:

```js
const match = results.find((t) => t.listenerProtocol === protocol && t.agentTriggerKind && t.deletionScope);
centralScopes.set(protocol, match?.deletionScope);
```

packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/index.tsx
```diff
@@ -371,6 +386,19 @@ export function BIFocusFlowDiagram(props: BIFocusFlowDiagramProps) {
         return choice === withService ? "service" : undefined;
     };
 
+    const tryAgentTrigger = async (usage: AgentUsage) => {
```
kaumini, line 389: This swallows failures. It is invoked as `void tryAgentTrigger(usage)` (line 1079) with no try/catch, so a rejected `executeCommand` becomes an unhandled promise rejection and the user gets no feedback at all — they click Try It and the menu just closes.

`deleteAgentTrigger` directly below already has the pattern to copy: try/catch with `console.error(">>> agent focus: ...")` plus `showErrorMessage`. Worth matching it here so the two sibling actions behave the same way when the RPC fails.

packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/agentUsages.ts
dan-niles: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/965/commits/26dd32bbe1d69790629a914edf4ede281e6e4b4b

packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/index.tsx
dan-niles: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/965/commits/26dd32bbe1d69790629a914edf4ede281e6e4b4b

kaumini (approved):

PR #966, opened 2026-08-31
[Agent Builder] Fix event trigger deletion for agents

dan-niles:
Changes include:

- You can now remove an agent from one event trigger without deleting the whole service
- Triggers now show in a consistent order every time
- Fixed the agent node shifting position while a refresh was in progress
- Fixed tab selection jumping to the wrong agent after a rename or edit
- Fixed the agent diagram loading off-center on first open
- Closing the Add Tool or Add Memory panel without saving no longer reloads the diagram

### DIFF
--- a/packages/ballerina-visualizer/src/MainPanel.tsx
+++ b/packages/ballerina-visualizer/src/MainPanel.tsx
@@ -212,6 +212,8 @@ const MainPanel = () => {
     const navKeyRef = useRef<number>(0);
     const remountKeyRef = useRef<number>(0);
     const previousNavTargetRef = useRef<string | undefined>(undefined);
+    const agentFocusTargetRef = useRef<string | undefined>(undefined);
+    const agentFocusIdRef = useRef<number>(0);
 
     useSuppressAgentStatusOrb(viewHidesAgentStatusOrb(activeView) || !!viewError);
     useTraceAnimationBridge();
@@ -349,8 +351,15 @@ const MainPanel = () => {
                             if ((await fetchProductMode(rpcClient)) === ProductMode.AGENT_BUILDER) {
                                 const { AgentBuilderOverview } = await import("./views/BI/AgentBuilderOverview");
                                 if (isStaleNavigation()) return;
-                                const agentFocus = value.documentUri && value.position
-                                    ? { path: value.documentUri, startLine: value.position.startLine, requestId: navKey }
+                                const agentFocusTarget = value.documentUri && value.position
+                                    ? `${value.documentUri}::${value.position.startLine}`
+                                    : undefined;
+                                if (agentFocusTarget !== agentFocusTargetRef.current) {
+                                    agentFocusTargetRef.current = agentFocusTarget;
+                                    agentFocusIdRef.current += 1;
+                                }
+                                const agentFocus = agentFocusTarget
+                                    ? { path: value.documentUri, startLine: value.position.startLine, requestId: agentFocusIdRef.current }
                                     : undefined;
                                 setViewComponent(
                                     <AgentBuilderOverview projectPath={value.projectPath} agentFocus={agentFocus} />
--- a/packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/agentUsages.test.ts
+++ b/packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/agentUsages.test.ts
@@ -568,11 +568,20 @@ const scatteredChannels = {
 } as unknown as CDModel;
 
 describe("rail ordering", () => {
-    it("keeps rows of the same channel together", () => {
-        const labels = findAgentUsages(scatteredChannels, { filePath: AGENTS_BAL, startLine: 4 })
-            .map((usage) => usage.label);
+    const labelsOf = (model: CDModel) =>
+        findAgentUsages(model, { filePath: AGENTS_BAL, startLine: 4 }).map((usage) => usage.label);
 
-        expect(labels).toEqual(["onAssigned", "onReopened", "Agent Chat", "main"]);
+    it("groups rows by channel, and orders the channels by name", () => {
+        expect(labelsOf(scatteredChannels)).toEqual(["Agent Chat", "onAssigned", "onReopened", "main"]);
+    });
+
+    it("does not depend on the order the design model happens to list services in", () => {
+        const reversed = {
+            ...scatteredChannels,
+            services: [...(scatteredChannels.services ?? [])].reverse(),
+        } as unknown as CDModel;
+
+        expect(labelsOf(reversed)).toEqual(labelsOf(scatteredChannels));
     });
 });
 
--- a/packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/agentUsages.test.ts
+++ b/packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/agentUsages.test.ts
@@ -16,10 +16,11 @@
  * under the License.
  */
 
-import { CDFunction, CDModel, CDResourceFunction } from "@wso2/ballerina-core";
+import { AgentUsageTrigger, CDFunction, CDModel, CDResourceFunction } from "@wso2/ballerina-core";
 import {
     AgentTriggerScopes,
     agentCallerProtocols,
+    clearAgentCallFromHandler,
     deleteEachResolved,
     findAgentUsages,
     findListenerPosition,
@@ -611,6 +612,12 @@ describe("try it on a rail row", () => {
         });
     });
 
+    it("drops an attached listener the model carries no declaration for", () => {
+        const row = httpRows({ ...supportApi, attachedListeners: ["http-listener", "gone"] })[0];
+
+        expect(row.tryIt?.listener).toBe("httpListener");
+    });
+
     it("aims each endpoint of one service at its own resource", () => {
         const rows = httpRows({
             ...supportApi,
@@ -702,7 +709,7 @@ describe("triggers that only Ballerina Central knows about", () => {
         return { client: { getServiceDesignerRpcClient: () => ({ searchTriggers }) } as never, searchTriggers };
     };
 
-    it("asks only about the protocols that actually call this agent", () => {
+    it("asks only about the protocols that actually call this agent, and says nothing when the agent is not in the model", () => {
         expect(agentCallerProtocols(slackModel, agent)).toEqual(["slack"]);
         expect(agentCallerProtocols(slackModel, { filePath: AGENTS_BAL, startLine: 99 })).toEqual([]);
     });
@@ -827,6 +834,23 @@ describe("event handler rows", () => {
         expect(rows[0].trigger?.orphansService).toBe(true);
     });
 
+    it("does not orphan the service when a sibling handler reaches a different agent", () => {
+        const service = issuesService(["onOpened"], ["runAgentOnOpened"]);
+        const crossAgentService = {
+            ...service,
+            remoteFunctions: [
+                service.remoteFunctions[0],
+                { ...service.remoteFunctions[1], connections: [SUPPORT_AGENT_UUID] },
+                service.remoteFunctions[2],
+            ],
+        };
+
+        const rows = eventRows(crossAgentService);
+
+        expect(rows.map((row) => row.label)).toEqual(["onOpened"]);
+        expect(rows[0].trigger?.orphansService).toBe(false);
+    });
+
     it("names the helpers that reach the agent so the offload target can be resolved", () => {
         const rows = eventRows(issuesService(["onOpened"], ["runAgentOnOpened", "init"]));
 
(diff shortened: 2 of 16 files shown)


kaumini (commented):
**1 critical, 6 major** inline below

packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/agentUsages.ts
```diff
@@ -112,18 +114,29 @@ function triggerFor(model: CDModel, service: CDService): AgentUsageTrigger {
...
         ...trigger,
         entryPoint: { label, documentUri: location.filePath, position: toPosition(location) },
-        orphansService:
-            (service.resourceFunctions?.length ?? 0) + (service.remoteFunctions?.length ?? 0) === 1 &&
-            (service.functions?.length ?? 0) === 0,
+        orphansService: scope === "ENTRY_POINT_BODY"
+            ? entryPoints.filter((fn) => fn.connections?.includes(uuid)).length === 1
+            : entryPoints.length === 1 && (service.functions?.length ?? 0) === 0,
+        scope,
+        helpers: scope === "ENTRY_POINT_BODY" ? agentHelpers(service, uuid) : undefined,
```
kaumini, line 142: **Critical — over-broad `orphansService` destroys another agent's wiring**

For `ENTRY_POINT_BODY` this counts only entry points wired to *this* agent's uuid. I probed it against the real implementation: a service where `onOpened` reaches agent A and `onClosed` reaches **agent B** still returns `orphansService: true`. The modal then offers "Delete Trigger and Service", the user consents on the basis of a false claim ("No other trigger on X uses the agent"), and `deleteComponentAt(serviceName, trigger.position)` takes the whole service — silently removing agent B's handler. No warning and no partial-failure message; it looks like it worked.

The same hole covers siblings that carry `workflows` or plain hand-written code. The old `ENTRY_POINT` rule (`entryPoints.length === 1 && functions.length === 0`) was conservative precisely to avoid this.

Suggestion — require that no sibling entry point carries any other wiring:

```ts
const wired = entryPoints.filter((fn) => (fn.connections?.length ?? 0) > 0 || (fn.workflows?.length ?? 0) > 0);
orphansService = wired.length === 1 && wired[0].connections?.includes(uuid);
```

That still can't see a sibling holding non-connection logic, so it may be worth keeping a stricter guard for the "and Service" option specifically.

packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/agentUsages.ts
```diff
@@ -361,3 +371,87 @@ export async function resolveTriggerScopes(
...
+export function startLineOf(node: FlowNode): number {
+    return node.codedata?.lineRange?.startLine?.line ?? 0;
+}
+
+export function namesHelper(node: FlowNode, helper: string): boolean {
+    const expression = node.properties?.expression?.value;
+    if (typeof expression !== "string") {
+        return false;
+    }
+    return new RegExp(`(^|[^\\w.])(self\\.)?${helper}\\s*\\(`).test(expression);
```
kaumini: **Major — identifier interpolated into a RegExp unescaped**

Three failure modes, all reproduced by running the expression:

- a name containing `(` throws `SyntaxError: Unterminated group` → surfaces as the generic "Some changes may have been applied" even though nothing was applied;
- a name containing `+` matches a **different** member — `processAdd+` matched `start self.processAddd(payload)`, which would delete the wrong statement from the handler;
- it matches inside string literals: `namesHelper('log:printInfo("calling runAgent(x)")', "runAgent")` returns `true`.

Generator-produced names (`runAgentOnOpened`) are safe today, so this is latent rather than live — but `agentHelpers` returns *any* service function whose `connections` include the agent, so a hand-written or quoted identifier reaches here too.

Escaping the interpolation is the one-line fix. The better fix is to stop matching on expression text: the flow node's `codedata` carries the symbol, and the design model already tells you which functions reach the agent. Relying on a single `properties.expression.value` string also means any node kind that stores the call elsewhere yields "no agent call found in …".

packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/agentUsages.ts
```diff
@@ -361,3 +371,87 @@ export async function resolveTriggerScopes(
...
+        throw new Error(`no agent call found in ${entryPoint.label}`);
+    }
+
+    for (const node of calls) {
+        await rpcClient.getBIDiagramRpcClient().deleteFlowNode({ filePath: entryPoint.documentUri, flowNode: node });
+    }
+
+    await deleteEachResolved(
+        rpcClient,
+        helpers.filter((helper) => calls.some((node) => namesHelper(node, helper.symbol))),
```
kaumini: **Major — a shared helper is deleted out from under a sibling handler**

`agentHelpers` is service-wide, so `helpers` here is not scoped to this entry point. If two handlers offload to the same helper (`runAgent` rather than `runAgentOnOpened`), removing one handler's call deletes the helper the other still calls, breaking compilation of the sibling.

Latent on the generated path since the generator emits one helper per handler — live the moment a helper is shared or hand-written. Worth guarding on "no remaining entry point references this helper" before deleting it.

packages/bi-diagram/src/components/nodes/AgentNode/AgentNodeWidget.tsx
```diff
@@ -879,13 +879,16 @@ export function AgentNodeWidget(props: AgentNodeWidgetProps) {
     };
 
     const canDeleteTrigger = (usage: AgentUsage) =>
-        !readOnly && Boolean(usage.trigger) && Boolean(agentNode?.onDeleteTrigger);
+        Boolean(usage.trigger) && Boolean(agentNode?.onDeleteTrigger);
 
     const canTryTrigger = (usage: AgentUsage) =>
-        !readOnly && Boolean(usage.tryIt) && Boolean(agentNode?.onTryTrigger);
+        Boolean(usage.tryIt) && Boolean(agentNode?.onTryTrigger);
```
kaumini, line 890: **Major — readOnly affordances now render enabled but inert**

Dropping `!readOnly` here (and from `showsAddTile` below) is the right call for geometry stability — drawing the same elements is what stops the node shifting mid-refresh. And the actions are still safely guarded: `handleUsageMenuClick` and `onAddTriggerClick` both bail on `readOnly`.

The problem is the affordance. During every refresh the ⋮ and the add tile look fully clickable and swallow the click. Everywhere else in this widget readOnly gets visual feedback — `cursor: not-allowed` at lines 1302 and 1354, muted stroke at 1486. Giving these the same treatment keeps the footprint identical while making the state honest.

This is the highest-frequency defect in the PR: it fires on every refresh, not just on an edge case.

packages/bi-diagram/src/components/nodes/AgentNode/AgentNodeWidget.tsx
```diff
@@ -879,13 +879,16 @@ export function AgentNodeWidget(props: AgentNodeWidgetProps) {
...
 
     const hasUsageMenu = (usage: AgentUsage) => canTryTrigger(usage) || canDeleteTrigger(usage);
 
+    const deleteTriggerLabel = (usage: AgentUsage) =>
+        usage.trigger?.scope === "ENTRY_POINT" ? "Delete Endpoint" : "Delete Trigger";
```
kaumini: **Major (maintainability) — Endpoint/Trigger label logic duplicated across two packages**

This derives the noun from `AgentTriggerDeletionScope`, and `confirmTriggerDeletion` in `FocusFlowDiagram/index.tsx:380-383` derives the same noun from the same enum for the modal's buttons. The two must agree or the menu item and the dialog button disagree with each other.

They agree today, but the copies live in different packages and track a shared enum in `ballerina-core`, so drift is a matter of time. One exported helper next to the type — `triggerScopeNoun(scope)` — removes the coupling.

Related: `triggerFor` never sets `scope`, so `trigger.scope` is `undefined` for SERVICE-scoped triggers while it's populated for member-scoped ones. Both label sites happen to fall through to the right answer, but a half-populated field is a trap for the next reader.

packages/ballerina-visualizer/src/MainPanel.tsx
```diff
@@ -349,8 +351,15 @@ const MainPanel = () => {
...
-                                    ? { path: value.documentUri, startLine: value.position.startLine, requestId: navKey }
+                                const agentFocusTarget = value.documentUri && value.position
+                                    ? `${value.documentUri}::${value.position.startLine}`
+                                    : undefined;
+                                if (agentFocusTarget !== agentFocusTargetRef.current) {
+                                    agentFocusTargetRef.current = agentFocusTarget;
+                                    agentFocusIdRef.current += 1;
+                                }
+                                const agentFocus = agentFocusTarget
+                                    ? { path: value.documentUri, startLine: value.position.startLine, requestId: agentFocusIdRef.current }
```
kaumini, line 362: **Major — keying `requestId` on the target breaks repeat navigation**

Deduping on `documentUri::startLine` fixes refresh-reapplies-focus, but it also means a genuine repeat navigation produces the same `requestId`. Repro: go-to-agent X → user clicks agent Y's tab → go-to-agent X again. The target is unchanged, so `requestId` is unchanged, `appliedFocusRef.current === agentFocus.requestId` short-circuits in `AgentBuilderOverview`, and the tab never moves back.

(Reasoned from the code, not executed — worth a manual check.)

There's also an inconsistency: this PR concluded `startLine` is too unstable to identify an agent (hence `agentKey` moving to `path::name`), yet the focus request still matches on `startLine` at `AgentBuilderOverview/index.tsx:107`. An edit that shifts the agent's line makes focus silently miss. Using the identifier on both sides would make the two paths agree.

packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/agentUsages.test.ts
```diff
@@ -585,12 +611,6 @@ describe("try it on a rail row", () => {
         });
     });
 
-    it("drops an attached listener the model carries no declaration for", () => {
-        const row = httpRows({ ...supportApi, attachedListeners: ["http-listener", "gone"] })[0];
-
-        expect(row.tryIt?.listener).toBe("httpListener");
-    });
```
kaumini, line 592: **Major (process) — a passing test was deleted with no replacement**

I re-added this test to the branch and ran it: **it still passes**. `tryItFor` is untouched by this PR, and `httpRows` builds a single-service model so the new `groupByChannel` ordering doesn't affect `[0]`. So this looks like a straight loss of live coverage rather than a test invalidated by the change.

Could you restore it, or say what made it obsolete?

(The other deletion in this hunk — "says nothing when the agent is not in the model" — was folded into the test above as an extra assertion, which is fine, though that test's name no longer describes its second half.)

packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/agentUsages.ts
dan-niles: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/966/commits/e7c5a2654d

packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/agentUsages.ts
dan-niles: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/966/commits/e7c5a2654d

packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/agentUsages.ts
dan-niles: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/966/commits/e7c5a2654d

packages/bi-diagram/src/components/nodes/AgentNode/AgentNodeWidget.tsx
dan-niles: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/966/commits/e7c5a2654d

packages/bi-diagram/src/components/nodes/AgentNode/AgentNodeWidget.tsx
dan-niles: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/966/commits/e7c5a2654d

packages/ballerina-visualizer/src/views/BI/FocusFlowDiagram/agentUsages.test.ts
dan-niles: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/966/commits/e7c5a2654d

packages/ballerina-visualizer/src/MainPanel.tsx
dan-niles: This is not reproducible in the UI.

https://github.com/user-attachments/assets/2fa7cbe6-119f-45e2-8a5c-dde20955655c

kaumini (approved):

PR #969, opened 2026-08-31
Run the scheduled E2E suite on Windows alongside Linux

NipunaRanasinghe:
## Purpose
The E2E suite runs on Linux only, so a regression that appears only on Windows is caught by no automated test. This PR enables the Windows E2E run in the scheduled (6-hourly) E2E workflow. Enabling it for PR builds, behind the `Checks/Run Ballerina UI Tests` label, is left for a follow-up.

`e2e-scheduled.yml` is the workflow targeted here because it is the only one that runs the E2E suite on a schedule today — `schedule.yml` passes `runFrontendTests: false` and deliberately does not run E2E.

## Goals
- `e2e-scheduled.yml` runs the suite on `windows-latest` as well as `ubuntu-latest`.
- Windows results are visible (step summary, artifacts, `e2e-metrics` history) without gating the workflow, since the suite's Windows pass rate is not yet known.
- Linux behaviour, artifact names and gating are unchanged.

## Approach
**`.github/actions/run-e2e-group/action.yml`** — made platform-aware rather than forked into a second action:
- A new first step folds `runner.os` into a lowercase platform label (`linux` / `windows`) that replaces the hard-coded `linux` in all four artifact names. On a Linux runner the names are byte-identical to before.
- The `apt-get` install (xvfb, ffmpeg, gnome-keyring, libsecret, dbus-x11) is gated to `runner.os == 'Linux'`. `npx playwright install` moved to its own step and runs on both.
- Both test steps take a Windows guard clause that runs `pnpm exec playwright test` directly: the Windows runner has its own desktop session, so there is no virtual display to create and no `ffmpeg` X11 screen grab to wrap the run in. The Linux branch is unchanged. The suite captures its own failure screenshots (`setup.ts`) and session video (`test.list.ts`) on both platforms — these are not Playwright's `use.video`/`use.screenshot`, which are both `off` — and traces keep working on Windows via `BI_E2E_TRACE` below; only the whole-session `record.mp4` is Linux-only.
- The `--last-failed` re-run step is skipped on Windows for a first attempt (`runner.os != 'Windows'` added to that clause only). It is a second whole invocation whose condition is independent of the retry count, so it would re-execute up to ten platform failures against what is left of the 60-minute budget — the same cost `BI_E2E_RETRIES: 0` removes from the first pass. The `steps.check-results.outcome == 'success'` clause is kept, so a manual "Re-run failed jobs" on a Windows leg still resumes via `--last-failed`; without it a re-run would execute nothing, since the first-attempt step is already skipped on re-runs.

**`.github/workflows/e2e-scheduled.yml`**:
- `Prepare` publishes a `platforms` output (`["linux","windows"]`) so the `E2E` matrix and `Report`'s two loops read one list.
- `E2E` becomes a platform × group matrix (8 legs). `continue-on-error: ${{ matrix.os == 'windows' }}` keeps `needs.E2E.result` at `success` when only Windows fails, so the existing failure notification does not fire on it.
- Added `git config --system core.longpaths true` on Windows (it covers git's own file operations — here the recursive submodule checkout, whose nested paths run past the 260-character limit), and `shell: bash` on the two `run:` steps that had none — the default shell on a Windows runner is pwsh, which parses neither `set -euo pipefail`/`source` nor the bare `@wso2/...` argument (pwsh reads a leading `@` as splatting).
- `BI_E2E_RETRIES: 0` on the Windows legs only. `playwright.config.js` otherwise retries a failed test twice, and the suite is serial (`workers: 1`) with a 20-minute per-test timeout — while the blockers under *Learning* are open, those retries spend the 60-minute job budget re-running tests that fail for a platform reason rather than a flaky one, and a leg killed by the timeout uploads no artifact at all. Zero retries reaches more distinct tests per run, which is the pass rate this phase exists to measure. Linux gets an empty value, which leaves the config default of 2 in place. `maxFailures: 10` is deliberately untouched: truncating a badly-failing group early is what keeps a leg inside the job timeout and producing an artifact.
- `BI_E2E_TRACE: retain-on-first-failure` on the Windows legs only. `playwright.config.js` sets `trace: 'on-first-retry'`, which captures nothing once retries are 0 — and the trace is the one diagnostic that depends on a retry, since the screenshots and video are the suite's own. Linux gets an empty value and keeps `on-first-retry`.
- `Report` downloads into `e2e-results/<platform>/<group>` and aggregates once per platform. The per-platform root matters: `aggregate-e2e-results.js` derives the group name from the first path segment under its root, so the platform cannot share that segment without landing in every history row's `group` field. A Windows-only aggregation failure becomes a `::warning::` instead of failing the job, mirroring the matrix's `continue-on-error`, and a Windows artifact that will not download after 3 attempts is skipped with a warning rather than failing `Report` — otherwise that one path would still fire the notification on a leg that is not a gate. The two warnings are distinct: results that ran and failed, versus no results produced at all (the legs died before uploading), which is the expected shape while Windows is new.

**`packages/ballerina-extension/e2e-test/playwright.config.js`** — `trace` reads `BI_E2E_TRACE`, defaulting to the existing `on-first-retry`. One line; no change to any run that does not set the variable.

**`.github/scripts/aggregate-e2e-results.js`** — reads `E2E_OS` and adds an `os` field to every NDJSON history row and to the step-summary heading. Rows written before this carry no `os`.

**`.github/workflows/README.md`** — new paragraphs under *Scheduled E2E testing* covering the two-platform matrix, why Windows does not gate and where to flip it, and the retry/timeout reasoning behind `BI_E2E_RETRIES: 0`; plus the workflow and composite-action table rows.

Four places let Windows through, and making it a gate means reverting all four: the `continue-on-error` expression on the `E2E` job, the `windows` branch in `Report`'s download loop, the `windows` branch in `Report`'s aggregation loop, and the `runner.os != 'Windows'` guard on `run-e2e-group`'s re-run step. Turning `BI_E2E_RETRIES` and `BI_E2E_TRACE` back off is a separate call, and should happen once the Windows pass rate is high enough that a retry distinguishes flakiness rather than masking a platform failure.

## UI Component Development
N/A — CI configuration only, no UI change.

## User stories
As a developer, a Windows-only regression in the extension is surfaced by the scheduled E2E run instead of reaching users unnoticed.

## Release note
N/A — CI only, no user-facing change.

## Documentation
`.github/workflows/README.md` is updated in this PR. No product documentation impact.

## Training
N/A

## Certification
N/A — CI configuration change with no impact on certification exams.

## Marketing
N/A

## Automation tests
- Unit tests
  > N/A — no product code changed.
- Integration tests
  > This PR *is* test infrastructure. Verified locally: both YAML files and the JS parse; the per-platform aggregation was exercised against fixtures — Linux-pass/Windows-fail exits 0 and emits 8 correctly `os`-tagged history rows; Linux-fail exits 1; missing Windows artifacts neither gate nor leave an empty stats file behind (an empty one would make the `Persist` step attempt an empty commit).
  > The gating branches were re-checked after each round of review changes, running the `Aggregate results` step extracted from the workflow itself rather than a hand-copy, over six scenarios: Linux-pass/Windows-fail exits 0, Linux-fail/Windows-pass exits 1, both-pass exits 0, Windows-absent exits 0 or 1 following Linux alone, and an empty `PLATFORMS_JSON` (the `Prepare`-failed shape) exits 1 instead of passing vacuously. Each emits the correct warning and writes only the platforms that produced rows.
  > Verified on a real Windows run: [33469593941](https://github.com/ballerina-platform/ballerina-vscode/actions/runs/33469593941), a `workflow_dispatch` on this branch. All 8 legs executed, all 8 uploaded artifacts, and both platforms aggregated. It also confirms that sourcing a checked-out `.sh` on a Windows runner works — the legs ran `retry.sh` twice before reaching the tests.
  > The `continue-on-error` behaviour is confirmed separately, by [33495692184](https://github.com/ballerina-platform/ballerina-vscode/actions/runs/33495692184) on a branch off this one. Run 33469593941 could not confirm it: with the re-run step skipped on Windows, no step in those legs ever failed, so the job succeeded on its own and the job-level `continue-on-error` was never reached. Once `Surface Windows first-attempt failure` made a Windows test failure fail its step, that run showed the mechanism end to end — all four Windows jobs report `conclusion: failure` and show red, while `needs.E2E.result` still evaluated to `success`, so `Notify on failure` was skipped and the workflow passed. A Windows failure is therefore visible without gating, which is exactly the posture this PR is aiming for.

## Security checks
- Ran FindSecurityBugs plugin and verified report? N/A — no Java code changed.

## Samples
N/A

## Related PRs
None.

## Migrations (if applicable)
None. Existing `e2e-metrics` history rows remain valid; they simply have no `os` field, which a reader should treat as `linux`.

## Test environment
GitHub-hosted `ubuntu-latest` and `windows-latest` runners; Node 22.x, pnpm 10.10.0.

## Learning
The Windows legs do fail, as expected. The first run measured:

```
## E2E flakiness report (linux)
Groups reported: 4 · Total: 136 · Passed: 120 · Failed: 0  · Skipped: 16 · Flaky: 11

## E2E flakiness report (windows)
Groups reported: 4 · Total: 136 · Passed: 45  · Failed: 20 · Skipped: 71 · Flaky: 0
```

Windows sits at a 33% pass rate. The workflow is green because that is what the non-gating path is for — the failures are recorded in the step summary and the `e2e-metrics` history, and reported as a `::warning::` rather than failing `Report`.

Of the 20 failures, 12 are `EBUSY: resource busy or locked, rmdir ...\data\test_project` raised from `wipeAndRecreateDataFolder`, which is the `SIGKILL` teardown blocker below — note that it surfaces in the *next* test's setup, not in teardown, where the failure is caught and only warned about. The remaining 8 are locator timeouts (5) and assertion failures (3), not yet triaged. The 71 skips are `maxFailures` truncating groups once the `EBUSY` cascade starts, so fixing that one blocker should move the pass rate substantially on its own.

The audit found these blockers, all test-code rather than CI, and out of scope for this phase:
- `e2e-test/e2e-playwright-tests/utils/helpers/setup.ts` — `execFileSync('bal', ...)` cannot spawn `bal.bat` without `shell: true`.
- `e2e-test/e2e-playwright-tests/file-integration/directory.spec.ts` — a hard-coded `/tmp/wso2/bi/sample` path.
- `e2e-test/e2e-playwright-tests/utils/helpers/setup.ts` — a `zip -r` shell-out for failure snapshots; it is caught, so it only loses snapshots.
- `test.list.ts` teardown kills Electron with `SIGKILL`, which orphans its children on Windows and leaves file locks that make the subsequent `fs.rmSync` of the project folder fail.
- In `submodules/wso2-vscode-extensions`: `playwright-vscode-tester/src/utils.ts` gates the VS Code download on a macOS-only bundle name, so VS Code is re-downloaded on every launch. Pre-existing on Linux too, but the Windows `.zip` unpack is slower and brings the 60-minute job timeout closer.

What would make these expensive rather than merely noisy is the interaction with the run's shape: the suite is serial, each test may take 20 minutes, and Playwright retries a failure twice on top of a second `--last-failed` pass. A group with a handful of platform failures would spend its 60-minute budget on re-runs and could be killed before uploading anything, recording nothing at all — the opposite of the point. That is what `BI_E2E_RETRIES: 0` and the re-run guard are for, and the run bears it out: the Windows legs finished in 14–19 minutes against Linux's 14–19, and all four uploaded their artifacts. `.github/workflows/README.md` carries the full reasoning.

Running Windows non-blocking first is what makes it safe to land these one at a time and watch the pass rate move, rather than holding the whole platform back until every one is fixed.

## Summary

- Extend scheduled E2E testing to Linux and Windows runners.
- Add platform-aware artifacts and conditional Linux setup.
- Run eight platform-and-group matrix jobs.
- Use Windows-compatible test execution with zero Playwright retries.
- Keep Windows failures non-blocking during the initial rollout.
- Aggregate results separately by platform and include `os` in history records.
- Update trace handling and workflow documentation.

### DIFF
--- a/.github/actions/run-e2e-group/action.yml
+++ b/.github/actions/run-e2e-group/action.yml
@@ -60,6 +60,12 @@ runs:
       env:
         # Not under test-results/ (Playwright's outputDir, wiped every invocation).
         PLAYWRIGHT_JSON_OUTPUT_FILE: e2e-reports/e2e-results-first.json
+        # Passed through env, not interpolated into the run: block below — a `${{ }}`
+        # expression is substituted into the script text before the shell parses it, so a
+        # caller-supplied group would be executed rather than matched. Both callers pass a
+        # literal matrix value today; this keeps that from being load-bearing. Same reason
+        # setup-ballerina/action.yml passes gradlePropertiesRef this way.
+        TEST_GROUP: ${{ inputs.group }}
       run: |
         # Windows runs VS Code headed on the runner's own desktop, so there is no xvfb
         # display to create and no ffmpeg X11 screen grab to wrap the run in. The suite
@@ -68,13 +74,13 @@ runs:
         # only the whole-session record.mp4 is Linux-only.
         if [ "$RUNNER_OS" = "Windows" ]; then
           cd packages/ballerina-extension
-          pnpm exec playwright test -c e2e-test/playwright.config.js --grep ${{ inputs.group }}
+          pnpm exec playwright test -c e2e-test/playwright.config.js --grep "$TEST_GROUP"
           exit
         fi
         export $(dbus-launch)
         export DISPLAY=:98.0
         cd packages/ballerina-extension
-        xvfb-run --server-num 98.0 -s "-ac -screen 0 1920x1080x24" bash -c "ffmpeg -y -nostdin -f x11grab -r 30 -s 1920x1080 -i :98.0 -c:v libx264 -preset ultrafast -pix_fmt yuv420p -movflags +faststart -hide_banner -loglevel error record.mp4 & pnpm exec playwright test -c e2e-test/playwright.config.js --grep ${{ inputs.group }}"
+        xvfb-run --server-num 98.0 -s "-ac -screen 0 1920x1080x24" bash -c 'ffmpeg -y -nostdin -f x11grab -r 30 -s 1920x1080 -i :98.0 -c:v libx264 -preset ultrafast -pix_fmt yuv420p -movflags +faststart -hide_banner -loglevel error record.mp4 & pnpm exec playwright test -c e2e-test/playwright.config.js --grep "$TEST_GROUP"'
 
     # Not run automatically on Windows: with maxFailures 10 and a 20-minute per-test
     # timeout, a second --last-failed pass over platform failures can spend what is left
@@ -88,17 +94,19 @@ runs:
       env:
         # run_attempt-suffixed so a second re-run doesn't overwrite the first's report.
         PLAYWRIGHT_JSON_OUTPUT_FILE: e2e-reports/e2e-results-rerun-${{ github.run_attempt }}.json
+        # See the first-attempt step for why this is an env var and not an interpolation.
+        TEST_GROUP: ${{ inputs.group }}
       run: |
         rm -rf packages/ballerina-extension/e2e-test/data/new-project
         if [ "$RUNNER_OS" = "Windows" ]; then
           cd packages/ballerina-extension
-          pnpm exec playwright test -c e2e-test/playwright.config.js --grep ${{ inputs.group }} --last-failed
+          pnpm exec playwright test -c e2e-test/playwright.config.js --grep "$TEST_GROUP" --last-failed
           exit
         fi
         export $(dbus-launch)
         export DISPLAY=:98.0
         cd packages/ballerina-extension
-        xvfb-run --server-num 98.0 -s "-ac -screen 0 1920x1080x24" bash -c "ffmpeg -y -nostdin -f x11grab -r 30 -s 1920x1080 -i :98.0 -c:v libx264 -preset ultrafast -pix_fmt yuv420p -movflags +faststart -hide_banner -loglevel error record.mp4 & pnpm exec playwright test -c e2e-test/playwright.config.js --grep ${{ inputs.group }} --last-failed"
+        xvfb-run --server-num 98.0 -s "-ac -screen 0 1920x1080x24" bash -c 'ffmpeg -y -nostdin -f x11grab -r 30 -s 1920x1080 -i :98.0 -c:v libx264 -preset ultrafast -pix_fmt yuv420p -movflags +faststart -hide_banner -loglevel error record.mp4 & pnpm exec playwright test -c e2e-test/playwright.config.js --grep "$TEST_GROUP" --last-failed'
 
     - name: Upload Ballerina e2e Test results
       uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
--- a/.github/actions/run-e2e-group/action.yml
+++ b/.github/actions/run-e2e-group/action.yml
@@ -62,8 +62,9 @@ runs:
         PLAYWRIGHT_JSON_OUTPUT_FILE: e2e-reports/e2e-results-first.json
       run: |
         # Windows runs VS Code headed on the runner's own desktop, so there is no xvfb
-        # display to create and no ffmpeg X11 screen grab to wrap the run in. Playwright's
-        # own per-test videos, screenshots and traces are still produced and uploaded;
+        # display to create and no ffmpeg X11 screen grab to wrap the run in. The suite
+        # captures its own failure screenshots (setup.ts) and session video (test.list.ts)
+        # on both platforms, and BI_E2E_TRACE keeps traces working where retries are 0;
         # only the whole-session record.mp4 is Linux-only.
         if [ "$RUNNER_OS" = "Windows" ]; then
           cd packages/ballerina-extension
@@ -75,9 +76,15 @@ runs:
         cd packages/ballerina-extension
         xvfb-run --server-num 98.0 -s "-ac -screen 0 1920x1080x24" bash -c "ffmpeg -y -nostdin -f x11grab -r 30 -s 1920x1080 -i :98.0 -c:v libx264 -preset ultrafast -pix_fmt yuv420p -movflags +faststart -hide_banner -loglevel error record.mp4 & pnpm exec playwright test -c e2e-test/playwright.config.js --grep ${{ inputs.group }}"
 
+    # Not run automatically on Windows: with maxFailures 10 and a 20-minute per-test
+    # timeout, a second --last-failed pass over platform failures can spend what is left
+    # of the 60-minute budget and get the leg killed before it uploads anything, which is
+    # the same cost BI_E2E_RETRIES: 0 removes from the first attempt. The re-run still
+    # happens on a manual "Re-run failed jobs" (check-results success), which is the
+    # resume path it exists for.
     - name: Run Failed Ballerina e2e Test (Re-run)
       shell: bash
-      if: steps.run-tests-first.outcome == 'failure' || steps.check-results.outcome == 'success'
+      if: (steps.run-tests-first.outcome == 'failure' && runner.os != 'Windows') || steps.check-results.outcome == 'success'
       env:
         # run_attempt-suffixed so a second re-run doesn't overwrite the first's report.
         PLAYWRIGHT_JSON_OUTPUT_FILE: e2e-reports/e2e-results-rerun-${{ github.run_attempt }}.json
--- a/.github/actions/run-e2e-group/action.yml
+++ b/.github/actions/run-e2e-group/action.yml
@@ -2,10 +2,11 @@ name: 'Run E2E test group'
 description: >-
   Runs one matrix group of the Ballerina e2e Playwright suite (first attempt, then a
   --last-failed re-run if needed) and uploads its results/recordings/project-data
-  artifacts. Shared by reusable-build.yml's BalE2ETest job and e2e-scheduled.yml's E2E
-  job — see "Scheduled E2E testing" in .github/workflows/README.md for the design
-  rationale. Assumes the workspace is already prepared (deps installed, Ballerina
-  distribution installed) by the calling job.
+  artifacts. Runs on both Linux and Windows runners; every artifact name carries the
+  platform so the two legs never collide. Shared by reusable-build.yml's BalE2ETest job
+  and e2e-scheduled.yml's E2E job — see "Scheduled E2E testing" in
+  .github/workflows/README.md for the design rationale. Assumes the workspace is already
+  prepared (deps installed, Ballerina distribution installed) by the calling job.
 inputs:
   group:
     description: 'Matrix group name, e.g. group1'
@@ -13,6 +14,14 @@ inputs:
 runs:
   using: "composite"
   steps:
+    # Artifact names are lowercase ('linux', 'windows'). GitHub expressions have no
+    # lowercase function, so runner.os ('Linux'/'Windows') is folded here once rather
+    # than passed in as a second input the caller could contradict.
+    - name: Resolve platform label
+      shell: bash
+      id: platform
+      run: echo "label=$(echo "$RUNNER_OS" | tr '[:upper:]' '[:lower:]')" >> "$GITHUB_OUTPUT"
+
     - name: Check for previous test results
       shell: bash
       if: github.run_attempt > 1
@@ -26,16 +35,22 @@ runs:
       uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4
       continue-on-error: true
       with:
-        name: Ballerina-e2e-test-results-linux-${{ inputs.group }}-${{ env.PREVIOUS_ATTEMPT }}
+        name: Ballerina-e2e-test-results-${{ steps.platform.outputs.label }}-${{ inputs.group }}-${{ env.PREVIOUS_ATTEMPT }}
         # Restored e2e-reports/ is deliberately kept, not discarded — see README.
         path: packages/ballerina-extension
 
-    - name: Install packages
+    # Linux only: the Windows runner already has an interactive desktop session, a
+    # working keyring, and no need for a virtual display or dbus.
+    - name: Install display and keyring packages
       shell: bash
+      if: runner.os == 'Linux'
       run: |
         sudo apt-get update -y
         sudo apt-get install -y xvfb ffmpeg gnome-keyring libsecret-1.0 dbus-x11
-        cd packages/ballerina-extension && npx playwright install
+
+    - name: Install Playwright browsers
+      shell: bash
+      run: cd packages/ballerina-extension && npx playwright install
 
     - name: Run Ballerina e2e Test (First Attempt)
       shell: bash
@@ -46,6 +61,15 @@ runs:
         # Not under test-results/ (Playwright's outputDir, wiped every invocation).
         PLAYWRIGHT_JSON_OUTPUT_FILE: e2e-reports/e2e-results-first.json
       run: |
+        # Windows runs VS Code headed on the runner's own desktop, so there is no xvfb
+        # display to create and no ffmpeg X11 screen grab to wrap the run in. Playwright's
+        # own per-test videos, screenshots and traces are still produced and uploaded;
+        # only the whole-session record.mp4 is Linux-only.
+        if [ "$RUNNER_OS" = "Windows" ]; then
+          cd packages/ballerina-extension
+          pnpm exec playwright test -c e2e-test/playwright.config.js --grep ${{ inputs.group }}
+          exit
+        fi
         export $(dbus-launch)
         export DISPLAY=:98.0
         cd packages/ballerina-extension
@@ -59,6 +83,11 @@ runs:
         PLAYWRIGHT_JSON_OUTPUT_FILE: e2e-reports/e2e-results-rerun-${{ github.run_attempt }}.json
       run: |
         rm -rf packages/ballerina-extension/e2e-test/data/new-project
+        if [ "$RUNNER_OS" = "Windows" ]; then
+          cd packages/ballerina-extension
+          pnpm exec playwright test -c e2e-test/playwright.config.js --grep ${{ inputs.group }} --last-failed
+          exit
+        fi
         export $(dbus-launch)
         export DISPLAY=:98.0
         cd packages/ballerina-extension
@@ -68,7 +97,7 @@ runs:
       uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
       if: always()
       with:
-        name: Ballerina-e2e-test-results-linux-${{ inputs.group }}-${{ github.run_attempt }}
+        name: Ballerina-e2e-test-results-${{ steps.platform.outputs.label }}-${{ inputs.group }}-${{ github.run_attempt }}
         path: |
           packages/ballerina-extension/test-results/**
           packages/ballerina-extension/e2e-reports/**
@@ -79,7 +108,7 @@ runs:
       uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
       if: always()
       with:
-        name: Ballerina-e2e-test-recording-linux-${{ inputs.group }}-${{ github.run_attempt }}
+        name: Ballerina-e2e-test-recording-${{ steps.platform.outputs.label }}-${{ inputs.group }}-${{ github.run_attempt }}
         path: |
           packages/ballerina-extension/e2e-test/test-resources/videos/**
           packages/ballerina-extension/e2e-test/test-resources/screenshots/**
@@ -91,7 +120,7 @@ runs:
       uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
       if: always()
       with:
-        name: Ballerina-e2e-test-project-linux-${{ inputs.group }}-${{ github.run_attempt }}
+        name: Ballerina-e2e-test-project-${{ steps.platform.outputs.label }}-${{ inputs.group }}-${{ github.run_attempt }}
         path: |
           packages/ballerina-extension/e2e-test/data/new-project/**
         retention-days: 5
--- a/.github/actions/run-e2e-group/action.yml
+++ b/.github/actions/run-e2e-group/action.yml
@@ -82,6 +82,17 @@ runs:
         cd packages/ballerina-extension
         xvfb-run --server-num 98.0 -s "-ac -screen 0 1920x1080x24" bash -c 'ffmpeg -y -nostdin -f x11grab -r 30 -s 1920x1080 -i :98.0 -c:v libx264 -preset ultrafast -pix_fmt yuv420p -movflags +faststart -hide_banner -loglevel error record.mp4 & pnpm exec playwright test -c e2e-test/playwright.config.js --grep "$TEST_GROUP"'
 
+    # The first-attempt step is continue-on-error, and the re-run below — the only step
+    # that turns a test failure into a job failure — is skipped on Windows. Without this,
+    # no Windows step ever fails, the job-level continue-on-error never fires, and
+    # removing that expression to start gating would change nothing. Re-asserting the
+    # outcome here keeps the four reversions in the README order-independent. Remove this
+    # together with the re-run guard below.
+    - name: Surface Windows first-attempt failure
+      shell: bash
+      if: runner.os == 'Windows' && steps.run-tests-first.outcome == 'failure'
+      run: exit 1
+
     # Not run automatically on Windows: with maxFailures 10 and a 20-minute per-test
     # timeout, a second --last-failed pass over platform failures can spend what is left
     # of the 60-minute budget and get the leg killed before it uploads anything, which is
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -15,7 +15,7 @@ workflow because the extension manifest owns the shared extension/LS version.
 | `reusable-build.yml` | `workflow_call` only | Reusable build pipeline (ballerina-only) |
 | `devBuild.yml` | manual + `workflow_call` | Builds a custom branch as a timestamped pre-release VSIX. It creates workflow artifacts only: no GitHub release and no marketplace publication. `schedule.yml` reuses this workflow after stamping the nightly branch. |
 | `schedule.yml` | nightly cron + manual | Syncs the `builds/nightly` branch, runs the LS multi-branch pack/test/Windows-build matrix, calls `devBuild.yml`, and moves the `nightly` tag after every job passes. Manual runs can select a source branch and otherwise behave exactly like scheduled nightlies, including notifications. The VSIX remains a workflow artifact; no GitHub Release is created. See [Versioning](#versioning) and [The nightly branch](#the-nightly-branch). |
-| `e2e-scheduled.yml` | every 6h cron + manual | Runs the Playwright E2E suite against the nightly VSIX without rebuilding, and tracks flakiness over time. See [Scheduled E2E testing](#scheduled-e2e-testing). |
+| `e2e-scheduled.yml` | every 6h cron + manual | Runs the Playwright E2E suite on Linux and Windows against the nightly VSIX without rebuilding, and tracks flakiness over time. See [Scheduled E2E testing](#scheduled-e2e-testing). |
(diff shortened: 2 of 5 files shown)


.github/workflows/e2e-scheduled.yml
```diff
@@ -58,22 +63,47 @@ jobs:
...
+      # Zero retries covers more distinct tests per run, which is the pass rate this phase
+      # is here to measure. Empty on Linux, which leaves the config default of 2 in place.
+      BI_E2E_RETRIES: ${{ matrix.os == 'windows' && '0' || '' }}
     steps:
+      # Only affects git's own file operations — here, the recursive submodule checkout,
+      # whose nested paths run past the 260-character legacy limit.
+      - name: Enable long paths for Git
+        if: runner.os == 'Windows'
+        shell: bash
+        run: git config --system core.longpaths true
```
samithkavishke, line 100: **1. Blocking: `core.autocrlf` is not disabled, so this leg likely dies before the first test.**

Long paths are handled here, but not line endings. Windows runners ship Git for Windows' default `core.autocrlf=true`, and the repo has no `.gitattributes`, so `actions/checkout` writes `.github/scripts/retry.sh` with CRLF. The very next `run:` block sources it under `bash -eo pipefail` (L104-L111), and `setup-ballerina` sources the same file on Windows too — this workflow now takes that branch because it passes `gradlePropertiesRef: nightly` (https://github.com/ballerina-platform/ballerina-vscode/blob/99a251932f2db94645f83a8db194135a7c3a5578/.github/actions/setup-ballerina/action.yml#L31-L32).

With CRLF, bash fails one of two ways: it does not parse the function at all (`retry() {\r` — `{\r` is not the reserved word `{`), or it defines it with `\r` glued onto every expansion, so `local max_attempts="$1"` becomes `6\r`, `seq 1 "6\r"` errors, and `retry` always returns 1. Either way the step exits 1 and the job never reaches the tests.

Worth saying there is no existing evidence this path works: this PR is the first place in the repo that sources a checked-out `.sh` from a Windows runner. `schedule.yml`'s Windows job uses `cmd`/`gradlew.bat` and calls `setup-ballerina` *without* the ref, so it never hits it.

```suggestion
      - name: Configure Git for Windows
        if: runner.os == 'Windows'
        shell: bash
        run: |
          # Only affects git's own file operations — here, the recursive submodule
          # checkout, whose nested paths run past the 260-character legacy limit.
          git config --system core.longpaths true
          # Checkout would otherwise write .github/scripts/*.sh with CRLF, which
          # bash cannot source.
          git config --system core.autocrlf false
```

A repo-level `.gitattributes` with `*.sh text eol=lf` would fix it more durably, if you would rather it not depend on a workflow step.

.github/actions/run-e2e-group/action.yml
```diff
@@ -46,6 +61,15 @@ runs:
         # Not under test-results/ (Playwright's outputDir, wiped every invocation).
         PLAYWRIGHT_JSON_OUTPUT_FILE: e2e-reports/e2e-results-first.json
       run: |
+        # Windows runs VS Code headed on the runner's own desktop, so there is no xvfb
+        # display to create and no ffmpeg X11 screen grab to wrap the run in. Playwright's
+        # own per-test videos, screenshots and traces are still produced and uploaded;
+        # only the whole-session record.mp4 is Linux-only.
```
samithkavishke, line 74: **2. This claim is not true today — a failing Windows leg produces no visual diagnostics at all.**

`playwright.config.js` sets neither `use.video` nor `use.screenshot`, and both default to `off`; traces are retry-gated at `trace: 'on-first-retry'` (https://github.com/ballerina-platform/ballerina-vscode/blob/99a251932f2db94645f83a8db194135a7c3a5578/packages/ballerina-extension/e2e-test/playwright.config.js#L9-L14). Combined with `BI_E2E_RETRIES: 0` on the Windows legs, a first-attempt failure is never retried, so `on-first-retry` never fires either.

Net effect on Windows: no `record.mp4` (by design), no video, no screenshot, **and** no trace — only the JSON report. That is the opposite of what this phase needs, since the point of the rollout is diagnosing Windows-only failures.

Either give the Windows legs `trace: 'retain-on-failure'` (cheap, and the only artifact that survives zero retries), or drop the claim from this comment and from `README.md:244`.

.github/actions/run-e2e-group/action.yml
```diff
@@ -59,6 +83,11 @@ runs:
         PLAYWRIGHT_JSON_OUTPUT_FILE: e2e-reports/e2e-results-rerun-${{ github.run_attempt }}.json
       run: |
         rm -rf packages/ballerina-extension/e2e-test/data/new-project
+        if [ "$RUNNER_OS" = "Windows" ]; then
+          cd packages/ballerina-extension
+          pnpm exec playwright test -c e2e-test/playwright.config.js --grep ${{ inputs.group }} --last-failed
```
samithkavishke: **3. `BI_E2E_RETRIES: 0` does not remove the re-run cost it was added to remove.**

The reasoning in the env-var comment holds for Playwright's *per-test* retries, but this step runs a second whole invocation regardless — its `if:` (`steps.run-tests-first.outcome == 'failure' || steps.check-results.outcome == 'success'`) is independent of the retry setting.

With `maxFailures: 10` and a 20-minute per-test timeout, this `--last-failed` pass can re-execute up to ten tests that are failing for platform reasons, past whatever is left of the 60-minute budget. So the leg is still killed by `timeout-minutes: 60` and still uploads nothing — precisely the failure mode `BI_E2E_RETRIES: 0` was introduced to avoid. If the goal is "cover more distinct tests per run", this step needs skipping (or capping) on Windows too.

.github/workflows/e2e-scheduled.yml
```diff
@@ -166,10 +210,41 @@ jobs:
           E2E_SOURCE_TAG: nightly
           E2E_SOURCE_SHA: ${{ needs.Prepare.outputs.nightlySha }}
           EXPECTED_GROUPS_JSON: ${{ needs.Prepare.outputs.groups }}
+          PLATFORMS_JSON: ${{ needs.Prepare.outputs.platforms }}
         run: |
           set -o pipefail
-          node .github/scripts/aggregate-e2e-results.js e2e-results e2e-run-stats.jsonl \
-            | tee -a "$GITHUB_STEP_SUMMARY"
+          gate_status=0
+          for platform in $(echo "$PLATFORMS_JSON" | jq -r '.[]'); do
```
samithkavishke: **4. The aggregation gate can now pass without aggregating anything.**

`Report` runs with `if: always()`, so it also executes when `Prepare` failed and `needs.Prepare.outputs.platforms` is empty. `for platform in $(echo "" | jq -r '.[]')` then iterates zero times, `gate_status` stays `0`, and the step exits 0. Before this PR the script always ran once and exited 1 with *"No e2e-results\*.json found under …; the artifact contract is broken"*.

The run is still notified via `needs.E2E.result != 'success'`, so this is not a missed alert — but the step-level contract check is silently gone. An empty-list guard that fails the step would restore it. Same zero-iteration behaviour in the download loop at L182.

.github/workflows/e2e-scheduled.yml
```diff
@@ -166,10 +210,41 @@ jobs:
...
+              | tee -a "$GITHUB_STEP_SUMMARY"; then
+              # Mirrors the E2E job's continue-on-error: Windows results are summarised and
+              # persisted, but a Windows-only failure must not fail this job either, or the
+              # notification below would fire on a leg that is not a gate yet.
+              if [ "$platform" = "windows" ]; then
+                # The script writes the stats file only when it read at least one test, so
+                # its absence separates "the suite ran and failed" from "the legs produced
+                # nothing" — the latter being the expected shape while Windows is new.
+                if [ -s "e2e-run-stats-${platform}.jsonl" ]; then
+                  echo "::warning::Windows e2e results contain failures; recorded but not gating."
```
samithkavishke: **5. Nit: this warning can name the wrong cause.**

`aggregate-e2e-results.js` exits 1 for four distinct reasons — failed tests, unreadable reports, `missingGroups`, and `truncatedGroups`. Since `EXPECTED_GROUPS_JSON` always lists all four groups, any absent Windows artifact (the expected shape while Windows is new, as the comment two lines up says) takes this branch and reports "contain failures" even when everything that ran passed.

`"Windows e2e aggregation reported problems"` would be accurate, or the script could signal which of the four it hit.

.github/workflows/e2e-scheduled.yml
NipunaRanasinghe: Checked this one against a real run before changing anything, and I don't think it holds — `core.autocrlf` is not converting on these runners.

`schedule.yml`'s `ls-windows-build` runs on `windows-latest` and calls `setup-ballerina` **without** the ref, so it takes the working-tree branch — which reads a checked-out text file with `sed` and then anchors it:

```
version=$(sed -n 's/^ballerinaLangVersion=//p' "${properties}")
if [[ ! "${version}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then ... exit 1
```

If checkout were writing CRLF, `version` would be `2201.13.4\r`, `$` would not match, and that step would fail. It doesn't. In run 33331106182 (job 99309827283) the step passes and the next step resolves `BALLERINA_VERSION=2201.13.4` cleanly. The checkout log also shows no `core.autocrlf` being set either way — the runner image ships Git with conversion off.

So the CRLF mechanism you describe is real, but the premise that checkout produces CRLF isn't true here. I've left the step as-is rather than adding a config line whose comment would assert a problem that doesn't exist.

Agreed that `.gitattributes` with `*.sh text eol=lf` is the more durable mechanism if we want insurance against an image change or a self-hosted Windows runner later — happy to do that as its own PR, since it affects every consumer of the repo, not just this workflow. And the `workflow_dispatch` Windows run settles it either way before this merges.

.github/actions/run-e2e-group/action.yml
NipunaRanasinghe: **Traces: correct.** `trace: 'on-first-retry'` plus `BI_E2E_RETRIES: 0` means no trace is ever captured on Windows. Fixed by parameterising the config and giving the Windows legs a retry-independent mode:

```js
trace: process.env.BI_E2E_TRACE || 'on-first-retry',
```

with `BI_E2E_TRACE: retain-on-first-failure` on the Windows legs only (Playwright 1.55.1, so it's available). Linux is unchanged.

**Video and screenshots: not correct.** These don't come from `use.video`/`use.screenshot` — you're right that both are `off`, but the suite captures them itself and always has: `test.list.ts:242-243` saves `test_<ts>.webm` into `test-resources/videos`, and `setup.ts:111-116` writes `captureFailureScreenshot` PNGs into `test-resources/screenshots`. Both directories are already in the recordings artifact, and neither is retry-gated, so both survive on Windows.

The comment was still misleading in attributing them to Playwright, so I've reworded it here and at `README.md:244` to say which code produces what.

.github/actions/run-e2e-group/action.yml
NipunaRanasinghe: Valid — the retry setting doesn't touch this step, and a second `--last-failed` pass over up to ten platform failures is the same budget problem in a different place.

Skipped it on Windows for a first attempt:

```yaml
if: (steps.run-tests-first.outcome == 'failure' && runner.os != 'Windows') || steps.check-results.outcome == 'success'
```

Kept the `check-results.outcome == 'success'` clause so a manual "Re-run failed jobs" on a Windows leg still resumes via `--last-failed` — that's the path this step exists for, and dropping it entirely would make a re-run execute nothing at all (the first-attempt step is already skipped on a re-run).

Noted in the README that this guard comes off together with the `continue-on-error` when Windows becomes a gate, since a gating platform wants the re-run back.

.github/workflows/e2e-scheduled.yml
NipunaRanasinghe: Valid, and a regression from this PR — before it the script ran unconditionally and exited 1. Added the guard to both loops:

```bash
platforms=$(echo "$PLATFORMS_JSON" | jq -r '.[]')
if [ -z "$platforms" ]; then
  echo "::error::Prepare published no platform list; nothing to aggregate."
  exit 1
fi
```

Verified against fixtures: empty `PLATFORMS_JSON` now exits 1 with that error, and the five real gate scenarios are unaffected.

.github/workflows/e2e-scheduled.yml
NipunaRanasinghe: Valid — with `EXPECTED_GROUPS_JSON` always listing four groups, a partly-missing Windows platform takes that branch even when everything that ran passed.

Reworded to `"Windows e2e aggregation reported problems (see the report above); recorded but not gating."` — the step summary directly above already itemises which of the four it was, so the warning points there rather than guessing. Left the script's exit contract alone.

.github/actions/run-e2e-group/action.yml
```diff
@@ -46,19 +61,40 @@ runs:
...
         PLAYWRIGHT_JSON_OUTPUT_FILE: e2e-reports/e2e-results-first.json
       run: |
+        # Windows runs VS Code headed on the runner's own desktop, so there is no xvfb
+        # display to create and no ffmpeg X11 screen grab to wrap the run in. The suite
+        # captures its own failure screenshots (setup.ts) and session video (test.list.ts)
+        # on both platforms, and BI_E2E_TRACE keeps traces working where retries are 0;
+        # only the whole-session record.mp4 is Linux-only.
+        if [ "$RUNNER_OS" = "Windows" ]; then
+          cd packages/ballerina-extension
+          pnpm exec playwright test -c e2e-test/playwright.config.js --grep ${{ inputs.group }}
```
samithkavishke: Non-blocking, and pre-existing in shape rather than introduced here: the Windows branches add two more `${{ inputs.group }}` interpolations directly into `run:` blocks (this line and the re-run at L95). Not exploitable — the value comes from a static matrix in both callers — but it's the exact pattern `setup-ballerina` deliberately avoids, with a comment explaining why:

```yaml
env:
  # Passed through env, not interpolated directly into the script below: this
  # input could in principle carry attacker-controlled text from a future caller...
  GRADLE_PROPERTIES_REF: ${{ inputs.gradlePropertiesRef }}
```

Since all four occurrences would move together, this is worth doing as a small follow-up on this action rather than expanding the diff here.

.github/workflows/README.md
```diff
@@ -237,6 +237,49 @@ Ballerina distribution version to what the nightly LS actually shipped with, not
...
+**Windows runs but does not gate.** The Windows legs set `continue-on-error`, and `Report` turns a
+Windows-only aggregation failure into a warning, so `needs.E2E.result` stays `success` and the chat
+notification does not fire on a Windows failure. Windows results still reach the step summary and the
+`e2e-metrics` history — `aggregate-e2e-results.js` runs once per platform, and every history row now
+carries an `os` field (absent on rows written before this). It is not a gate yet because parts of
+the suite are still Linux-only — a `zip` shell-out for failure snapshots, a `bal` invocation that
+does not resolve `bal.bat`, a hard-coded `/tmp` path in one spec, and `SIGKILL` teardown that leaves
+Windows file locks behind. Those are tracked separately.
+
+There are three places that let Windows through, and making it a gate means reverting all three: the
```
samithkavishke: Nit: this says "three places" and then the next sentence adds the `runner.os != 'Windows'` guard as a fourth thing to revert at the same time, while the PR description counts four. Reads fine, but "four places" folding the guard into the list would save the next person reconciling the two.

.github/workflows/e2e-scheduled.yml
NipunaRanasinghe: Settled by a real run now, rather than by my inference — this doesn't reproduce.

`workflow_dispatch` on this branch: [run 33469593941](https://github.com/ballerina-platform/ballerina-vscode/actions/runs/33469593941). Step results for `Run Ballerina e2e Test (windows group1)` (job 99736485157):

```
success  Enable long paths for Git
success  Checkout repository
success  Download nightly VSIX          <- sources .github/scripts/retry.sh
success  Setup Rush
success  Build playwright-vscode-tester
success  Install Ballerina distribution <- sources retry.sh again (nightly ref branch)
success  Run e2e test group
```

Both steps that `source` the checked-out `retry.sh` on a Windows runner succeeded, and the leg went on to execute 136 tests. Under the CRLF failure you describe, the first `source` would have failed and the job would have stopped there — `retry` would either not parse or `seq 1 "6\r"` would error out.

For completeness, the Windows legs did fail — 45 passed, 20 failed, 71 skipped — but the failures are `EBUSY: resource busy or locked, rmdir ...\data\test_project` (12 of 20, the orphaned-Electron file locks from the `SIGKILL` teardown), locator timeouts and assertion failures. None are shell parse errors.

So the mechanism you described is real and worth knowing about; the premise that `actions/checkout` writes CRLF just doesn't hold on these runners. I've left the step as-is.

Still happy to add `.gitattributes` with `*.sh text eol=lf` as a separate PR if you want insurance against a runner-image change or a self-hosted Windows runner later — it affects every consumer of the repo, so it seemed wrong to fold into this one.

samithkavishke (dismissed):
Approve with minor comments

.github/actions/run-e2e-group/action.yml
NipunaRanasinghe: Agreed on both the observation and the sequencing — deferring, since fixing it properly means moving all four `--grep` sites plus the four artifact-name interpolations together, and half of those lines are untouched by this PR.

For the record on exposure: `inputs.group` comes from a literal matrix in both callers (`["group1"..."group4"]` in `e2e-scheduled.yml`, `[group1, group2, group3, group4]` in `reusable-build.yml`), so there is no path for caller-controlled text today. It's a consistency gap with `setup-ballerina`, not a live hole.

Happy to raise the follow-up on `run-e2e-group` — env-passing the group and quoting the `--grep` argument, which also fixes the unquoted expansion inside the `bash -c "..."` on the Linux branches. Will link it here once opened, unless you would rather fold it into whichever PR next touches this action.

.github/workflows/README.md
NipunaRanasinghe: Fixed in f86f2dd — my inconsistency: I updated the PR description to "four" and left the README at "three" plus a trailing sentence.

Now reads as one list:

> There are four places that let Windows through, and making it a gate means reverting all four: the `continue-on-error` expression on the `E2E` job, the `windows` branch in `Report`'s download loop, the `windows` branch in `Report`'s aggregation loop, and the `runner.os != 'Windows'` guard on `run-e2e-group`'s re-run step — a gating platform wants the `--last-failed` pass back.

.github/actions/run-e2e-group/action.yml
NipunaRanasinghe: Folded into this PR after all — 2a92acf. It turned out to sit inside steps this PR already rewrites (both `run:` blocks got the Windows guard clause), and this PR added two of the four sites itself, so leaving them for later meant shipping new instances of the pattern and cleaning them up afterwards.

All four `--grep` sites now take `TEST_GROUP` from the step `env:`. The artifact `name:` fields are left as-is — they are action inputs under `with:`, not shell, so there is no substitution-into-script there.

One thing worth flagging, because the obvious version of this fix does not actually work. Env-passing alone leaves the two Linux sites injectable:

```bash
bash -c "... --grep \"$TEST_GROUP\""
```

The **outer** shell expands `$TEST_GROUP` into the string before `bash -c` re-parses it, so a `"` in the value still breaks out. I checked with a probe standing in for the playwright call:

```
TEST_GROUP='g"; touch /tmp/PWNED; #'   ->  argc=2 [--grep] [g]   pwned=YES
```

The fix is to single-quote the `bash -c` body so the outer shell never touches it and the inner shell expands the variable from the inherited environment:

```bash
bash -c 'ffmpeg ... & pnpm exec playwright test -c e2e-test/playwright.config.js --grep "$TEST_GROUP"'
```

Same probe against the line as it now stands in the file, for both `"` and `'` payloads:

```
g"; touch /tmp/PWNED; #   ->  argc=2 [--grep] [g"; touch /tmp/PWNED; #]   pwned=no
g'; touch /tmp/PWNED; #   ->  argc=2 [--grep] [g'; touch /tmp/PWNED; #]   pwned=no
```

No behaviour change for the real values — `group1`..`group4` pass through as a single argument either way.

.github/actions/run-e2e-group/action.yml
```diff
@@ -46,19 +61,40 @@ runs:
...
+    # Not run automatically on Windows: with maxFailures 10 and a 20-minute per-test
+    # timeout, a second --last-failed pass over platform failures can spend what is left
+    # of the 60-minute budget and get the leg killed before it uploads anything, which is
+    # the same cost BI_E2E_RETRIES: 0 removes from the first attempt. The re-run still
+    # happens on a manual "Re-run failed jobs" (check-results success), which is the
+    # resume path it exists for.
     - name: Run Failed Ballerina e2e Test (Re-run)
       shell: bash
-      if: steps.run-tests-first.outcome == 'failure' || steps.check-results.outcome == 'success'
+      if: (steps.run-tests-first.outcome == 'failure' && runner.os != 'Windows') || steps.check-results.outcome == 'success'
```
samithkavishke, line 104: **A Windows leg reports green no matter how many tests fail.**

The first-attempt step carries `continue-on-error: true` (L59), so on Linux the only thing that turns a test failure into a *job* failure is this `--last-failed` re-run step, which has no `continue-on-error`. Adding `runner.os != 'Windows'` here removes that last conversion on Windows.

Concretely, a `windows group2` leg where all 10 tests fail: first attempt fails → swallowed by `continue-on-error` → this step skipped → three `if: always()` uploads → **job green**.

Two consequences:

1. `continue-on-error: ${{ matrix.os == 'windows' }}` on the `E2E` job is now dead for the case it was written for. It can only ever fire for infra failures such as the 60-minute timeout. The pass rate this whole phase exists to measure is visible only in the `Report` step summary — the Windows matrix legs show a green check either way, which is a live observability gap, not just a cosmetic one.
2. It makes the flip-to-gating list order-sensitive. Removing the `continue-on-error` expression first — the first of the four reversions the README lists — changes nothing, because with this guard still in place no Windows step ever fails. The guard *is* one of the four, so this is documented rather than hidden, but the dependency between them isn't.

Cheapest fix that keeps the re-run skipped: re-assert the outcome in a Windows-only step right after the first attempt, and let the job-level `continue-on-error` do the job it was written for.

```yaml
- name: Surface Windows first-attempt failure
  shell: bash
  if: runner.os == 'Windows' && steps.run-tests-first.outcome == 'failure'
  run: exit 1
```

That also makes the four reversions order-independent. If you would rather not add a step, spelling the ordering out in the README would close the second half of this.

NipunaRanasinghe: Correct on both counts, and the second one caught a false claim of mine. Fixed in 085e7e4 with your suggested step verbatim.

On point 1 — it is worse than a cosmetic gap. I had written in the PR description that run 33469593941 *confirmed* `continue-on-error` keeps `needs.E2E.result` at `success` when only Windows fails. It confirmed no such thing. Checking the job now:

```
job conclusion: success
  success  Run e2e test group
  success  Complete job
```

Every step green on a leg with 20 test failures. With the re-run skipped, nothing failed, so the job succeeded on its own merits and the job-level `continue-on-error` was never reached. I had verified the outcome and mistaken it for verifying the mechanism. The description is corrected.

I did not verify your framing that removing the guard would make the leg show red — that depends on whether a job-level `continue-on-error` renders a failed job green in the API, which neither of us has a run to prove. It does not matter for the fix: with the surface step in place a Windows test failure now fails a step, so the next run exercises `continue-on-error` for real either way.

On point 2, one wrinkle worth recording. Placing the step before the re-run means it is not independent of the guard: after a flip, a first-attempt failure would fail this step and skip the re-run. So it does make your case work — removing `continue-on-error` alone now gates, which it did not before — but it creates a pair rather than four fully independent items. I have documented that in the README rather than leaving it implicit:

> That guard pairs with the `Surface Windows first-attempt failure` step just above it, which exists only because the guard removes the one step that would otherwise fail the job; drop both together.

If you would rather have four genuinely independent reversions, the alternative is dropping the re-run guard and letting the `--last-failed` pass run on Windows again — but that reintroduces the budget problem it was added for, so the pair seems the better trade.

samithkavishke (approved):
LGTM

PR #970, opened 2026-08-31
Durable workflow designer: agent creation form, reviewer roles, data events, and palette colouring

rdulmina:
## Purpose

Six fixes to the durable workflow designer: the durable agent creation form, reviewer-role editing, data event binding, and the colouring of the node palette and the agent canvas.

Fixes https://github.com/wso2/product-integrator/issues/2256
Fixes https://github.com/wso2/product-integrator/issues/2232
Fixes https://github.com/wso2/product-integrator/issues/2222
Fixes https://github.com/wso2/product-integrator/issues/2259
Fixes https://github.com/wso2/product-integrator/issues/2260
Fixes https://github.com/wso2/product-integrator/issues/2261

### Durable agent creation asks for the model and the system prompt (#2256)

The creation form asked only for a name, and generation reused that name as the agent's role and its (hidden) description as the instructions — so a new agent's prompt read as an identifier and its instructions were always empty, and the model was picked silently from whatever provider the package happened to declare.

The form now asks for **Model**, **Role**, **Instructions** and an optional **Input Data Type**. Model is declared as an `ai:ModelProvider`-typed expression, which is what makes the webview render it as the model-provider selector — the same field the Chat Agent Service creation form gets from the `ai:Agent` signature. Role and Instructions are required prompt-or-expression fields.

```ballerina
final workflow:DurableAgent claimAgent = check new ({
    systemPrompt: {
        role: string `Expense claim assistant`,
        instructions: string `Process expense claims end to end. Validate each claim with
        validateClaim first and reject invalid claims with a clear reason.`
    },
    model: wso2ModelProvider,
    inputType: ClaimPayload
});
```

Choosing the WSO2 default still binds to a named module-level variable rather than inlining `check ai:getDefaultModelProvider()` in the config, since the agent box's model circle is a dropdown over provider variables; an existing `ai:Wso2ModelProvider` is reused rather than a second declared. The field ships empty on purpose — resolving providers to prefill it would make a node-template fetch compile the package, which is enough to leave the `ballerina:workflow` compiler plugin unable to initialize on a later compilation in the same session.

### Reviewer roles are editable as text (#2232)

A reviewer-role field holds a `string|string[]` but offered only the expression mode, so a single role had to be hand-written with its quotes — while Await Human Task, whose `userRoles` is derived from the parameter and so expands one mode per union member, let it be typed as plain text. The three role fields now declare the same three modes, and the value is rendered through `Property#toSourceCode()` so text and list modes reach source as Ballerina rather than as their own wire syntax.

### A string-valued field opens in its text mode (#2260)

With no mode chosen, a field carrying a value fell to the last mode it declares — the expression one. Right for a reference, a call or a list; wrong for a plain string, which came back quoted. A string literal the text mode can edit now opens there and renders unquoted; every other shape still lands on the expression mode. Scoped to a *leading* text mode, since a record-typed field reports every value as compatible and would otherwise open its record editor on a bare reference.

### Agent forms stop misreporting syntax (#2261)

Diagnostics for a node that writes into the durable agent's declaration read the generated source back as a statement. That edit is a list entry or a record field, so read alone it parses as a broken variable declaration — saving a gated activity reported errors describing the reader rather than anything wrong with the entry. Those node kinds now carry no diagnostics.

### Data event sends bind by the channel's declared response (#2222)

A data event channel declares an optional `response`: a channel with one hands that type back from `sendData`, a channel without answers nothing. The send node assumed a `string` turn token either way, so a one-way channel produced a binding for a value that never arrives. The response is now read off the declaration — the declared type when there is one, `_ = check ...` / `error? ...` when there is not.

Register Data Event already carried a `responseType` property, but the default form layout lifts one type field into a slot of its own and skips the rest, so it never rendered and every channel came out one-way; that node keeps template order so both type fields show. The send node's Data Event dropdown also offered a hardcoded `chat` channel whenever the target agent declared none — an agent with no declared channels now offers no options.

### Palette and agent canvas colouring (#2259)

- Starting a durable agent and sending it a data event take the work-doing colours; reading a result back reads differently, so the two halves of a turn do not look alike. Sending a data event reads as pushing the flow into a workflow waiting on it.
- The three prebuilt activities arrive as one node kind with no icon of their own, so nothing matched them and all three were the default foreground. Colour keys on the symbol instead: the two API calls read as the activity family, the email send as the other sends.
- Await Human Task takes Call Activity's colour (asking a person to act is the other way a workflow gets work done) and Sleep takes Await Data Event's (both are the workflow suspending on something outside it). That empties the group added for the two of them, so yellow means error handling again.
- Registered activities, human tasks and data events on the agent canvas drew raw icons and were the default foreground while the same things carry a colour in the palette. They go through `NodeIcon` now — one source for the glyph and its colour — so both surfaces read alike. Glyphs are unchanged.
- An activity gated by a review carries the shield the chat agent puts on a gated tool, reusing that component rather than a copy. It sits at the circle's bottom-right because the remove button owns the top-right here, so the position is a prop defaulting to what the chat agent already used.

## Checks

- [x] `flow-model-generator-ls-extension`: `SourceGeneratorTest` + `NodeTemplateTest` pass (505 tests), re-run against a clean-tree baseline for comparison.
- [x] Webview: `tsc --noEmit` clean on `ballerina-visualizer`; its `src/utils` jest suites pass.
- [x] New source-generation tests for durable agent creation — a selected provider with an input type, the WSO2 default reusing an existing `ai:Wso2ModelProvider`, and the WSO2 default declaring one.
- [x] New reviewer-role coverage — `workflow_operations.bal` now declares a single role and a list of roles, exercised by `workflow_activity_reviewer_roles.json`.
- [ ] The palette/canvas colouring changes are visual and were not re-verified in this pass; they carry no automated coverage.

## Summary

- Expand durable agent creation with model, role, instructions, and optional input data type fields.
- Support text, list, and expression modes for reviewer roles.
- Reopen compatible string literals in text mode.
- Suppress misleading diagnostics for edits within durable agent declarations.
- Handle declared data-event response types correctly and remove invalid channel fallbacks.
- Reuse valid WSO2 default providers and report provider declarations accurately.
- Improve form ordering, node colours, icons, approval badges, and badge interactions.
- Add source-generation, reviewer-role, provider-detection, string-compatibility, and approval-badge tests.


packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
```diff
@@ -745,6 +745,24 @@ public static void addRetryPolicyFormProperties(NodeBuilder nodeBuilder, String
...
+    private static Property buildReviewerRolesSubProperty(String label, String description) {
+        return WorkflowUtil.addRoleFieldTypes(
+                        new Property.Builder<Void>(null)
+                                .metadata()
+                                    .label(label)
+                                    .description(description)
+                                    .stepOut())
+                .value("")
+                .editable(true)
+                .optional(false)
```
dulajdilshan: The description explicitly says an empty value allows any role, and `retryPolicyExpression()` already serializes an empty value as `[]`. Making this sub-property required prevents that documented unrestricted configuration from being saved. Please make it optional (and add/retain a save-to-source case for an empty ManualRetry role).

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentBuilder.java
```diff
@@ -82,62 +96,134 @@ public void setConcreteConstData() {
...
+        // Object model: the agent IS the workflow — only the module-level declaration is
+        // generated. It is started from other artifacts via `<name>.run(...)` or through the
+        // management API, and its capabilities all live on the declaration's config.
+        String modelValue = sourceBuilder.getProperty(DurableAgentRunBuilder.MODEL_KEY)
+                .map(property -> property.value() == null ? "" : property.value().toString().trim())
+                .orElse("");
+        if (modelValue.isEmpty()) {
+            // No model reached the builder (a client that predates the field, say): fall back to
+            // whatever provider the package declares, and to the WSO2 default when it declares none.
+            String existingProvider = resolveExistingModelProvider(sourceBuilder);
```
dulajdilshan, line 185: This searches every module but returns only the bare variable name. If the first provider is declared in a different module, the generated declaration cannot resolve that unqualified name (and the symbol may not be public). Restrict the search to the declaration's module, or emit a valid qualified/imported reference after checking visibility. `resolveWso2ModelProvider()` has the same cross-module issue; please cover a multi-module package in source-generation tests.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -586,14 +589,53 @@ public static List<Option> durableAgentOptions(
...
+     * @param workspaceManager the workspace manager to resolve the project from
+     * @param filePath         the file the statement is generated into
+     * @param targetAgent      the agent variable name to scope to, or {@code null} for all agents
+     * @param eventName        the channel name, unquoted
+     * @return the response type as written, or empty for a one-way or unknown channel
+     */
+    public static Optional<String> declaredAgentEventResponseType(
+            org.ballerinalang.langserver.commons.workspace.WorkspaceManager workspaceManager, Path filePath,
+            String targetAgent, String eventName) {
+        String response = declaredAgentEvents(workspaceManager, filePath, targetAgent).get(eventName);
```
dulajdilshan: The new response-type lookup always scans the default module. For an agent and send node in a non-default package module, the event is not found and `sendData` is generated as one-way even when the channel declares a response. Resolve the module containing `filePath` (or the agent declaration) rather than `getDefaultModule()`, and add a non-default-module response-channel test.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
```diff
@@ -1190,6 +1203,13 @@ export function DurableAgentRunNodeWidget(props: DurableAgentRunNodeWidgetProps)
                                 <div className="connector-icon">{renderCapabilityIcon(item)}</div>
                             </foreignObject>
 
+                            {/* The same shield the chat agent puts on a gated tool. It sits at the
+                                circle's bottom-right rather than its top-right, because the remove
+                                button already owns that corner here. */}
+                            {item.kind === "activity" && isApprovalGated(item) && (
+                                <ApprovalBadge background={NODE_BG_COLOR} x="88.5" y="31.5" />
```
dulajdilshan: `ApprovalBadge` is a sibling `foreignObject` layered over the activity circle; the circle is the element with `onCapabilityClick(item)`. Clicking the visible badge therefore does not open the activity editor. Please put the click handler on a shared SVG parent or forward it through the badge so the entire activity affordance has the same behavior.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -760,6 +807,83 @@ public static String quoteIfBareRole(String value) {
...
+     * Whether the node writes into the durable agent's declaration instead of emitting a statement.
+     * A caller that reads generated source back as a statement has nothing to read for these — the
+     * edit is a list entry or a record field, and parsing it on its own describes a broken
+     * statement rather than anything wrong with the edit.
+     *
+     * @param nodeKind the node kind to test
+     * @return whether the node's source belongs to the agent declaration
+     */
+    public static boolean editsAgentDeclaration(NodeKind nodeKind) {
+        return AGENT_DECLARATION_NODES.contains(nodeKind);
```
NipunaRanasinghe: `AGENT_DECLARATION_NODES` is a `Set.of(...)`, so `contains(null)` throws NPE rather than returning false. `DiagnosticRequest.call()` passes `flowNodeObj.codedata().node()`, which Gson leaves null whenever the client sends a `codedata.node` string that isn't in the `NodeKind` enum (version skew between extension and LS). That used to be harmless — the null went into `NodeBuilder.getNodeFromKind`, which falls back via `getOrDefault` — but now the whole diagnostics request fails.

Suggest `nodeKind != null && AGENT_DECLARATION_NODES.contains(nodeKind)`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -586,14 +589,53 @@ public static List<Option> durableAgentOptions(
...
         try {
             project = workspaceManager.loadProject(filePath);
         } catch (Exception e) {
-            LOGGER.log(Level.WARNING, "Skipping declared agent event options: failed to load the project of "
+            LOGGER.log(Level.WARNING, "Skipping declared agent events: failed to load the project of "
                     + filePath, e);
-            return List.of();
+            return events;
         }
         Module module = project.currentPackage().getDefaultModule();
```
NipunaRanasinghe: This only looks at the default module, but `durableAgentOptions` above iterates `currentPackage.modules()`. So if you pick an agent declared in `modules/<sub>/`, the Agent dropdown offers it but the Data Event dropdown comes back empty. `eventName` is then `""`, and `requireValue(..., EVENT_NAME_KEY, ...)` throws on save — with the old `chat` default gone, the Send Data Event node can't be created for those agents at all.

Scanning the same module set as `durableAgentOptions` fixes this and the response-type issue in `DurableAgentUpdateBuilder`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentUpdateBuilder.java
```diff
@@ -178,20 +175,45 @@ public Map<Path, List<TextEdit>> toSource(SourceBuilder sourceBuilder) {
...
-                .name(variableName)
-                .whiteSpace()
-                .keyword(SyntaxKind.EQUAL_TOKEN);
-        if (checkError) {
-            sourceBuilder.token().keyword(SyntaxKind.CHECK_KEYWORD);
+        // What the send answers with is the channel's own business: a channel that declares a
+        // `response` hands that type back, and one that declares none answers nothing (the field's
+        // default is `()`). So the binding is read off the declaration rather than assumed — binding
+        // a value a one-way channel never produces does not compile.
+        Optional<String> responseType = WorkflowUtil.declaredAgentEventResponseType(
```
NipunaRanasinghe: `declaredAgentEventResponseType` returns `Optional.empty()` for three different situations: the channel is genuinely one-way, the channel wasn't found, or the project couldn't be read (see the catch/log path in `WorkflowUtil`). Only the first means "no response".

So if the agent lives in a non-default module, `loadProject` fails, the user hand-types `eventName` (the field is editable), or `targetAgent(...)` doesn't match, we emit `_ = check agent.sendData(...)` for a channel that declares `response: string`, which doesn't compile. The old unconditional `string` was wrong the other way, but it never dropped a real response.

Worth separating "channel not found" from "channel declares no response" and keeping the old shape for the former.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -760,6 +807,83 @@ public static String quoteIfBareRole(String value) {
...
+     * The role value as Ballerina source. The field is multi-mode, so the raw value is a string in
+     * expression mode, a string template in text mode, and a list of entries in list mode; going
+     * through {@link Property#toSourceCode()} renders each of those, and {@link #quoteIfBareRole}
+     * then quotes a bare word that arrived without a template wrapper (a value read back from
+     * source, say) while leaving a list or a reference alone.
+     *
+     * @param property the role property, or {@code null}
+     * @return the role expression, or an empty string when nothing was entered
+     */
+    public static String roleSource(Property property) {
```
NipunaRanasinghe, line 889: This ends in `quoteIfBareRole(source)`, which only leaves a value alone if it contains one of `[(){}:.`. Now that the Expression entry is typed `string|string[]` and offered as "an expression producing either", a user who switches Reviewer Roles to Expression and types `financeRoles` (a module-level `string[]`) gets `userRoles: "financeRoles"` — a literal role name, silently, and with diagnostics suppressed for these node kinds there's no error either.

Also, the new javadoc says `quoteIfBareRole` leaves "a list or a reference alone" — it doesn't leave a bare identifier reference alone.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -760,6 +807,83 @@ public static String quoteIfBareRole(String value) {
...
+     * full type on the trailing expression entry. The role fields on the activity and agent forms are
+     * hand-built with no parameter symbol to derive from, so they declare the same three modes here
+     * instead of collapsing to expression-only — otherwise the same value is edited two different
+     * ways depending on which form it is opened from.
+     *
+     * @param builder the property builder to add the role input modes to
+     * @param <T>     the builder's step-out target
+     * @return the same builder, for fluent chaining
+     */
+    public static <T> Property.Builder<T> addRoleFieldTypes(Property.Builder<T> builder) {
```
NipunaRanasinghe, line 871: This appends TEXT, REPEATABLE_LIST, EXPRESSION, but the UI only ever shows the first and last: `FieldFactory.tsx` does `newRenderingTypes = [types[0], types[types.length - 1]]`, and both the editor elements and the `ModeSwitcherProvider` get that trimmed array.

So the "build a list of roles item by item" mode is unreachable on all three role fields (and on the retry sub-property in `ActivityCallBuilder`) — users still have to hand-write `["finance", "manager"]` in Expression mode.

packages/ballerina-side-panel/src/components/editors/FieldFactory.tsx
```diff
@@ -155,6 +155,16 @@ export const FieldFactory = (props: FieldFactoryProps) => {
         if (!props.field.value) {
             return props.field.types[0];
         }
+        // With no mode chosen, a valued field falls to the last mode — the expression one. That is
+        // right for a reference, a call or a list, but not for a plain string read back from
+        // source: it opened as an expression and showed its own quotes. So when the field leads
+        // with a text mode and the value is a string literal that mode can edit, start there —
+        // text mode renders it unquoted. Every other shape still lands on the expression mode.
+        const primaryType = props.field.types[0];
+        if (primaryType?.fieldType === "TEXT" && typeof props.field.value === "string"
```
NipunaRanasinghe, line 164: `getIsValueCompatible` for TEXT mode is just `startsWith('"') && endsWith('"')` (or the `string \`...\`` wrapper). An expression like `"Hello " + name + "!"` passes that check, so a TEXT-primary field opens in Text mode, `serializeValue` strips the outer quotes, and the user sees `Hello " + name + "!` as literal text. Saving re-serialises it into a string — the expression is silently rewritten.

Tightening the check to a single complete string literal (no unescaped interior quote) would avoid it.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/diagnostics/DiagnosticRequest.java
```diff
@@ -96,6 +97,14 @@ public JsonElement call() {
             return null;
         }
 
+        // The analysis below reads the edited range back as a statement, which a node writing into
+        // the durable agent's declaration does not produce: its edit is a list entry or a record
+        // field, and read on its own it parses as a broken variable declaration. Reporting that
+        // would describe the reader, not the edit, so these kinds carry no diagnostics.
+        if (WorkflowUtil.editsAgentDeclaration(flowNodeObj.codedata().node())) {
```
NipunaRanasinghe, line 111: This drops *every* diagnostic for the six agent-declaration kinds, not just the misleading parse error. Real problems now save with no form error: a Reviewer Roles value naming an undefined variable, a `requiresApproval` predicate pointing at a function that doesn't exist, an undefined type in Register Data Event's `responseType`.

The narrower fix would be to keep the analysis and read the edit back in its actual syntactic position (list entry / record field) instead of as a statement. If that's too big for this PR, fine, but it's worth a TODO — the roles issue below depends on these diagnostics to be visible.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentBuilder.java
```diff
@@ -171,4 +260,38 @@ private static String resolveExistingModelProvider(SourceBuilder sourceBuilder)
...
+
+    // Narrower than resolveExistingModelProvider: only an `ai:Wso2ModelProvider` variable answers
+    // for an explicit "Default WSO2 Model Provider" pick. Binding that choice to, say, an existing
+    // OpenAI provider variable would silently run the agent on a model the user did not choose.
+    private static String resolveWso2ModelProvider(SourceBuilder sourceBuilder) {
+        try {
+            Package currentPackage = PackageUtil
+                    .loadProject(sourceBuilder.workspaceManager, sourceBuilder.filePath).currentPackage();
+            PackageUtil.getCompilation(currentPackage);
+            for (Module module : currentPackage.modules()) {
```
NipunaRanasinghe: This walks every module and returns the first `ai:Wso2ModelProvider` variable name unqualified. If the only such variable is in `modules/<sub>/` and the agent is being created in the default module, we emit `model: wso2ModelProvider` referencing a symbol that isn't visible there, so the package doesn't compile — the exact thing the comment above says this code prevents.

Scoping the search to the module containing `sourceBuilder.filePath` would fix it. `resolveExistingModelProvider` has the same shape; pre-existing, but the new empty-model fallback path reaches it.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -638,20 +678,27 @@ private static void collectDeclaredEventNames(
...
-                        if (!raw.isEmpty()) {
-                            names.add(raw);
-                        }
+                        name = raw;
+                    } else if ("response".equals(fieldName)) {
+                        response = raw;
                     }
                 }
+                if (!name.isEmpty()) {
+                    events.put(name, response);
```
NipunaRanasinghe: Minor: `put` is last-wins on a duplicate channel name, where the old `LinkedHashSet` kept the first. `targetAgent` is null at template time (`codedata().parentSymbol()` is absent for a statement node in a workflow), so if two agents both declare a channel named e.g. `chat`, they collapse to the later agent's `response` type and `declaredAgentEventResponseType` can hand the send statement the wrong binding.

`putIfAbsent` would at least preserve the previous behaviour.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
```diff
@@ -1190,6 +1203,13 @@ export function DurableAgentRunNodeWidget(props: DurableAgentRunNodeWidgetProps)
                                 <div className="connector-icon">{renderCapabilityIcon(item)}</div>
                             </foreignObject>
 
+                            {/* The same shield the chat agent puts on a gated tool. It sits at the
+                                circle's bottom-right rather than its top-right, because the remove
+                                button already owns that corner here. */}
+                            {item.kind === "activity" && isApprovalGated(item) && (
```
NipunaRanasinghe: The badge is gated on `kind === "activity"`, but `DurableAgentRegisterToolBuilder` also emits `requiresApproval` into a tool entry and `CodeAnalyzer.collectDeclaredCapabilities` maps it into the `tools` capability values. So a tool registered with approval gating renders identically to an ungated one — same "gating reads differently on different surfaces" problem the rest of this section is fixing. Human tasks too, if they can carry it.

Badge coordinates themselves look right — correct mirror of the existing `(88.5, -0.5)` for `cx=80 cy=24 r=22`, and no collision with the remove button.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentUpdateBuilder.java
```diff
@@ -139,7 +136,7 @@ public void setConcreteTemplateData(TemplateContext context) {
                 .codedata()
                     .kind(ParameterData.Kind.REQUIRED.name())
                     .stepOut()
-                .value(eventOptions.get(0).value())
+                .value(eventOptions.isEmpty() ? "" : eventOptions.get(0).value())
```
NipunaRanasinghe, line 139: Defaulting to `""` when the options list is empty is what turns the module-scope gap in `declaredAgentEvents` into a hard failure — `requireValue` on `EVENT_NAME_KEY` then throws on save. Fixing the lookup scope covers this, but it may still be worth keeping the field usable (or surfacing why the list is empty) when no channel can be resolved.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
rdulmina: Agreed — fixed in 5b1d392aab.

`buildReviewerRolesSubProperty` built the field with `.optional(false)`, which feeds `hasIncompleteRequiredFormFields` and blocks submit, while `retryPolicyExpression` already treats a blank value as the unrestricted policy:

```java
case MANUAL_RETRY_VALUE -> {
    String rolesValue = WorkflowUtil.roleSource(properties.get(RETRY_USER_ROLES_KEY));
    // ManualRetry is the reviewer role(s); an empty list means any role may decide.
    yield rolesValue.isBlank() ? "[]" : rolesValue;
}
```

It is `.optional(true)` now.

On the save-to-source case: `to_source/config/activity_call_manual_retry_to_source.json` already pinned `retryPolicy = []` with the role property absent, and it is retained. I added `activity_call_manual_retry_empty_roles_to_source.json` for the shape the form actually submits — the property present but empty — since that goes through `roleSource` rather than the absent-property path. Both generate:

```ballerina
int result = check context->callActivity(sendNotification, {...}, retryPolicy = []);
```

The rest of the diff on this file is the resulting `optional` flip in nine goldens.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentBuilder.java
rdulmina: Confirmed and fixed in 5b1d392aab — though the remedy needed one adjustment to your suggestion, so flagging it.

You are right that scanning every module and returning a bare name is wrong: the name only resolves inside its own module, and need not be public. Both `resolveExistingModelProvider()` and `resolveWso2ModelProvider()` had it.

The adjustment is *which* module to restrict to. It is not the module the user is editing — the declaration never lands there. `SourceBuilder.resolvePath` maps `DURABLE_AGENT` to `WORKFLOW_BAL` against the project root:

```java
case WORKFLOW, DURABLE_AGENT -> WORKFLOW_BAL;
...
Path resolvedPath = workspaceManager.projectRoot(inputPath).resolve(defaultFile);
```

so the declaration always goes into the default module. `sourceBuilder.filePath` is already that resolved target (the constructor assigns `resolvePath(...)`), so both lookups now share:

```java
private static Module declaringModule(SourceBuilder sourceBuilder) {
    ...
    return sourceBuilder.workspaceManager.module(sourceBuilder.filePath).orElse(null);
}
```

which lands on the module the declaration is written into. A package whose only provider lives elsewhere now reads as having none, and `wso2ModelProvider` is declared alongside the agent.

Multi-module coverage added: `to_source/source/workflow_submodule/` puts the package’s only provider in `modules/ops` and leaves the default module without one. `durable_agent_creation_submodule_to_source.json` pins the result —

```ballerina
final ai:Wso2ModelProvider wso2ModelProvider = check ai:getDefaultModelProvider();
final workflow:DurableAgent opsClaimAgent = check new ({ ... model: wso2ModelProvider });
```

Before the fix that same fixture produced `model: opsModel`, which does not resolve from `workflow.bal`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
rdulmina: Confirmed and fixed in 5b1d392aab.

`declaredAgentEvents` hardcoded the default module, so an agent declared in a non-default module was never seen and `declaredAgentEventResponseType` came back empty — indistinguishable from a one-way channel:

```suggestion
        // The agent is declared in the module the send statement is generated into, which is not
        // necessarily the default one: scanning only the default module left a send node in a
        // non-default module unable to find its channel, so the send came out one-way even when
        // the channel declares a response.
        Module module = workspaceManager.module(filePath)
                .orElseGet(() -> project.currentPackage().getDefaultModule());
```

Non-default-module tests added against the new `workflow_submodule` fixture, whose agent lives in `modules/ops` and declares one channel with a response and one without:

- `durable_agent_send_data_submodule_to_source.json` → `string eventToken = check opsAgent.sendData(instanceId, "opsMessage", "hello");`
- `durable_agent_send_data_submodule_one_way_to_source.json` → `_ = check opsAgent.sendData(instanceId, "opsNotice", "hello");`

One related thing I did **not** change, since it is outside this line: `WorkflowUtil.findAgentDeclaration` (~line 441) has the same limitation — its comment still reads "Scans the default module" — so *editing* an agent declared in a non-default module fails the same way. Happy to fix it here or track it separately, whichever you prefer.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
rdulmina: Confirmed and fixed in 5b1d392aab.

The badge is indeed a sibling `foreignObject` painted after the circle, and it carries pointer events, so it swallowed the click. The icon immediately above it already guards against exactly this:

```jsx
<foreignObject x="68" y="12" width="44" height="44" style={{ pointerEvents: "none" }}>
```

I did not take that route for the badge, because it needs pointer events for its own "Requires Approval" tooltip — `pointerEvents: none` would restore the click and silently kill the tooltip. Instead the circle’s handler is forwarded down, so the whole affordance behaves the same:

```jsx
<ApprovalBadge
    background={NODE_BG_COLOR}
    x="88.5"
    y="31.5"
    onClick={() => onCapabilityClick(item)}
/>
```

`ApprovalBadge` takes an optional `onClick` and sets `cursor: pointer` only when one is supplied. `AgentNodeWidget` and `AgentCallNodeWidget` pass none, so their badges are unchanged.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
rdulmina: Confirmed and fixed in e7eea198ee — `Set.of()` throws on a null probe rather than answering `false`, and `codedata.node()` is null on any kind this LS does not know.

```java
// Set.of() throws on a null probe, and Gson leaves node() null whenever the client sends a
// codedata.node this LS does not know (version skew), so the guard is load-bearing.
return nodeKind != null && AGENT_DECLARATION_NODES.contains(nodeKind);
```

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
rdulmina: Confirmed and fixed in e7eea198ee — `events.putIfAbsent(name, response)` now, restoring the first-wins behaviour of the `LinkedHashSet` this replaced.

This matters more after the scope change on the other thread: now that the lookup scans every module, two agents declaring the same channel name are more likely to meet, and `targetAgent` is indeed null at template time.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentBuilder.java
rdulmina: Already fixed — 5b1d392aab, pushed before this review landed, so the thread reads against the older code.

Both `resolveWso2ModelProvider()` and `resolveExistingModelProvider()` now share one lookup scoped exactly as you suggest:

```java
private static Module declaringModule(SourceBuilder sourceBuilder) {
    ...
    return sourceBuilder.workspaceManager.module(sourceBuilder.filePath).orElse(null);
}
```

`sourceBuilder.filePath` is already the resolved target — the constructor assigns `resolvePath(...)`, and `DURABLE_AGENT` maps to `WORKFLOW_BAL` against the project root — so this lands on the module the declaration is actually written into.

Covered by `durable_agent_creation_submodule_to_source.json` against a new multi-module fixture whose only provider sits in `modules/ops`: the generated declaration now declares its own `wso2ModelProvider` instead of emitting `model: opsModel`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
rdulmina: Right, and this survived my earlier fix — worth spelling out since the thread shows as outdated.

5b1d392aab had narrowed the scan from the default module to the module owning `filePath`, which fixed the same-module case but not the one you describe: `durableAgentOptions` iterates `currentPackage.modules()`, so the Agent dropdown could still offer an agent whose Data Event dropdown came back empty.

e7eea198ee takes your suggestion and scans the same module set:

```java
// Every module, matching the set durableAgentOptions offers agents from: scoping this
// narrower than the Agent dropdown means picking an agent the dropdown offered can leave the
// Data Event dropdown empty, and eventName then fails requireValue on save. Only channel
// metadata is read here, never a bare name that would have to resolve from somewhere.
for (Module module : project.currentPackage().modules()) {
```

The last sentence is the distinction from the model-provider lookup, which has to stay module-scoped precisely because it answers with a bare name.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentUpdateBuilder.java
rdulmina: Agreed on the diagnosis — the root cause is the module-scope gap, fixed in e7eea198ee by scanning the same module set as `durableAgentOptions`, so the list is no longer empty for a submodule agent.

I have left the `""` default and `requireValue` as they are: with the scope fixed I could not construct a case where the list is empty but the agent is real, and inventing a fallback value for a channel that genuinely does not exist would move a loud failure to a quiet one. If you would still like the field to stay usable with a surfaced reason when nothing resolves, I would rather do that as its own change than fold it in here.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentUpdateBuilder.java
rdulmina: Good catch — fixed in e7eea198ee. The three situations really were collapsed into one empty Optional, and only one of them means one-way.

```java
if (!events.containsKey(eventName)) {
    // Unknown channel: the declaration could not be read, the agent did not match, or the
    // name was hand-typed. Answering "one-way" here would drop a declared response and emit
    // a call that does not compile, so fall back to the turn token the send always returns.
    return Optional.of(UNRESOLVED_RESPONSE_TYPE);
}
```

`UNRESOLVED_RESPONSE_TYPE` is `string`, i.e. exactly the old shape, so an unresolvable channel errs the way you describe as harmless — naming a value that does arrive — while a channel actually read as declaring no `response` still generates the one-way form. The non-default-module trigger is separately gone with the scope fix.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
rdulmina: Both points confirmed and fixed in e7eea198ee.

`financeRoles` contains none of `[(){}:.`, so it fell through to `quoteIfPlain` and became a role literal of the same spelling. The mode is now what decides:

```java
// In expression mode the value IS the expression: a bare `financeRoles` names a module-level
// variable, and quoting it would rewrite that reference into a role literal of the same
// spelling. Only a value that arrived without an expression mode selected can be a bare role
// name needing quotes.
if (isExpressionModeSelected(property)) {
    return source;
}
return quoteIfBareRole(source);
```

You are right about the javadoc too — `quoteIfBareRole` does not leave a bare identifier alone. It now says a *qualified or called* reference, and notes that is why the mode is checked first.

packages/ballerina-side-panel/src/components/editors/FieldFactory.tsx
rdulmina: Confirmed and fixed in e7eea198ee — this was a real defect in the text-mode change, thanks for catching it.

The check was leading-quote plus trailing-quote, which `"Hello " + name + "!"` satisfies. Tightened to a single complete literal:

```ts
function isSingleStringLiteral(value: string): boolean {
    if (value.length < 2 || !value.startsWith(String.fromCharCode(34)) || !value.endsWith(String.fromCharCode(34))) {
        return false;
    }
    for (let i = 1; i < value.length - 1; i++) {
        if (value[i] === "\\") { i++; continue; }
        if (value[i] === String.fromCharCode(34)) { return false; }
    }
    return true;
}
```

An unescaped quote before the last character means the literal ends early and the rest is expression syntax, so the field opens in expression mode instead. Escaped interior quotes still qualify. `ballerina-side-panel` suites pass (167 tests).

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
rdulmina: Confirmed and fixed in e7eea198ee — and thanks for checking the coordinates.

The badge is keyed on the declared value rather than the capability kind now, so a gated tool renders like a gated activity:

```jsx
{isApprovalGated(item) && (
    <ApprovalBadge
        background={NODE_BG_COLOR}
        x="88.5"
        y="31.5"
        onClick={item.kind === "tool" ? undefined : () => onCapabilityClick(item)}
    />
)}
```

One deliberate difference from a plain kind check: the circle beneath a tool is intentionally not clickable (`item.kind === "tool" ? undefined : ...`), so the badge mirrors it rather than taking the handler unconditionally — otherwise a tool badge would be clickable when its circle is not, which is the inconsistency the other thread was about.

On human tasks: `CodeAnalyzer.collectDeclaredCapabilities` maps `name/roles/title/description/resultType/timeout` for `humanTasks` and no `requiresApproval`, so there is nothing to badge there today. `isApprovalGated` is kind-agnostic now, so it will light up if that field is ever added.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/diagnostics/DiagnosticRequest.java
rdulmina: Confirmed, and taking the TODO you offered rather than the rework — flagging that plainly so it is not mistaken for a fix.

You are right that the early return drops everything, not just the bogus parse error. Reading the edit back in its actual syntactic position means teaching the diagnostics path how each of the six kinds is embedded (list entry vs record field) and building the enclosing context for each, which is a larger change than the rest of this PR and easy to get subtly wrong. e7eea198ee records it at the suppression site:

```java
// TODO: this suppresses EVERY diagnostic for these kinds, not just the bogus parse error, so
// real problems reach source unreported — a Reviewer Roles value naming an undefined
// variable, a requiresApproval predicate pointing at a function that does not exist, an
// undefined type in Register Data Event responseType. The narrow fix is to keep the
// analysis and read the edit back in its actual syntactic position (list entry / record
// field) instead of as a statement.
```

Worth noting the interaction you spotted with the roles thread: that one is now fixed at the source (expression-mode values are no longer quoted), so it no longer depends on these diagnostics being visible. Happy to raise this as its own issue.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
rdulmina: Confirmed — and I have deliberately not fixed it here, so flagging rather than quietly leaving it.

`FieldFactory` does trim to first and last:

```ts
const newRenderingTypes = props.field.types.length === 1
    ? [props.field.types[0]]
    : [props.field.types[0], props.field.types[props.field.types.length - 1]];
```

so with `[TEXT, REPEATABLE_LIST, EXPRESSION]` the list mode never renders.

Two reasons for leaving it: that trim governs every multi-mode field in the product, and changing it is far wider than this PR; and it is not something this PR introduced — Await Human Task `userRoles` has the same three-mode shape from union expansion, and its list mode is equally unreachable. So the parity these fields were given in #2232 does hold; what is wrong is that the list mode works nowhere.

That leaves an honest gap: the fields advertise a mode users cannot reach. The options are to make `FieldFactory` render all declared modes, or to stop declaring `REPEATABLE_LIST` until it can. I would rather do either as its own change with its own testing than fold it in here — tell me which you prefer and I will raise it.

rdulmina: Taking your suggestion — `REPEATABLE_LIST` is gone in e58bc29357.

Removed in the shared `addRoleFieldTypes`, so it drops from all four role fields at once (the two agent builders, `DurableAgentRegisterToolBuilder` and `DurableAgentAddActivityBuilder`, plus the retry sub-property in `ActivityCallBuilder`):

```java
public static <T> Property.Builder<T> addRoleFieldTypes(Property.Builder<T> builder) {
    return builder
            .type().fieldType(Property.ValueType.TEXT).ballerinaType(ROLE_TYPE).stepOut()
            .type().fieldType(Property.ValueType.EXPRESSION).ballerinaType(ROLE_UNION_TYPE).stepOut();
}
```

The fields now declare exactly the two modes `FieldFactory` renders, so nothing advertises an editor that cannot be opened. `ROLE_LIST_TYPE` and the list-item template went with it, and the javadoc records the reason so the mode does not get added back:

> No `REPEATABLE_LIST` mode: `FieldFactory` renders only the first and last declared mode (`[types[0], types[types.length - 1]]`), so a list mode declared between them never reaches the user.

Worth being clear this narrows what the form *offers*, not what it accepts — the expression entry still carries the full `string|string[]`, which is where a list was being written anyway, so no existing value changes shape. Await Human Task derives the same two rendered modes from its parameter, so the parity these fields were given in #2232 still holds.

Golden churn is 13 files, pure deletion of the `REPEATABLE_LIST` entries and their templates. LS suites and the full `bi-diagram` suite pass.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
```diff
@@ -1190,6 +1203,21 @@ export function DurableAgentRunNodeWidget(props: DurableAgentRunNodeWidgetProps)
...
+                                button already owns that corner here. Keyed on the declared
+                                `requiresApproval` rather than the capability kind, because a
+                                registered tool carries it too and used to render as ungated. The
+                                click mirrors the circle underneath, which a tool does not have. */}
+                            {isApprovalGated(item) && (
+                                <ApprovalBadge
+                                    background={NODE_BG_COLOR}
+                                    x="88.5"
+                                    y="31.5"
+                                    onClick={item.kind === "tool" ? undefined : () => onCapabilityClick(item)}
```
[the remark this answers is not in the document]
rdulmina: Fair ask — the path instruction does apply and there is an established L2 pattern (`DiagnosticsPopUp.test.tsx`). Added `packages/bi-diagram/src/test/ApprovalBadge.test.tsx` in e58bc29357, covering the contract this PR actually changed:

- renders the shield with its "Requires Approval" tooltip
- forwards a click to the handler the host supplies
- stays inert when the host supplies none — the tool case, where the circle underneath is not clickable either — and does not throw
- cursor is `pointer` only when clickable

4 tests, passing; the full `bi-diagram` suite is green (11 suites, 96 tests, 14 snapshots).

Two parts of the suggestion I did **not** implement, with reasons rather than silence:

**Gated/ungated capability and absent/`false` `requiresApproval` at L2.** That decision lives in `isApprovalGated`, local to `DurableAgentRunNodeWidget`, and rendering that widget is not a cheap component render — it needs a real `DiagramEngine` (`engine.getModel().registerListener`) and a model with wired `in`/`out` ports for four `PortWidget`s. I looked at extracting the predicate to make it unit-testable and decided against it: the chat agent widgets take a real boolean `tool.requiresApproval`, while a durable capability carries `requiresApproval` as declared *source text* (so a predicate function name gates it too). Those are different rules, not duplication, and extracting a shared helper would misrepresent that. The source-text semantics are covered LS-side by `durable_agent_register_tool_gated_to_source.json` and friends.

**Icon changes.** Those route glyph and colour through `NodeIcon` rather than adding behaviour, and the whole-diagram render snapshots in `Diagram.flowModel.render.test.tsx` already cover them — they passed unchanged.

Happy to add the widget-level render test if you think the engine setup is worth it, but it looked like more mock than assertion.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -760,6 +831,100 @@ public static String quoteIfBareRole(String value) {
...
+     * a list is entered in the expression mode, whose type is the full union.
+     *
+     * @param builder the property builder to add the role input modes to
+     * @param <T>     the builder's step-out target
+     * @return the same builder, for fluent chaining
+     */
+    public static <T> Property.Builder<T> addRoleFieldTypes(Property.Builder<T> builder) {
+        return builder
+                .type().fieldType(Property.ValueType.TEXT).ballerinaType(ROLE_TYPE).stepOut()
+                .type().fieldType(Property.ValueType.EXPRESSION).ballerinaType(ROLE_UNION_TYPE).stepOut();
```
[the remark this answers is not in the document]
rdulmina: REPEATABLE_LIST is filtered from the frontend; hence, having Text and Expression is sufficient for now

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentUpdateBuilder.java
```diff
@@ -178,20 +175,45 @@ public Map<Path, List<TextEdit>> toSource(SourceBuilder sourceBuilder) {
...
+                        .name(ActivityCallBuilder.NIL_UNCHECKED_RESULT_TYPE)
+                        .whiteSpace()
+                        .name(variableName)
+                        .whiteSpace()
+                        .keyword(SyntaxKind.EQUAL_TOKEN);
+            }
+        } else {
+            String resultType = responseType.get();
+            sourceBuilder.token()
+                    .name(checkError ? resultType : resultType + "|error")
```
NipunaRanasinghe: Blocker:

`sendData` does not return the channel's response. From `module-ballerina-workflow`, `ballerina/durable_agent.bal`:

```ballerina
public isolated function sendData(string instanceId, string eventName, anydata data)
        returns string|error
```

with the doc comment "A correlation token for `getDataResult`/`waitForDataResult`, or an error". The return is always `string|error` whatever the channel declares — the `response` type is what you read back later via `getDataResult`, not what the send hands you.

So both branches here are wrong unless `response` happens to be `string`:

- `events: [{name: "claimFiled", request: json, response: ClaimResult}]` generates `ClaimResult eventToken = check claimAgent.sendData(...)` — type mismatch, does not compile.
- A one-way channel takes the empty-Optional path and generates `error? eventToken = agent.sendData(...)` (line 200), which does not compile either, since the call still yields `string|error`.

This is a regression: `staging/5.14.0` had `String resultType = STRING_TYPE;` with the comment "sendData always returns the turn's correlation token; the answer is read via Get Agent Data Result". The class javadoc at line 46 still says "Always generates `string <token> = ...` — the correlation token", so the file now documents the old behaviour while the code does something else. `DurableAgentDataResultBuilder.TOKEN_DOC` and `CodeAnalyzer.findSendDataEventForToken` also still treat that binding as the token.

Tests don't catch it because every new golden uses `response: string`, which coincides with the token type, and `to_source` goldens are text comparisons that never compile the output. A golden with a non-string `response` would fail today.

Suggest restoring `string` as the binding type. If the response type should be surfaced anywhere, it belongs on the Get Agent Data Result node, where the read actually happens.

**Correction to my earlier review:** on the thread you fixed in e7eea198ee I wrote that the old unconditional `string` "was wrong in the opposite direction". That was wrong — the unconditional `string` was correct. The only real problem was the empty Optional collapsing three different situations, which mattered for *diagnostics*, not for the binding type.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -586,46 +589,104 @@ public static List<Option> durableAgentOptions(
...
+            // Unknown channel: the declaration could not be read, the agent did not match, or the
+            // name was hand-typed. Answering "one-way" here would drop a declared response and emit
+            // a call that does not compile, so fall back to the turn token the send always returns.
+            return Optional.of(UNRESOLVED_RESPONSE_TYPE);
+        }
+        String response = events.get(eventName);
+        if (response == null || response.isBlank() || NIL_RESPONSE_TYPE.equals(response)) {
+            return Optional.empty();
+        }
+        return Optional.of(response);
```
NipunaRanasinghe: Related to the blocker on `DurableAgentUpdateBuilder`, and it goes away if that one is fixed by dropping the response type from the binding.

This hands back the `response` type as raw source text read from whichever file declares the agent — and since `declaredAgentEvents` now scans every module, that file can be any of them. `DurableAgentUpdateBuilder` emits it verbatim. Unlike a property-sourced value it carries no `imports`, so `SourceBuilder.getProperty`'s import collection never sees it.

Ballerina imports are per-file, so an agent in `workflow.bal` declaring `response: mypkg:Reply` (prefix imported there) generates `mypkg:Reply eventToken = ...` into `main.bal`, where the prefix is undefined and no import edit is added.

Worth noting the commit message's argument — "Only channel metadata is read here, never a bare name that would have to resolve from somewhere" — holds for the channel name but not for the `response` type, which is exactly a name that has to resolve.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentBuilder.java
```diff
@@ -150,25 +236,77 @@ private static String defaultModelProviderDeclaration() {
...
+            Module module = declaringModule(sourceBuilder);
+            if (module == null) {
+                return null;
+            }
+            for (Symbol symbol : module.getCompilation().getSemanticModel().moduleSymbols()) {
+                if (symbol.kind() != SymbolKind.VARIABLE) {
+                    continue;
+                }
+                VariableSymbol variable = (VariableSymbol) symbol;
+                if (isWso2ModelProviderType(variable.typeDescriptor()) && variable.getName().isPresent()) {
```
NipunaRanasinghe: This matches on the type alone, so it also accepts a provider the user built with an explicit endpoint. In a package declaring

```ballerina
final ai:Wso2ModelProvider claimModel = check new ("http://localhost:9099", "test-token");
```

picking "Default WSO2 Model Provider" binds the agent to that hand-configured endpoint rather than `ai:getDefaultModelProvider()` — which is the failure the method's own javadoc says it exists to prevent ("would silently run the agent on a model the user did not choose"). `durable_agent_creation_default_model_to_source.json` encodes this behaviour (`model: claimModel`), so the golden currently asserts the wrong thing.

Also, when several such variables exist the pick is arbitrary — `moduleSymbols()` order is unspecified.

Checking the initializer (a `getDefaultModelProvider()` call, not a `new` with arguments) would narrow it to what the label promises.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentBuilder.java
```diff
@@ -82,62 +96,134 @@ public void setConcreteConstData() {
...
+        }
+        String modelProviderDeclaration = "";
+        if (isDefaultModelProviderExpr(modelValue)) {
+            // Bind the WSO2 default to a named module-level variable instead of inlining
+            // `check ai:getDefaultModelProvider()` in the config: the agent box's model circle is a
+            // dropdown over module-level provider variables, so an inline expression would leave it
+            // with nothing to point at. Referencing a name without declaring it is what used to
+            // leave the generated package failing to compile on `wso2ModelProvider`.
+            String existingWso2Provider = resolveWso2ModelProvider(sourceBuilder);
+            if (existingWso2Provider != null) {
```
NipunaRanasinghe, line 196: The LS's decision to declare a provider and the webview's decision to configure one have drifted apart.

`FunctionForm.onSubmit` probes `projectHasModelProvider()` — any `MODEL_PROVIDER` node — *before* generation, and skips `configureWso2ModelProvider()` when one exists. The LS here declares a fresh `wso2ModelProvider` whenever no `ai:Wso2ModelProvider` is found in the declaring module.

Those two conditions no longer coincide. In a package whose only provider is, say, an `ai:OpenAiProvider`, the webview probe returns true, so no `Config.toml` entries are written, but the LS still emits `final ai:Wso2ModelProvider wso2ModelProvider = check ai:getDefaultModelProvider();`, which needs them. Before this PR the two agreed, because `resolveExistingModelProvider` accepted any provider in any module. The module scoping in 5b1d392aab widens the gap.

Either the webview needs the same "is there a WSO2 provider in this module" question, or the LS needs to report back whether it declared one.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/Configurations.tsx
```diff
@@ -61,6 +61,29 @@ const customSqlHighlightStyle = HighlightStyle.define([
...
+ * mode would show `Hello " + name + "!` as literal text, then re-quote it on save and silently
+ * rewrite the expression. An unescaped quote before the last character means the literal ends early
+ * and the rest is expression syntax.
+ */
+function isSingleStringLiteral(value: string): boolean {
+    if (value.length < 2 || !value.startsWith('"') || !value.endsWith('"')) {
+        return false;
+    }
+    for (let i = 1; i < value.length - 1; i++) {
+        if (value[i] === "\\") {
```
NipunaRanasinghe, line 76: Minor: the backslash skip can consume the closing quote, so an unterminated literal passes. For `"abc\"` (6 chars) the loop reaches the `\` at index 4, skips index 5 — the final `"` — then exits and returns `true`. Same for `"\"`. Text mode then opens it and re-quotes on save.

Low impact, since such a value can't come from valid source, but it defeats the guard at exactly its tail. Bounding the skip with `i + 1 < value.length - 1` closes it.

The rest of this function is a solid fix for the concatenation case.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentUpdateBuilder.java
rdulmina: Fixed in 9eeea2345b — you were right, and the regression was one step worse than described.

I checked the signature rather than the doc: `ballerina/workflow` **0.8.3 and 0.9.0** both declare

```ballerina
public isolated function sendData(string instanceId, string eventName, anydata data)
        returns string|error
```

so both branches were wrong, as you said. The extra part: `durable_agent_send_data_no_check_to_source.json` is not a new golden — it existed on `staging/5.14.0` asserting the correct `string|error eventToken = ...`, and this PR changed it to `error? eventToken = ...`. So the bug was written *into* a golden that had been right.

Restored the unconditional `string` binding, so `checkError` picks `string` vs `string|error` and nothing is read off the declaration. The javadoc at line 46 is accurate again, and `DurableAgentDataResultBuilder.TOKEN_DOC` / `CodeAnalyzer.findSendDataEventForToken` no longer disagree with the generator.

All five send-data goldens now bind the token:
- `..._to_source` → `string eventToken = check ...` (unchanged, it was right by coincidence)
- `..._no_check_to_source` → `string|error eventToken = ...` (back to staging's text)
- `..._one_way_to_source`, `..._submodule_one_way_to_source` → `string eventToken = check ...`

I kept the two one-way goldens rather than deleting them — a channel declaring no `response` is exactly the case that broke, so it is worth asserting it binds like any other. Also took your correction on the earlier thread; no diagnostics depended on the empty Optional, it only fed the binding type.

Verified: `flow-model-generator-ls-extension` 2053 tests, 0 failures.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
rdulmina: Fixed in 9eeea2345b — and you were right that it goes away with the blocker.

`declaredAgentEventResponseType` had exactly one caller (the send binding), so once that stopped reading the declaration the helper was dead. Removed it along with `UNRESOLVED_RESPONSE_TYPE` and `NIL_RESPONSE_TYPE`, and `declaredAgentEvents` is back to collecting names only (`LinkedHashSet<String>`), so the `response` value is never read out of the syntax tree at all. That closes the import problem at the source rather than trying to carry `imports` along with a raw type name.

**I deliberately did not revert the every-module scan.** Your note is what decided it — the "only channel metadata is read here" argument does hold for the channel name, and only fails for the response type. Since the response type is gone, the widening from 5b1d392aab stands on the reasoning it was accepted for: scoping the channel lookup narrower than the Agent dropdown lets a user pick an agent the dropdown offered and get an empty Data Event dropdown, which then fails `requireValue` on save. The comment now says "never a type that would have to resolve from somewhere" so it states what is actually true.

Net: `WorkflowUtil` loses ~60 lines.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentBuilder.java
rdulmina: Fixed in 9eeea2345b. Confirmed both halves before changing anything — the fixture really is a hand-configured provider:

```ballerina
// to_source/source/workflow_function/agent.bal:5
final ai:Wso2ModelProvider claimModel = check new ("http://localhost:9099", "test-token");
```

and `durable_agent_creation_default_model_to_source.json` really did assert `model: claimModel`. So picking "Default WSO2 Model Provider" bound the agent to that endpoint and token.

`resolveWso2ModelProvider` now walks the module's declarations and requires **both** the type and a `getDefaultModelProvider()` initializer:

```java
if (!isDefaultModelProviderInitializer(varDecl.initializer().get())
        || !(varDecl.typedBindingPattern().bindingPattern()
                instanceof CaptureBindingPatternNode capture)) {
    continue;
}
return capture.variableName().text();
```

`isDefaultModelProviderInitializer` strips a leading `check`/`checkpanic` and matches `getDefaultModelProvider()` with any module prefix, so an alias for `ballerina/ai` still works while `check new (...)` does not. Walking the syntax tree in source order also settles your second point — the pick is deterministic where `moduleSymbols()` ordering was unspecified.

On the golden: rather than just relaxing it, I added a real default provider to the fixture **after** `claimModel`, so the assertion now proves the narrowing instead of merely tolerating it — source order reaches the hand-configured one first, so the old type-only match would take `claimModel` and fail. Expectation is `model: defaultClaimModel`.

One wrinkle worth flagging: I first added that declaration mid-file and it broke four unrelated goldens (`durable_agent_add_activity`, `register_tool`, `register_toolkit`, `register_tool_gated`) because they carry absolute line numbers into `agent.bal`. Appended it at end of file instead, so no existing line moves.

Verified: 2053 tests, 0 failures.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentBuilder.java
rdulmina: Fixed in 9eeea2345b, taking your second option — but implemented one layer down from where you put it, so flagging the divergence.

Confirmed the drift exactly as you describe: `projectHasModelProvider()` asks `searchNodes({ filePath: projectPath, query: { kind: "MODEL_PROVIDER" } })` — any provider, project-wide — while the LS asks "any `ai:Wso2ModelProvider` in the declaring module". And narrowing that lookup for the thread above makes the gap *wider*, since the LS now declares a provider in strictly more cases.

I did not take the "webview asks the same question" option: re-deriving the LS's decision in TypeScript is what drifted in the first place, and a module-scoped re-derivation would drift again on the next change to either side. So the LS's answer travels instead:

- `UpdatedArtifactsResponse.declaredDefaultModelProvider?: boolean`
- computed in `bi-diagram/rpc-manager.ts` from the edits the LS actually returned, before they are applied
- `FunctionForm.onSubmit` acts on that and the pre-generation probe is gone

```ts
if (sourceCode.declaredDefaultModelProvider) {
    await configureWso2ModelProvider();
}
```

The divergence from "the LS needs to report back": the flag is derived in the rpc-manager from `model.textEdits` rather than added to the LS's response payload. Same guarantee — it reflects what was emitted, not a guess about what would be — without changing the source-generation protocol. The marker requires both `Wso2ModelProvider` and `getDefaultModelProvider()`, so a user-configured provider (no Config.toml needed) does not trigger it and the `ai` import prefix is not pinned. Six unit tests in `src/test-support/defaultModelProvider.test.ts` cover the reuse, hand-configured, aliased-prefix and empty-edit cases.

Happy to move it into the LS response proper if you would rather the flag be explicit there.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/Configurations.tsx
rdulmina: Fixed in 9eeea2345b. Traced both of your cases by hand and they behave exactly as you said — `"abc\"` reaches the `\` at index 4, skips index 5 (the closing quote), exits the loop and returns `true`; same for `"\"`.

Took your suggested bound, expressed as an explicit early return so the reason is on the line:

```ts
if (value[i] === "\\") {
    if (i + 1 >= value.length - 1) {
        // The escape consumes the closing quote, so the literal never terminates:
        // `"abc\"` would otherwise skip past its own last character and pass.
        return false;
    }
    i++;
    continue;
}
```

Also added `stringTemplateCompatibility.test.ts`, driving it through the public `getIsValueCompatible` seam — both of your inputs, plus the cases that must keep passing (`"say \"hi\""`, `"a\\b"`, `string \`hello\``) and the concatenation case. I checked the test actually earns its place: with the guard removed, "rejects a literal whose closing quote is escaped" fails and the other five still pass.

Agreed such a value cannot come from valid source, so this is a guard-hardening rather than a user-visible fix.

NipunaRanasinghe (approved):
LGTM

NipunaRanasinghe (commented):
@hasithaa appreciate if you could review and approve from the functionality POV

hasithaa (approved):

PR #971, opened 2026-08-31
Improve libraryGetTool with proper error handling and artifact managements

SasinduDilshara:
Fixes https://github.com/wso2/product-integrator/issues/2366

libraryGetTool reliability:
- Propagate the AI SDK abort signal through selectRequiredFunctions, getRequiredFunctions, getSuggestedFunctions and getRequiredTypesFromLibJson so a cancelled run actually cancels the underlying generateObject calls, which previously received a fresh AbortController().signal that never fired.
- Distinguish "no library matched" from "the fetch failed": LibraryGetTool rethrows instead of returning [], and execute() returns a distinct, actionable message for each so the agent never treats an internal failure as evidence to invent API.
- De-duplicate and trim requested library names before fetching.
- Emit the requested libraryNames on the tool_call event, and fix the [AI Request Complete] log to count the hallucination-filtered list.
- De-duplicate type definitions when merging healthcare type libraries.

Deprecated alternative listeners:
- A service type hosted by several listeners writes one into the `on new ...` clause and lists the rest as alternatives. A deprecated alternative (mcp:Listener, superseded by mcp:StreamableHttpListener) lost its deprecation reason because alternatives were carried as bare name strings.
- Carry an AlternativeListener { name, deprecated? } end to end (Java producer model, wire shape, and TS renderer) and render the reason in the catalog note.
- Update the renderer unit test and the LS-extension schema test.

- Improve `libraryGetTool` reliability with abort-signal propagation, input cleanup, clear failure handling, and accurate tool-call reporting.
- Distinguish unmatched libraries from lookup failures and prevent duplicate healthcare type definitions.
- Preserve alternative-listener deprecation metadata across Java models, wire schemas, and TypeScript catalog notes.
- Update renderer and language-server schema tests for structured alternative listeners and deprecation details.

### DIFF
--- a/packages/ballerina-extension/src/features/ai/agent/tools/library-get.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/tools/library-get.ts
@@ -69,18 +69,20 @@ export async function LibraryGetTool(
     generationType: GenerationType,
     eventHandler: CopilotEventHandler,
     toolModelUsage: ToolModelUsage,
-    toolCallId: string
+    toolCallId: string,
+    abortSignal?: AbortSignal
 ): Promise<Library[]> {
     try {
         // Emit tool_call event with ID from AI SDK
         eventHandler({
             type: "tool_call",
             toolName: LIBRARY_GET_TOOL,
+            toolInput: { libraryNames: params.libraryNames },
             toolCallId
         });
 
         const startTime = Date.now();
-        const { libraries, usage } = await selectRequiredFunctions(params.userPrompt, params.libraryNames, generationType);
+        const { libraries, usage } = await selectRequiredFunctions(params.userPrompt, params.libraryNames, generationType, abortSignal);
         console.log(
             `[LibraryGetTool] Fetched ${libraries.length} libraries: ${libraries
                 .map((lib) => lib.name)
@@ -96,7 +98,7 @@ export async function LibraryGetTool(
     } catch (error) {
         console.error(`[LibraryGetTool] Error fetching libraries: ${error}`);
 
-        // Emit error result with same ID
+        // Emit error result with same ID so the UI closes this tool call's spinner
         eventHandler({
             type: "tool_result",
             toolName: LIBRARY_GET_TOOL,
@@ -104,7 +106,9 @@ export async function LibraryGetTool(
             toolCallId
         });
 
-        return [];
+        // Rethrow rather than return []: an empty result is a legitimate outcome of selection, and the
+        // caller must be able to tell "nothing matched" from "the fetch failed" to report each honestly.
+        throw error;
     }
 }
 
@@ -136,19 +140,46 @@ name, description, type definitions (records, objects, enums, type aliases), cli
 
 `,
         inputSchema: LibraryGetToolSchema,
-        execute: async (input: { libraryNames: string[]; userPrompt: string }, context?: { toolCallId?: string }) => {
+        execute: async (
+            input: { libraryNames: string[]; userPrompt: string },
+            context?: { toolCallId?: string; abortSignal?: AbortSignal }
+        ) => {
             // Extract toolCallId from AI SDK context
             const toolCallId = context?.toolCallId || `fallback-${Date.now()}`;
 
+            // The model occasionally repeats a name; a duplicate would be fetched, selected over
+            // and rendered twice.
+            const libraryNames = [...new Set(input.libraryNames.map((name) => name.trim()).filter(Boolean))];
+
             console.log(
-                `[LibraryGetTool] Called with ${input.libraryNames.length} libraries: ${input.libraryNames.join(
+                `[LibraryGetTool] Called with ${libraryNames.length} libraries: ${libraryNames.join(
                     ", "
                 )} and prompt: ${input.userPrompt} [toolCallId: ${toolCallId}]`
             );
-            const rawLibraries: Library[] = await LibraryGetTool(input, generationType, eventHandler, toolModelUsage, toolCallId);
-            const afterContent = toSyntaxString(rawLibraries);
-
-            return afterContent;
+            try {
+                const rawLibraries: Library[] = await LibraryGetTool(
+                    { libraryNames, userPrompt: input.userPrompt },
+                    generationType,
+                    eventHandler,
+                    toolModelUsage,
+                    toolCallId,
+                    context?.abortSignal
+                );
+                if (rawLibraries.length === 0) {
+                    // An empty string here reads as a blank tool result the model can make nothing of.
+                    return `No relevant functions, clients, services, or types were found in the requested libraries (${libraryNames.join(", ")}) for this query. Verify each name is in 'organization/libraryName' format and appeared in the LibrarySearchTool results; if so, retry with a more specific userPrompt or select different libraries.`;
+                }
+                return toSyntaxString(rawLibraries);
+            } catch (error) {
+                // A cancelled run must keep unwinding so the SDK can wind the step down.
+                if (context?.abortSignal?.aborted) {
+                    throw error;
+                }
+                // Tell the model the fetch FAILED — returning nothing here is indistinguishable from
+                // "these libraries matched nothing", and the agent is forbidden to invent API in that gap.
+                const message = error instanceof Error ? error.message : String(error);
+                return `Fetching library documentation for (${libraryNames.join(", ")}) failed: ${message}. This was an internal failure, not evidence the libraries are empty or missing — retry this tool call, and if it keeps failing, say so instead of inventing API details.`;
+            }
         },
     });
 }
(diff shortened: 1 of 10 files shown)


packages/ballerina-extension/src/features/ai/agent/tools/library-get.ts
```diff
@@ -69,18 +69,20 @@ export async function LibraryGetTool(
...
         eventHandler({
             type: "tool_call",
             toolName: LIBRARY_GET_TOOL,
+            toolInput: { libraryNames: params.libraryNames },
             toolCallId
         });
 
         const startTime = Date.now();
-        const { libraries, usage } = await selectRequiredFunctions(params.userPrompt, params.libraryNames, generationType);
+        const { libraries, usage } = await selectRequiredFunctions(params.userPrompt, params.libraryNames, generationType, abortSignal);
```
RNViththagan, line 85: The same dead signal is still on the healthcare path: `healthcare-library.ts:196` passes `new AbortController().signal` to `generateObject`, `getRelevantLibrariesAndFunctions` (line 166) calls `selectRequiredFunctions` without a signal, and that tool's `execute` doesn't take `context.abortSignal` at all. So cancelling a FHIR generation still lets the Sonnet selection call and the Haiku calls run to completion.

Shall we thread it through there too, same as here? Happy for it to be a follow-up if you'd rather keep this PR tight.

packages/ballerina-extension/src/features/ai/agent/tools/library-get.ts
```diff
@@ -96,15 +98,17 @@ export async function LibraryGetTool(
...
             type: "tool_result",
             toolName: LIBRARY_GET_TOOL,
             toolOutput: [],
             toolCallId
         });
 
-        return [];
+        // Rethrow rather than return []: an empty result is a legitimate outcome of selection, and the
+        // caller must be able to tell "nothing matched" from "the fetch failed" to report each honestly.
+        throw error;
```
RNViththagan, line 119: Now that this rethrows, the `tool_result` event just above still goes out with `toolOutput: []`, so the UI shows an empty result while the model gets the "fetch failed" message from `execute`. It also fires on a user abort, before the abort unwinds.

Shall we mark that event as failed (or skip it when `abortSignal.aborted`) so the UI and the model agree on what happened?

packages/ballerina-extension/src/features/ai/agent/tools/library-get.ts
```diff
@@ -136,19 +140,46 @@ name, description, type definitions (records, objects, enums, type aliases), cli
...
+        execute: async (
+            input: { libraryNames: string[]; userPrompt: string },
+            context?: { toolCallId?: string; abortSignal?: AbortSignal }
+        ) => {
             // Extract toolCallId from AI SDK context
             const toolCallId = context?.toolCallId || `fallback-${Date.now()}`;
 
+            // The model occasionally repeats a name; a duplicate would be fetched, selected over
+            // and rendered twice.
+            const libraryNames = [...new Set(input.libraryNames.map((name) => name.trim()).filter(Boolean))];
```
RNViththagan: Tiny one: this trims whitespace but not case, so `ballerina/http` and `Ballerina/http` both survive and the second one goes to the LS as an unresolvable name. Shall we normalise case here as well, or is a mixed-case name ever legit?

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
```diff
@@ -71,24 +71,25 @@ const TYPE_CONSTRUCTOR = 'Constructor';
...
  * name, its parameter names and its return type — so it scales with how much the query matches, not with
  * how large the library is. At the previous 8192 a broad query against a large connector ("wire up GitHub
- * issues, PRs, releases and webhooks") could exceed it, and the failure is the worst shape available: the
+ * issues, PRs, releases and webhooks") could exceed it, and the failure was the worst shape available: the
  * JSON is truncated mid-token, `generateObject` rejects it against the schema, the throw unwinds to
- * `LibraryGetTool`'s catch, and the agent is handed `[]` — indistinguishable from a library that matched
- * nothing. It then has no API documentation and a system prompt forbidding it to invent any.
+ * `LibraryGetTool`'s catch, and the agent was handed `[]` — indistinguishable from a library that matched
+ * nothing, with no API documentation and a system prompt forbidding it to invent any. The tool now reports
+ * a fetch failure explicitly, but the fetch still fails — the cap is what keeps it from failing at all.
```
RNViththagan: Nit: "The tool now reports a fetch failure explicitly" describes this PR's change rather than the code, so it'll read oddly in a few months. Maybe keep just the last clause ("the cap is what keeps it from failing at all")?

packages/ballerina-extension/src/features/ai/agent/tools/library-get.ts
SasinduDilshara: For now, let's do not update the healthcare flow. Lets do that later

packages/ballerina-extension/src/features/ai/agent/tools/library-get.ts
SasinduDilshara: Fixed in 5a6f5cad90. The catch now emits the `tool_result` with `failed: true`, so the UI shows the same thing the model is told. When the abort signal has already fired it skips the emit entirely, since AgentExecutor's `tool-error` handler emits the failed result for that `toolCallId`; emitting here as well gave the UI two results for one call.

packages/ballerina-extension/src/features/ai/agent/tools/library-get.ts
SasinduDilshara: Fixed in 5a6f5cad90. Names are now lower-cased as well as trimmed before the dedupe. Ballerina org and package names are lower-case, so a mixed-case name is only ever the same library mis-cased by the model.

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
SasinduDilshara: Fixed in 5a6f5cad90. Reworded so it describes the code rather than the change: the tool reports a fetch failure to the model, but a reported failure is still a failure, and the cap is what keeps it from failing at all.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/service/ServiceAspects.java
```diff
@@ -278,11 +287,16 @@ private static List<String> alternativeListenerNames(TriggerScope scope) {
...
             // canonical `Listener` for an unnamed one — and an alternative identical to the primary is not
             // an alternative at all.
-            if (!name.equals(primary) && !names.contains(name)) {
-                names.add(name);
+            if (name.equals(primary) || !seen.add(name)) {
+                continue;
+            }
+            AlternativeListener entry = new AlternativeListener(name);
+            if (alternative.deprecated() != null && !alternative.deprecated().isBlank()) {
+                entry.setDeprecationNote(alternative.deprecated());
```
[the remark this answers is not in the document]
SasinduDilshara: Fixed in 5a6f5cad90. Alternatives are now collected in a `LinkedHashMap` keyed by rendered name, and a later duplicate's `deprecated` note is merged onto the existing entry when that entry has none, so the note survives regardless of which document entry came first. Document order is preserved.

packages/ballerina-extension/src/features/ai/utils/libs/library-types.ts
```diff
@@ -421,7 +431,12 @@ export interface Service {
     // choice the reader may want to make differently. `ballerina/mcp` is the corpus case — all four of its
     // service types are listed under both `StreamableHttpListener` and `Listener` — and it is the only one,
     // so the field is absent for every other library.
-    alternativeListeners?: string[];
+    //
+    // Each entry is an `AlternativeListener`, not a bare name: a superseded listener is still offered here,
+    // and its `deprecated` reason travels with it. mcp's `Listener` is deprecated in favour of
+    // `StreamableHttpListener`, and it is exactly the transport that lands here rather than in the
+    // `on new …` clause — so the name alone would present a retired listener as an equal choice.
+    alternativeListeners?: AlternativeListener[];
```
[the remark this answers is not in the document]
SasinduDilshara: Not changing this. The flow-model-generator jar ships inside the extension and is added to the LS classpath via `LS_EXTENSIONS_PATH`, so the Java producer and the TypeScript renderer are always the same version. There is no deployment where an older extension reads a newer server or vice versa, so dual-shape handling here would be dead code.

RNViththagan (approved):
LGTM

IsuruMaduranga (approved):

NipunaRanasinghe (approved):

PR #973, opened 2026-09-01
Sanitize malformed tool-call inputs before sending agent history to Anthropic

RNViththagan:
Related to https://github.com/wso2/product-integrator/issues/2206

When a streamed tool call's JSON is invalid — most often a large `file_batch_edit` write truncated at the output-token cap — the AI SDK cannot parse it and keeps the raw text as a string on the `tool-call` part; when it parses but fails the tool's schema, the SDK keeps the parsed array/null/number instead. Anthropic requires `tool_use.input` to be an object and rejects every later request in the thread with `tool_use.input: Input should be an object`, permanently bricking the conversation (and the raw provider text is what the user sees in chat).

- Add a `resilience/messageSanitization` module: `repairToolCallInputs` replaces any non-object tool-call input with an object (parses a string if that yields a plain object, otherwise `{}`), behind a `sanitizeMessages` umbrella so later repair passes can compose without touching call sites
- Repair in place: the malformed input is corrupt stored state (the parts are the chat store's own objects, held by reference), so fixing them heals every alias at once and the next thread save persists the clean value, rather than re-coercing the same corruption on every send
- Return the repair count so callers can surface or count dropped calls (telemetry is a follow-up)
- Call it in `AgentExecutor`'s `prepareStep`, before `addCacheControlToMessages`, and at the end of `populateHistoryForAgent`
- Add a jest suite (`src/test-support/messageSanitization.test.ts`) covering every input shape, in-place repair across aliases, the count, and the no-op case

The server-side `normalizeToolUseInputs` fix in the LLM proxy only covers proxied requests, so this still occurs for users on their own Anthropic key — the case reported in #2206. Ported and extended from the approved-but-unmerged wso2/vscode-extensions#2378.

### DIFF
--- a/packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
@@ -1,0 +1,75 @@
+/**
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com) All Rights Reserved.
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Repair passes that keep a model-message history provider-valid before it is sent to Anthropic.
+ * `sanitizeMessages` is the single entry point; individual repairs compose under it, so new failure
+ * modes can be handled by adding a pass rather than touching call sites.
+ */
+
+function isMalformedToolCallPart(part: any): boolean {
+    return part?.type === 'tool-call' && typeof part.input === 'string';
+}
+
+function coerceInput(raw: string): Record<string, unknown> {
+    try {
+        const parsed = JSON.parse(raw);
+        if (parsed && typeof parsed === 'object' && !Array.isArray(parsed)) {
+            return parsed;
+        }
+    } catch {
+        // unparseable
+    }
+    return {};
+}
+
+/**
+ * Coerce every malformed (string) tool-call input to an object, in place. Returns the number
+ * repaired; clean histories are a no-op.
+ *
+ * When a streamed tool call's JSON is invalid — most often truncated at the output-token cap on a
+ * large file write — the AI SDK cannot parse it and keeps the raw text as a string on the
+ * `tool-call` part. Anthropic requires `tool_use.input` to be an object and rejects replay with
+ * `tool_use.input: Input should be an object`, which bricks the thread on every later request.
+ */
+export function repairToolCallInputs(messages: any[]): number {
+    let repaired = 0;
+    for (const message of messages ?? []) {
+        if (!Array.isArray(message?.content)) {
+            continue;
+        }
+        for (const part of message.content) {
+            if (isMalformedToolCallPart(part)) {
+                part.input = coerceInput(part.input);
+                repaired++;
+            }
+        }
+    }
+    if (repaired > 0) {
+        console.warn(`[messageSanitization] Coerced ${repaired} malformed tool-call input(s) to objects.`);
+    }
+    return repaired;
+}
+
+/**
+ * Run every repair pass over a message history, in place, so it stays provider-valid. Call before
+ * sending history to the provider (prepareStep, history load). Add new passes here as needed.
+ */
+export function sanitizeMessages(messages: any[]): void {
+    repairToolCallInputs(messages);
+}
--- a/packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
@@ -23,18 +23,13 @@ import type { AssistantModelMessage, ModelMessage } from 'ai';
  * `sanitizeMessages` is the single entry point; individual repairs compose under it, so new failure
  * modes can be handled by adding a pass rather than touching call sites.
  *
- * Every pass is copy-on-write: only the messages it has to change are copied, and the input array
- * and its objects are never mutated. Replayed history is handed out by reference from the chat
- * store, so writing into it here would silently edit persisted state.
+ * Passes edit the history in place. The malformed input is corrupt stored state — the parts here
+ * are the chat store's own objects, held by reference on both the replay and the in-turn path — so
+ * repairing them heals the thread at the source: every alias is fixed at once, and the next thread
+ * save persists the clean value, rather than re-coercing the same corruption on every send.
  */
 
 type AssistantPart = Exclude<AssistantModelMessage['content'], string>[number];
-
-export interface SanitizedHistory {
-    messages: ModelMessage[];
-    /** Tool calls whose arguments had to be replaced with `{}` — each one a call that never ran. */
-    repaired: number;
-}
 
 function isPlainObject(value: unknown): value is Record<string, unknown> {
     return typeof value === 'object' && value !== null && !Array.isArray(value);
@@ -46,9 +41,10 @@ function isPlainObject(value: unknown): value is Record<string, unknown> {
  * parses but fails the tool's schema, the SDK keeps the parsed value instead, so an array, `null`
  * or a number can arrive here too. Anthropic requires `tool_use.input` to be an object and rejects
  * replay with `tool_use.input: Input should be an object`, which bricks the thread on every later
- * request. Either way the SDK ran nothing for the call.
+ * request. The paired tool-result already carries the parse error, so the model still learns the
+ * call failed; coercing the input just makes the history sendable again.
  */
-function coerceInput(raw: unknown): Record<string, unknown> {
+function coerceToObject(raw: unknown): Record<string, unknown> {
     if (typeof raw === 'string') {
         try {
             const parsed: unknown = JSON.parse(raw);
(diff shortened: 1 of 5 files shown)


packages/ballerina-extension/src/test-support/fixtures/message-sanitization/issue-2206.json
```diff
@@ -0,0 +1,64 @@
+{
+    "description": "product-integrator#2206: a streamed file_batch_edit call whose JSON is invalid (a '\">' typo, or truncation at the output-token cap) is kept as a string on the tool-call part; Anthropic then rejects every later request with tool_use.input: Input should be an object.",
```
SasinduDilshara: Change the filename

packages/ballerina-extension/src/test-support/messageSanitization.test.ts
```diff
@@ -0,0 +1,186 @@
...
+
+describe("repairToolCallInputs", () => {
+    it("coerces an unparseable string input to an empty object", () => {
+        const messages = [assistantWithMalformedCall()];
+        const repaired = repairToolCallInputs(messages);
+        const call = (messages[0].content as any[])[1];
+        expect(repaired).toBe(1);
+        expect(typeof call.input).toBe("object");
+        expect(call.input).toEqual({});
+    });
```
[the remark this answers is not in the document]
RNViththagan: Decided to keep the plain unit suite instead of a fixture corpus here — it already pins the sanitizer's full contract (parseable/unparseable/truncated/empty/array inputs, repair counts, non-tool-call parts untouched), so a fixture added no unique coverage for this fix.

SasinduDilshara (dismissed):

packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+/**
+ * Repair passes that keep a model-message history provider-valid before it is sent to Anthropic.
+ * `sanitizeMessages` is the single entry point; individual repairs compose under it, so new failure
+ * modes can be handled by adding a pass rather than touching call sites.
+ */
+
+function isMalformedToolCallPart(part: any): boolean {
+    return part?.type === 'tool-call' && typeof part.input === 'string';
```
samithkavishke: **Non-blocking.** A string input isn't the only bad shape the SDK produces. In `parseToolCall`'s catch, `input = parsedInput.success ? parsedInput.value : toolCall.input` — so when the JSON *parses* but fails schema validation, `input` is the parsed value. A model emitting `"[]"`, `"[1,2]"`, `"null"` or `"42"` yields an array/null/number input, this predicate skips it, and Anthropic still rejects the thread with `Input should be an object`.

Suggest inverting the check: repair anything that is not a plain non-array object. (The suite covers the array-as-*string* case, which hides this gap.)

packages/ballerina-extension/src/features/ai/utils/ai-utils.ts
```diff
@@ -82,6 +83,8 @@ export function populateHistoryForAgent(chatHistory: any[]): ModelMessage[] {
             });
         }
     }
+    // Keep replayed history provider-valid (e.g. coerce malformed tool-call inputs).
+    sanitizeMessages(messages);
```
samithkavishke, line 87: **Non-blocking.** `populateHistoryForAgent` copies `entry.content` by reference, and `getChatHistoryForLLM` returns references into `generation.modelMessages`. So `sanitizeMessages` mutates objects owned by the chat store — discarding the raw truncated text from in-memory state without going through a store mutator or marking it dirty, so in-memory and on-disk diverge until the next unrelated save. It self-heals (the read path re-sanitizes on reload), but a sanitizer being an implicit writer to persisted state is worth either a copy or an explicit store-side repair.

packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
```diff
@@ -0,0 +1,75 @@
...
+function coerceInput(raw: string): Record<string, unknown> {
+    try {
+        const parsed = JSON.parse(raw);
+        if (parsed && typeof parsed === 'object' && !Array.isArray(parsed)) {
+            return parsed;
+        }
+    } catch {
+        // unparseable
+    }
+    return {};
```
samithkavishke: **Nit.** Unparseable inputs collapse to `{}`, so the replayed call looks like it was made with no arguments. The SDK pairs these with a `tool-error` part, so the model isn't badly misled — but preserving a truncated raw excerpt under one reserved key would make the "my big write got cut off" case legible.

packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
```diff
@@ -0,0 +1,75 @@
...
+        }
+        for (const part of message.content) {
+            if (isMalformedToolCallPart(part)) {
+                part.input = coerceInput(part.input);
+                repaired++;
+            }
+        }
+    }
+    if (repaired > 0) {
+        console.warn(`[messageSanitization] Coerced ${repaired} malformed tool-call input(s) to objects.`);
```
samithkavishke, line 76: **Nit.** The repair is observable only via `console.warn`. Since this fires on the token-cap truncation path that previously bricked threads, a telemetry counter would tell you how often it actually happens in the field.

packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Coerce every malformed (string) tool-call input to an object, in place. Returns the number
+ * repaired; clean histories are a no-op.
+ *
+ * When a streamed tool call's JSON is invalid — most often truncated at the output-token cap on a
+ * large file write — the AI SDK cannot parse it and keeps the raw text as a string on the
+ * `tool-call` part. Anthropic requires `tool_use.input` to be an object and rejects replay with
+ * `tool_use.input: Input should be an object`, which bricks the thread on every later request.
+ */
+export function repairToolCallInputs(messages: any[] | null | undefined): number {
```
samithkavishke: **Nit.** `any[]` throughout (also L25, L73). Defensible for a defensive sanitizer, but typing against `ModelMessage[]` with local narrowing would catch a future SDK rename of `input`/`tool-call` at compile time instead of silently no-opping.

packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
madushajg, line 58: Falling back to `{}` leaves the history claiming a call the model never made (`file_batch_edit({})`) with nothing telling it the arguments were dropped. We already have the idiom for this: `getChatHistoryForLLM` injects `<system-reminder>` blocks for interrupted generations (`chatStateStorage.ts:903-908, 913-918`) saying the described actions were NOT performed. Worth doing the same for a coerced tool call so the model knows to redo the edit instead of inferring it from the tool result.

packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+    return repaired;
+}
+
+/**
+ * Run every repair pass over a message history, in place, so it stays provider-valid. Call before
+ * sending history to the provider (prepareStep, history load). Add new passes here as needed.
+ */
+export function sanitizeMessages(messages: any[] | null | undefined): void {
+    repairToolCallInputs(messages);
```
madushajg: `repairToolCallInputs` returns a count and the umbrella discards it, so neither call site can tell that data loss happened. Suggest return `repairToolCallInputs(messages);` and a number return type — one line, and it's the hook for a user-facing notice or a telemetry event (`sendTelemetryEvent` is already imported at `AgentExecutor.ts:56`, though the `TM_EVENT_*` constant lives in the submodule so the event itself could be a follow-up)

packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
RNViththagan: Fixed in 810f4048c4. Checked the SDK path you pointed at (`ai@6.0.103`: `parsedInput.success ? parsedInput.value : toolCall.input`) — the repair now covers anything that is not a plain object, and the suite has actual-array, `null`, number and `undefined` cases alongside the string ones.

packages/ballerina-extension/src/features/ai/utils/ai-utils.ts
RNViththagan: Fixed in 810f4048c4. The sanitizer is now copy-on-write: it returns a new history in which only the messages it had to change are copied, and never writes into the array or objects it was given. Both call sites send what comes back, and a test asserts the source stays untouched.

packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
RNViththagan: Agreed — tracking as a follow-up. `sanitizeMessages` now returns the repair count so there is something to send; the event constant lives outside this repo, so wiring the event lands separately.

packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
RNViththagan: Done in 810f4048c4. Typed against `ModelMessage` / `AssistantModelMessage` from `ai`, narrowing on `role` and `part.type`; no `any` left in the module or the test.

packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
RNViththagan: Added in 810f4048c4, following the `getChatHistoryForLLM` idiom: each assistant message with a coerced call is followed by a `<system-reminder>` naming the tool(s), saying the arguments could not be read and the call was NOT performed. It goes after the paired tool result so the tool_use/tool_result adjacency Anthropic requires is kept, and it is injected at send time only — never persisted.

packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
RNViththagan: Done — `sanitizeMessages` returns `{ messages, repaired }`. Callers use `messages`; `repaired` is the hook for the telemetry follow-up.

madushajg (dismissed):

packages/ballerina-extension/src/features/ai/utils/ai-utils.ts
RNViththagan: Reconsidered and went the other way — deliberate in-place mutation (dc169da8d1), not a copy. The malformed input is corrupt *stored* data: the parts here are the chat store's own objects by reference (getChatHistoryForLLM spreads generation.modelMessages; populateHistoryForAgent copies content by ref), so mutating repairs every alias at once and the next thread save persists the clean value — the thread is healed at the source instead of re-coerced on every send. The in-memory/on-disk divergence you flagged is real but self-heals: in-memory is correct immediately (that is what gets sent), the next saveThread persists it, and the read path re-sanitizes on reload regardless. Reopened for your take.

packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
RNViththagan: Added this, then removed it again (dc169da8d1). The SDK already pairs the malformed call with an error tool-result carrying the parse failure, so the model does learn the call failed — the reminder was redundant with that. It also conflicts with the in-place repair we settled on: inserting a history message cannot be done in place without splicing phantom messages into the stored modelMessages (they would persist and re-accumulate on every load). So we rely on the existing error tool-result. Reopened in case you would still prefer the explicit reminder — happy to add it back as a send-only (non-persisted) message if so.

packages/ballerina-extension/src/features/ai/agent/resilience/messageSanitization.ts
RNViththagan: Done, and simplified past the original suggestion: with in-place mutation there is no wrapper to return, so sanitizeMessages now returns the repair count directly as a number (dc169da8d1). That is the hook for the telemetry follow-up.

SasinduDilshara (approved):

madushajg (approved):

PR #975, opened 2026-09-01
Stop labelling defaultable record fields as optional

pasindufernando1:
## Purpose

Fixes wso2/product-integrator#2210

In the record configuration dialog, every field that was not required rendered as `(Optional)` — the check was `param.optional || param.defaultable`. A field with a default (`string host = "localhost"`) therefore read the same as a truly optional one (`string password?`), which is misleading: leaving a defaultable field out still applies a value.

## Approach

The language server already sends `optional` and `defaultable` as separate flags on the record config model (see the `find_matching_type` test fixtures), so no LS change was needed — only the renderers were conflating them.

- New `utils/optionality.ts` with `getOptionalityLabel()`, which labels only `optional` fields.
- All five type renderers (`CustomType`, `RecordType`, `InclusionType`, `ArrayType`, `UnionType`) now go through it. `UnionType`'s label element is also gated on `param.optional` alone so a defaultable union field no longer renders an empty span.

For the `postgresql:DatabaseConfiguration` from the issue:

| field | before | after |
| --- | --- | --- |
| `string host = "localhost"` | (Optional) | *no label* |
| `string username = "postgres"` | (Optional) | *no label* |
| `int port = 5432` | (Optional) | *no label* |
| `string password?` | (Optional) | (Optional) |
| `postgresql:Options options?` | (Optional) | (Optional) |
| `sql:ConnectionPool connectionPool?` | (Optional) | (Optional) |

`ParameterBranch` also grouped fields by `isRequiredParam`, so defaultable fields stayed hidden inside the collapsed **Optional fields** section even once they were unlabelled. It now groups by `isOptionalParam` (also in `utils/optionality.ts`): only `field?` goes under that section, and defaultable fields render in the main list — unchecked, since auto-selecting them would force `host: "localhost"` and friends into the generated record literal.

For the record in the issue, the tree now reads:

```
[x] DatabaseConfiguration
    [x] database  string
    [ ] host      string
    [ ] username  string
    [ ] port      int
    v Optional fields
        [ ] password        string (Optional)
        [ ] options         (Optional) Options
        [ ] connectionPool  (Optional) ConnectionPool
```

`isRequiredParam` itself is untouched — it still drives checkbox locking and `updateFieldsSelection`, where defaultable fields correctly count as not required.

The legacy copies of this component tree (`HelperPane/RecordConstructView`, `HelperView/ConfigurePanel`, `statement-editor/ParameterTree`) are unreachable — nothing sets `configurePanelData` and `HelperPane/ConfigureRecordPage` has no importers — so they were left alone.

## Testing

- Added `utils/optionality.test.ts` (8 cases, covering both helpers) — passing.
- `tsc --noEmit -p tsconfig.spec.json` clean for the changed files.

## Summary

- Fixed record field classification so only truly optional fields display `(Optional)`.
- Kept defaultable fields in the main parameter list while preserving required-field behavior.
- Added shared optionality helpers in `@wso2/ballerina-core`.
- Applied the helpers across integrator forms, connector parameter trees, and statement-editor parameter trees.
- Added unit tests for optionality labels and field grouping.
- Updated an end-to-end test wait time and test fixture declaration.


packages/ballerina-visualizer/src/views/BI/HelperPaneNew/Components/RecordConstructView/utils/optionality.ts
```diff
@@ -0,0 +1,31 @@
...
+
+// Only `field?` is optional. A defaultable field (`field = <default>`) reads as a normal field, so it
+// carries no label — calling it optional wrongly implies no value is applied when it is left out.
+export function getOptionalityLabel(param: FormField): string {
+    return param.optional ? " (Optional)" : "";
+}
+
+// Grouping mirrors the label: only `field?` belongs under the collapsible "Optional fields" section. A
+// defaultable field is an ordinary field with a fallback, so it stays in the main list.
+export function isOptionalParam(param: FormField): boolean {
```
dulajdilshan: [Medium] Update the active ParameterTree too — Connector/action parameter forms still render defaultable fields as `(Optional)` because their active renderers retain the old predicate. Apply the same optional-only label and grouping there.

pasindufernando1: Done in 0f47ccd. Moved `getOptionalityLabel`/`isOptionalParam` into `@wso2/ballerina-core` (structurally typed, so the `FormField` and `TypeField` trees share one definition) and applied them in `HelperView/ConfigurePanel` (connection wizards, FlowDiagram panel) and `statement-editor/Parameters/ParameterTree` (connector/action forms). Test moved to `ballerina-core/src/test`.

Left `views/BI/HelperPane/RecordConstructView` alone — its only entry, `HelperPane/ConfigureRecordPage`, has no importers. Happy to update or delete it if you'd rather.

Note #993 is separate, not a duplicate: it splits `advanced` from `optional` on `Property` for the side panel's Advanced section. `TypeField` has no `advanced`, so it doesn't reach these renderers.

madushajg (approved):

PR #976, opened 2026-09-01
Collapse readonly intersections in the LS; fix GraphQL delete crash

### LINKED ISSUE #1839: Intersection Types are not Showing Up Properly in Config Editor
### Description

Config Editor does not show the proper type of a field, if the field is an intersection type. It just uses `"intersection"` as the value.

Image

### Steps to Reproduce

- Create a GraphQL service
- Add a GraphQL Field
- Go to "Advanced Configurations"
- Click on "Field Configuration" box to open the Config Editor
- Check the type of the `cacheConfig`, which populates as `"intersection"`, but the field expects a record value

### Version

5.0.0

_No response_

dulajdilshan:
## Purpose
Fixes wso2/product-integrator#1839.

`readonly & T` fields — e.g. `graphql:GraphqlResourceConfig.cacheConfig`, which is
`readonly & ServerCacheConfig` — reach the record config editor as
`{typeName: "intersection", members: [{readonly}, {record}]}`. The editor dispatches on `typeName`
and has no `intersection` renderer, so the field renders as a non-expandable row labelled
`intersection` and generates `{ cacheConfig: "intersection" }` →
`BCE2066: incompatible types: expected 'ServerCacheConfig?', found 'string'`.

Also fixes an uncaught runtime error that takes down the webview when a GraphQL operation is deleted.

## Goals
- No record config response carries a collapsible `intersection`: `readonly & T` at any depth reaches
  the editor as `T` and round-trips to a valid record literal.
- Deleting a GraphQL query/mutation/subscription refreshes the panel instead of crashing.

## Screenshots

image

## Approach

**Intersections — collapsed in the language server, not the webview.**
- New `IntersectionNormalizer` replaces an intersection with its single non-`readonly` member (record,
  union, array, map, tuple), merging the wrapper's field identity (`name`, `typeInfo`, `optional`,
  `defaultable`, `selected`) onto it. It never mutates its input — `Type.fromSemanticSymbol` memoizes
  and hands out shared instances — and shares untouched subtrees.
- Applied in `serializeRecordConfig` (the choke point for `recordConfig`, `findMatchingType` and
  `updateRecordConfig`) and in `TypeSymbolAnalyzerFromTypeModel.analyze` *before* the expression walk.
  The second call site fixes a related bug: `updateTypeConfig` never descended into an
  intersection-typed field, so re-opening saved source silently dropped its values.
- `RecordValueGenerator` keeps an `intersection` branch as a net for payloads posted back by an older
  webview; the webview's own `normalizeIntersections` stays for the same reason
  (`useDistributionLanguageServer` / `langServerPath` can serve these endpoints from an older LS).
- Scoped on purpose: `misc/diagram-util` is untouched, so `visibleVariableTypes`, `ballerinaSymbol/*`
  and `ballerinaConnector/*` keep rendering an intersection as `A & B`, which
  `HelperView/VariablePanel/utils.tsx` relies on. The `visible_variable_types` fixtures are unchanged,
  which is the proof the change is correctly scoped.

**GraphQL operation delete.** Deleting a field shortens the service, so the line range the panel holds
ends past where the service now does; `findNonTerminalNode` threw `IndexOutOfBoundsException` and
failed the request. Nothing awaits the delete handler, so it escaped as an unhandled rejection —
`[object Object]` in the dev overlay.
- `getServiceFromSource` falls back to the service *containing* the requested start line: an edit
  inside a service moves its end but never its start. `findNonTerminalNode` returns `null` rather than
  throwing, with guards added where callers did not already gate on the node kind (without one,
  `getListenerFromCode` would have returned a blank listener as a success, and saving that form would
  overwrite the real listener config). Writes are not retargeted — only the read path recovers.
- Webview: the delete handler and both fetches handle errors, the last known-good model survives a
  failed refresh, the listener lookup no longer throws for a project with no listeners, and the
  refresh effect depends on the range's values instead of an object literal rebuilt on every render
  (which is why the error appeared three times).

Known limitations, unchanged by this PR: `interceptors`
(`readonly & (Interceptor & readonly | (Interceptor & readonly)[])`) now collapses to a union of
`object` leaves — it renders honestly, but `object` has no record-literal form, so selecting it still
generates `object {}`; that needs an expression/variable picker. A genuine `Foo & Bar` has no single
member to stand in for it and is deliberately left as a leaf.

## UI Component Development
No new UI components; the change is in the data the existing record editor renders.
- [x] Matches with the native VSCode look and feel.

## User stories
As a developer configuring a GraphQL field, I can expand `cacheConfig`, set `enabled`/`maxAge`/`maxSize`,
save, and re-open the value with my entries intact. Deleting a GraphQL operation no longer breaks the view.

## Release note
Fixed intersection-typed fields (`readonly & T`) rendering as an unusable `intersection` row in the
config editor and generating invalid source. Fixed an error when deleting a GraphQL operation.

## Documentation
N/A — bug fix, no behaviour worth documenting beyond the release note.

## Training
N/A

## Certification
N/A — no user-facing concept changes.

## Marketing
N/A

## Automation tests
- Unit tests
  > `IntersectionNormalizerTest` (11 new, `flow-model-generator-core`): merge precedence, fixpoint,
  > ambiguous `Foo & Bar` left alone, identity/idempotence, no mutation of the input, structure
  > sharing, array/map/union/inclusion slots. Visualizer: 472/472 across 32 suites, including a new
  > `GraphqlServiceEditor` suite (delete refreshes, a failing refresh raises no unhandled rejection,
  > no listeners) — all three fail against `main` and pass here. `tsc --noEmit` and the test
  > typecheck are clean.
- Integration tests
  > New LS fixtures: `record_config/config6,7` (a `readonly & X` field, and a `readonly & record`
  > type definition), `update_record_config/config5` (values inside a `readonly & X` field survive the
  > round trip — fails without this change), `record_value_gen/config31,32`,
  > `get_sm_from_source/graphql_service_with_stale_end_line` (an end line past EOF resolves to the
  > same service, asserted against the exact-range fixture's output).
  > `find_matching_type/config8,10` regenerated: the intersection wrappers collapse.
  > Pre-existing failures unrelated to this PR, verified identical on `main`: `UpdateRecordConfigTest`
  > (3), `FindMatchingTypeTest` (6), `GetListenerModelTest` (1) — stdlib version drift against the
  > local distribution (`crypto` 2.10.1 vs the 2.12.1 pinned in the fixtures).

## Security checks
- Ran FindSecurityBugs plugin and verified report? no — not run; the change adds no I/O, no
  deserialization of untrusted input and no new dependencies.

## Samples
N/A

## Related PRs
N/A

## Migrations (if applicable)
N/A

## Test environment
macOS (darwin 25.5.0), JDK 21.0.5 (Temurin), Ballerina 2201.13.5, Node 22.22.3.

## Learning
N/A

## Summary

- Normalize `readonly & T` intersections before record configuration rendering.
- Preserve field metadata and support nested record expansion and valid record-value round trips.
- Improve intersection value generation and add comprehensive normalization and invariant tests.
- Recover from stale GraphQL service ranges and verify the requested service.
- Refresh the GraphQL editor after operation deletion and handle refresh failures without unhandled errors.
- Clear stale service data when the service no longer exists.
- Update the service model request contract with optional `originalName` support.


packages/ballerina-visualizer/src/views/GraphQLDiagram/GraphqlServiceEditor.tsx
```diff
@@ -241,34 +249,40 @@ export function GraphqlServiceEditor(props: GraphqlServiceEditorProps) {
...
                     lineRange: reqLineRange,
                 },
-            })
-            .then((res) => {
-                console.log("Service Model: ", res.service);
-                setServiceModel(res.service);
             });
-        getProjectListeners();
+            console.log("Service Model: ", res.service);
+            if (res.service) {
```
pasindufernando1: Minor: this guard keeps the previous model when the LS returns a response with no service, which is a different case from the rejection the `catch` below handles.

If the GraphQL service is deleted or renamed while the panel is open, the refresh comes back with `service: undefined`, the panel keeps rendering the old operations, and a subsequent `onDeleteFunction` calls `removeStatement` with `model.codedata.lineRange` values that no longer correspond to anything. The previous unconditional `setServiceModel(res.service)` cleared the panel back to the spinner.

The `catch` already covers the failed-refresh case the comment describes, so dropping the `if` may be enough.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/ServiceModelGeneratorService.java
```diff
@@ -188,16 +188,47 @@ public ServiceModelGeneratorService() {
...
+        }
         return modulePartNode.findNode(TextRange.from(start, end - start), true);
     }
 
+    /**
+     * Finds the service declaration containing the given range's start. An edit inside a service moves its end
+     * but never its start, so the start a client recorded still falls inside the service it was taken from even
+     * once the end has gone stale. Services do not nest, so at most one can contain it.
+     */
+    private static Optional<ServiceDeclarationNode> findServiceContaining(Codedata codedata, Document document) {
```
pasindufernando1, line 221: Minor: the match is on the start **line** only. The javadoc's premise ("an edit inside a service moves its end but never its start") holds for edits inside that service, but not for edits *above* it in the same file.

With two services in one `.bal`, removing lines above the open service can make the stale start line fall inside a *different* service, and that service's model is then returned as a success. `getServiceModelFromCode` is shared with ServiceDesigner, ServiceConfigureView and AIChatAgent, so the panel would render another service's operations — while the write path (deliberately not retargeted) still goes to the original range.

Comparing the resolved service's `startLine.offset` and/or its identifier against the requested codedata would make the fallback safe.

pasindufernando1 (commented):
Nice cleanup overall — the normalizer's structure-sharing/memoization and the `applyModifications(..., filePath)` fix both look right.

Four **minor** comments inline, none of them blocking.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/type/RecordValueGenerator.java
```diff
@@ -60,6 +61,64 @@ private static void generateValue(JsonObject json, StringBuilder builder, int in
...
+            // No single member to build a value from; keep the unrepresentable-type fallback of the default branch.
+            builder.append("\"%s\"".formatted(json.get("typeName").getAsString()));
+            return;
+        }
+
+        // The wrapper is the field slot, so its name and selection stand - the same merge
+        // IntersectionNormalizer applies on the way out. The record and array branches skip an unselected node,
+        // which for a record field would emit the field name with no value at all, and isAmbiguousUnionModulePrefix
+        // keys off the field's name.
+        JsonObject member = shape.deepCopy();
```
pasindufernando1, line 94: Minor: the comment says this is "the same merge `IntersectionNormalizer` applies", but only `selected` and `name` are copied onto the shape member — not `defaultValue` or `typeInfo`.

Two consequences:
- A wrapper carrying `defaultValue` and no `value` (e.g. `{name: "x", typeName: "intersection", defaultValue: "5", members: [readonly, {typeName: "int"}]}`) used to emit `5` from the default branch; it now delegates to the member, which has no `defaultValue`, so `generateDefaultValue` emits `0`.
- `isAmbiguousUnionModulePrefix` needs *both* `name` and `typeInfo`, so copying `name` alone still can't make it fire for a `readonly & task:DatabaseConfig` field — the comment implies it can.

Reachability is narrow (this is a compat net for payloads from an older webview), but mirroring `IntersectionNormalizer.merge`'s `preferMember` for `value`/`defaultValue` and carrying `typeInfo` would keep the two in step.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/type/IntersectionNormalizer.java
```diff
@@ -0,0 +1,320 @@
...
+            merged.typeInfo = wrapper.typeInfo;
+        }
+        if (wrapper.displayAnnotation != null) {
+            merged.displayAnnotation = wrapper.displayAnnotation;
+        }
+        merged.optional = wrapper.optional;
+        merged.defaultable = wrapper.defaultable;
+        merged.isRestType = wrapper.isRestType;
+        merged.selected = wrapper.selected;
+        merged.documentation = preferMember(member.documentation, wrapper.documentation);
```
pasindufernando1, line 152: Minor: this lets the *type's* documentation win over the *field slot's*. `Type.fromSemanticSymbol` sets `type.documentation` from the `TypeDefinitionSymbol`'s doc, while `getRecordType` puts the field's own doc on the wrapper via `subType.setDocumentation(documentationMap.get(name))`.

So for

```ballerina
# Per-field caching.
readonly & ServerCacheConfig cacheConfig;
```

the tooltip (`CustomType`/`RecordType` render `param.documentation`) shows `ServerCacheConfig`'s type-level description instead of the field doc — whereas a plain `ServerCacheConfig cacheConfig;` field renders the field doc correctly.

Per the AGENTS.md rule added in this PR ("the wrapper is the field slot, so its name, `typeInfo` and flags win"), `documentation` belongs with the field slot rather than the shape. The TS `mergeWrapper` has the same ordering, so both sides would need the flip.

packages/ballerina-visualizer/src/views/GraphQLDiagram/GraphqlServiceEditor.tsx
dulajdilshan: Agreed — guard dropped in 7035c94; `setServiceModel(res.service)` is unconditional again.

You're right that the two cases are different, and the `catch` is the one the comment was actually about — a failed request says nothing about whether the service is still there, so keeping the model is correct there. A success with no service says the opposite, and holding the old operations is worse than the spinner: each of them stays wired to a `codedata.lineRange` that now points at other text, and `onDeleteFunction` edits at exactly that range.

Comments split accordingly, and `GraphqlServiceEditor.test.tsx` grew a case for it ("clears the panel when the refresh finds no service").

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/core/ServiceModelGeneratorService.java
dulajdilshan: Confirmed and fixed in 7035c94 — thanks, this was reachable.

I added `get_sm_from_source/source/sample20/main.bal` (`/first` on lines 4-13, `/second` on 15-20) and asked for a stale range starting at line 8 with the end past EOF. It came back as `/first`, exactly as you describe. Worth noting the trigger is text *added* above rather than removed: removal only shortens the gap, whereas growth pushes an earlier service down over the recorded start line.

Your two suggested discriminators didn't work out, though:

- **`startLine.offset`** — a service node's range includes its metadata, so the recorded start is rarely the node's start. In `graphql_service_with_stale_end_line.json` the request starts at line 20 offset 0 while the service starts at line 17 offset 0; both offsets are 0 here and in the wrong-service case, so the comparison neither discriminates nor survives.
- **The artifact identifier** — `ModuleNodeTransformer` derives it from an annotation, then the attach point, then the type descriptor, with module-specific labels (`"GraphQL Service - /graphql"`). Recomputing that in this extension would duplicate those rules, and a mismatch would silently kill the recovery.

So the identity used is the attach point, which both sides already agree on with no new naming logic: `Utils.getPath(serviceNode.absoluteResourcePath())` is literally what `extractServicePathInfo` reports as `properties.basePath.value`. The panel sends it back as `codedata.originalName` and the fallback refuses a service that doesn't match. A request that names no service keeps the plain containment behaviour, so `ServiceEditView` / `ServiceConfigureView` / `AIChatAgent` / `DiagramWrapper` and every existing fixture are untouched.

Two fixtures: `graphql_service_with_stale_range_in_other_service` (refused) and `graphql_service_named_by_attach_point` (a stale range still resolves to the named service). The write path stays unretargeted.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/type/RecordValueGenerator.java
dulajdilshan: Thanks — splitting this one.

**`defaultValue`: agreed, fixed in 7035c94.** `generateIntersectionValue` now mirrors the default branch's order (`value` → `defaultValue` → delegate), with `record_value_gen/config33` pinning it.

Reachability is nil rather than narrow, for the record: nothing on the record-config path ever sets `defaultValue` — `Type.fromSemanticSymbol` only sets it for enum reordering, and `GeneratorUtils` sets it on connector *parameters*, which never reach `recordValue`. The key appears in no `record_config/` or `record_value_gen/` fixture. Still worth closing, since it is free.

**`typeInfo`: the member already carries it.** `getIntersectionType` never calls `setTypeInfo` on the intersection; the type-reference member goes through `getEnumType(TypeReferenceTypeSymbol, ...)`, which ends in `setTypeInfo(typeRef.getName(), symbol, type)` on the *member*, and `shape.deepCopy()` keeps it. So for `readonly & task:DatabaseConfig` the member has `typeInfo` and lacks only `name` — copying `name` alone is exactly what makes `isAmbiguousUnionModulePrefix` fire, and copying the wrapper's `typeInfo` would be copying a null. `record_value_gen/config31,32` show the same shape: `typeInfo` on the member, none on the wrapper. `IntersectionNormalizer.merge` is consistent with this — it copies `wrapper.typeInfo` only when non-null.

The comment was overstating things either way, so it now says what is actually carried and why.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/type/IntersectionNormalizer.java
dulajdilshan: I checked this against the exact case and the field doc wins today — the member's `documentation` is null on this path, so `preferMember` falls through to the wrapper's.

Reproduction, in `record_config/source/proj/types.bal`:

```ballerina
# Type level doc for ServerCacheConfig.
type ServerCacheConfig record {| boolean enabled = true; int maxAge = 60; int maxSize = 15; |};

type ReadonlyHolder record {|
    # Field doc for cacheConfig.
    readonly & ServerCacheConfig cacheConfig;
    # Field doc for plainCache.
    ServerCacheConfig plainCache;      // control
|};
```

`typesManager/recordConfig` for `ReadonlyHolder` returns `fields[0].documentation = "Field doc for cacheConfig."` — identical to the plain-field control, not the type-level description.

Why: `type.documentation = typeDocumentation.get()` only runs in the `TypeDefinitionSymbol` branch (`Type.java:424-432`). The intersection member is reached as `IntersectionTypeSymbol` → `TypeReferenceTypeSymbol` → `getEnumType(...)` → `fromSemanticSymbol(typeRef.typeDescriptor())` → `getRecordType`, which never sets `documentation`. So `member.documentation` is null and the field doc the wrapper got from `subType.setDocumentation(documentationMap.get(name))` survives.

That said, your reading of the rule is right — `documentation` describes the field slot, not the value, so grouping it with `value`/`defaultValue` is arguably the wrong bucket even though it is currently harmless. Say the word and I'll flip it to wrapper-wins on both sides (2 lines + 2 test updates); otherwise I'd leave it, since the behaviour is already what you want.

packages/ballerina-visualizer/src/views/GraphQLDiagram/GraphqlServiceEditor.tsx
```diff
@@ -241,34 +249,47 @@ export function GraphqlServiceEditor(props: GraphqlServiceEditorProps) {
...
+            const res = await rpcClient.getServiceDesignerRpcClient().getServiceModelFromCode({
                 filePath: reqFilePath,
                 codedata: {
                     lineRange: reqLineRange,
+                    // Name the service being refreshed, so the server's stale-range recovery cannot
+                    // answer with a different service that has come to enclose this range. Sent back
+                    // exactly as the server reported it; absent on the first fetch, whose range is
+                    // the one the panel was opened with and so is not stale.
+                    originalName: serviceModel?.properties?.basePath?.value,
                 },
```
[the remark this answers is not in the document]
dulajdilshan: Skipping this one — the premise doesn't hold. `serviceIdentifier` is not the attach point; it is the artifact's display name.

It arrives as `location.identifier` → `dir.name` → the LS artifact name, which `ModuleNodeTransformer` builds as `entryPointMap.get(module) + " - " + attachPoint` (graphql is not in `attachPointNamedModules`, so `usesAttachPointAsName` is false and it falls through to `serviceNameWithPath`). The committed fixture shows the result:

```
architecture-model-generator-ls-extension/src/test/resources/artifacts/config/graphql.json:100
  "name": "GraphQL Service - /graphql"
```

The server compares `codedata.originalName` against `Utils.getPath(serviceNode.absoluteResourcePath())`, i.e. `/graphql`. Sending `"GraphQL Service - /graphql"` matches nothing, so `isRequestedService` would refuse **every** fallback and switch off the stale-range recovery this PR adds — the GraphQL delete refresh would stop working. Stripping the `"<label> - "` prefix in the webview to recover the path would hard-code that naming rule in a second place, which is exactly the coupling this check was designed to avoid.

On the initial request there is also nothing to identify with: the panel has no attach point until the server sends one. That is not a gap, though — the mount range and `serviceIdentifier` come from the same artifact scan (`index.tsx:471` builds `lineRange` from the same `position`), so it is not a range the panel has edited out from under itself, and a range that fails to resolve exactly falls back to the containment behaviour that predates this PR. `getServiceModelFromCode`'s four other callers send no name either, for the same reason.

The test asserting `originalName` is `undefined` on the first call and `SERVICE_NAME` on the refresh is pinning that intent deliberately.

packages/ballerina-visualizer/src/views/BI/HelperPaneNew/Components/RecordConstructView/utils/intersection.ts
```diff
@@ -18,34 +18,164 @@
...
+        merged.displayName = wrapper.displayName;
+    }
+    if (member.hide === undefined && wrapper.hide !== undefined) {
+        merged.hide = wrapper.hide;
+    }
+    if (member.position === undefined && wrapper.position !== undefined) {
+        merged.position = wrapper.position;
+    }
+    // The member describes the shape, so anything it states about the value stands.
+    if (member.documentation === undefined && wrapper.documentation !== undefined) {
```
NipunaRanasinghe, line 100: **Low — `mergeWrapper` and `IntersectionNormalizer.merge` disagree on what counts as absent. CONFIRMED.**

`IntersectionNormalizer.merge`'s `preferMember` (IntersectionNormalizer.java:158-160) treats a null *or empty-string* `documentation`/`value`/`defaultValue` on the member as absent and falls back to the wrapper's. `mergeWrapper` here only checks `member.x === undefined` (lines 100-107), so a member carrying an explicit empty string keeps it instead of picking up the wrapper's — the same `readonly & T` payload collapses differently depending on which language server produced it, which is exactly what the comment above this function and AGENTS.md rule 7 say must not happen. Narrow trigger (an explicit empty string, not merely an absent field) and only live on the older-LS compat path, so real-world impact is likely low.

NipunaRanasinghe (approved):

packages/ballerina-visualizer/src/views/BI/HelperPaneNew/Components/RecordConstructView/utils/intersection.ts
dulajdilshan: Will address this in a future PR

PR #980, opened 2026-09-01
Make the orb invite a hover affordance with no close button

RNViththagan:
Related to https://github.com/wso2/product-integrator/issues/2227

The idle orb offered a "How can I help?" input with a ✕ to dismiss it, but clicking ✕ did nothing while the pointer was still on the orb: the dismiss set a flag that `showInvite` immediately overrode with `|| hovered`, so the same hover re-opened what was just closed.

Rather than patch the flag, this drops the ✕ entirely and makes the invite a pure hover affordance — there is nothing to dismiss when hover is the only way in and out.

- Remove the ✕ button and the `inviteDismissed` state; the invite shows on hover and hides on leave
- Bridge the dead gap between the orb and the box (`InviteHitBridge`, padding + equal negative margin) so a pointer travelling from the orb to the input never reads as a leave — no timed grace needed
- Pin the invite open while the input has focus or holds typed text, so moving the mouse away mid-sentence never discards a draft
- Hide via an interruptible opacity/transform transition (160ms) instead of an unmount, so returning mid-fade reverses smoothly
- Add `AgentStatusOrb/index.test.tsx` covering the gap-crossing, the focus/content pin, and the immediate fade-out

## Summary

- Updated the orb invite to behave as a hover-only affordance.
- Added `InviteHitBridge` to preserve the invite while the pointer crosses the gap between the orb and input.
- Keeps the invite visible when the input has focus or content.
- Replaced unmounting with an interruptible 160ms opacity and transform transition.
- Replaced the close button with a conditional clear button.
- Added tests for pointer transitions, focus and content pinning, draft handling, and fade behavior.

### DIFF
--- a/packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
+++ b/packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
@@ -253,6 +253,8 @@ export function AgentStatusOrb() {
     const [status, setStatus] = useState<AgentRunStatus | null>(null);
     const statusRef = useRef<AgentRunStatus | null>(null);
     const [hovered, setHovered] = useState(false);
+    /** Tab-to-orb has no hover, so this is what lets the keyboard reach the invite. */
+    const [orbFocused, setOrbFocused] = useState(false);
     const [anchor, setAnchor] = useState<Anchor>(loadAnchor);
     /** Orb top-left in px while dragging/snapping; null when docked at an anchor. */
     const [dragPos, setDragPos] = useState<{ x: number; y: number } | null>(null);
@@ -310,6 +312,7 @@ export function AgentStatusOrb() {
             setMiniOpen(false);
             miniPromptRef.current = undefined;
             setHovered(false);
+            setOrbFocused(false);
         }
     }, [orbHidden]);
 
@@ -322,22 +325,30 @@ export function AgentStatusOrb() {
     const dragging = dragPos !== null && !snapping;
     // Active states keep the pill visible the whole time. Idle keeps the invitation input
     // in the tree and only ever fades it, so the transition always has a start value and a
-    // focused input is never pulled out from under the keyboard. Hover brings it in; focus
-    // or an unsent draft holds it while the pointer wanders. The mini chat replaces it, and
-    // it stays hidden mid-drag and mid-snap: `anchor` isn't committed until the snap timer
-    // fires, so showing it earlier would open the mini chat at the stale anchor.
+    // focused input is never pulled out from under the keyboard. Hover or keyboard focus on
+    // the orb brings it in; focus in the input or an unsent draft holds it while the pointer
+    // wanders. The mini chat replaces it, and it stays hidden mid-drag and mid-snap: `anchor`
+    // isn't committed until the snap timer fires, so showing it earlier would open the mini
+    // chat at the stale anchor.
     const inviteHosted = !orbHidden && status?.state === "idle";
     const inviteVisible =
-        inviteHosted && !dragging && !snapping && !miniOpen && (hovered || inviteFocused || inviteText.length > 0);
+        inviteHosted &&
+        !dragging &&
+        !snapping &&
+        !miniOpen &&
+        (hovered || orbFocused || inviteFocused || inviteText.length > 0);
 
-    // The input unmounts here without ever blurring, and a draft left behind would hold the
-    // box open the moment the orb came back with the pointer nowhere near it.
+    // A draft left behind when the orb hides, a run starts or the mini chat takes over would
+    // hold the box open the moment it came back with the pointer nowhere near it. In the first
+    // two cases the input also unmounts without ever blurring.
     useEffect(() => {
-        if (!inviteHosted) {
+        if (!inviteHosted || miniOpen) {
             setInviteText("");
+        }
+        if (!inviteHosted) {
             setInviteFocused(false);
         }
-    }, [inviteHosted]);
+    }, [inviteHosted, miniOpen]);
 
     if (orbHidden) {
         return null;
@@ -514,6 +525,8 @@ export function AgentStatusOrb() {
                 onPointerDown={handlePointerDown}
                 onPointerMove={handlePointerMove}
                 onPointerUp={handlePointerUp}
+                onFocus={() => setOrbFocused(true)}
+                onBlur={() => setOrbFocused(false)}
                 title={label ? `WSO2 Integration Intelligence — ${label}` : "WSO2 Integration Intelligence"}
                 aria-label={label ? `WSO2 Integration Intelligence: ${label}. Click to open the mini chat, double-click for the chat panel.` : "Click to open the WSO2 Integration Intelligence mini chat, double-click for the chat panel"}
             >
--- a/packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
+++ b/packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
@@ -142,19 +142,25 @@ const LabelPill = styled.div`
     cursor: pointer;
 `;
 
+interface InviteVisibility {
+    visible: boolean;
+}
+
 /**
  * Carries the box's hit area across the gap the wrapper leaves untouchable, so a pointer
  * travelling from the orb to the input never counts as having left the widget. The negative
- * margin hands the padding back to the layout, leaving the box where it was.
+ * margin hands the padding back to the layout, leaving the box where it was. Only while the
+ * box is showing: hidden, it must neither catch a hover nor swallow a click meant for the
+ * diagram beneath.
  */
-const InviteHitBridge = styled.div`
-    pointer-events: auto;
+const InviteHitBridge = styled.div<InviteVisibility>`
     display: flex;
     padding: ${WIDGET_GAP}px;
     margin: -${WIDGET_GAP}px;
+    pointer-events: ${(props: InviteVisibility) => (props.visible ? "auto" : "none")};
 `;
 
-const InviteBox = styled.div<{ visible: boolean }>`
+const InviteBox = styled.div<InviteVisibility>`
     display: flex;
     align-items: center;
     gap: 4px;
@@ -163,11 +169,31 @@ const InviteBox = styled.div<{ visible: boolean }>`
     border-radius: 14px;
     padding: 5px 6px;
     box-shadow: 0 2px 10px rgba(0, 0, 0, 0.3);
-    opacity: ${(props: { visible: boolean }) => (props.visible ? 1 : 0)};
-    transform: translateX(${(props: { visible: boolean }) => (props.visible ? "0" : "6px")});
-    transition: opacity ${INVITE_FADE_MS}ms ease, transform ${INVITE_FADE_MS}ms ease;
+    opacity: ${(props: InviteVisibility) => (props.visible ? 1 : 0)};
+    transform: translateX(${(props: InviteVisibility) => (props.visible ? "0" : "6px")});
+    visibility: ${(props: InviteVisibility) => (props.visible ? "visible" : "hidden")};
+    /* visibility waits for the fade-out so the box leaves the tab order only once it is gone. */
+    transition:
+        opacity ${INVITE_FADE_MS}ms ease,
+        transform ${INVITE_FADE_MS}ms ease,
+        visibility 0s linear ${(props: InviteVisibility) => (props.visible ? 0 : INVITE_FADE_MS)}ms;
     @media (prefers-reduced-motion: reduce) {
         transition: none;
+    }
+`;
+
+const InviteClear = styled.button`
+    display: flex;
+    align-items: center;
+    background: transparent;
+    border: none;
+    color: var(--vscode-descriptionForeground);
+    cursor: pointer;
+    padding: 4px;
+    border-radius: 4px;
+    &:hover {
+        color: var(--vscode-foreground);
+        background: var(--vscode-toolbar-hoverBackground);
     }
 `;
 
@@ -235,10 +261,6 @@ export function AgentStatusOrb() {
     const snapTimerRef = useRef<ReturnType<typeof setTimeout> | undefined>(undefined);
     const [inviteText, setInviteText] = useState("");
     const [inviteFocused, setInviteFocused] = useState(false);
-    /** Rendered — true through the fade-out, after the invite is no longer wanted. */
-    const [inviteMounted, setInviteMounted] = useState(false);
-    const [inviteVisible, setInviteVisible] = useState(false);
-    const inviteFadeTimerRef = useRef<ReturnType<typeof setTimeout> | undefined>(undefined);
     /** The current view opts out of the floating orb. */
     const [orbSuppressed, setOrbSuppressed] = useState(false);
     /** Mini chat overlay toggled by clicking the orb. */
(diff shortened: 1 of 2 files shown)


SasinduDilshara (dismissed):

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
```diff
@@ -289,22 +302,50 @@ export function AgentStatusOrb() {
...
     // until the snap timer fires — showing the invite/label meanwhile would open the
     // mini chat at the stale anchor. Keep them hidden until the orb settles.
-    const showInvite = state === "idle" && !dragging && !snapping && !miniOpen && (!inviteDismissed || hovered);
+    const showInvite =
+        !orbHidden &&
+        status?.state === "idle" &&
+        !dragging &&
+        !snapping &&
+        !miniOpen &&
+        (hovered || inviteFocused || inviteText.length > 0);
```
samithkavishke: **Blocking**

**`inviteText` is never cleared when the invite goes away, so it re-opens the box unbidden.**

`inviteText.length > 0` is now a permanent pin on `showInvite`, but nothing resets the text when the orb hides or leaves idle — contrast `miniOpen`/`miniPromptRef`, which the `orbHidden` effect at L292-297 deliberately clears for exactly this reason ("the mini resurfaces unprompted as soon as the orb returns").

Reproduced against this commit with a throwaway test: hover the orb, type `stale draft`, open the AI panel (orb unmounts), close it again → the invite comes back **mounted at opacity 1 with `stale draft` still in it**, with the pointer nowhere near the orb. Same path via a state change: type into the invite, let a run start (`state !== "idle"` hides it), and when the agent returns to idle the box pops open again on its own.

Before this change the dismiss button gave the user a way out of that; now there is none. Suggest clearing `inviteText` in the `orbHidden` effect (and/or whenever `showInvite` goes false without a submit).

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
```diff
@@ -150,7 +163,12 @@ const InviteBox = styled.div`
     border-radius: 14px;
     padding: 5px 6px;
     box-shadow: 0 2px 10px rgba(0, 0, 0, 0.3);
-    animation: ${fadeIn} 0.25s ease-out;
+    opacity: ${(props: { visible: boolean }) => (props.visible ? 1 : 0)};
```
samithkavishke: **Non-blocking**

(Judgement call — this is a real bug, not a style point, and the fix is one line, so it is a strong non-blocking rather than a nit. I stopped short of blocking because the bad window is 160 ms, the failure is self-healing on a second click, and nothing the user typed is lost. Happy to see it promoted if you'd rather not ship a silent misclick.)

**The faded-out box still swallows clicks for `INVITE_FADE_MS`.**

`opacity: 0` does not remove an element from hit testing, and the box stays mounted (`inviteMounted`) for 160 ms after the pointer leaves while `InviteHitBridge` keeps `pointer-events: auto`. The `Wrapper` is deliberately `pointer-events: none` so the overlay never eats clicks on the diagram — this hands ~250 px (230 px input + padding + the 10 px bridge) of that back, invisibly.

Scenario: pointer leaves the orb heading for a node that sits under where the invite was, and clicks inside 160 ms → the click lands on the invisible box instead of the node. Fix is `pointer-events: ${p => p.visible ? "auto" : "none"}` on `InviteBox`/the bridge (and `visibility: hidden` at the end of the fade also drops it out of the tab order — see the focus note below).

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
```diff
@@ -231,7 +234,11 @@ export function AgentStatusOrb() {
     const dragStateRef = useRef<{ startX: number; startY: number; wasDrag: boolean } | null>(null);
     const snapTimerRef = useRef<ReturnType<typeof setTimeout> | undefined>(undefined);
     const [inviteText, setInviteText] = useState("");
-    const [inviteDismissed, setInviteDismissed] = useState(false);
+    const [inviteFocused, setInviteFocused] = useState(false);
+    /** Rendered — true through the fade-out, after the invite is no longer wanted. */
+    const [inviteMounted, setInviteMounted] = useState(false);
```
samithkavishke: **Non-blocking**

**On the three new `useState`s: `inviteMounted` and `inviteVisible` are derived state, and I'd push back on them.**

Neither carries information `showInvite` doesn't already have — they are `showInvite` delayed by a frame and by 160 ms. The cost of keeping them as state is real and shows up in the other comments on this PR: two extra render passes per hover, an `rAF` whose timing is load-bearing (L336), a `setTimeout` that must be cleared in two places, a window where the box is mounted-but-invisible-and-clickable, and `inviteFocused` needing a compensating effect (L324) to stay in sync.

The cheaper shape is to stop unmounting: keep the box mounted whenever the orb is (it is a single input), and drive it entirely from `showInvite` in CSS —

```
opacity: visible ? 1 : 0;
pointer-events: visible ? auto : none;
visibility: visible ? "visible" : "hidden"; /* transition-delay ${INVITE_FADE_MS}ms when hiding */
```

That deletes both states, the ref, the timer and the `rAF`, and the fade-in works because the element was already in the tree at opacity 0 rather than because a frame boundary happened to land in the right place. `inviteFocused` is the one addition I'd keep as state — it's genuinely event-sourced — though see L324 for the `document.activeElement` desync.

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
```diff
@@ -289,22 +302,50 @@ export function AgentStatusOrb() {
...
+        }
+    }, [showInvite]);
+
+    // Mount a frame before turning visible so the transition has an opacity to leave from,
+    // and hold the box in the tree until the fade-out has run.
+    useEffect(() => {
+        clearTimeout(inviteFadeTimerRef.current);
+        if (showInvite) {
+            setInviteMounted(true);
+            const frame = requestAnimationFrame(() => setInviteVisible(true));
```
samithkavishke: **Non-blocking**

**A single `requestAnimationFrame` is not reliably enough to get a transition start.**

React commits the mount inside the pointer-event task; the `rAF` callback then runs in that *same* frame, before style recalc and paint. The browser therefore sees only the final `opacity: 1` for the first paint and there is no `0 → 1` transition — the box just appears. The test can't catch this because jsdom's `rAF` is timer-backed and `advanceTimersByTime` forces the two into separate ticks, so `settle()` always sees the intermediate state.

If the mount/unmount approach is kept, this needs a double `rAF` (or a forced reflow, e.g. reading `offsetHeight`) rather than one. Not needed at all under the always-mounted approach suggested on L239.

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
```diff
@@ -289,22 +302,50 @@ export function AgentStatusOrb() {
...
+        status?.state === "idle" &&
+        !dragging &&
+        !snapping &&
+        !miniOpen &&
+        (hovered || inviteFocused || inviteText.length > 0);
+
+    // Removing a focused input fires no blur, so the pin would outlive the box it
+    // belongs to and reopen it with the pointer nowhere near.
+    useEffect(() => {
+        if (!showInvite) {
```
samithkavishke: **Non-blocking**

**`setInviteFocused(false)` here desyncs the flag from real DOM focus, because the input is still mounted and still focused.**

The comment says "removing a focused input fires no blur" — true, but the removal happens 160 ms later, and this effect fires immediately. During that window the input holds actual keyboard focus while `inviteFocused` says it doesn't, and no `onFocus` will ever re-fire to correct it (the element never lost focus).

Scenario: the invite has focus but is empty, a run starts (`state !== "idle"` → `showInvite` false → flag forced false) and finishes inside 160 ms. `showInvite` flips back true against the still-mounted, still-focused input; `hovered` is false and the text is empty, so the box now hangs on nothing and vanishes from under a focused input — the user's next keystrokes go into an element that isn't there.

Reading focus from the element (`document.activeElement === inputRef.current`, or just `:focus-within` on the wrapper via CSS) avoids needing this effect at all.

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
```diff
@@ -430,23 +471,24 @@ export function AgentStatusOrb() {
...
-                </InviteBox>
+            {inviteMounted && (
+                <InviteHitBridge>
+                    <InviteBox visible={inviteVisible}>
+                        <InviteInput
+                            value={inviteText}
+                            onChange={(event) => setInviteText(event.target.value)}
+                            onKeyDown={(event) => {
+                                if (event.key === "Enter") {
+                                    submitInvite();
```
samithkavishke: **Non-blocking**

**No dismissal path survives the removal of the close button.**

`Enter` submits, but there is no `Escape`. Once a character is typed, `inviteText.length > 0` holds the box open indefinitely (see L320) and the only ways out are to submit or to manually select-all-and-delete. The removed `InviteDismiss` was the one affordance that covered "I opened this by accident" — worth restoring as a key handler:

```
if (event.key === "Escape") { setInviteText(""); event.currentTarget.blur(); }
```

That also gives keyboard users a way to leave the widget, which the hover-only model otherwise doesn't.

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
```diff
@@ -289,34 +319,45 @@ export function AgentStatusOrb() {
...
+    const inviteVisible =
+        inviteHosted && !dragging && !snapping && !miniOpen && (hovered || inviteFocused || inviteText.length > 0);
+
+    // The input unmounts here without ever blurring, and a draft left behind would hold the
+    // box open the moment the orb came back with the pointer nowhere near it.
+    useEffect(() => {
+        if (!inviteHosted) {
+            setInviteText("");
+            setInviteFocused(false);
+        }
```
madushajg: This reset covers the orb hiding and a run starting, but not the mini chat. `miniOpen` suppresses `inviteVisible` without clearing `inviteText`, so: hover → type → click the orb → move the pointer away → close the mini chat, and the box re-opens on its own with a stale draft and the pointer nowhere near the orb, the exact case the comment above says must not happen. Suggest `if (!inviteHosted || miniOpen)`, plus a test alongside the other two resurfacing cases. If carrying the draft across a mini-chat visit is intentional, worth saying so here.

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
```diff
@@ -289,34 +319,45 @@ export function AgentStatusOrb() {
...
+    const dragging = dragPos !== null && !snapping;
+    // Active states keep the pill visible the whole time. Idle keeps the invitation input
+    // in the tree and only ever fades it, so the transition always has a start value and a
+    // focused input is never pulled out from under the keyboard. Hover brings it in; focus
+    // or an unsent draft holds it while the pointer wanders. The mini chat replaces it, and
+    // it stays hidden mid-drag and mid-snap: `anchor` isn't committed until the snap timer
+    // fires, so showing it earlier would open the mini chat at the stale anchor.
+    const inviteHosted = !orbHidden && status?.state === "idle";
+    const inviteVisible =
+        inviteHosted && !dragging && !snapping && !miniOpen && (hovered || inviteFocused || inviteText.length > 0);
```
madushajg: `hovered` is only ever set by Wrapper's `onMouseEnter`, so with the old always-rendered invite gone the input is now unreachable by keyboard: it's `visibility: hidden` until a pointer arrives, and focusing `OrbButton` doesn't reveal it. Could we OR in an orb-focus flag (`onFocus/onBlur` on `OrbButton`) so Tab-to-orb also brings the invite in? The mini chat is still reachable via Enter on the orb, so this isn't a hard block — just a lost affordance.

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
RNViththagan: Fixed in 3ce6fe6741. The draft is now wiped whenever the orb hides or a run starts, so nothing can hold the box open when the orb comes back. An unsent draft still keeps the box up while the pointer is away — that part is intended. Both repros are now tests.

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
RNViththagan: Fixed. The bridge's `pointer-events` is gated on visibility, and the box goes `visibility: hidden` at the end of the fade so it also leaves the tab order. A test checks it stops taking pointer events the moment the pointer leaves.

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
RNViththagan: Took this. The box stays mounted the whole time the orb is idle and is driven from CSS off a single `inviteVisible` boolean. `inviteMounted`, `inviteVisible` state, the timer and the rAF are gone; `inviteFocused` remains as the one event-sourced flag.

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
RNViththagan: Gone with the always-mounted change — the element is already in the tree at opacity 0, so the transition has a real start value.

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
RNViththagan: Resolved by the always-mounted change. The flag is now only reset when the input actually unmounts (orb hides, or state leaves idle), where no blur can fire — never while it is still in the tree.

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
RNViththagan: Added `Escape` (clears and blurs) and an in-input clear button that appears only once there is text. Both clear without closing; with the text gone the box just fades once the pointer moves off.

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
RNViththagan: Good catch — fixed in 9ad5695fb7. The draft is now also wiped when the mini chat opens (`!inviteHosted || miniOpen`), and your repro is a test. Carrying it across a mini-chat visit was not intended; handing the draft *into* the mini chat on an orb click is product-integrator#2217 and will land separately, before this clear.

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
RNViththagan: Added in 9ad5695fb7. `OrbButton` now reports focus into an `orbFocused` flag that is OR'd into visibility, so Tab-to-orb reveals the invite and Shift+Tab reaches the input; the flag is reset with `hovered` when the orb hides. Tests cover both. One consequence worth knowing: the mini chat does not take focus on open, so after a mouse click on the orb the button keeps focus and an empty invite will show once the mini closes — the draft is gone by then, so it is just the same reveal a focused orb gets from the keyboard.

madushajg (approved):

SasinduDilshara (approved):

PR #982, opened 2026-09-01
Stop the agent leaving a tab open for every file it edits

RNViththagan:
Related to https://github.com/wso2/product-integrator/issues/2231

Every `.bal` file the agent edited popped open as an editor tab. `addToIntegration` applies edits through `workspace.applyEdit`, which loads each file as an in-memory document and leaves it dirty with no editor; the `workspace.saveAll()` that followed saves only documents that have an editor, so those were never saved — and VS Code opens a tab for any dirty editor-less document after ~800 ms unless short-delay autosave covers it. A generation touching twenty files buried the editor in twenty tabs.

- Save exactly the documents the edit touched, straight after `applyEdit` (replacing `saveAll()`) — the document is clean before VS Code looks, in every autosave mode, and the user's own unrelated unsaved editors are no longer flushed. This is the fix.
- Mark the bulk edit `{ isRefactoring: true }`, so `files.refactoring.autoSave` saves as part of the edit where VS Code honours it (edits spanning more than one resource, e.g. a new file)
- As a safety net, close any tab the edit itself opened — never one the user already had open, and never one whose document is still unsaved
- Run the save and the tab cleanup only after `applyEdit` succeeds (a rejected edit now throws, from #1047), so a failed edit cannot save the user's pre-existing changes
- Makes the disk read-back in `verifyPersisted` (#1047) deterministic: the edited file is on disk before `addToIntegration` returns, whereas `saveAll()` never saved a document without an editor, so that check could only pass when the user's autosave happened to run first
- Match documents by full URI rather than `fsPath`, so a `git:` view of the same path cannot stand in for the real document (from review)
- Warn when a save fails — a failed save is the one case where VS Code surfaces a tab regardless of autosave
- Extract the tab helpers to `edit-tabs.ts` and cover them with `editTabs.test.ts`

Keeps the `applyEdit` write path deliberately: it is what keeps the Language Server's `file://` view (and therefore the review diff and diagrams) in sync. The LS does watch `.bal` files on disk, but its workspace manager ignores watcher events for files the client already syncs, so a plain disk write to a file the editor holds would silently desync it. VS Code's own chat editing (Copilot agent mode) takes the same approach: edit the model without opening an editor, then save the file immediately.

Note: verified via unit tests; still to confirm with a live check that no tab appears with autosave off.

## Summary

- Improved `.bal` file edit handling in the AI panel.
- Enabled refactoring autosave for bulk edits.
- Saved only documents changed by the edit.
- Closed tabs opened by edits when autosave is disabled, while preserving existing and unsaved tabs.
- Added reusable tab-management helpers and unit tests.
- Preserved `workspace.applyEdit` to keep the Language Server file view synchronized.

### DIFF
--- a/packages/ballerina-extension/src/rpc-managers/ai-panel/edit-tabs.ts
+++ b/packages/ballerina-extension/src/rpc-managers/ai-panel/edit-tabs.ts
@@ -1,0 +1,74 @@
+/**
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com) All Rights Reserved.
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { TabInputText, Uri, window, workspace } from 'vscode';
+
+/**
+ * Keeping a live workspace edit out of the user's tab bar.
+ *
+ * The agent writes `.bal` files through `workspace.applyEdit`, which is what keeps open editors,
+ * the diagrams and the Language Server's `file://` view in step — nothing else notifies the LS,
+ * since the language client registers no file watcher. The cost is that VS Code materialises a
+ * document for every file the edit touches and surfaces one the user never opened as a tab, so a
+ * generation writing twenty files buries the editor in twenty tabs.
+ */
+
+function textTabs(): { tab: unknown; uri: Uri }[] {
+    return window.tabGroups.all
+        .flatMap((group) => group.tabs)
+        .filter((tab) => tab.input instanceof TabInputText)
+        .map((tab) => ({ tab, uri: (tab.input as TabInputText).uri }));
+}
+
+function documentFor(uri: Uri) {
+    return workspace.textDocuments.find((document) => document.uri.fsPath === uri.fsPath);
+}
+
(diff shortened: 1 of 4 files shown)


packages/ballerina-extension/src/rpc-managers/ai-panel/edit-tabs.ts
```diff
@@ -0,0 +1,74 @@
...
+function textTabs(): { tab: unknown; uri: Uri }[] {
+    return window.tabGroups.all
+        .flatMap((group) => group.tabs)
+        .filter((tab) => tab.input instanceof TabInputText)
+        .map((tab) => ({ tab, uri: (tab.input as TabInputText).uri }));
+}
+
+function documentFor(uri: Uri) {
+    return workspace.textDocuments.find((document) => document.uri.fsPath === uri.fsPath);
+}
```
madushajg, line 43: `documentFor` matches documents by `fsPath` alone, ignoring the URI scheme. This can cause non-file documents like `git:/ws/main.bal` to collide with the actual `file:` document and return the wrong match.

This breaks both call sites:

* `saveEditedDocuments`: the matched `git:` document isn't dirty, so the real edited document is never saved.
* `closeTabsOpenedByEdit`: the same lookup reports the document as clean, bypassing the "never close a dirty tab" guard and potentially discarding unsaved changes.

Since tab identity already uses `uri.toString()`, `documentFor` should use the same matching strategy:

```ts
function documentFor(uri: Uri) {
    return workspace.textDocuments.find(
        (document) => document.uri.toString() === uri.toString()
    );
}
```

This also keeps the document lookup consistent with the existing tab matching logic.

SasinduDilshara (dismissed):

packages/ballerina-extension/src/rpc-managers/ai-panel/edit-tabs.ts
RNViththagan: Good catch, thanks. `documentFor` now matches on `uri.toString()`, same as the tab lookup, and there are two new jest cases with a clean `git:` view sharing the file's path — both fail on the old `fsPath` match. Fixed in 952ca37.

SasinduDilshara (dismissed):

SasinduDilshara (approved):

madushajg (approved):

PR #984, opened 2026-09-02
Resolve Central submodule functions against their own module

### LINKED ISSUE #2199: Submodule functions of Central-published packages resolve to the root module (edifact libraries) — follow-up to #1747
### Description

Follow-up to #1747, which is only **partly** fixed in the latest pack.

What now works:

- Submodules of the **current project** (`modules/<name>/`) — their functions appear in the Call Function palette and can be added.
- Submodules of **another integration / library in the same workspace** (sibling package) — same.

What still does not work: submodules of a package **pulled from Ballerina Central**. The generated EDI libraries (`ballerinax/edifact.d03a.*`) are the concrete case.

The submodule *is* now listed in the palette (e.g. `edifact.d03a.supplychain.mORDERS` → `fromEdiString`), but selecting it **opens the form of the package-root function of the same name**, not the submodule's. The parameters/return type shown belong to the root module's generic `fromEdiString`, so the inserted call does not match the submodule API and the node cannot be saved.

Both the root module and every `m<MSG>` submodule of a `libgen` package export a `fromEdiString`/`toEdiString`, so every EDI library is affected — the typed per-message API published to Central remains unreachable from low-code, which is the same user impact as #1747.

### Steps to Reproduce

1. Create an integration and add `ballerinax/edifact.d03a.supplychain` (0.9.0) as a dependency.
2. Add a **Call Function** node and search for `fromEdiString`.
3. Under `edifact.d03a.supplychain.mORDERS`, select `fromEdiString`.
4. The form that opens is the **root** `edifact.d03a.supplychain:fromEdiString` form (generic, schema-name based) instead of the `mORDERS` one; completing it and saving fails.

Expected: the submodule's own function signature is used, the call is emitted as `edifact.d03a.supplychain.mORDERS:fromEdiString(...)` with the matching import, and Save succeeds — i.e. the same behaviour that now works for project-local and sibling-package submodules.

For contrast, a submodule-only function (`getSchema`, `transformFromEdiString`, `transformToEdiString` exist only in `m<MSG>`, not at the root) is a useful discriminator: if those also misbehave, the problem is submodule resolution in general; if only the name-colliding ones do, it is a root-vs-submodule disambiguation problem.

### Possible cause

For Central-resolved packages the function is resolved against the package's **default (root) module** semantic model, ignoring the submodule in the codedata:

`packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/FunctionDataBuilder.java`

- `resolvePackageAndSemanticModel()` — the workspace branch was fixed for #1747 (its comment even notes that `resolvedPackage()` "initially selects its default module. In particular, this would make a function in a sibling package's submodule invisible"), but the Central branch still ends in `this.resolvedPackage(PackageUtil.resolveModulePackage(org, packageName, version))`.
- `resolvedPackage(Package)` then derives the semantic model from `resolvedPackage.getDefaultModule().moduleId()`, so `moduleInfo.moduleName()` (the submodule) is dropped and a name lookup finds the root-module function.

The shipped index itself is fine — `central-index.sqlite` carries the submodules as separate `Package` rows with their own functions (`edifact.d03a.supplychain.mORDERS` → `fromEdiString`, `toEdiString`, `getSchema`, `fromEdiStringWithSchema`, `toEdiStringWithSchema`, `transformFromEdiString`, `transformToEdiString`), and `DatabaseManager.getFunction(...)` filters on `p.module_name`. So the fix belongs on the package/semantic-model resolution path, not in the index.

### Version

Latest pack containing the #1747 submodule support (WSO2 Integrator VS Code extension 1.0.1)

- `ballerinax/edifact.d03a.supplychain` 0.9.0 (from Ballerina Central; root module + 30 `m<MSG>` submodules, all in `package.json` `export`)
- Related: #1747

dulajdilshan:
## Purpose

Fixes https://github.com/wso2/product-integrator/issues/2199

Also addresses editor changes for https://github.com/wso2/product-integrator/issues/2393

Previously: Renders the function from the default module
image

Now: Renders the correct form for the function in the submodule
image

## Problem

`FunctionDataBuilder.resolvedPackage(Package)` derived its semantic model from `resolvedPackage.getDefaultModule()`, dropping `moduleInfo.moduleName()` before the symbol lookup. Only the Central branch of `resolvePackageAndSemanticModel()` goes through that setter — the workspace branch was made module-aware in #1747 — and the module-aware fallback in `PackageUtil.getSemanticModel(ModuleInfo)` never runs, because `semanticModel` is non-null by then.

So a submodule symbol was either not found or silently shadowed by a same-named root function. `edifact.d03a.supplychain.mORDERS:fromEdiString` rendered the root's `(string, EDI_NAME) returns anydata` form instead of the submodule's `(string ediText) returns EDI_ORDERS_...`, and the resulting call did not compile.

## Approach

Derive the semantic model from the module matching `moduleInfo`, falling back to the default module when nothing matches (new `getTargetModuleId`).

The derivation happens in `build()`/`buildChildNodes()` rather than in `resolvedPackage(Package)`, so the two setters may be called in either order — deriving it in the setter would silently pin the default module for a caller that supplies the package first. The match itself goes through a new shared `PackageUtil.findModule(Package, String)`, which accepts both the qualified `<package>.<part>` form and the bare `<part>` form that some requests carry.

## Automation tests

- 4 new `node_template` configs driving `getNodeTemplate`: the reported submodule case, a submodule-only function, the root-module function of the same package (fallback guard), and a submodule-only function from a second package (`ballerina/workflow`), whose package name has no dots.
- `FunctionDataBuilderModuleResolutionTest` (5 cases) for the branches the LSP API cannot distinguish from outside — submodule-only resolution, the unknown-submodule fallback, and setter-order independence.
- All new tests were confirmed to fail with the production change reverted.
- Fixture: `ballerinax/edifact.d03a.supplychain:0.9.0` added to `build-config/ballerina_dependencies` (3.4 MB). Its root and every `m<MSG>` submodule export `fromEdiString` with different arity *and* return type, so assertions discriminate structurally rather than on doc text.

## Release note

Fixed the Call Function form showing the package root's function when a submodule function of a package pulled from Ballerina Central is selected.

## Security checks

- Ran FindSecurityBugs plugin and verified report? no — language-server internal refactor, no new I/O or user input handling

## Documentation

N/A — bug fix, no behaviour to document beyond the form now matching the selected function.

## Test environment

JDK 21.0.5 (Temurin), Ballerina 2201.13.4, macOS 26.5.2 (arm64).

## Notes

`ConnectionActionProvider.buildTemplates` is a second caller reaching the same branch and stays uncovered — neither provisioned package has a client class in a submodule. Its own `findSemanticModel`, and `FunctionDataBuilder.checkLocalModule` (which compares `packageName + "." + moduleNamePart()` and so never matches a default module), are two remaining copies of the same matching logic; both should move onto `PackageUtil.findModule` separately, since collapsing them changes which semantic model wins for a workspace-local module.

The unknown-submodule fallback stays silent by design here: it degrades to the root overload rather than reporting an error, which `testUnknownSubmoduleFallsBackToDefaultModule` pins. Making it fail loudly is a behaviour change worth its own issue.

## Summary

- Fixes Central package function resolution for submodule functions.
- Selects the semantic model that matches `moduleInfo`, with root-module fallback.
- Adds shared module lookup for qualified and bare module names.
- Defers semantic-model selection until module information is finalized.
- Adds tests and EDIFACT fixtures for submodule overloads, submodule-only functions, fallback behavior, naming patterns, and setter order.
- Adds the EDIFACT supply-chain package and related dependency configuration for test coverage.


pasindufernando1 (dismissed):
LGTM. Verified the core fix reaches the reported scenario — `CallBuilder.setConcreteTemplateData` never sets `resolvedPackage`, so resolution took the Central branch that pinned `getDefaultModule()`; deferring derivation to `build()` is the right fix, and the other `resolvedPackage(...)` callers set `semanticModel(...)` explicitly.

All of the below is low priority / non-blocking:

1. `FunctionDataBuilder.java:203` — `getTargetModuleId` only matches the full `<pkg>.<part>` module name, but bare-part names also flow through `ModuleInfo.moduleName()` (e.g. `new_local_connection.json` uses `module: "mod"`), and those still fall back to the package root. Mirroring `PackageUtil.getSemanticModelIfMatched` (`PackageUtil.java:585-593`) would cover both forms.
2. `FunctionDataBuilder.java:209` — the silent `getDefaultModule()` fallback makes an unknown submodule resolve to the root overload with no diagnostic when `resolvedPackage` is set, while the same input throws `IllegalStateException` when it isn't. Restricting the fallback to names not prefixed with `packageName + "."` would keep the failure visible.
3. `FunctionDataBuilder.java:347` — stale comment: the ordering rationale still claims `resolvedPackage()` "initially selects its default module", no longer true after this change. The staleness argument for checking the workspace first still holds, so worth rewording rather than leaving it to be read as dead code.
4. `FunctionDataBuilder.java:200` — fourth copy of the module-matching loop. Worth noting `checkLocalModule()` in this same class compares `packageName + "." + moduleNamePart()`, which yields `"pkg."` for a default module and so never matches it; a shared `PackageUtil.findModule(Package, String)` would fix that one too.
5. `FunctionDataBuilderModuleResolutionTest.java:73` — `VERSION = "0.9.0"` is hardcoded. Tests are offline-first, so a bump in `build-config/ballerina_dependencies/Ballerina.toml` fails `@BeforeClass` with a misleading "fixture missing" error. `flow-model-generator-ls-extension/build.gradle:138-158` already derives versions from `Dependencies.toml` for the copilot corpus.
6. `node_template` outputs don't carry import statements, so none of the new configs assert that the `mORDERS:` prefix comes with `import ballerinax/edifact.d03a.supplychain.mORDERS;` — the actual "did not compile" symptom. A `to_source` config for the same call would close that loop.

Minor doc nits: the PR description says 4 new `node_template` configs (there are 5) and describes a setter-ordering requirement that the code inverts — `testSetterOrderDoesNotAffectModuleResolution` asserts order-independence. `Ballerina.toml:305` says 30 `m<MSG>` submodules; the lock has 29 plus the root.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/FunctionSearchCommand.java
```diff
@@ -156,11 +172,131 @@ protected List<Item> search() {
...
+    static List<PackageCoordinate> topUpOrder(Map<PackageCoordinate, Set<ModuleCoordinate>> importedModulesByPackage,
+                                              Set<ModuleCoordinate> pagedModules,
+                                              Set<PackageCoordinate> pagedPackages) {
+        Set<PackageCoordinate> partiallyPaged = new LinkedHashSet<>();
+        Set<PackageCoordinate> unpaged = new LinkedHashSet<>();
+        importedModulesByPackage.forEach((importedPackage, modules) -> {
+            if (pagedModules.containsAll(modules)) {
+                return;
+            }
+            if (pagedPackages.contains(importedPackage)) {
```
samithkavishke: **[HIGH]** `pageIsFull` is computed as `centralResults.size() >= limit`, but `CentralSearchUtil.searchFunctions()` can return fewer than `limit` rows even when more matches exist — it gives up after `MAX_FETCH_ITERATIONS` (3) regardless of remaining matches.

For a broad query where `allowedOrgs` post-filtering discards most rows in each of the 3 fetch iterations, `filteredResults` ends up under `limit` even though Central holds more matches. `pageIsFull` is then `false`, so `isTopUpWorthwhile` (line 331) skips any imported package whose org is in `allowedOrgs` (e.g. `ballerinax`) — the exact truncation bug this PR exists to fix is silently defeated for broad queries.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/FunctionSearchCommand.java
```diff
@@ -156,11 +172,131 @@ protected List<Item> search() {
...
+    }
+
+    /**
+     * The imported packages worth asking Central about, most-likely-to-pay-off first: those with rows already on the
+     * page but not for every module the project imports, then those with no rows at all. A package whose every
+     * imported module is already listed is not returned at all.
+     *
+     * @param importedModulesByPackage the imported modules, grouped by their package
+     * @param pagedModules             the modules the general page covers
+     * @param pagedPackages            the packages the general page covers
```
samithkavishke: **[MEDIUM]** `withImportedPackageFunctions` only tops up when `offset == 0`, but the extra rows it prepends aren't reflected in the offset used for later pages. A topped-up function ranked below the limit by Central's own ordering can be fetched again on page 1 (`offset = limit`), duplicating it across pages.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/FunctionSearchCommand.java
```diff
@@ -156,11 +172,131 @@ protected List<Item> search() {
...
+        importedModulesByPackage.forEach((importedPackage, modules) -> {
+            if (pagedModules.containsAll(modules)) {
+                return;
+            }
+            if (pagedPackages.contains(importedPackage)) {
+                partiallyPaged.add(importedPackage);
+            } else {
+                unpaged.add(importedPackage);
+            }
+        });
```
samithkavishke: **[MEDIUM]** `MAX_TOPPED_UP_PACKAGES` caps top-up to 5 imported packages per search with no signal when more are skipped.

A project importing functions from 6+ packages outside `{ballerina, ballerinax, currentOrg}` (all "always worth topping up" per `isTopUpWorthwhile`) has `topUpOrder`'s alphabetical `TreeMap` ordering silently pick which 5 get topped up — the rest stay invisible in search with no indication anything was truncated.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/FunctionSearchCommand.java
```diff
@@ -75,6 +77,17 @@ class FunctionSearchCommand extends SearchCommand {
...
     private static final Set<String> PAGINATED_SECTION_ORGS = Set.of(STANDARD_LIBRARY_ORG, EXTENDED_LIBRARY_ORG);
+    // A searched page is a window onto a result set that a reindexed Central makes far larger: a function-name query
+    // now matches one row per declaring module rather than one per package, so a package with thirty modules can fill
+    // a page on its own and push another package's rows past the end of it. These bound the top-up that keeps an
+    // imported package's functions reachable. Central has no module filter, so a topped-up package arrives whole and
+    // the modules the project does not import are dropped below; the per-package budget therefore has to cover the
+    // package rather than the handful of rows that survive it. A d03a EDI package declares a little over two hundred
+    // functions whose names contain "fromEdiString" across its thirty modules, and a broad query against a tighter
+    // budget would be truncated before it reached the one module the project actually imports.
+    private static final int IMPORTED_PACKAGE_FUNCTION_LIMIT = 250;
```
samithkavishke: **[MEDIUM]** `IMPORTED_PACKAGE_FUNCTION_LIMIT = 250` is a hardcoded budget that appears tuned to the observed `d03a` fixture (~200 rows across 29 submodules), not derived generally.

A libgen package with more submodules/matches than `d03a`, or a broader query against it, would have its top-up request truncated at 250 rows before reaching the specific submodule the project imports — reproducing the same "wrong overload / symbol not found" bug this PR fixes, just for a larger package.

NipunaRanasinghe (commented):
Two findings inline. Separately: the PR description's Purpose/Approach sections only cover the `FunctionDataBuilder` change, but the diff also carries the Central registry URL switch, the `moduleName` deserialization on `SymbolResponse.Symbol`, and the whole imported-package top-up in `FunctionSearchCommand` (~460 lines of the 2588). The top-up is an independent feature with its own risk profile and already has four open review comments of its own; splitting it into a second PR would let the submodule resolution fix merge on its own evidence. At minimum the description needs updating to describe what is actually in the diff.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/FunctionSearchCommand.java
```diff
@@ -156,11 +173,157 @@ protected List<Item> search() {
...
+     * @param pagedPackages            the packages the general page covers
+     * @return the packages to query, in the order to query them
+     */
+    static List<PackageCoordinate> topUpOrder(Map<PackageCoordinate, Set<ModuleCoordinate>> importedModulesByPackage,
+                                              Set<ModuleCoordinate> pagedModules,
+                                              Set<PackageCoordinate> pagedPackages) {
+        Set<PackageCoordinate> partiallyPaged = new LinkedHashSet<>();
+        Set<PackageCoordinate> unpaged = new LinkedHashSet<>();
+        importedModulesByPackage.forEach((importedPackage, modules) -> {
+            if (pagedModules.containsAll(modules)) {
```
NipunaRanasinghe: **[MEDIUM]** Coverage is decided per module, so a single row from module M on the page suppresses the top-up for every other matching function in M.

`topUpOrder` drops a package when `pagedModules.containsAll(modules)`, and `mergeImportedPackageFunctions` additionally filters out `pagedModules.contains(...)` rows. Both treat "M appeared once" as "M is fully represented".

Concretely: a project imports only `ballerinax/edifact.d03a.supplychain.mORDERS`. The user types `fromEdi`. Central's general page returns one `mORDERS` row (say `fromEdiStringWithSchema`) plus 29 rows from unrelated packages, filling the page. `pagedModules` now contains `mORDERS`, so `containsAll` is true, the package is skipped, and `fromEdiString` itself — the function the user is after, and the one the reported issue is about — is never fetched. The page being full is exactly the condition the top-up exists to handle, and one arbitrary row from the module is enough to disable it.

The cheaper signal is module presence; the correct one is whether the module's matching functions were truncated, which the page cannot tell you. Topping up a module that already has rows and deduplicating on `(module, symbol name)` rather than on module would close this, at the cost of not being able to skip a package on the strength of one row.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/CentralSearchUtil.java
```diff
@@ -319,6 +319,60 @@ public List<SearchResult> searchFunctionsByOrg(String query, int limit, int offs
...
+     * from a single query: an EDI package declares upwards of two hundred functions whose names contain
+     * {@code fromEdiString} across its thirty modules. {@code limit} is the caller's budget for that.</p>
+     *
+     * @param query       the search query string (empty to list all of the package's functions)
+     * @param limit       the maximum number of the package's functions to return
+     * @param org         the organization that published the package
+     * @param packageName the package name to scope the search to
+     * @return the package's matching functions, or null if the request failed
+     */
+    public List<SearchResult> searchFunctionsInPackage(String query, int limit, String org, String packageName) {
```
NipunaRanasinghe: **[LOW]** `searchFunctionsInPackage` is `searchFunctionsByOrg` (line 284) plus a composed `q`, a fixed offset, and a post-filter — same `queryMap` keys, same null/empty guards, same `symbolType` check, same `catch (RuntimeException) -> null`. Two copies will drift; the `q`-only-when-non-empty difference between them is already one such divergence.

It reduces to a delegation:

```java
public List<SearchResult> searchFunctionsInPackage(String query, int limit, String org, String packageName) {
    if (packageName == null || packageName.isEmpty()) {
        return new ArrayList<>();
    }
    String scopedQuery = query == null || query.isEmpty() ? packageName : query + " " + packageName;
    List<SearchResult> results = searchFunctionsByOrg(scopedQuery, limit, 0, org);
    if (results == null) {
        return null;
    }
    return results.stream()
            .filter(result -> packageName.equals(result.packageInfo().packageName())
                    && org.equals(result.packageInfo().org()))
            .collect(Collectors.toCollection(ArrayList::new));
}
```

`toSearchResult` puts `symbol.name()` in `Package.packageName()` and `symbol.organization()` in `Package.org()`, so the filter is equivalent to the current `symbol`-level one. The `org == null || org.isEmpty()` guard is already inside `searchFunctionsByOrg`, which returns an empty list for it.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/FunctionSearchCommand.java
dulajdilshan: Confirmed, and worse than a theoretical risk — fixed in 278a0627.

Measured against the reindexed dev registry:

```
q=fromEdi -> Central count=1298
fetch loop scans 180 rows, keeps 8 after the org filter
page=8 < limit=20  ->  pageIsFull=false  ->  top-up skipped
```

So the previous commit fixed this for organizations *outside* `allowedOrgs` and left it standing for those inside — which is exactly where `ballerinax/edifact.d03a.supplychain` sits, i.e. the package the issue was reported against.

`pageIsFull` and `isTopUpWorthwhile` are both gone. Every imported module is now asked about; nothing is inferred from the page's length. `FunctionSearchTopUpMergeTest.testShortPageStillTopsUp` pins it, and I confirmed it fails if the length gate is put back.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/FunctionSearchCommand.java
dulajdilshan: Correct, and this was the sharpest of the round — fixed in 278a0627, in the shape you suggested.

Both places treated "M appeared once" as "M is fully represented": `topUpOrder`'s `pagedModules.containsAll(modules)` skip, and the merge's `!pagedModules.contains(...)` filter, which would have discarded the rows even if the request had been made.

Now:
- no module is skipped on the strength of a row — presence only reorders (modules absent from the page are asked about first, since they are likelier to be missing something);
- deduplication is by `(module, function name)` via a small `FunctionCoordinate` record, so a module already holding a row contributes the functions it does not.

Your `fromEdi` → `fromEdiStringWithSchema` on the page → `fromEdiString` never fetched scenario is `testModuleWithARowOnThePageStillToppedUp`; the no-repeat half is `testFunctionAlreadyOnThePageNotRepeated`. Both were confirmed to fail against the old behaviour.

You noted the cost is losing the ability to skip a package on one row. Accepted — it was the wrong signal. The request count is bounded by the budget instead.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/FunctionSearchCommand.java
dulajdilshan: Fair — it was fitted to the one package family I could measure, and 278a0627 removes the need for it.

The registry does support scoping to a module: naming the full dotted module name in `q` narrows the match.

| query | count | distinct modules |
|---|---|---|
| `fromEdiString edifact.d03a.finance` | 210 | 30 |
| `fromEdiString edifact.d03a.finance.mINVOIC` | **7** | **1** |

I could not verify this when the constant was written — dev was returning 504 to every query — so I sized the budget to cover a whole package and got the number you flagged.

The top-up is now per module, and `IMPORTED_MODULE_FUNCTION_LIMIT = 50`. A module holds a handful of matching functions (7 by function name, 13 for its whole surface) regardless of how large the package around it is, so the budget no longer scales with package size.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/CentralSearchUtil.java
dulajdilshan: Agreed, and adopted in 278a0627 — thanks for writing out the delegation, it was right.

The method is now module-scoped (see the thread on `IMPORTED_PACKAGE_FUNCTION_LIMIT`), so it ended up as:

```java
public List<SearchResult> searchFunctionsInModule(String query, int limit, ModuleCoordinate module) {
    if (module == null || module.moduleName().isEmpty()) {
        return new ArrayList<>();
    }
    String scopedQuery = query == null || query.isEmpty()
            ? module.moduleName() : query + " " + module.moduleName();
    List<SearchResult> results = searchFunctionsByOrg(scopedQuery, limit, 0, module.org());
    if (results == null) {
        return null;
    }
    return results.stream()
            .filter(result -> module.equals(result.packageInfo().coordinate()))
            .collect(Collectors.toCollection(ArrayList::new));
}
```

Query keys, guards, symbol-type check and the null-on-failure contract now have one definition. The `q`-only-when-non-empty divergence you spotted is gone with the duplicate.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/FunctionSearchCommand.java
dulajdilshan: Valid. Half of it is fixed in 278a0627; the other half I'd rather not fake.

**Fixed:** deduplication is now by `(module, function name)` rather than by module, so a function already listed on page 1 is no longer also prepended to it.

**Not fixed:** a topped-up function can still be served again by the general stream on a later page. The search API is stateless — page 3 is a fresh request that knows nothing of what page 1 prepended — so closing this needs either per-session state in the LS or an exclusion parameter Central does not have.

Scope of what remains, measured on dev: the row has to be from an organization the general page carries at all (a partner-org package is filtered out of every page, so it cannot duplicate), and the user has to page past the row's natural rank. In the case I measured, the imported module's rows sat at ranks 526–532 of the org-filtered stream for `q=fromEdi`.

Happy to file it if you'd like it tracked rather than left in the thread.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/FunctionSearchCommand.java
dulajdilshan: Accurate description. It is a deliberate bound rather than an oversight, so I've left the cap in — but 278a0627 changes what it counts and how it is spent, which is worth restating.

The unit is now the **module**, not the package (`MAX_TOPPED_UP_MODULES = 5`), because the top-up is module-scoped. The cap exists to bound request fan-out: every imported module is now asked about unconditionally, so without it a project importing twenty modules would issue twenty-one Central requests per keystroke.

On "silently pick which 5" — the order is deterministic and no longer alphabetical-only: modules with no rows on the page come first (likeliest to be missing something), then those with some, each group keeping the stable order `ImportedModules` produces. So the budget goes to the modules most likely to need it, and the same project gets the same answer every time. `testUnpagedModulesGetTheBudgetFirst` and `testRequestBudgetIsCapped` pin both halves.

On the missing signal: you're right that there is none, and the search response has no field for "results truncated" today — adding one is an API change across every search kind, not something I'd slip into this PR. Worth its own issue if the team wants it; say the word and I'll open one.

NipunaRanasinghe (approved):

samithkavishke (approved):
LGTM

PR #985, opened 2026-09-02
Fix the module imports of the BI Inteligence for package's modules

SasinduDilshara:
Fix the module imports of WI Intelligence for the package's modules

Fixes https://github.com/wso2/product-integrator/issues/2279

## Summary

- Fixed library resolution for `ballerinax/aws` submodules, including `ballerinax/aws.sns`.
- Prevented failed library requests from stopping other libraries from loading.
- Preserved requested module identity and returned module-specific catalog data and documentation.
- Added regression tests for unresolved libraries and submodule resolution.

### DIFF
--- a/packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
+++ b/packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
@@ -952,9 +952,22 @@ async function getExternalRecords(
 
         let library = cachedLibraries.find((lib) => lib.name === libName);
         if (!library) {
-            const result = (await langClient.getCopilotFilteredLibraries({
-                libNames: [libName]
-            })) as { libraries: Library[] };
+            // A failed fetch of ONE external dependency must not cost the caller the libraries it
+            // already has. Without this catch, a rejected request unwound uncaught to LibraryGetTool's
+            // catch-all, which answers the WHOLE tool call with `[]` — `ballerinax/aws.sns` was lost
+            // because its `ConnectionConfig.auth` field links to `ballerinax/aws.auth`, whose fetch the
+            // language server answered with a JSON-RPC error. Skipping degrades to the same shape as a
+            // gracefully-empty response below: the referencing field still renders, with its Special
+            // Agent Note naming the owning module, so only this record's definition is missing.
+            let result: { libraries: Library[] };
+            try {
+                result = (await langClient.getCopilotFilteredLibraries({
+                    libNames: [libName]
+                })) as { libraries: Library[] };
+            } catch (error) {
+                console.warn(`Library ${libName} could not be fetched: ${error}. Skipping.`);
+                continue;
+            }
             if (result.libraries && result.libraries.length > 0) {
                 library = result.libraries[0];
             } else {
--- a/packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
+++ b/packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
@@ -352,10 +352,8 @@ export function libraryContains(library: string, libraries: string[]): boolean {
 }
 
 export async function getMaximizedSelectedLibs(libNames: string[]): Promise<Library[]> {
-    const result = (await langClient.getCopilotFilteredLibraries({
-        libNames: libNames
-    })) as { libraries: Library[] };
-    const normalizedLibraries: Library[] = result.libraries.map(lib => {
+    const libraries = await fetchFilteredLibraries(libNames);
+    const normalizedLibraries: Library[] = libraries.map(lib => {
             return {
                 name: lib.name,
                 description: lib.description,
@@ -370,6 +368,49 @@ export async function getMaximizedSelectedLibs(libNames: string[]): Promise<Libr
         });
 
     return normalizedLibraries;
+}
+
+/**
+ * Fetches the full catalogs of the given libraries in one request, degrading to one request per library
+ * when the batch is rejected.
+ *
+ * The language server already contains failures per library, so a rejected batch means the request as a
+ * whole failed. Retrying each library on its own lets every library that can still be served reach the
+ * caller instead of one bad entry costing the whole batch, e.g. `[salesforce, aws.sns]` losing both.
+ * Retries run one at a time: each compiles a package, so fanning them out would only contend for the
+ * language server. When no library can be served the original error propagates, so a dead language server
+ * still surfaces as a failure rather than an empty catalog.
+ */
+async function fetchFilteredLibraries(libNames: string[]): Promise<Library[]> {
+    const fetchLibraries = async (names: string[]): Promise<Library[]> => {
+        const result = (await langClient.getCopilotFilteredLibraries({
+            libNames: names
+        })) as { libraries: Library[] };
+        return result.libraries ?? [];
+    };
+
+    try {
+        return await fetchLibraries(libNames);
+    } catch (error) {
+        if (libNames.length <= 1) {
+            throw error;
+        }
+        console.warn(`Batch fetch of libraries [${libNames}] failed: ${error}. Retrying each library individually.`);
+        const libraries: Library[] = [];
+        let anyFetched = false;
+        for (const libName of libNames) {
+            try {
+                libraries.push(...await fetchLibraries([libName]));
+                anyFetched = true;
+            } catch (libError) {
+                console.warn(`Library ${libName} could not be fetched: ${libError}. Skipping.`);
+            }
+        }
+        if (!anyFetched) {
+            throw error;
+        }
+        return libraries;
+    }
 }
 
 export async function toMaximizedLibrariesFromLibJson(
--- a/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/CopilotLibraryManager.java
+++ b/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/CopilotLibraryManager.java
@@ -34,6 +34,7 @@ import io.ballerina.flowmodelgenerator.core.copilot.service.ServiceLoader;
 import io.ballerina.flowmodelgenerator.core.copilot.util.SymbolProcessor;
 import io.ballerina.modelgenerator.commons.ModuleInfo;
 import io.ballerina.modelgenerator.commons.PackageUtil;
+import io.ballerina.projects.Module;
 import io.ballerina.projects.Package;
 
 import java.io.IOException;
@@ -121,7 +122,10 @@ public class CopilotLibraryManager {
      * Documentation is included only for packages of the organizations listed in
      * {@link #DOC_WHITELIST_ORGS}.
      *
-     * @param libraryNames Array of library names in "org/package_name" format to filter
+     * @param libraryNames Array of library names in "org/module" format to filter. A package's default
+     *                     module is addressed by the package name itself; a
+     *                     non-default exported module by its full module name ("ballerinax/aws.auth"),
+     *                     which resolves through its owning package ("ballerinax/aws")
      * @return List of Library objects with complete information
      */
     public List<Library> loadFilteredLibraries(String[] libraryNames) {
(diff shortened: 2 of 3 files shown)


packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
```diff
@@ -952,9 +952,22 @@ async function getExternalRecords(
...
-            })) as { libraries: Library[] };
+            // A failed fetch of ONE external dependency must not cost the caller the libraries it
+            // already has. Without this catch, a rejected request unwound uncaught to LibraryGetTool's
+            // catch-all, which answers the WHOLE tool call with `[]` — `ballerinax/aws.sns` was lost
+            // because its `ConnectionConfig.auth` field links to `ballerinax/aws.auth`, whose fetch the
+            // language server answered with a JSON-RPC error. Skipping degrades to the same shape as a
+            // gracefully-empty response below: the referencing field still renders, with its Special
+            // Agent Note naming the owning module, so only this record's definition is missing.
+            let result: { libraries: Library[] };
+            try {
```
RNViththagan, line 1004: This guards the external-records path nicely, but the main fetch in `getMaximizedSelectedLibs` (around line 354) is still unguarded, and that's the call every top-level requested library goes through in one batch. So if the LS fails for one of, say, `[salesforce, aws.sns]`, we still lose both.

Shall we cover that call the same way? Or better, handle it per library on the LS side so one bad library never fails the whole batch (left a note on that in `CopilotLibraryManager`).

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/CopilotLibraryManager.java
```diff
@@ -144,43 +148,58 @@ public List<Library> loadFilteredLibraries(String[] libraryNames, Map<String, St
...
-            ModuleInfo moduleInfo = new ModuleInfo(org, packageName, org + "/" +
-                    packageName, null);
-
-            // Resolve the package once; the README loader below reuses this same Package
-            // to avoid a second (potentially network-bound) resolution.
+            // Resolve the owning package once; the README loader below reuses this same Package
+            // to avoid a second (potentially network-bound) resolution. The requested name may be a
+            // package ("ballerinax/aws.sns") or a non-default exported module ("ballerinax/aws.auth"
+            // lives in "ballerinax/aws") — the coordinate cross-package type links carry. Resolution
+            // failure is contained per library, so one bad entry cannot abort the rest of the batch.
```
RNViththagan: Small correction here: only the resolution step is contained. Everything after it (`getSemanticModel`, `processModuleSymbols`, `loadAllServices`, the doc readers) still runs unguarded inside the loop, so a compile error or NPE on one resolved library would still fail the whole request.

Shall we wrap the per-library body in a try/catch that logs and `continue`s, so this comment actually holds?

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/CopilotLibraryManager.java
```diff
@@ -216,6 +237,69 @@ public List<Library> loadFilteredLibraries(String[] libraryNames, Map<String, St
...
+     * lookup surfaces the 404), and for a non-final {@link #resolveModule} candidate — a module-only
+     * name such as {@code aws.auth} — that is a routine outcome, not an error.
+     */
+    private static Optional<Package> resolvePackageSafely(String org, String name, String pinnedVersion) {
+        try {
+            return pinnedVersion == null || pinnedVersion.isBlank()
+                    ? PackageUtil.getModulePackage(PackageUtil.getSampleProject(), org, name)
+                    : PackageUtil.getModulePackage(PackageUtil.getSampleProject(), org, name, pinnedVersion);
+        } catch (RuntimeException e) {
+            LOGGER.log(Level.FINE, "Package resolution failed for " + org + "/" + name, e);
```
RNViththagan: Catching every `RuntimeException` and logging at FINE means a Central outage looks exactly like "package doesn't exist" and the library just silently disappears from the catalog.

Shall we log this at WARNING at least? Then we can tell the two apart when someone reports a missing library.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/CopilotLibraryManager.java
```diff
@@ -216,6 +237,69 @@ public List<Library> loadFilteredLibraries(String[] libraryNames, Map<String, St
...
+                // export the requested module yields empty instead of a wrong catalog.
+                return candidate.equals(moduleName)
+                        ? Optional.of(new ResolvedModule(pkg, pkg.getDefaultModule()))
+                        : Optional.empty();
+            }
+            int lastDot = candidate.lastIndexOf('.');
+            if (lastDot < 0) {
+                return Optional.empty();
+            }
+            candidate = candidate.substring(0, lastDot);
```
RNViththagan, line 325: Just a thought on the stripping loop: plenty of real packages are dotted themselves (`health.fhir.r4.international401`, `aws.sns`...). If one of those fails to resolve transiently, we keep stripping and try `health.fhir.r4` → `health.fhir` → `health`, each a Central round-trip against a different real package, before giving up empty.

Not a correctness issue thanks to the exact-name check, just wasted calls, and we can't tell a wrong name from a flaky Central. Not a blocker, but worth keeping in mind if this ever shows up as slowness.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/CopilotLibraryManager.java
```diff
@@ -216,6 +237,69 @@ public List<Library> loadFilteredLibraries(String[] libraryNames, Map<String, St
...
+        }
+    }
+
+    /**
+     * One package-resolution attempt that reports failure as empty instead of throwing:
+     * {@code getModulePackage} throws for a name Central has no package under (its latest-version
+     * lookup surfaces the 404), and for a non-final {@link #resolveModule} candidate — a module-only
+     * name such as {@code aws.auth} — that is a routine outcome, not an error.
+     */
+    private static Optional<Package> resolvePackageSafely(String org, String name, String pinnedVersion) {
```
RNViththagan: There's already `PackageUtil.resolveModulePackage(org, name, version)` doing the same thing (same try/catch around `getModulePackage`), and the two already differ slightly: this one treats a blank version as latest, that one only `null`.

Shall we reuse that one (adding the blank check there if we need it) instead of having two copies to keep in sync?

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/CopilotLibraryManager.java
```diff
@@ -216,6 +237,69 @@ public List<Library> loadFilteredLibraries(String[] libraryNames, Map<String, St
...
+                Package pkg = optPackage.get();
+                for (Module module : pkg.modules()) {
+                    if (module.moduleName().toString().equals(moduleName)) {
+                        return Optional.of(new ResolvedModule(pkg, module));
+                    }
+                }
+                // Defensive: a package's default module always shares its name, so for the exact-name
+                // candidate this preserves the previous behaviour. A shortened candidate that does not
+                // export the requested module yields empty instead of a wrong catalog.
+                return candidate.equals(moduleName)
```
RNViththagan: I think this fallback can't actually be hit. `pkg.modules()` always includes the default module, and its name equals the package name, so for the exact-name candidate the loop above already returns.

Shall we just `return Optional.empty()` here? Keeps the method a bit simpler.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/CopilotLibraryManager.java
```diff
@@ -144,43 +148,58 @@ public List<Library> loadFilteredLibraries(String[] libraryNames, Map<String, St
...
-            Package pkg = optPackage.get();
+            Package pkg = resolution.get().pkg();
+            Module module = resolution.get().module();
+            String packageName = pkg.packageName().value();
+            boolean isDefaultModule = moduleName.equals(packageName);
+
+            // Create module info (use latest version by passing null). The default-module form is
+            // unchanged; a submodule gets the real (org, packageName, moduleName) triple.
+            ModuleInfo moduleInfo = isDefaultModule
+                    ? new ModuleInfo(org, moduleName, org + "/" + moduleName, null)
```
RNViththagan: Minor: `moduleName()` gets two different shapes here, `org/module` for the default module vs the bare `aws.auth` for a submodule. Nothing breaks today as far as I can see, but it's an easy trap for the next consumer of `ModuleInfo`. Shall we pick one shape for both?

packages/ballerina-core/src/rpc-types/ai-panel/interfaces.ts
```diff
@@ -425,6 +486,14 @@ export interface SemanticDiff {
 export interface SemanticDiffResponse {
     loadDesignDiagrams: boolean;
     semanticDiffs: SemanticDiff[];
+    /** Set when diff computation failed entirely; semanticDiffs is absent in that case. */
+    errorMsg?: string;
```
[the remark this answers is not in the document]
SasinduDilshara: This is outside the code changes

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/main/java/io/ballerina/copilotagent/extension/CopilotAgentService.java
```diff
@@ -56,6 +73,117 @@ public Class<?> getRemoteInterface() {
...
 
+    /**
+     * (Re)establishes the ai:// frozen baseline of a package from explicit file contents,
+     * atomically from the caller's perspective: any cached ai:// project is evicted, the
+     * package is reloaded, and the provided contents are applied before the response
+     * resolves. This replaces the fire-and-forget didClose/didOpen notification protocol,
+     * whose disk-read timing raced against the caller's first workspace edit.
+     */
+    @JsonRequest
+    public CompletableFuture<EnsureAiBaselineResponse> ensureAiBaseline(EnsureAiBaselineRequest request) {
```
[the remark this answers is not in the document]
SasinduDilshara: This is outside the code changes

packages/ballerina-extension/e2e-test/e2e-playwright-tests/type-editor/TypeEditorUtils.ts
```diff
@@ -407,6 +412,65 @@ export class TypeEditorUtils {
...
+     * call this before filling anything.
+     */
+    async openCreateFromScratchTab(): Promise<void> {
+        // The tab button carries no testid; `create-from-scratch-tab` is the id
+        // of the content it reveals, which is what confirms the switch landed.
+        const tabButton = this.webView.getByRole('button', { name: /Create from scratch/ }).first();
+        await this.waitForElement(tabButton, 30000);
+        await tabButton.click({ force: true });
+        await this.waitForElement(this.webView.locator('[data-testid="create-from-scratch-tab"]'), 30000);
+    }
```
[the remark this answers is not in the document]
SasinduDilshara: This is outside the code changes

packages/ballerina-visualizer/src/components/AgentStatusOrb/index.tsx
```diff
@@ -449,14 +372,23 @@ export function AgentStatusOrb() {
...
+        }
+        setMiniOpen((open) => !open);
+    };
+
+    const handleDoubleClick = () => {
+        if (dragPos !== null || !rpcClient) {
+            return;
         }
+        setMiniOpen(false);
+        rpcClient.getCommonRpcClient().executeCommand({ commands: [SHARED_COMMANDS.OPEN_AI_PANEL] });
```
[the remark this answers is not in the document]
SasinduDilshara: This is outside the code changes

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/CopilotLibraryManager.java
SasinduDilshara: To do this we need to update the packageUtil, which will affect the WI, So lets keep this in here for now

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
SasinduDilshara: Done in 2ea4f43. Went with both halves of the suggestion: the LS now contains failures per library (see the `CopilotLibraryManager` thread), and the batch call in `getMaximizedSelectedLibs` goes through a new `fetchFilteredLibraries` helper. It tries the batch first; if the batch is rejected it retries each library individually, skipping the ones that still fail, and rethrows the original error only when none can be served, so a dead LS still surfaces as a failure instead of an empty catalog.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/CopilotLibraryManager.java
SasinduDilshara: Good catch, fixed in 2ea4f43. The per-library body moved into `loadLibrary(libraryName, pinnedVersion)`, and the loop wraps that call in a try/catch that logs at WARNING and continues. Resolution, compilation, symbol processing, service loading and the doc readers are all inside the guard now, so the comment holds. The test javadoc in `CopilotLibraryFilterTest` was updated to match.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/CopilotLibraryManager.java
SasinduDilshara: Fixed in 2ea4f43. Shortened candidates (`aws.auth` -> `aws`) still log at FINE since failing there is routine, but `resolveModule` now keeps the exception from the exact-name attempt and, when nothing resolves, logs it at WARNING with that cause attached. A Central outage and a genuine 404 are now distinguishable in the log. `resolvePackageSafely` was folded into `resolveModule` to make this possible.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/CopilotLibraryManager.java
SasinduDilshara: Agreed on the cost, but we can't short-circuit reliably: `CentralAPIClient.getPackageVersions` throws the same `CentralClientException` for a 404 and for an I/O error (only the message text differs), and `RestClient` wraps it in a plain `RuntimeException`. Matching on message text felt too brittle. What we do have now: the exact-name failure is logged at WARNING with its cause (see the thread above), the extra lookups are bounded by the dot-segment count, and each lookup consults the local repository before Central, so a cached package never hits the network. Documented this trade-off in the `resolveModule` javadoc in 2ea4f43.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/CopilotLibraryManager.java
SasinduDilshara: You're right, it was unreachable. Replaced in 2ea4f43 with `return Optional.empty()` plus a WARNING naming the package that exists but doesn't export the requested module.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/CopilotLibraryManager.java
SasinduDilshara: Fixed in 2ea4f43, and thanks, the old shape was actually wrong: with `moduleName = "org/pkg"` the shared `CommonUtils.getImportStatement` would have produced `org/pkg.org/pkg`. Both cases now use `new ModuleInfo(org, packageName, moduleName, null)` with the bare module name, which is the same shape `ModuleInfo.from(ModuleID)` produces.

RNViththagan (approved):
LGTM

NipunaRanasinghe (dismissed):

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/copilot/CopilotLibraryManager.java
```diff
@@ -144,43 +148,58 @@ public List<Library> loadFilteredLibraries(String[] libraryNames, Map<String, St
...
             String pinned = pinnedVersions == null ? null : pinnedVersions.get(libraryName);
-            Optional<Package> optPackage = pinned == null || pinned.isBlank()
-                    ? PackageUtil.getModulePackage(PackageUtil.getSampleProject(), org, packageName)
-                    : PackageUtil.getModulePackage(PackageUtil.getSampleProject(), org, packageName, pinned);
-            if (optPackage.isEmpty()) {
+            Optional<ResolvedModule> resolution = resolveModule(org, moduleName, pinned);
+            if (resolution.isEmpty()) {
+                LOGGER.warning("No package resolvable for '" + libraryName + "'. Skipping.");
                 continue;
             }
```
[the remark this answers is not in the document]
SasinduDilshara: 'A compile failure or an NPE on one resolved library' If something happen like this it should resolve in semantic Model

IsuruMaduranga (approved):

NipunaRanasinghe (approved):

PR #989, opened 2026-09-02
Add import alias support 

samithkavishke:
## Description

Fixes: https://github.com/wso2/product-integrator/issues/2119

Two colliding modules currently produce source that does not compile, and a model that
cannot say which module a type came from. This PR fixes both, at every point where the
language server names a module.

---

### The problem

Ballerina derives an unaliased import's prefix from the module name's **last dot-segment**:

```ballerina
import ballerinax/googleapis.drive;   // binds `drive`
import ballerinax/ai.google.drive;    // binds `drive` too -> BCE2008 redeclared symbol
```

The language server derived prefixes the same way, in several places, independently. Two
failures follow from that, and they are not the same failure:

**1. Generated source does not compile.** Adding a connection whose module's natural
prefix is already bound by a different module emits a second bare import. The file no
longer compiles, and the user did nothing wrong.

**2. The model loses a module.** `Property.imports` is keyed by prefix. Two modules
ending in the same segment claimed one key, and the second `put` **displaced** the first
— that module's import vanished outright. Meanwhile the type text rendered both arms as
the same `drive:`, so nothing downstream could recover which module either arm meant:

```
return type   ai.google.drive:TextDataLoader | googleapis.drive:File

text          "drive:TextDataLoader|drive:File"        <- one qualifier, two modules
imports       {drive: ballerinax/googleapis.drive}     <- ai.google.drive is simply gone
```

---

### The core decision

**Render the type text and its imports map together, from one allocator, and key all
resolution on `org/module` — never on the prefix.**

That single decision is what the rest of the PR follows from:

- A prefix is **not derivable** from a module name. Two modules yield the same natural
  one, and only the component that rendered the text knows which of them got it.
  Therefore the prefix must be *carried*, not recomputed — and the imports map, which
  already travels alongside the text, is where it is carried.
- Because the map key **is** the qualifier the text uses, text and map cannot disagree.
  A consumer holding the text also holds the identity of everything in it.
- Because resolution keys on `org/module`, a prefix arriving from anywhere — the index,
  the frontend, an older client — is treated as *proposed*, not authoritative. It is
  a join key between text and module, nothing more.

A second decision, equally load-bearing:

**Aliasing and rewriting must happen together.** Emitting `import x/a.drive as drive2;`
while leaving references as `drive:Foo` is *worse* than the original bug: a compile error
becomes a silent mis-binding to whichever module kept `drive`. Every site that can alias
therefore also rewrites, and `isNaturalAmbiguous` refuses to alias where it cannot.

---

#### 1. A shared prefix resolver — `model-generator-commons`

`ModuleAliasResolver` and `ModulePrefixContext` moved out of
`service-model-generator/.../util` into `model-generator-commons`, where both generators
can reach them (~310 lines deleted from their old home, no logic change in the move).
`Utils.java` shed its private copy of the same walk.

Joined by two new classes:

- **`ImportPrefixReader`** — reads what a file *actually* binds, from its syntax tree.
  Org-less imports now match, so a local submodule's own `mod:Person` is no longer
  rewritten to something else.
- **`TypeQualifierAllocator`** — hands out one distinct qualifier per module named in a
  single type signature, and records what each stands for.

Allocation policy, in one place and applied everywhere: **natural segment → camelCase
(`googleapisDrive`) → numbered suffix (`github2`)**. First claimant keeps the natural
segment, so anything that names no colliding module renders exactly as it does today.

#### 2. Type rendering — `CommonUtils`, `FunctionDataBuilder`

`CommonUtils.getTypeSignature` gained an allocator-threading overload. Where it used to
truncate `ballerinax/ai.google.drive:1.0.0:TextDataLoader` to `drive:TextDataLoader`, it
now asks the allocator for the qualifier, and the allocator records `org/module ->
qualifier` as it goes.

`FunctionDataBuilder` builds one allocator per return type and one per parameter, then
records imports through the new `qualifiedImports(...)`.

> **Authority is split deliberately here.** The symbol walk (`getImportStatements`) still
> decides *which* modules are importable; the allocator supplies *only the qualifier*.
> The allocator reads text, and text names modules that are never imported — a dependent
> type's `array:Type` resolves to `ballerina/lang.array`. Keying the map off the
> allocator would have started emitting `import ballerina/lang.array;` into user source.
> With the split, the import **set** is byte-identical to before; only a contested key
> differs.

Also fixes an eagerly evaluated `Optional.orElse` in the union and typedesc branches: the
fallback renders the whole union, error members included, and its result was discarded
but its **side effects on the allocator were not**.

#### 3. Model shape — `Property`

`Property.Builder.imports(...)` stopped re-deriving the key from the module name. It now
reads the key from the encoded statement when one is carried, and — critically — a key
already standing for a *different* module gets a free one instead of displacing it. Map
changed to `LinkedHashMap` so the emitted import order is deterministic.

#### 4. Generated source — `SourceBuilder`

The centre of the change. `SourceBuilder` now holds a **`ModulePrefixContext` per target
file**, seeded from that file's bound prefixes and built lazily. `prefixFor(org, module)`
is the single decision point for the whole operation, so two modules resolved in one
request cannot both claim a prefix.

- `imports` became `Map<String, String>` (was `Set<String>` + `HashSet`) — a prefix now
  has to be stored, not recomputed.
- `TokenBuilder.expressionWithType(...)` was **deleted** so no site can bypass
  requalification.
- `requalifiedType` / `requalifiedValue` rewrite a property's authored text onto the
  file's actual prefixes. Values are rewritten only for new nodes, so editing an existing
  node never disturbs source the user has already written.
- Pre-declared lang libs return their own prefix rather than `""` (otherwise `string:`
  would be dropped from `string:length(s)`).

Ten node builders now ask `sourceBuilder.importQualifier()` instead of deriving a prefix
from `codedata.module()`.

#### 5. Expression editor, insertion, references

- **`ExpressionEditorContext`** — the probe is compiled against the real document, so it
  builds its own `ModulePrefixContext`, requalifies `ballerinaType`, and emits pending
  imports with the right `as` clause. Otherwise the diagnostics you see are from a
  compilation of the wrong type.
- **`ExpressionEditorService`** — a search result is inserted under the prefix the target
  file binds, resolved *lazily* per codedata kind (an eager resolve NPEs on
  `kind: CURRENT`, where `module` is null).

#### 6. Type editor and cast probe

- **`TypesManager`** — external imports now carry `as <prefix>` where the natural one is
  taken, and every generated snippet is rewritten onto those prefixes in the same pass.
- **`AmbiguousTypeCastResolver`** — reads the document's bound prefix instead of deriving
  one, for the same reason as the expression editor probe.

#### 7. Data mapper

The output type is resolved by the qualifier its text actually uses, keyed off the
imports map's own key, with a version-stripping failover for the org-less local case.

#### 8. Frontend (`packages/ballerina-extension`)

- **`to-syntax-string.ts`** — the *only* place a frontend-derived prefix is final,
  because its output goes to the LLM as library documentation and never reaches the
  language server. It got the same three-step policy, mirrored, plus a re-assignment
  across the merged set in `collectFunctionExternalLinks` (the per-type passes cannot
  see each other).
- **`rpc-manager.ts`** — unrelated to aliasing, surfaced by it: `updateImports` reported
  `importStatementOffset` from the *untrimmed* statement while sending the trimmed one,
  drifting the expression editor's cursor on every accepted completion.

---

### What deliberately did **not** change

- **No request/response model change.** The qualifier rides in the existing comma-joined
  string as `prefix=org/module`. A bare `org/module` still parses. No new field, no
  removed field, no `typeMembers`-style side channel.
- **No index schema change and no mandatory rebuild.** Only a signature naming two
  colliding modules produces a string that differs at all.
- **No frontend request-path change.** Several frontend sites still mint a prefix
  naively (`AgentDefinitionDesigner`, `AIAgentSidePanel`, `modification-utils`,
  `statement-editor`). They do not need to know about each other or about the target
  file: the backend takes the incoming prefix as a *join key* and resolves on
  `org/module`. **The frontend proposes; the backend disposes.**
- **Import set unchanged.** Only contested keys differ.

---

### Testing

New fixtures use `ballerinax/github` and `ballerinax/trigger.github` — the pair that
actually collides, both already in the offline cache. Expectations were captured from the
implementation, then the `updateConfig` hooks restored.

| Fixture | File already has | Asserts |
|---|---|---|
| `new_connection-github-existing-alias` | `trigger.github` + `github as ghClient` | reuses the existing alias; **no import emitted** |
| `new_connection-github-collision` | `trigger.github` | allocates `import ballerinax/github as github2;` + `github2:Client` |
| `github-both-imports` | nothing | **both modules decided in one operation**, before either import is written |
| `github_collision` (function_call_template) | `github` | insertion returns `triggerGithub:Listener` |

Between them these cover both allocation branches — numbered (`github2`, undotted module,
no camelCase available) and camelCase (`triggerGithub`, dotted) — and the three distinct
decision points: source generation, code generation, insertion.

~25 existing snapshots changed **import order only** (`LinkedHashMap` determinism).
120/120 node templates showed zero model drift, which is the check that the import set
really is unchanged.

Full suite last green at **2045 / 0 failures**; that run predates the final four commits,
whose targeted suites (`SourceGeneratorTest`, `NodeTemplateTest`, `FunctionCallTemplateTest`,
all `DataMap*`) are green. **A full re-run before merge is worth doing.**

---

### Known gaps / follow-ups

- **Index regeneration is deliberately out of scope** for this PR. It needs no code or
  schema change; regenerating only improves the intra-signature collision case.
- **`to-syntax-string.ts` has no test coverage.** It is the one place the allocation rule
  exists twice, and only the Java side is tested.
- **`ActivityGenerator:361`** — `needsModuleImport` still tests against the natural
  prefix. Low risk; flagged rather than changed.
- **`PARAM_FOR_TYPE_INFER`** now takes imports from `paramForTypeInfer.typeSymbol()`
  rather than the enclosing parameter's. Correct, but beyond the strict minimum —
  happy to revert if reviewers prefer a tighter diff.

## Summary

- Added shared module-prefix resolution and deterministic alias allocation.
- Prevented prefix collisions in generated models, types, expressions, references, data mappings, and frontend documentation.
- Requalified generated types, values, function calls, annotations, and configuration fields.
- Reused existing aliases and target-file imports.
- Improved handling of local, foreign, same-package, and org-less modules.
- Added import encoding, parsing, and syntax-tree prefix utilities.
- Preserved deterministic import ordering.
- Updated import offset calculation to use trimmed statements.
- Added tests for collisions, aliases, shared prefixes, import handling, expression rewriting, and annotation rendering.


packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/TypesManager.java
```diff
@@ -1042,24 +1044,33 @@ private static void addImportsToTextEdits(Map<String, String> imports, ModulePar
...
+                importStmts.add(getImportStmt(importSignature));
             }
         });
 
+        // Rewrite the generated snippets onto the prefixes decided above. This has to happen for every module that
+        // got an alias: an aliased import whose references still name the natural prefix would bind them to
+        // whichever module kept it, which is worse than not aliasing at all. requalify leaves an ambiguous natural
+        // prefix alone, and isNaturalAmbiguous is what stops such a module from being aliased in the first place.
+        if (prefixes.hasAliases()) {
+            textEdits.replaceAll(textEdit ->
```
pasindufernando1: **High — this requalify pass is a no-op for `createMultipleTypes`.**

`createMultipleTypes` (~L662) calls `addImportsToTextEdits(...)` *before* `textEdits.add(...codeSnippets...)`, so `textEdits.replaceAll(...)` here runs on an empty list.

Create two types in one request whose modules end in the same segment (`googleapis.drive` + `ai.google.drive`): the aliased `import ballerinax/ai.google.drive as aiGoogleDrive;` is emitted, but the appended snippet still says `drive:Foo` and binds to whichever module kept `drive` — exactly the silent mis-binding this comment says is worse than not aliasing.

`updateType` (~L640) and `createGraphqlClassType` (~L682) append the snippet first and are fine. Fix: move the `addImportsToTextEdits` call after the snippet is added.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/expressioneditor/ExpressionEditorContext.java
```diff
@@ -209,21 +211,26 @@ public LineRange generateStatement(String stmtStart, String stmtEnd) {
...
-            Map<String, String> imports = property.importStatements();
-            if (imports != null && !imports.isEmpty()) {
-                for (String importStmt : imports.values()) {
-                    Optional<TextEdit> textEdit = getImport(importStmt.split(":")[0]);
-                    if (textEdit.isPresent()) {
-                        textEdits.add(textEdit.get());
-                        lineOffset++;
-                    }
+            // Add the import statements of the dependent types, each under the prefix resolved above.
+            for (Map.Entry<String, String> pending : prefixes.pendingImports().entrySet()) {
```
pasindufernando1: **High — org-less pending import produces `import /pkg.sub;`.**

`ModulePrefixContext.prefixFor` keys `pendingImports` as `(org == null ? "" : org) + "/" + module`. `CommonUtils.getImportStatements` emits a bare `pkg.sub` for a module of the current package, so the key becomes `/pkg.sub`. `withAliasClause("/pkg.sub", "sub")` returns it unchanged (natural prefix, no `as`) and `getImport` builds `import /pkg.sub;`.

The probe document then fails to parse, so every diagnostic the user sees comes from the malformed probe. The replaced code did `getImport(importStmt.split(":")[0])` → `import pkg.sub;`. Reachable whenever a property's type depends on a local submodule the file hasn't imported yet.

Fix: strip a leading `/`, or emit from the map's value rather than the composite key.

packages/ballerina-extension/src/features/ai/utils/libs/to-syntax-string.ts
```diff
@@ -441,14 +488,17 @@ function collectFunctionExternalLinks(params: Parameter[], returnType?: Type): E
...
         const key = `${l.recordName}::${l.libraryName}`;
         if (seen.has(key)) {
             return false;
         }
         seen.add(key);
         return true;
     });
+    // Reassign across the merged set: the per-type passes above cannot see each other, so two libraries
+    // sharing a natural prefix would each have kept it.
+    return assignModulePrefixes(distinct);
```
pasindufernando1, line 609: **Medium — the merged reassignment never reaches parameter rendering.**

`renderParam` (~L508) still calls `collectExternalLinks(param.type)` itself, so this merged result is only used for the return type and `buildSpecialAgentNote`.

For `foo(googleapis.drive:File a, ai.google.drive:TextDataLoader b)` both params render as `drive:` (each per-type pass keeps the natural prefix) while the return type uses `aiGoogleDrive` — the LLM sees one qualifier standing for two modules, the bug this hunk is meant to close.

Fix: pass `allExternalLinks` into `renderParam` and use it.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/FunctionDataBuilder.java
```diff
@@ -1263,6 +1281,41 @@ private boolean isSameAsUserModule() {
...
+        Map<String, String> imports = new LinkedHashMap<>();
+        for (String entry : importStatements.split(",")) {
+            String signature = entry.trim().split(":")[0];
+            if (signature.isEmpty() || imports.containsValue(signature)) {
+                continue;
+            }
+            String qualifier = qualifierBySignature.get(signature);
+            String module = signature.contains("/")
+                    ? signature.substring(signature.indexOf('/') + 1) : signature;
+            imports.put(qualifier != null ? qualifier
```
pasindufernando1: **Medium — unguarded `put` can displace a module's import.**

`imports.put(qualifier, signature)` overwrites whatever already holds that key. The type text is rendered with `ignoreError=true` (error members dropped), but `getImportStatements`/`analyzeTypeSymbolForImports` is not — so for `github:Client|trigger.github:SomeError` this map holds only `ballerinax/github -> "github"` while `importStatements` lists both modules.

`importStatements` comes from a `HashSet` join, so in the order where `ballerinax/trigger.github` is processed first it takes key `github` via the fallback and is then overwritten by `ballerinax/github` — one module's import is lost, the exact failure this function exists to prevent.

Fix: fall back to `allocatePrefix` when the key is already held by a different signature.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/ImportPrefixReader.java
```diff
@@ -0,0 +1,208 @@
...
+        boolean anyOrg = org == null || org.isBlank();
+        for (ImportDeclarationNode importDeclarationNode : node.imports()) {
+            String moduleName = moduleNameOf(importDeclarationNode);
+            if (!module.equals(moduleName)) {
+                continue;
+            }
+            // An import written without an organization is a module of the file's own package, and its module name
+            // identifies it within the file on its own. The model records such a module with the organization, so
+            // matching strictly on it would miss the very import that already binds the prefix.
+            if (anyOrg || importDeclarationNode.orgName().isEmpty()
```
pasindufernando1: **Medium — an org-less import now matches any organization.**

For `org=ballerinax, module=ai.google`, a file carrying the same-package `import ai.google;` is reported as already binding that module: `prefixFor` returns `google`, records no pending import and adds no alias. But `SourceBuilder.removeExistingImports` / `CommonUtils.importExists` are org-aware and still emit `import ballerinax/ai.google;`.

Two imports then bind `google` → BCE2008. Before this relaxation the external module would have been aliased to `aiGoogle`.

Fix: limit the org-less match to the case where the requested org equals the current package's org.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
```diff
@@ -549,31 +551,32 @@ public Metadata.Builder<Builder<T>> metadata() {
...
-                    if (pathParts.length < 2) {
-                        continue;
-                    }
-                    String packageName = pathParts[1];
-                    this.imports.put(CommonUtils.getDefaultModulePrefix(packageName), importStatement);
-                }
+                // A key already standing for a different module gets a free one rather than displacing it, which
+                // would drop that module's import.
+                this.imports.put(existing == null ? entry.getKey()
+                        : ModuleAliasResolver.allocatePrefix(entry.getKey(), this.imports.keySet()),
```
pasindufernando1: **Low — `allocatePrefix` is handed the qualifier, not the module name.**

`entry.getKey()` is the qualifier, so `defaultAlias` is a no-op and the loser of a key collision always gets `drive2` instead of the policy alias `aiGoogleDrive` that `TypeQualifierAllocator` and `ModulePrefixContext` pick for the same module.

Worse, the re-keyed qualifier no longer matches the qualifier the property's type *text* uses, so `requalifyAuthored` maps the text's `drive:` onto the first module — a silent mis-binding in place of the dropped import.

Fix: derive the new key from the entry's module name so all three allocators agree.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/SourceBuilder.java
```diff
@@ -347,32 +353,158 @@ public SourceBuilder acceptImport() {
...
+     * <p>
+     * A node read back from source is left alone. Its values were copied out of the file and so already use whatever
+     * that file binds, while its imports map is rebuilt from the model -- so the map's qualifiers and the value's
+     * qualifiers are not the same vocabulary, and rewriting by them would corrupt text that is already correct.
+     * </p>
+     *
+     * @param property the property whose value is being written
+     * @return the value to write
+     */
+    public String requalifiedValue(Property property) {
```
pasindufernando1, line 501: **Low — `requalify` rewrites prefixes inside string literals.**

`ModuleAliasResolver.requalify` rewrites every `(?<![\\w.])prefix(?=:)` occurrence with no awareness of string literals, and `requalifiedValue` applies it to every property value of a new node — including text the user typed into the form.

With an imports entry `http -> ballerina/http` and a target file that already binds `http` to another module, the value `"http://localhost:9090"` is written out as `"http2://localhost:9090"`.

Fix: restrict the substitution to identifier positions outside string literals.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/TypesManager.java
```diff
@@ -1042,24 +1044,33 @@ private static void addImportsToTextEdits(Map<String, String> imports, ModulePar
                     return;
```
pasindufernando1: **Low — the same-package branch returns without registering the module.**

It adds `import pkg.github;` and returns without registering the module with `prefixes`, so the prefix that org-less import will bind never enters `claimed`.

If the same request also needs `ballerinax/github`, that external module is judged to have a free natural prefix and both imports are emitted unaliased, both binding `github`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/main/java/io/ballerina/flowmodelgenerator/extension/ExpressionEditorService.java
```diff
@@ -292,7 +314,38 @@ private void applyModuleImport(String filePathString, String moduleId, String im
...
+        try {
+            Path filePath = Path.of(filePathString);
+            PackageUtil.loadProject(this.workspaceManagerProxy.get(), filePath);
+            Optional<Document> document = this.workspaceManagerProxy.get().document(filePath);
+            ModulePartNode rootNode = document
+                    .map(doc -> doc.syntaxTree().rootNode() instanceof ModulePartNode node ? node : null)
+                    .orElse(null);
+            // A null root falls back to the module's natural prefix, which is what this method promises without a
+            // file to read. An import is emitted for the result, so allocating a free prefix is the right answer.
+            return ImportPrefixReader.resolve(rootNode, codedata.org(), codedata.module(), null);
```
pasindufernando1: **Low — allocating a prefix on a path that writes no import.**

The `IMPORTED` branch emits no import, but `boundPrefix` calls `ImportPrefixReader.resolve`, which *allocates* a fresh prefix when the file doesn't import the module — contradicting that method's own contract ("for a caller that is going to write the import"). The inserted template then names an unbound prefix (`github2:Sym`) in the expression editor.

Fix: use `ImportPrefixReader.boundPrefix`, which never allocates.

pasindufernando1 (commented):
Code review of the import-aliasing work. 9 findings inline: 2 high, 3 medium, 4 low. All are prefix-collision cases where the aliasing path silently mis-binds a prefix or drops an import.

Also noted, not filed inline: `TypeQualifierAllocator.imports()`, `TypeQualifierAllocator.hasReallocation()` and `ModulePrefixContext.resolvedPrefixFor(...)` have no production caller; `ModulePrefixContext.pendingImports()` returns `Map.copyOf`, discarding the `LinkedHashMap` ordering relied on elsewhere for deterministic import output (harmless in its single current consumer).

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/ModulePrefixContext.java
```diff
@@ -129,6 +154,87 @@ public String resolveNatural(String naturalPrefix) {
...
+     * </p>
+     *
+     * @param org    the organization name; blank for a module of the current package
+     * @param module the module name
+     * @return {@code org/module}, with {@code as <prefix>} appended where the prefix is a followable rename
+     */
+    public String importSignatureFor(String org, String module) {
+        String prefix = prefixFor(org, module);
+        String signature = org == null || org.isBlank() ? module : org + "/" + module;
+        return isNaturalAmbiguous(ModuleAliasResolver.selfPrefix(module))
```
NipunaRanasinghe: This gate means the TypesManager path doesn't actually fix the collision — it just avoids making it worse.

`isNaturalAmbiguous` is true exactly when two registered modules share a natural segment, which is the headline scenario. So if a type-editor operation names both `googleapis.drive` and `ai.google.drive`, neither gets an `as` clause, both are emitted bare, and the redeclared-symbol error is still there.

I agree with the reasoning in the javadoc — an alias you can't follow through is worse than a compile error. Two things though:

1. The PR description says "external imports now carry `as <prefix>` where the natural one is taken". That only holds when the *target file* already binds the prefix, not when the collision happens inside one operation. Worth rewording so nobody later assumes this path is covered.
2. There's no fixture for a TypesManager collision. All four new fixtures go through SourceBuilder/insertion, which rewrite by authored qualifier and so do alias correctly. A test pinning the current TypesManager behaviour would make this gap visible instead of leaving it latent.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/ModulePrefixContext.java
```diff
@@ -129,6 +154,87 @@ public String resolveNatural(String naturalPrefix) {
...
+                ? signature
+                : ModuleAliasResolver.withAliasClause(signature, prefix);
+    }
+
     /** The modules that still need an import statement, as {@code org/module} -> resolved prefix. */
     public Map<String, String> pendingImports() {
         return Map.copyOf(pendingImports);
```
NipunaRanasinghe: `Map.copyOf` throws away the insertion order this map was built to keep. `pendingImports` is a `LinkedHashMap`, but `Map.copyOf` returns an immutable map with unspecified iteration order.

The only consumer, `ExpressionEditorContext.generateStatement`, iterates this to emit import text edits, so the emitted order ends up arbitrary. That sits oddly against §3 of the description, which switched `Property.imports` to `LinkedHashMap` specifically for deterministic import order, and against the ~25 snapshots rebaselined for it.

```java
return Collections.unmodifiableMap(new LinkedHashMap<>(pendingImports));
```

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/TypeQualifierAllocator.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /**
+     * The modules named in the signature, as {@code qualifier -> org/module}. Every key occurs in the rendered text,
+     * and no two keys are the same, so the text and this map together identify every type the signature names.
+     *
+     * @return the accumulated imports, empty when the signature named no module outside the current one
+     */
+    public Map<String, String> imports() {
+        return Map.copyOf(importByQualifier);
```
NipunaRanasinghe: Same as in `ModulePrefixContext.pendingImports()` — `importByQualifier` is a `LinkedHashMap` but `Map.copyOf` doesn't preserve its order. Use `Collections.unmodifiableMap(new LinkedHashMap<>(importByQualifier))` if the ordering is meant to survive.

Separately: this class has no unit tests, and neither do `CommonUtils.encodeImportStatements` / `parseImportStatements`. That pair is the PR's backward-compatibility contract — "a bare `org/module` still parses", no index rebuild needed. A round-trip test plus one asserting a legacy bare entry still parses would protect the claim the PR leans on most. `model-generator-commons` already has a test source set and a working `testng.xml`, so there's somewhere to put them.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/SourceBuilder.java
NipunaRanasinghe, line 501: This widens the regex rewrite from types to arbitrary values, which is a bigger surface than it looks.

`ModuleAliasResolver.requalify` is a plain textual substitution: `(?<![\w.])(prefix)(?=:)`. On type text that's fine, since every `x:` there is a qualifier. But values also contain `ident:` as a mapping-constructor field key, so with `drive` aliased, a value of `{drive: 1}` becomes `{aiGoogleDrive: 1}`, and a string literal containing `drive:` gets rewritten too.

It's bounded — only fires when an alias is actually in effect, so only in collision cases — but nothing scopes the rewrite to qualifier positions, and this PR is what expands it from types to values. A test pinning `{drive: 1}` and `"drive:x"` would at least make the trade-off explicit.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
```diff
@@ -549,31 +551,32 @@ public Metadata.Builder<Builder<T>> metadata() {
...
-            String[] importList = importStatements.split(",");
-            for (String importStatement : importList) {
-                if (importStatement.trim().isEmpty()) {
-                    continue;
-                }
-                String[] parts = importStatement.split(":");
-                if (parts.length < 1) {
+            for (Map.Entry<String, String> entry : parsed.entrySet()) {
+                String existing = this.imports.get(entry.getKey());
+                if (Objects.equals(existing, entry.getValue()) || this.imports.containsValue(entry.getValue())) {
```
NipunaRanasinghe, line 569: The `containsValue` arm can drop a qualifier the text actually uses.

It skips an entry whose module is already in the map under a *different* key. If the incoming type text is qualified with the new key, the map no longer holds that key, and the lookups keyed on the text's qualifier — `SourceBuilder.getExpressionBodyText` and `requalifyAuthored` — miss it.

Narrow: it needs two `imports(...)` calls that disagree on one module's qualifier. But it's the same failure this PR set out to fix, a module's entry disappearing from a prefix-keyed map. If the intent is "first key wins, later duplicates are redundant", a comment saying so would settle it. Otherwise the later key should map to the same module rather than being dropped.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/test/java/io/ballerina/servicemodelgenerator/extension/util/ModuleAliasResolverTest.java
```diff
@@ -20,6 +20,9 @@
 
 import io.ballerina.compiler.syntax.tree.ModulePartNode;
 import io.ballerina.compiler.syntax.tree.SyntaxTree;
+import io.ballerina.modelgenerator.commons.ImportPrefixReader;
```
NipunaRanasinghe: Every class this file tests has now moved to `model-generator-commons`, but the test itself stayed in `service-model-generator`, so the test module reaches past its own boundary to test its dependency.

Moving it into the commons test source set would fix the layering and give the new classes an obvious home — `TypeQualifierAllocator` and `ModulePrefixContext.requalifyAuthored` currently have no unit tests, and `requalifyAuthored` is the rewrite every `SourceBuilder` site now depends on.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/Utils.java
```diff
@@ -1447,74 +1448,6 @@ public static boolean importExists(ModulePartNode node, String org, String modul
         });
     }
```
NipunaRanasinghe, line 1450: Nit: double blank line left behind where the moved methods were removed.

packages/ballerina-extension/src/features/ai/utils/libs/to-syntax-string.ts
samithkavishke: Addressed

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/expressioneditor/ExpressionEditorContext.java
samithkavishke: Addressed

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/FunctionDataBuilder.java
samithkavishke: Addressed

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
samithkavishke: Addressed

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/ImportPrefixReader.java
samithkavishke: Addressed

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/TypesManager.java
samithkavishke: Addressed

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/main/java/io/ballerina/flowmodelgenerator/extension/ExpressionEditorService.java
samithkavishke: Addressed

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/ModulePrefixContext.java
```diff
@@ -0,0 +1,317 @@
...
+        Set<String> contestedNaturals = new HashSet<>();
+        for (Map.Entry<String, String> entry : importsByAuthored.entrySet()) {
+            // A version tail is dropped, and an entry with no organization is a module of the current package.
+            String[] parts = entry.getValue().split(":")[0].split("/", 2);
+            String org = parts.length == 2 ? parts[0] : "";
+            String module = parts.length == 2 ? parts[1] : parts[0];
+            if (module.isEmpty()) {
+                continue;
+            }
+            String resolved = prefixFor(org, module);
```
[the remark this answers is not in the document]
samithkavishke: Addressed

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/util/Utils.java
samithkavishke: Addressed

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/test/java/io/ballerina/servicemodelgenerator/extension/util/ModuleAliasResolverTest.java
samithkavishke: Addressed

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
samithkavishke: Addressed

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/ModulePrefixContext.java
samithkavishke: Addressed

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/ModulePrefixContext.java
samithkavishke: Addressed

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/TypeQualifierAllocator.java
samithkavishke: Addressed

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/SourceBuilder.java
samithkavishke: Addressed

samithkavishke: Addressed

packages/ballerina-extension/src/features/ai/utils/libs/to-syntax-string.ts
```diff
@@ -441,21 +488,28 @@ function collectFunctionExternalLinks(params: Parameter[], returnType?: Type): E
...
+ *
+ * `externalLinks` must be the whole signature's, from {@link collectFunctionExternalLinks}. Collecting them
+ * per parameter gives each one its own empty `taken` set, so two libraries whose names end in the same segment
+ * would both keep that segment — and the return type, which does use the merged set, would then disagree with
+ * the parameters about which module the one qualifier stands for.
  */
-function renderParam(param: Parameter): string {
-    const externalLinks = collectExternalLinks(param.type);
+function renderParam(param: Parameter, externalLinks: ExternalLinkInfo[]): string {
     const typeName = applyPrefixToTypeName(param.type.name, externalLinks);
```
RNViththagan: Passing the whole signature's links here does fix the aliasing, but `applyPrefixToTypeName` matches on the bare record name (`\bFile\b`), so every link in the merged set now runs over every parameter. When the two colliding libraries share a record name, the second replacement lands inside the first one's output:

```
foo(googleapis.drive:File a, ai.google.drive:File b)
-> function foo(drive:aiGoogleDrive:File a, drive:aiGoogleDrive:File b)
```

On staging this came out as `drive:File a, drive:File b` — ambiguous, but valid. (The return type already had this hole: `drive:drive:File`.)

Shall we keep the merged allocation but apply only the type's own links — filter `externalLinks` to the `recordName`/`libraryName` pairs in `param.type.links` before the replace, and the same for the return type? Making the regex skip an already-qualified name, like `qualifyDeclaredType` does below, would be a cheap second guard.

packages/ballerina-extension/src/features/ai/utils/libs/to-syntax-string.ts
```diff
@@ -113,6 +113,55 @@ export function deriveModulePrefix(libraryName: string): string {
...
+}
+
+/**
+ * Gives every distinct library its own prefix.
+ *
+ * Ballerina derives an unaliased import's prefix from the module name's last dot-segment, so two libraries
+ * whose names end in the same segment - "ballerinax/googleapis.drive" and "x/ai.google.drive" - both derive
+ * `drive`. Rendering both records as `drive:Name` names one module twice and would not compile. Distinct
+ * libraries therefore fall back to the camelCase alias and then to a numbered suffix, matching what the
+ * language server emits for a colliding import.
```
RNViththagan, line 138: One thing the alias still needs is a way to reach the model. The Special Agent Note only says `TextDataLoader FROM ai.google.drive package`, and the system prompt (`agent/prompts.ts:177`) tells the model to alias dotted packages by the last segment — `import org/package.one as one;`. So from

```
function foo(drive:File a, aiGoogleDrive:TextDataLoader b); // Special Agent Note: File FROM googleapis.drive package, TextDataLoader FROM ai.google.drive package
```

it will write `import ballerinax/ai.google.drive as drive;` and `aiGoogleDrive:` is unbound. Since this output never reaches the LS, the note is the only channel.

Shall we have the note carry the alias when it differs from the natural one — `TextDataLoader FROM ai.google.drive package (import as aiGoogleDrive)` — or emit a `// Requires: import ballerinax/ai.google.drive as aiGoogleDrive;` line the way the annotation directives already do?

packages/ballerina-extension/src/features/ai/utils/libs/to-syntax-string.ts
```diff
@@ -126,15 +175,13 @@ function collectExternalLinks(type: Type): ExternalLinkInfo[] {
...
-    return type.links
-        .filter((link): link is Link & { libraryName: string } =>
-            link.category === "external" && !!link.libraryName
-        )
-        .map((link) => ({
-            recordName: link.recordName,
-            libraryName: link.libraryName,
-            modulePrefix: deriveModulePrefix(link.libraryName),
-        }));
+    return assignModulePrefixes(
```
RNViththagan, line 246: This per-type pass is also what `renderRecord` uses, once per field, so a record's fields have the same gap Pasindu raised for the parameters:

```
type Cfg record {
    drive:File f1;           // File FROM googleapis.drive package
    drive:TextDataLoader f2; // TextDataLoader FROM ai.google.drive package
};
```

Shall we collect over all fields once, like `collectFunctionExternalLinks`, and render each field from the shared set?

packages/ballerina-extension/src/features/ai/utils/libs/to-syntax-string.ts
```diff
@@ -441,21 +488,28 @@ function collectFunctionExternalLinks(params: Parameter[], returnType?: Type): E
...
-    return links.filter((l) => {
+    const distinct = links.filter((l) => {
         const key = `${l.recordName}::${l.libraryName}`;
         if (seen.has(key)) {
             return false;
         }
         seen.add(key);
         return true;
     });
+    // Reassign across the merged set: the per-type passes above cannot see each other, so two libraries
```
RNViththagan, line 607: Not blocking, just a question. Allocation is per signature, so one library can be `drive` in `foo(...)` and `aiGoogleDrive` in `bar(...)` a few lines apart in the same document. For an LLM reader a per-document (or per-library) allocation would be simpler and consistent, and the import hint could then be stated once per library. Was per-signature a deliberate mirror of the Java side?

packages/ballerina-extension/src/test-support/toSyntaxStringModulePrefix.test.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/** The rendered `function foo(...)` line. */
+function signatureOf(library: Library): string {
+    const line = toSyntaxString([library]).split("\n").find((l) => l.startsWith("function foo("));
+    expect(line).toBeDefined();
+    return line as string;
+}
+
+describe("toSyntaxString module prefixes", () => {
```
RNViththagan, line 83: Two cases worth pinning here, since this is the one place the policy exists twice: a same-named record in two colliding libraries (the double-prefix case above), and the numbered-suffix path CodeRabbit asked for — `ballerinax/github` + `ballerinax/trigger.github` → `github`, `triggerGithub`; two undotted modules → `github2`.

packages/ballerina-extension/src/features/ai/utils/libs/to-syntax-string.ts
samithkavishke: Addressed

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/SourceCodeGenerator.java
```diff
@@ -451,25 +509,28 @@ private String generateDocs(String docs, String indent) {
...
+                paramName = paramName + "?";
+            }
 
-        // Annotation and type descriptor
-        List<AnnotationAttachment> copyOfAnnotAttachments = getAnnotationAttachments(member);
-        String annotatedTypeDesc = generateAnnotatedTypeDescriptor(member.type(), copyOfAnnotAttachments);
+            // Default value
+            String defaultValue =
+                    (withDefaultValue && member.defaultValue() != null && !member.defaultValue().isEmpty())
+                            ? " = " + member.defaultValue() : "";
```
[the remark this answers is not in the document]
samithkavishke: Addressed

packages/ballerina-extension/src/test-support/toSyntaxStringModulePrefix.test.ts
samithkavishke: Addressed

packages/ballerina-extension/src/features/ai/utils/libs/to-syntax-string.ts
samithkavishke: Addressed

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/FunctionDataBuilder.java
```diff
@@ -880,38 +887,41 @@ private Map<String, ParameterData> getParameters(ParameterSymbol paramSymbol,
...
-            paramType = getTypeSignature(typeSymbol);
+            paramType = getTypeSignature(typeSymbol, false, paramAllocator);
             placeholder = DefaultValueGeneratorUtil.getDefaultValueForType(typeSymbol);
             optional = false;
         } else {
             if (paramForTypeInfer != null) {
                 if (paramForTypeInfer.paramName().equals(paramName)) {
+                    // The inferred type's text is rendered elsewhere, so this parameter's imports are the ones
+                    // its own type symbol names.
+                    String inferredImports = getImportStatements(paramForTypeInfer.typeSymbol());
```
NipunaRanasinghe: **Medium — CONFIRMED. `PARAM_FOR_TYPE_INFER`'s text and its imports are still built through two different, un-reconciled paths.**

`paramForTypeInfer.type()` (used as `paramType` two lines below) comes from `CommonUtils.getTypeSignature(semanticModel, ..., true)` at line 657 — the 3-arg overload, which forwards `allocator = null` (traced through `CommonUtils.java:236→200→186`), so it renders every module qualifier under its bare natural segment. `inferredImports` on this line comes from the plain symbol-walk (`getImportStatements`), never through `qualifiedImports`/`TypeQualifierAllocator`. Every sibling branch in this same method (`REQUIRED` at line 898, and others) threads a `paramAllocator` through both the text and the imports so the two agree — this branch is the one left out. If the inferred type itself unions two modules colliding on their natural segment, the rendered text collapses both to the same qualifier while the parsed imports would allocate distinct keys for the two `org/module` entries: text and imports disagree, which is the "one qualifier, two modules" failure this PR otherwise fixes.

packages/ballerina-extension/src/features/ai/utils/libs/to-syntax-string.ts
```diff
@@ -113,6 +113,121 @@ export function deriveModulePrefix(libraryName: string): string {
...
+ *
+ * A library keeps its natural prefix whenever nothing else has claimed it, so single-library output is
+ * unchanged.
+ */
+function assignModulePrefixes(links: { recordName: string; libraryName: string }[]): ExternalLinkInfo[] {
+    if (documentPrefixes) {
+        // A document-wide allocation is in effect, so the answer is already decided; see `allocateForDocument`.
+        // Falling back to a local allocation for a library the pre-pass somehow missed keeps this total.
+        const resolved = links.map((link) => ({ ...link, modulePrefix: documentPrefixes!.get(link.libraryName) }));
+        if (resolved.every((link) => link.modulePrefix !== undefined)) {
```
NipunaRanasinghe, line 148: **Low — CONFIRMED. The pre-pass fallback isn't itself collision-safe against the document-wide allocation it's a fallback for.**

When `documentPrefixes` is set but some link in this call isn't found in it (line 148 false), execution falls through to the local allocation loop below with a fresh, empty `taken` set — not seeded from `documentPrefixes.values()`. If that fallback ever fires, the locally-derived natural prefix could collide with a prefix `documentPrefixes` already assigned to a different module elsewhere in the same document — the same collision bug, inside the code meant to be its own safety net. The comment at line 145-146 frames this as a defensive case for a library the pre-pass "somehow missed," so it may be rare, but seeding `taken` from `documentPrefixes.values()` would close it.

packages/ballerina-extension/src/test-support/toSyntaxStringModulePrefix.test.ts
```diff
@@ -0,0 +1,263 @@
...
+        expect(signature).toContain("drive:File a");
+        expect(signature).toContain("aiGoogleDrive:TextDataLoader b");
+        expect(signature).toContain("returns aiGoogleDrive:TextDataLoader");
+
+        // The decisive property: whatever the qualifiers are, no two libraries may share one.
+        const qualifiers = [...signature.matchAll(/(\w+):(File|TextDataLoader)/g)].map((m) => m[1]);
+        expect(new Set(qualifiers).size).toBe(2);
+    });
+
+    it("agrees between a parameter and the return type naming the same library", () => {
```
NipunaRanasinghe: **Low — CONFIRMED. This test would pass under the pre-fix allocator too.**

`deriveModulePrefix` is a pure function of `libraryName` alone, and both the parameter and the return type here reference the identical library (`"googleapis.drive"`). Asserting `paramQualifier === returnQualifier` holds by construction — it would be true even with no allocator, no `taken` set, and no cross-pass reconciliation, since both sites derive from the same one input. The property this test's own name claims to check (agreement between two sites) can't distinguish fixed from broken allocation; the real regression — two *different* libraries colliding — is already covered more strongly by the preceding test with exact expected strings.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/expressioneditor/ExpressionEditorContext.java
```diff
@@ -209,21 +210,28 @@ public LineRange generateStatement(String stmtStart, String stmtEnd) {
...
 
         if (property != null) {
-            // Append the type if exists
-            String ballerinaType = property.propertyType().ballerinaType();
+            // The statement is compiled against this document, so the type has to name its modules by the prefixes
+            // this document binds -- not by the qualifiers the model authored it with. Two dependent modules whose
+            // names end in the same segment, ballerinax/github and ballerinax/trigger.github, are both authored as
+            // `github:`; imported verbatim they redeclare that prefix and every diagnostic reported from this
+            // statement is an artefact of the probe rather than of the user's expression.
+            ModulePrefixContext prefixes = ModulePrefixContext.from(documentContext.document()
```
NipunaRanasinghe, line 219: **Medium — PLAUSIBLE, not fully confirmed. This is the one caller of `ModulePrefixContext.from` that doesn't thread the current module.**

`ModulePrefixContext.from(rootNode)` — the single-arg factory — classifies every module as `EXTERNAL` per its own javadoc (`ModulePrefixContext.java:70-72`), since `classify()` needs a non-null `currentModule` to ever return `SAME_MODULE`. Every other caller of this class in the diff (`TypesManager`, `AmbiguousTypeCastResolver`) passes `ModuleInfo.from(...)` instead. If `property.importStatements()` can ever carry an entry for the file's own module here, this path would register a spurious import and requalify a same-module reference away from empty — turning a reference that needs no qualifier into one that needs an alias. I didn't trace how the LS index populates `importStatements()` for this specific probe to confirm it can actually happen, so flagging as worth a second look rather than asserting it's live.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/FunctionDataBuilder.java
samithkavishke: Addressed

packages/ballerina-extension/src/test-support/toSyntaxStringModulePrefix.test.ts
samithkavishke: Addressed

packages/ballerina-extension/src/features/ai/utils/libs/to-syntax-string.ts
samithkavishke: Addressed

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/expressioneditor/ExpressionEditorContext.java
samithkavishke: Addressed, same-module entry  no longer gets a spurious self-import/alias.

NipunaRanasinghe (dismissed):

NipunaRanasinghe (approved):

PR #992, opened 2026-09-03
Add diagnostics hints and prompt updates for lock statements

SasinduDilshara:
Add diagnostics hints and prompt updates for lock statements

FIxes https://github.com/wso2/product-integrator/issues/2365

## Summary

- Added concurrency-safety rules to the AI agent system prompt.
- Added diagnostic hints for isolation, lock, and service-concurrency diagnostics.
- Refactored diagnostic processing to separate errors from non-error concurrency hints.
- Updated tool descriptions and result types to expose concurrency hints.
- Added tests for prompt integration, diagnostic mappings, routing, and hint formatting.

### DIFF
--- a/packages/ballerina-extension/src/features/ai/agent/tools/diagnostic-hints.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/tools/diagnostic-hints.ts
@@ -1,0 +1,297 @@
+import type { DiagnosticEntry, Diagnostics } from '@wso2/ballerina-core';
+import * as path from 'path';
+
+/**
+ * Diagnostic entry enriched with a resolving hint
+ */
+export interface EnrichedDiagnostic extends DiagnosticEntry {
+    hint?: string;
+}
+
+/**
+ * Service-concurrency hint codes (severity: Hint, not Error).
+ *
+ * These are emitted by the compiler's IsolationAnalyzer when a service's
+ * resource/remote methods cannot be dispatched concurrently because the
+ * service and/or the method could not be declared or inferred `isolated`.
+ * The code still compiles, so they must never be counted as compilation
+ * errors — they are surfaced separately as `concurrencyHints`.
+ */
+export const CONCURRENCY_HINT_CODES: ReadonlySet<string> = new Set([
+    "BCH2003", // concurrent calls will not be made to this method since the service and the method are not 'isolated'
+    "BCH2004", // concurrent calls will not be made to this method since the service is not an 'isolated' service
+    "BCH2005", // concurrent calls will not be made to this method since the method is not an 'isolated' method
+]);
+
+const SERVICE_ISOLATION_FIX =
+    "To enable concurrent dispatch, make the service isolated: (1) declare every mutable service field `private` " +
+    "and access it via `self` only inside `lock { }` blocks (clone values crossing the lock boundary with " +
+    "`.clone()`/`.cloneReadOnly()`); (2) ensure field initializers are isolated expressions (fresh literals/constructors); " +
+    "(3) ensure each resource/remote method only calls `isolated` functions and accesses no non-final module-level mutable state. " +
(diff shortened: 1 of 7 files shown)


packages/ballerina-extension/src/features/ai/agent/tools/diagnostics-utils.ts
```diff
@@ -110,7 +60,7 @@ export async function checkCompilationErrors(
             // `ballerinax/.config` — omitting "client". As a workaround, we detect this and
             // instruct the agent to use the correct quoted form `import ballerinax/'client.config;`
             // instead of attempting to resolve the dependency automatically.
-            const enrichedDiagnosticsTry = transformDiagnosticsToEnriched(diagnostics);
+            const enrichedDiagnosticsTry = transformDiagnostics(diagnostics).errors;
```
RNViththagan, line 63: Small one: the early return just below goes out before `concurrencyHints` and the hint note are computed, so any BCH hints in the same run are dropped for that turn. Not a big deal since the agent has to fix the import and call the tool again anyway, but shall we pass them through here too so both return paths look the same?

packages/ballerina-extension/src/features/ai/agent/tools/diagnostics.ts
```diff
@@ -45,7 +45,8 @@ For single-package projects, omit the packagePath parameter.
 
 The tool analyzes the entire Ballerina package and returns:
 - Compilation errors with file location, error message, and diagnostic code
-- Resolving hints for common error codes to help fix issues quickly
+- Resolving hints for common error codes (including isolation/lock errors like BCE3943-BCE3966) to help fix issues quickly
+- A separate concurrencyHints list (BCH2003-BCH2005) when service methods will not be dispatched concurrently — these are NOT errors; only act on them if the task involves concurrency or shared state
```
RNViththagan, line 49: Related question: `AgentExecutor.handleStreamFinish` (around line 852) still forwards only `finalDiagnostics.diagnostics` in the end-of-turn `diagnostics` event, so the new `concurrencyHints` never reach that event. If it's meant to stay errors-only that's totally fine, just want to make sure it's a deliberate choice rather than a miss.

packages/ballerina-extension/src/test-support/isolationDiagnosticHints.test.ts
```diff
@@ -0,0 +1,320 @@
+/**
```
RNViththagan, line 1: CodeRabbit's header note applies to the two test files as well. `activeFileContext.test.ts` next door has the 2026 one to copy.

packages/ballerina-extension/src/features/ai/agent/tools/diagnostic-hints.ts
```diff
@@ -0,0 +1,297 @@
...
+ * guessing at the compiler's isolation/lock terminology.
+ *
+ * The isolation/lock/concurrency entries were derived from the ballerina-lang
+ * compiler sources (IsolationAnalyzer.java, DiagnosticErrorCode.java,
+ * compiler.properties) and the language server's quick-fix code actions
+ * (AddIsolatedQualifierCodeAction, AddLockCodeAction, AddReadonlyCodeAction,
+ * CloneValueCodeAction, ConvertToReadonlyCloneCodeAction). The comment above each
+ * entry quotes the compiler's message template for that code.
+ */
+export const DIAGNOSTIC_HINTS: Readonly<Record<string, string>> = {
```
RNViththagan, line 49: Not for this PR, just thinking ahead: this catalog is a hand-copied mirror of the compiler's codes and messages, and nothing here will tell us when ballerina-lang renumbers or rewords one (the tests check the catalog against itself). Shall we at least note which ballerina-lang version it was derived from, so whoever bumps the LS knows to re-check? Longer term, the LS code actions (`AddIsolatedQualifier` / `AddLock` / `CloneValue`) already know how to fix each code and could be the source of truth instead.

PR #1003, opened 2026-09-04
Fix imported modules missing from type, function and all-kinds search

dulajdilshan:
## Purpose

Fixes https://github.com/wso2/product-integrator/issues/2172

ballerina-lang-server PR: https://github.com/ballerina-platform/ballerina-language-server/pull/994

image

image

Type search only ever returned rows from the bundled search-index.sqlite, so a connector published too recently to have been indexed yielded nothing at all in the Type Browser. Dependency collection also scanned a single module, and the indexed lookup itself was organization-blind and effectively unpaginated.

All of the following produce the same user-visible symptom - "a type from a module I imported doesn't show up" - so they are fixed together:

  - Fall back to the compiled semantic model for imported modules absent from
    the index, ranking every missing module's matches together before paging so
    a strong match isn't pushed to a later page by the module visit order.
  - Collect dependencies from every module of the package rather than only the
    one the request position sits in, so a connector imported by a submodule is
    discoverable.
  - Identify modules by organization and name together (ModuleCoordinate).
    Package.name has no uniqueness constraint - the shipped index holds
    ballerina/np vs ballerinax/np and xlibb/solace vs ballerinax/solace - so
    name-only matching merged unrelated vendors' rows and could report an
    unindexed module as indexed, suppressing the fallback entirely.
  - Share the pagination window fairly across modules using per-module quotas
    instead of a bare LIMIT/OFFSET with no ORDER BY, which let one large module
    crowd every other import off the page.
  - Clamp limit and offset at every database entry point that binds them;
    SQLite treats a negative LIMIT as unlimited, and SearchCommand.parseIntParam
    does a bare Integer.parseInt with no bounds check, so a negative limit from
    the client reached SQLite intact and returned the whole matched set at once.
  - Drain the text-query path as three pools with stable capacities - imported
    and indexed, the rest of the index, then live-compiled - so a page boundary
    can't permanently drop rows once the indexed pool runs dry.
  - Give every search ordering a deterministic tie-breaker, organization
    included, so ties don't reorder arbitrarily between pages.

FunctionSearchCommand and AllKindsSearchCommand carried the same dependency collection and organization defects and get the same treatment, plus the clamp.

Scope limits, stated explicitly rather than left to be inferred from the list above:

  - Per-module fair-share paging is **not** extended to the all-kinds path.
    `searchAllTypes` keeps its global `LIMIT`/`OFFSET`; all-kinds received the
    org-aware classification and the clamp only. Fair-share paging there needs
    its own design, since the all-kinds window is already split per result kind.
  - A live-compilation fallback and the tiered query path are deliberately not
    extended to functions; those also need their own design.

Known cost, not measured: the type-search fallback calls `PackageUtil.getCompilation` for imported modules absent from the index, inside the request. `Package.getCompilation()` caches per `Package` instance, so the cost is paid once per unindexed dependency per resolution rather than per keystroke, and a fully indexed project never reaches that path at all. On a project whose connectors are all unindexed, though, the compiles are serial and in-request, and a re-resolution discards the cache. Every other caller of that method (`IndexGenerator`, `SearchIndexGenerator`) is an offline tool. First-open latency for that case has not been benchmarked, so it is recorded here as a known cost rather than as a validated one.

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary

- Fix type, function, and all-kinds search for imported connector modules.
- Fall back to compiled models when imported modules are absent from the search index.
- Collect dependencies across package modules and identify them by organization and name.
- Improve pagination with per-module quotas, stable ordering, and bounded values.
- Add organization-aware filtering and tiered text-search results.
- Add tests for fallback behavior, filtering, pagination, and limit handling.


packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/SearchDatabaseManager.java
```diff
@@ -109,6 +111,9 @@ private SearchDatabaseManager() {
     public List<SearchResult> searchFunctions(String q, int limit, int offset) {
         List<SearchResult> results = new ArrayList<>();
         String sanitizedQuery = sanitizeQuery(q);
+        // SQLite treats a negative LIMIT as unlimited, so clamp the unchecked client values.
+        int safeLimit = Math.max(limit, 0);
+        int safeOffset = Math.max(offset, 0);
```
NipunaRanasinghe: **Medium — the all-kinds path is named in the title but got neither the clamp nor the fair share.**

The clamp landed here, in `searchTypes` (602-603) and in `searchTypesExcludingPackages` (853-854), but `searchAllTypes` still binds the raw values at 1209-1210, and so do `searchConnectors` (298-299) and `searchConnectorsByPackage` (491-492). `AllKindsSearchCommand:120` passes `limit`/`offset` straight from the client query map into `searchAllTypes`, so a negative `limit` on the all-kinds search still hits SQLite's unlimited-LIMIT behavior and returns the whole matched set in one response — the exact defect the PR body describes. `searchAllTypes` also keeps its bare global `LIMIT ?/OFFSET ?`, so all-kinds received the org-aware classification but not the pagination treatment the description says it got. Either clamp those three and adjust the body's claim about all-kinds, or say explicitly that all-kinds paging is out of scope.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/FunctionSearchCommand.java
```diff
@@ -111,9 +106,9 @@ protected List<Item> defaultView() {
         if (offset == 0) {
             WorkspaceFunctionNodeBuilder.buildSubmoduleWorkspaceNodes(
                     rootBuilder, project, position, query, functionsDoc);
-            if (!moduleNames.isEmpty()) {
+            if (!importedModules.isEmpty()) {
                 searchResults.addAll(
-                        dbManager.searchFunctionsByPackages(moduleNames, List.of(), Integer.MAX_VALUE, 0));
+                        dbManager.searchFunctionsByPackages(importedModules, List.of(), Integer.MAX_VALUE, 0));
```
NipunaRanasinghe, line 111: **Low — the new fair-share allocation is a no-op on the only production path that calls it.**

This asks for `Integer.MAX_VALUE` rows from offset 0, so every quota comes back equal to the module's full count and the ranges are always "take everything". `searchFunctionsByPackages` still pays for the allocation: `fetchPerPackageFunctionCounts` is a second round trip (`COUNT(*) ... GROUP BY p.org, p.name` over `Function`) plus a `ROW_NUMBER` window, where the pre-PR version was a single statement. `search()` never calls this method and `fetchPopularItems` passes a fixed two-module set, so nothing in the product exercises the quota split for functions. A `limit >= total` short-circuit inside `searchFunctionsByPackages` would keep the behavior and skip the count query.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/SearchDatabaseManager.java
```diff
@@ -633,62 +678,214 @@ public List<SearchResult> searchTypes(String q, int limit, int offset) {
...
+     *
+     * @param modules the modules to search, each identified by organization and index module name
+     * @param limit   the maximum number of results to return
+     * @param offset  the number of results to skip
+     * @return a list of search results matching the criteria
+     * @throws RuntimeException if there is an error executing the search
      */
-    public List<SearchResult> searchTypesByPackages(List<String> packageNames, int limit, int offset) {
-        if (packageNames.isEmpty()) {
+    public List<SearchResult> searchTypesByPackages(Set<ModuleCoordinate> modules, int limit, int offset) {
```
NipunaRanasinghe, line 538: **Low — six public methods here have no caller outside the test.**

`searchTypesByPackages(Set, int, int)`, `searchTypesByPackagesMatching`, `countIndexedTypes`, `countIndexedMatchingTypes`, `findIndexedModules` and `countTypesExcludingPackages` are referenced only from `SearchDatabaseManagerTypeSearchTest`; the product path is `indexedTypeCounts` + `searchTypesInRanges`, which `TypeSearchCommand:313` and `:330` call directly. The pre-PR `searchTypesByPackages(List<String>, ...)` did have a production caller, so this is new test-only surface on a public class, not carried-over dead code. Only `countTypesExcludingPackages` says so in its javadoc — either say it on all six, or drop the ones whose invariants `FairShareWindowTest` already covers at the level below.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/TypeSearchCommand.java
```diff
@@ -273,17 +275,238 @@ private void buildLibraryNodes(List<SearchResult> typeSearchList) {
...
+     * <i>wider</i> than they are: the index is queried by FTS token prefix, while this scores substring and
+     * Levenshtein matches too. A live module can therefore answer a query the indexed ones would not - searching
+     * {@code CompressionType} also surfaces a live {@code Compression}. Narrowing it to token prefixes would mean
+     * reimplementing FTS5 tokenization in Java for a pool that is meant to be forgiving, so the asymmetry is
+     * accepted; widening the indexed side instead is the change to make if the two must agree exactly.</p>
+     */
+    private List<LiveTypeMatch> collectLiveModuleTypes(Module module, String moduleName, Package dependencyPackage) {
+        SemanticModel semanticModel;
+        try {
+            semanticModel = PackageUtil.getCompilation(module.packageInstance()).getSemanticModel(module.moduleId());
```
NipunaRanasinghe: **Low — this compiles dependency packages inside a search request, which nothing on the LS request path did before.**

`PackageUtil.getCompilation` delegates to `Package.getCompilation()`, which caches per `Package` instance, so the cost is paid once per unindexed dependency per resolution rather than per keystroke. But it is paid inside the request, serially across every missing module, and a re-resolution throws the cache away. The only other callers that compile a dependency package are `IndexGenerator` and `SearchIndexGenerator`, which are offline tools. Worth stating in the PR whether the first Type Browser open was measured on a project with several unindexed connectors, since that is the case the fix exists for.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/java/io/ballerina/flowmodelgenerator/extension/SearchDatabaseManagerTypeSearchTest.java
```diff
@@ -0,0 +1,254 @@
...
+        long timeCount = page.stream().filter(r -> "time".equals(r.packageInfo().moduleName())).count();
+        Assert.assertEquals(osCount, 4, "os (5 types) must get its half of the window rather than be crowded out");
+        Assert.assertEquals(timeCount, 4, "time (15 types) must be capped at its fair share");
+    }
+
+    @Test(description = "Consecutive pages must tile: every row exactly once, none dropped or duplicated")
+    public void testPaginationTilesWithoutGapsOrOverlaps() {
+        Set<ModuleCoordinate> modules = Set.of(mod(BALLERINA, "os"), mod(BALLERINA, "time"));
+        int total = dbManager.countIndexedTypes(modules);
+        Assert.assertEquals(total, 20, "ballerina/os=5 + ballerina/time=15");
```
NipunaRanasinghe, line 181: **Low — exact counts asserted against a file that changes on unrelated PRs.**

`search-index.sqlite` is regenerated regularly — the last few commits touching it are "Update edifact and edi library indexes", "Update zip library index to 1.0.0", "Index zip library APIs". Literals like `total == 20` here, `61` at line 122, `osCount == 4`/`timeCount == 4` at 173-174, and `countIndexedMatchingTypes("Civil", modules) == 2` at 223 will fail the moment one of those packages is re-indexed, and LS Tests gates the merge, so the breakage lands on whoever refreshes the index. `testTierCapacitiesSumToGlobalMatchCount` is the shape that survives a refresh because it compares the tiers against each other; the exact-count assertions could derive their expectations from `countIndexedTypes` the same way. The JSON config expectations already work this way, so this is the repo's convention — flagging the cost, not asking for a rewrite.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/SearchDatabaseManager.java
dulajdilshan: Valid, and the reach is slightly wider than described. Clamped all three: `searchAllTypes`, `searchConnectors` and `searchConnectorsByPackage`.

Two details from verifying it:

`searchAllTypes` splits the window per result kind before binding — `int functionsLimit = limit / 2` — so a negative limit made the two inner CTE `LIMIT`s unlimited as well, not just the outer one. The clamp therefore has to precede the split, which is where it now sits.

It also wasn't confined to the all-kinds path. `ConnectorSearchCommand` reaches the other two with the same unchecked values (`:143` and `:171`), and `SearchCommand.parseIntParam` is a bare `Integer.parseInt` with no bounds check, so whatever integer the client sends arrives intact.

On the second half of your point: you're right, and the PR body was overclaiming. I have **not** given all-kinds the fair-share treatment — `searchAllTypes` keeps its global `LIMIT`/`OFFSET`. I've rewritten that paragraph to say all-kinds received the org-aware classification and the clamp only, and to list per-module fair-share paging for all-kinds as out of scope with its own design needed, rather than leaving it reading as though it had landed.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/FunctionSearchCommand.java
dulajdilshan: Correct, and the count query was the part worth removing. `defaultView` asks for `Integer.MAX_VALUE` rows from offset 0, so every quota came back equal to the module's own count and every range was "take everything" — while still paying `fetchPerPackageFunctionCounts`, a second round trip, purely to size quotas that could not bind anything.

Implemented slightly differently from your suggestion, because a `limit >= total` short-circuit still needs `total` — which is the query we're trying to avoid. So the unbounded case is detected from the arguments alone and the window is built directly, skipping the count entirely:

```java
boolean unbounded = offset <= 0 && limit == Integer.MAX_VALUE;
...
if (unbounded) {
    Map<ModuleCoordinate, Integer> wholePool = new HashMap<>();
    for (ModuleCoordinate module : moduleList) {
        wholePool.put(module, Integer.MAX_VALUE);
    }
    ranges = new FairShareWindow.Ranges<>(Map.of(), wholePool);
} else {
    Map<ModuleCoordinate, Integer> counts = fetchPerPackageFunctionCounts(conn, modules, functionNames);
    ranges = FairShareWindow.rangesOf(counts, offset, limit);
}
```

The row set is unchanged: the filter is `rn > pkg_skip AND rn <= pkg_skip + pkg_take`, and `rn <= count` and `rn <= MAX_VALUE` admit the same rows because `rn` never exceeds the module's count. `ORDER BY` is untouched, and `SearchTest` confirms it — 114/114 with `custom_default*.json` and `function_pagination_page1/2/3.json` unchanged.

I left the `ROW_NUMBER` window in place. Removing it on this path would need a second SQL statement, which costs more in complexity than the window costs to evaluate.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/SearchDatabaseManager.java
dulajdilshan: Verified — all six resolve only to `SearchDatabaseManagerTypeSearchTest`, and the product path is `indexedTypeCounts` + `searchTypesInRanges` as you say. `findIndexedModules` is the one that reads as production from its javadoc, but the request path takes the same key set off the `indexedTypeCounts` map it already fetches, so it genuinely has no caller either.

Took the "say it on all six" option rather than deleting. Each now states that it is off the request path and which invariant it is kept for. On the `FairShareWindowTest` half of your point: that test covers the allocator arithmetic in isolation, over synthetic counts — it can prove the quotas tile, but not that the SQL translating those quotas into `rn > skip AND rn <= skip + take` tiles against the real index. These six are what cover the second half, so dropping them would trade a real assertion for a smaller public surface. That seems like the wrong side of the trade while the tiling is this new; happy to revisit once it has some mileage.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/TypeSearchCommand.java
dulajdilshan: Agreed on the characterisation, and no code change here — but the PR body was overclaiming, so I've corrected it.

To your actual question: no, first-open latency was not measured on a project with several unindexed connectors, so I won't assert a number. What the code guarantees is narrower than "fast" — `Package.getCompilation()` caches per `Package` instance, so the cost is once per unindexed dependency per resolution rather than per keystroke, and `buildImportedNodes` only reaches the compile path for modules absent from `indexedTypeCounts`, which is the empty set on a fully indexed project. The bad case is a project whose connectors are all unindexed, and there the compiles are serial, inside the request, with a re-resolution discarding the cache.

You're right that this is new company for `PackageUtil.getCompilation`: every other caller (`IndexGenerator`, `SearchIndexGenerator`) is an offline tool. The PR description now records it as a known, unmeasured cost rather than implying it was validated.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/java/io/ballerina/flowmodelgenerator/extension/SearchDatabaseManagerTypeSearchTest.java
dulajdilshan: The risk is real, and since you're flagging the cost rather than asking for a rewrite I've left the literals in place — but it's worth saying why rather than declining silently.

The literals are load-bearing in a way a derived expectation isn't. `total == 20` here and the `61` at line 122 are what would catch an off-by-one in the range tiling: if `searchTypesByPackages` dropped or duplicated a row at a module boundary, an assertion derived from `countIndexedTypes` would move with the bug and stay green, because both sides would read the same counts. `testTierCapacitiesSumToGlobalMatchCount` survives a refresh precisely because it compares the tiers against each other — which is also why it cannot see an error common to both tiers.

So the trade is: literals break on an unrelated index refresh but catch tiling bugs; derived assertions survive refreshes but go blind to exactly the class of bug this PR could introduce. I'd rather absorb the refresh cost while the tiling is new. When they do break, the fix is mechanical — re-read the counts and update five literals — and the assertion message names the package, so it shouldn't cost whoever refreshes the index much.

That said, you're the one who'd carry it if it breaks on someone else's PR. If you'd rather not, say so and I'll convert them in this PR.

NipunaRanasinghe (approved):
Two things left, neither of them blocking:

1. Nothing covers a submodule import on the live tier. `imported_local_module.json` uses `http` and `url`, and both are top-level modules. The keying looks right when I read it, but if it ever misses you get one log warning and no types for that module, which is easy to miss.
2. The exact index counts in the new DB tests (`total == 20`, `61`, `osCount == 4`) will break the next time `search-index.sqlite` is refreshed. LS tests block the merge, so whoever re-indexes is the one who hits it. Probably better as a follow-up issue than more changes here.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/search/FunctionSearchCommand.java
```diff
@@ -241,7 +238,7 @@ private void buildLibraryNodes(List<SearchResult> functionSearchList, boolean ca
                     .version(packageInfo.version())
                     .build();
             Category.Builder builder;
-            if (moduleNames.contains(packageInfo.moduleName())) {
+            if (importedModules.contains(packageInfo.coordinate())) {
```
pasindufernando1, line 241: Non-blocking: submodule imports get misclassified for Central-sourced results.

`CentralSearchUtil.toSearchResult(SymbolResponse.Symbol)` fills `SearchResult.Package.moduleName` with `symbol.name()`, which is the *package* name. `importedModules` now holds `packageName[.moduleNamePart]`, so for `import ballerinax/oraclefusion.erp;` the coordinate is `(ballerinax, oraclefusion.erp)` while Central returns `(ballerinax, oraclefusion)` — no match, and those functions land under Extended Library instead of Imported Functions. The old package-name-only comparison matched here.

The DB path (`searchFunctionsByPackages`) *does* match, so the same function can surface twice: once under `oraclefusion.erp` in Imported, once under `oraclefusion` in the library section. Only affects true submodule imports; root-module imports are unchanged.

pasindufernando1 (approved):
LGTM. Verified the `FairShareWindow` quota algorithm (exhaustive simulation: monotone, per-source capped, sums to `min(window, total)` — pages tile without gaps or duplicates), the two-tier skip arithmetic in `buildImportedNodes`, parameter binding order in all the rewritten SQL, and the compiler API assumptions (`EnumSymbol extends TypeDefinitionSymbol`, no overlap between `typeDefinitions()`/`classes()`/`enums()`, `moduleDependencies()` being direct-imports-only).

One non-blocking note inline about submodule imports.

PR #1004, opened 2026-09-04
Retrieve class members in Copilot library selection

ravindu25:
Fixes: https://github.com/wso2/product-integrator/issues/2283

## Problem

Copilot could not see the member APIs of types a connector hands back — things like a spreadsheet's `Sheet` or `Table`.

Before the agent fetches library docs, a smaller model picks which libraries are relevant. That picker was only ever shown a library's clients, module-level functions and services — never its classes. So a question like "read a cell" or "append a row" matched nothing it could see, no library was named, and the docs tool returned nothing at all.

The effect was worse than a missing detail. With no documentation and a prompt that forbids inventing APIs, Copilot treated real, supported APIs as unsupported: it reimplemented them by hand, asked whether an API that exists was available, and reached different designs for the same task across chats.

Reported while testing the `xlsx` connector, where 51 of its APIs live on classes and were invisible to the picker.

## Fix

- Classes and object types are now part of what the picker is shown, with their members — so a library can be chosen because of a method, not just a top-level function.
- Annotations are shown for the same reason. They are evidence only; a library still keeps all of them, so nothing that renders today can be dropped.
- A class the picker names is pulled into the docs whole, even when nothing else references it.
- The types a class's methods mention are now resolved too, so the docs no longer name a type they never define.
- Class members count toward the batching threshold that splits large libraries into their own requests.
- A class from an older language server, which omitted its type marker, renders as a class instead of being silently discarded.

## Testing

24 unit tests added; the library suite passes at 284.

## Summary

- Expanded Copilot library selection to include classes, class members, object types, and annotations.
- Included selected classes and types referenced by class methods in generated documentation.
- Updated batching thresholds to count class members.
- Preserved classes from older language-server versions without a type discriminator.
- Added defensive handling for malformed language-server output.
- Added 24 unit tests. The library test suite passes 284 tests.
- End-to-end testing remains pending until the language-server JAR is rebuilt.

### DIFF
--- a/packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
+++ b/packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
@@ -30,11 +30,14 @@ import { Client, GetTypeResponse, GetTypesRequest, GetTypesResponse, getTypesRes
 import { TypeDefinition, AbstractFunction, Type, RecordTypeDefinition, UnionTypeDefinition } from "./library-types";
 import {
     getClientFunctionCount,
+    getSelectableMemberCount,
     hasNothingToSelect,
     selectServices,
+    selectClassTypeDefs,
     toSelectionRequest,
     withRestoredServiceLibraries,
 } from "./library-selection";
+import { collectClassMemberTypeRefs, TYPE_CLASS } from "./class-typedefs";
 import { getAnthropicClient, ANTHROPIC_HAIKU } from "../ai-client";
 import { GenerationType } from "./libraries";
 // import { getRequiredTypesFromLibJson } from "../healthcare/healthcare";
@@ -186,9 +189,14 @@ async function getRequiredFunctions(
     // restating it; `withRestoredServiceLibraries` removes it for the libraries that do get asked about.
     const passthroughLibs = libraryList.filter(hasNothingToSelect);
     const selectableLibs = libraryList.filter((lib) => !passthroughLibs.includes(lib));
-    const passthroughResp: GetFunctionResponse[] = passthroughLibs.map((lib) => ({ name: lib.name }));
+    // A passthrough library names its own classes: nothing chose to drop them, and a library whose whole
+    // API is classes has no clients or functions for the closure to walk from.
+    const passthroughResp: GetFunctionResponse[] = passthroughLibs.map((lib) => ({
+        name: lib.name,
+        ...(lib.classes && lib.classes.length > 0 ? { classes: lib.classes.map((cls) => cls.name) } : {}),
+    }));
 
-    const largeLibs = selectableLibs.filter((lib) => getClientFunctionCount(lib.clients) >= 100);
+    const largeLibs = selectableLibs.filter((lib) => getSelectableMemberCount(lib) >= 100);
     const smallLibs = selectableLibs.filter((lib) => !largeLibs.includes(lib));
 
     console.log(
@@ -271,8 +279,10 @@ CRITICAL RULES:
 2. Your ONLY task is selection - include or exclude items, NEVER modify field values.
 3. Copy all field values EXACTLY as provided - preserve every character including backslashes and special characters.
 4. For resource functions: "accessor" and "paths" are SEPARATE fields - NEVER combine them.
-5. A library is relevant if ANY of its clients, functions, or services match the query. A service matches when its "doc" (what the service is for), its "listenerDoc" (how it is triggered), its name, or ANY ONE of its handlers under "methods", is what the query needs. List each matching service under the library's "services" field, copying its "listener" and "name" verbatim; omit the services that do not match. If a library matches ONLY via its services, still include the library in the output with empty/omitted clients and functions.
-6. "doc", "listenerDoc" and a handler's "doc" are evidence to reason over, never fields to copy: the response carries only "listener" and "name" for a service.`;
+5. A library is relevant if ANY of its clients, functions, services, classes, or annotations match the query. A service matches when its "doc" (what the service is for), its "listenerDoc" (how it is triggered), its name, or ANY ONE of its handlers under "methods", is what the query needs. List each matching service under the library's "services" field, copying its "listener" and "name" verbatim; omit the services that do not match. If a library matches ONLY via its services, still include the library in the output with empty/omitted clients and functions.
+6. "doc", "listenerDoc" and a handler's "doc" are evidence to reason over, never fields to copy: the response carries only "listener" and "name" for a service.
+7. "classes" are types the library hands back that carry callable methods (a spreadsheet's Sheet, a connection's Session). A class matches when its name, its "description", or ANY ONE of its methods under "functions" is what the query needs. List each matching class under the library's "classes" field as its NAME ONLY, copied verbatim — never as an object, and never narrowed to selected methods. A query about a class method is one of the most common kinds: if the query asks to do something to a thing the library returns, look here before concluding the library does not match.
+8. "annotations" are evidence only and have NO response field. Never emit them. If a library matches only via an annotation, include the library with its other fields empty/omitted.`;
 
     const getLibUserPrompt = `You will be provided with a list of libraries, clients, and their functions, and a user query.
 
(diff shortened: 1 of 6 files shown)


packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
```diff
@@ -716,6 +737,13 @@ function getOwnRecordRefs(functions: AbstractFunction[], allTypeDefs: TypeDefini
                 const foundTypes = addInternalRecord(member.type, ownRecords, allTypeDefs);
                 typesToProcess.push(...foundTypes);
             }
+        } else if (typeDef.type === TYPE_CLASS) {
```
madushajg: This arm tests `typeDef.type === TYPE_CLASS`, but every other new call site uses `isClassTypeDef()`, including `selectClassTypeDefs` which seeds unlabelled legacy classes into `ownRecords`, and `renderTypeDef`, which renders them. Result: for a class from an older language server (no `type` key), the closure treats it as a leaf, so `renderClass` prints `getSheet() returns Sheet` while Sheet is defined nowhere, the exact defect this PR fixes for labelled classes.

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
```diff
@@ -890,6 +918,11 @@ function getExternalTypeDefRefs(
             for (const member of unionDef.members) {
                 addExternalRecord(member.type, externalRecords);
             }
+        } else if (typeDef.type === TYPE_CLASS) {
```
madushajg: Same divergence in the external counterpart: an unlabelled legacy class contributes no external record refs, so a type its methods reach in another module is never fetched. Use `isClassTypeDef(typeDef)` here too.

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
```diff
@@ -183,9 +186,14 @@ async function getRequiredFunctions(
     // restating it; `withRestoredServiceLibraries` removes it for the libraries that do get asked about.
     const passthroughLibs = libraryList.filter(hasNothingToSelect);
```
madushajg, line 190: `hasNothingToSelect` wasn't extended to consider `lib.classes`, so a library whose entire API lives on classes (the xlsx-shaped case motivating this PR) short-circuits to passthrough, seeding every class without the model ever choosing. That also makes the new `classes` field and `getSelectableMemberCount` batching change inert for that exact library shape. Either count class members here or update the doc comment. It currently says, "no choice to make," which isn't true anymore.

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
```diff
@@ -267,8 +275,10 @@ CRITICAL RULES:
...
-6. "doc", "listenerDoc" and a handler's "doc" are evidence to reason over, never fields to copy: the response carries only "listener" and "name" for a service.`;
+5. A library is relevant if ANY of its clients, functions, services, classes, or annotations match the query. A service matches when its "doc" (what the service is for), its "listenerDoc" (how it is triggered), its name, or ANY ONE of its handlers under "methods", is what the query needs. List each matching service under the library's "services" field, copying its "listener" and "name" verbatim; omit the services that do not match. If a library matches ONLY via its services, still include the library in the output with empty/omitted clients and functions.
```
madushajg: Rules 5, 7, 8 each restate "include the library with empty/omitted X". Worth stating once given the prompt's length and the way this text tends to drift.

packages/ballerina-extension/src/features/ai/utils/libs/library-selection.ts
```diff
@@ -297,9 +322,81 @@ export function toSelectionRequest(lib: Library, includeFunctionDescriptions: bo
...
+export function toRequestClasses(typeDefs: TypeDefinition[] | undefined): MinifiedClassTypeDef[] | undefined {
+    if (!typeDefs) {
+        return undefined;
+    }
+    const classes: MinifiedClassTypeDef[] = [];
+    for (const typeDef of typeDefs) {
+        if (!isClassTypeDef(typeDef)) {
+            continue;
+        }
+        const members = toRequestClientFunctions(
```
madushajg, line 349: `ClassTypeDefinition.functions` is `any[]`; `class-typedefs.ts` guards every hop for exactly that reason (with a test). `toRequestClientFunctions` was written for the properly-typed `Client.functions` and guards nothing. A null member throws on `"accessor" in item`, a missing `parameters` throws on `.map`, a missing `return` throws on `item.return.type.name`. Any of these fails the entire generation, not just one library. Guard the members before handing off, or add `param?.name` / `item.return?.type?.name` checks inside `toRequestClientFunctions`.

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
```diff
@@ -267,8 +275,10 @@ CRITICAL RULES:
...
+6. "doc", "listenerDoc" and a handler's "doc" are evidence to reason over, never fields to copy: the response carries only "listener" and "name" for a service.
+7. "classes" are types the library hands back that carry callable methods (a spreadsheet's Sheet, a connection's Session). A class matches when its name, its "description", or ANY ONE of its methods under "functions" is what the query needs. List each matching class under the library's "classes" field as its NAME ONLY, copied verbatim — never as an object, and never narrowed to selected methods. A query about a class method is one of the most common kinds: if the query asks to do something to a thing the library returns, look here before concluding the library does not match.
```
SasinduDilshara: I think we only need to say, In which scenarios we filter the classes, 
* If the name matches
* If one of method matches etc

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
```diff
@@ -267,8 +275,10 @@ CRITICAL RULES:
...
+8. "annotations" are evidence only and have NO response field. Never emit them. If a library matches only via an annotation, include the library with its other fields empty/omitted.`;
```
SasinduDilshara: I think 1st sentence is LLM's own knowledge

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
```diff
@@ -283,16 +293,18 @@ ${JSON.stringify(libraryList)}
 To process the user query and filter the libraries, clients, services and functions, follow these steps:
 
 1. Analyze the user query to understand the specific requirements or needs.
-2. Review the provided libraries, clients, services and functions in Library_Context_JSON.
-3. Select only the libraries, clients, services and functions that directly match the query's needs.
-4. Exclude any irrelevant libraries, clients, services or functions.
-5. If no relevant functions and services are found, return an empty array for libraries.
+2. Review the provided libraries, clients, services, functions, classes and annotations in Library_Context_JSON.
+3. Select only the libraries, clients, services, functions and classes that directly match the query's needs.
```
SasinduDilshara, line 301: We need object types too

ravindu25: Object types are already in there. The LS maps TypeDescKind.OBJECT to TypeCategory.CLASS, so public type Sheet isolated object and public class Workbook both arrive tagged "Class" and land in the same classes array — Sheet and Table in xlsx are object types and they're covered.

Made it explicit in the rule text anyway so it isn't ambiguous to the model.

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
ravindu25: fixed

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
ravindu25: fixed

packages/ballerina-extension/src/features/ai/utils/libs/library-selection.ts
ravindu25: fixed

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
ravindu25: fixed

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
ravindu25: fixed

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
ravindu25: fixed

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
ravindu25: fixed

packages/ballerina-extension/src/features/ai/utils/libs/function-registry.ts
```diff
@@ -716,6 +737,13 @@ function getOwnRecordRefs(functions: AbstractFunction[], allTypeDefs: TypeDefini
                 const foundTypes = addInternalRecord(member.type, ownRecords, allTypeDefs);
                 typesToProcess.push(...foundTypes);
             }
+        } else if (typeDef.type === TYPE_CLASS) {
+            // A class's methods name types the reader needs. Without this arm a class was a leaf, and a
+            // type reachable only through one of its methods reached the prompt undefined.
+            for (const ref of collectClassMemberTypeRefs(typeDef)) {
+                const foundTypes = addInternalRecord(ref, ownRecords, allTypeDefs);
+                typesToProcess.push(...foundTypes);
+            }
```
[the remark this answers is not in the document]
ravindu25: fixed

SasinduDilshara (approved):

PR #1007, opened 2026-09-04
Surface output-limit truncation as an incomplete turn

ravindu25:
Fixes: https://github.com/wso2/product-integrator/issues/2285

## Issue

Copilot terminates abruptly part-way through a large single-file edit. The turn is presented as a normal completion — review card, feedback prompt and follow-up suggestions, with no indication anything was missed — while the project is left half-edited and failing to build.

Edits are written to disk as the turn progresses, not batched at the end. So when the model runs out of output room mid-edit, the changes already made stay on disk and the one that would have made them consistent never lands.

Reported against the xlsx connector: three consecutive turns died at the same ~200-line edit; asking for the change as several smaller edits completed it.

## Fix

The turn now recognises when the model was cut off at its output limit rather than having finished, and resumes itself instead of ending.

On truncation the model is handed back its own partial work along with a note stating that the cut-off change never reached the project, that earlier changes in the turn did, and that it should re-check the state of what it already edited before finishing the remaining work as several small edits. Generation then continues in the same turn, up to two resumes.

Nothing is surfaced to the user. The turn streams on as one continuous response, because the request was for a working generation, not for an explanation of why one stopped. The model may mention in its own words that it is completing the change in smaller steps, but that is its choice, not a UI state.

Supporting changes:

* Messages and token usage are carried across attempts, so a resumed turn is persisted, costed and replayed as the single turn it appears to be.
* Tool calls left unanswered by the truncation are stripped before the turn is re-sent, since an unmatched `tool_use` is rejected outright by the provider.
* The per-response output limit is raised to 64k, shared with the figure the context-usage widget reports as reserved so the two cannot disagree. This makes truncation rare rather than routine; the resume remains the safety net.
* A recovered turn is recorded as a completion carrying `generation.truncation_retries`; only a turn that exhausts its resumes reports as truncated. Since recovery is invisible, this is the only signal of how often the limit is reached.

Review state is left unchanged, so partial edits stay revertible.

## Verified

Reproduced at runtime by temporarily lowering the output cap. A single turn truncated twice, resumed automatically both times, and completed on the third attempt with a clean compile — no user-visible interruption. Token usage summed exactly across the three attempts, and the persisted turn retained the full message history.

## Summary

- Automatically resumes Copilot generation when the output limit truncates a response.
- Preserves messages, tool calls, review state, and cumulative token usage across retries.
- Allows up to two retries with a 64,000-token output limit.
- Cleans up incomplete tool calls before retrying.
- Records retry telemetry and reports truncation only when retries are exhausted.
- Adds unit tests for message cleanup, usage aggregation, and recovery-note generation.


packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
```diff
@@ -434,183 +447,248 @@ export class AgentExecutor extends AICommandExecutor<GenerateAgentCodeRequest> {
...
+                        );
+                        const resumed: ModelMessage[] = [
+                            ...dropDanglingToolCalls(attemptMessages),
+                            { role: 'user', content: TRUNCATION_RECOVERY_NOTE },
+                        ];
+                        carriedMessages.push(...resumed);
+                        allMessages.push(...resumed);
+                        carriedUsage = addUsage(carriedUsage, await totalUsage);
+                        continue;
+                    }
```
madushajg: Resume also fires on context-window overflow, where retrying can't help

`attemptFinishReason === 'length'` is not specific enough to gate the resume. Per `mapAnthropicStopReason` in @`ai-sdk/anthropic`, both `max_tokens` and `model_context_window_exceeded` unify to `'length'`. For a context-window overflow, this branch does the one thing guaranteed to fail: it pushes the partial assistant messages plus the recovery note onto `allMessages` and re-sends a strictly larger prompt, twice which can only overflow again, and the recovery note tells the model a false story (its output length was never the problem). `attemptRawFinishReason` is already captured here, gate on it instead, e.g. `attemptFinishReason === 'length' && attemptRawFinishReason !== 'model_context_window_exceeded'`, and let the context-window case fall straight through to `handleStreamFinish`. Reachable on Bedrock in particular, where `supportsCompaction()` is false so there's no server-side compaction to prevent the overflow in the first place.

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
```diff
@@ -434,183 +447,248 @@ export class AgentExecutor extends AICommandExecutor<GenerateAgentCodeRequest> {
...
-                if (this.config.abortController.signal.aborted) {
-                    console.log("[AgentExecutor] Detected abort after stream completion");
-                    const abortError = new Error('Aborted by user');
-                    abortError.name = 'AbortError';
-                    throw abortError;
+                    // Finalise against the whole turn, not just the last attempt's slice.
+                    streamContext.response = Promise.resolve({
+                        ...attemptResponse,
+                        messages: [...carriedMessages, ...attemptMessages],
+                    }) as StreamContext['response'];
```
madushajg: Overwriting `streamContext.response` breaks the abort/error handlers' invariant, causing double-persistence

Line 686 makes `streamContext.response` resolve to `[...carriedMessages, ...attemptMessages]`, but the abort handler at 701–702 and `handleStreamError` at line 941 both assume response holds only the current attempt's messages and prepend `carriedMessages` themselves. Once line 684 runs, that assumption is false. Anything thrown from `handleStreamFinish` on line 689 (`checkCompilationErrors`, `getProjectMetrics`, `emitReviewActions` all make LS calls) lands in the catch at 695 and a user who pressed Stop takes the abort path, persisting every carried message twice, including duplicate `tool_use` ids. That history is replayed verbatim by `getChatHistoryForLLM`, so the next turn in the thread 400s. Assign the combined array to a local and pass it to `handleStreamFinish` instead of mutating `streamContext.response`, or flag that carried messages are already merged in.

ravindu25: fixed

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
ravindu25: fixed

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
```diff
@@ -933,17 +1042,21 @@ Generation stopped by user. The last in-progress task was not saved. Any complet
             cost: `$${totalCost.toFixed(4)}`,
         });
 
-        // Send telemetry for generation complete
+        // A recovered turn is a real completion; `truncation_retries` records what it cost.
+        // Recovery is invisible in the UI, so this is the only signal that it happened.
```
RNViththagan, line 1052: This comment says recovery is invisible by design, which makes sense for a successful resume. But when retries are exhausted (endedTruncated true here) the project can still be left half-edited and non-compiling - the exact bug this PR sets out to fix - and nothing downstream differs: emitReviewActions and maybeScheduleFollowups(..., 'completed') a few lines down still run unconditionally, so the user gets the same review card and "completed" follow-ups as a normal turn. The only distinguishing signal is the telemetry event, which the user never sees. Shall we add a real user-facing indicator for this remaining case? An earlier commit on this branch had a turn_truncated event plus chat-UI rendering for exactly this before it was pulled out.

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
```diff
@@ -434,183 +448,251 @@ export class AgentExecutor extends AICommandExecutor<GenerateAgentCodeRequest> {
...
-                    const abortError = new Error('Aborted by user');
-                    abortError.name = 'AbortError';
-                    throw abortError;
+                    const attemptResponse = await response;
+                    const attemptMessages = (attemptResponse.messages ?? []) as ModelMessage[];
+
+                    // A truncated attempt never ran its last tool call, so the change it
+                    // carried is missing while earlier ones are already on disk. Hand the
+                    // model its partial work back with a note and let it finish.
+                    if (isResumableTruncation(attemptFinishReason, attemptRawFinishReason)
```
RNViththagan, line 674: Question on the premise here: with ai 6.0.103, doesn't the SDK already start another step itself once a tool call gets an error result (a truncated tool-call's JSON parses as an invalid input, which becomes a tool-error output, and the continuation check treats that as answered)? If so, this branch would only trigger for a truncation that happens outside a tool call - plain text, or a block the provider never closes - and dropDanglingToolCalls below would have nothing to drop in the client-tool case. Did the lowered-cap repro actually hit a dangling tool call, or a text-only cutoff? Want to make sure the recovery targets the failure mode it's meant to.

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
```diff
@@ -434,183 +448,251 @@ export class AgentExecutor extends AICommandExecutor<GenerateAgentCodeRequest> {
...
+                        && truncationRetries < MAX_TRUNCATION_RETRIES) {
+                        truncationRetries++;
+                        console.warn(
+                            `[AgentExecutor] Output limit reached (raw: ${attemptRawFinishReason ?? 'n/a'}) — ` +
+                            `resuming turn automatically (${truncationRetries}/${MAX_TRUNCATION_RETRIES})`
+                        );
+                        const resumed: ModelMessage[] = [
+                            ...dropDanglingToolCalls(attemptMessages),
+                            { role: 'user', content: TRUNCATION_RECOVERY_NOTE },
```
RNViththagan: Small one: this note gets pushed into carriedMessages, which becomes part of the turn's persisted history, and later turns in the same thread replay the full history verbatim. So once a single truncation happens, this note - a few hundred tokens - sits in every subsequent prompt for the rest of the thread. Worth trimming it out before persisting the final turn, or is keeping it around intentional?

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
ravindu25: You're right. I checked the persisted turn from the repro: both truncated attempts left an assistant message with only a text part and no tool-call, so the provider never closed the block, the SDK had nothing to feed back, and that's why the loop ended.

So if the block is closed with truncated JSON, repairToolCallInputs coerces the input, the tool-error counts as answered and the SDK continues on its own. The resume covers the other case, where no tool-call materialises at all. dropDanglingToolCalls is belt-and-braces for the narrow window in between — it didn't fire in the repro.

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
ravindu25: fixed

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
ravindu25: Deliberate — the user asked for a generation, not a report about why one stopped, so we resume rather than hand them a button. You're right that exhaustion still reads as a normal completion.

It is measurable either way though: truncation_retries is on both events, so a run that truncated and recovered lands as generation.completed with truncation_retries > 0, and one that gave up lands as generation.truncated. With 64k plus two resumes, exhaustion should be rare — if the numbers say otherwise, reinstating the indicator for just that case is easy, the earlier commit is in the branch history.

SasinduDilshara (approved):

RNViththagan (approved):
LGTM

NipunaRanasinghe (approved):

PR #1009, opened 2026-09-04
Fix summarise overflow config missing prompt editor and model provider dropdown

dan-niles:
## Summary
Fixes [#1622](https://github.com/wso2/product-integrator/issues/1622) — the summarise overflow configuration did not show the prompt editor or the model provider dropdown.

https://github.com/user-attachments/assets/81a9bdca-4679-48c2-858e-94e2859d97db

## Summary

- Add prompt editor and model provider selection for summarise overflow configuration.
- Support prompt editing and model provider selection for constructed record fields.
- Support node creation in a popup from the Record Configuration modal.
- Improve card-list sizing and connection selection layouts.
- Refresh expression diagnostics after field updates.
- Add tests for prompt-field editing and model updates.


packages/ballerina-visualizer/src/views/BI/HelperPaneNew/Views/RecordConfigModal.tsx
```diff
@@ -723,10 +764,14 @@ export function ConfigureRecordPage(props: ConfigureRecordPageProps) {
...
-                                onModelChange={handleModelChange}
-                            />
+                            <PromptFieldEditorContext.Provider
+                                value={{ openPromptEditor: (param, onChange) => setPromptEditTarget({ param, onChange }) }}
+                            >
+                                <RecordConfigView
+                                    recordModel={recordModel}
+                                    onModelChange={handleModelChange}
+                                />
+                            </PromptFieldEditorContext.Provider>
```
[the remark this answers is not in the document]
dan-niles: Added L2 coverage in `Views/PromptFieldEditing.test.tsx` for opening the prompt editor and the edit propagating to `onModelChange`; the RPC-backed source regeneration inside `ConfigureRecordPage` itself is a heavier lift and isn't covered.

MohamedSabthar (dismissed):

packages/ballerina-side-panel/src/components/editors/ExpandedEditor/ExpandedEditor.tsx
```diff
@@ -190,7 +191,8 @@ export const ExpandedEditor: React.FC<ExpandedPromptEditorProps> = ({
     error,
     formDiagnostics,
     inputMode,
-    readOnly
+    readOnly,
+    zIndex = 2001
```
pasindufernando1: **Blocker:** the `zIndex = 2001` default silently lowers every existing caller.

`ModalContainer`'s stylesheet sets `z-index: 2100`, but the inline `style={{ zIndex }}` added at L291 wins over the emotion class — so every pre-existing call site that passes no `zIndex` (`ExpressionEditor.tsx:938`, `TextAreaEditor.tsx:119`, `bi-diagram/NPPromptEditor.tsx:387`) drops from 2100 to 2001.

That puts them below `MainPanel.tsx:131`'s full-screen `Overlay` (`z-index: 2099`, `pointer-events: auto`), which renders whenever `modalStack.length > 0` — i.e. exactly the modal form path this PR newly exercises via `preferModal: true`. Repro: open a connection/agent create form through the modal path, then click expand on a prompt or expression field — the expanded editor paints behind the dim overlay (and behind the 2100 popup box), invisible and unclickable.

Suggest defaulting to `2100` (or omitting the inline style when the prop is undefined) and passing `2101` only from the new Record Configuration call site.

dan-niles: Fixed in 283b56cbe4 — default is now 2100, matching `ModalContainer`'s stylesheet.

Also found a related instance of the same class of bug while testing: `UseAgentTool.tsx` portals `AddAgentPopup` straight to `document.body` with a hardcoded `z-index: 2100`, bypassing `#visualizer-container`. That let it win the stacking tie against `ExpandedEditor` (same z-index, but outside the shared portal container) when expanding a prompt/expression field from the "Add Tool - Use Agent" flow. Fixed by portaling into `#visualizer-container` instead, so it shares the same stacking arena as `ExpandedEditor` and the `modalStack` popups.

SasinduDilshara (dismissed):

packages/ballerina-side-panel/src/components/editors/ExpandedEditor/ExpandedEditor.tsx
pasindufernando1: Addressed in 283b56c — default is 2100 now and the new call site passes 2101. Thanks!

pasindufernando1 (dismissed):
LGTM

packages/ballerina-visualizer/src/views/BI/HelperPaneNew/Views/RecordConfigModal.tsx
```diff
@@ -159,6 +169,34 @@ export function ConfigureRecordPage(props: ConfigureRecordPageProps) {
...
+        value && !value.startsWith("`") && !value.endsWith("`") ? `\`${value}\`` : value;
+
+    const handlePromptChange = (updatedValue: string) => {
+        if (!promptEditTarget) {
+            return;
+        }
+        promptEditTarget.param.value = wrapPrompt(updatedValue);
+        promptEditTarget.param.selected = true;
+        promptEditTarget.onChange();
+    };
```
madushajg, line 200: `handlePromptChange` runs on every keystroke (`RichTextTemplateEditor` fires `onChange` per doc change), so each character triggers a `getRecordSource` RPC and `handleModelChange` has no request sequencing, so a late-arriving earlier response can overwrite `localExpressionValue` with a stale expression. It also remounts the whole `ParameterBranch` tree each keystroke via the `key={JSON.stringify(recordModel)}` in `RecordConfigView`. Please debounce the `promptEditTarget.onChange()` propagation (or commit on close/blur) and add a request-id guard in `handleModelChange`.

packages/ballerina-visualizer/src/views/BI/HelperPaneNew/Views/RecordConfigModal.tsx
```diff
@@ -159,6 +169,34 @@ export function ConfigureRecordPage(props: ConfigureRecordPageProps) {
     // Diagnostics state
     const [formDiagnostics, setFormDiagnostics] = useState<any[]>(field?.diagnostics || []);
 
+    const [promptEditTarget, setPromptEditTarget] = useState<{ param: TypeField; onChange: () => void } | null>(null);
+
+    const sanitizePrompt = (value: string) =>
+        value && value.startsWith("`") && value.endsWith("`") ? value.slice(1, -1) : value;
+    const wrapPrompt = (value: string) =>
+        value && !value.startsWith("`") && !value.endsWith("`") ? `\`${value}\`` : value;
```
madushajg: `!startsWith("`") && !endsWith("`")` skips wrapping whenever the prompt starts or ends with a backtick. 

eg. an AI-accepted prompt ending in an inline-code span comes back `fromPromptMode.handleAcceptunwrapped` (that path bypasses `deserializeValue`) and gets written to `param.value` as-is, producing an invalid raw template in the generated source. Test for "already wrapped" instead: `value.length >= 2 && value.startsWith("`") && value.endsWith("`")` and wrap otherwise.

packages/ballerina-visualizer/src/views/BI/HelperPaneNew/Views/RecordConfigModal.tsx
dan-niles: Fixed in bfb0946: debounced the prompt editor's `onChange` propagation (300ms, flushed on close/save so the final edit isn't dropped) and added a request-id guard in `handleModelChange` to drop stale `getRecordSource` responses.

packages/ballerina-visualizer/src/views/BI/HelperPaneNew/Views/RecordConfigModal.tsx
dan-niles: Fixed in bfb0946: `wrapPrompt` now uses the same "already wrapped" check as `sanitizePrompt` (both ends must have a backtick, length >= 2), so it wraps correctly in your scenario instead of skipping.

madushajg (approved):

MohamedSabthar (approved):

SasinduDilshara (approved):

PR #1010, opened 2026-09-04
Fix back navigation not returning to originating package after agent definition creation

dan-niles:
## Summary

Fixes https://github.com/wso2/product-integrator/issues/2337

https://github.com/user-attachments/assets/e9baaa6f-15db-4ce2-8797-33cfcef85f47

Fixes the back button not working when creating an agent definition — choosing "Create as a library" did not redirect back to the package where agent definition creation was initiated.

## Test plan
- [ ] Initiate agent definition creation from a package, choose "Create as a library", then navigate back and verify it returns to the originating package

## Summary

- Fixes back navigation after creating an agent definition in a separate library package.
- Preserves the originating package path across history and view updates.
- Excludes agent creation views from back-navigation history and breadcrumbs.
- Preserves newly created types through the standard form change handler.
- Stabilizes designer response-type fields during refreshes.
- Increases the minimum library overview card width.


MohamedSabthar (dismissed):

packages/ballerina-visualizer/src/components/TopNavigationBar/index.tsx
```diff
@@ -249,6 +249,8 @@ export function TopNavigationBar(props: TopNavigationBarProps) {
             MACHINE_VIEW.BIDataMapperForm,
             MACHINE_VIEW.AddConnectionWizard,
             MACHINE_VIEW.EditConnectionWizard,
+            MACHINE_VIEW.AddAgent,
+            MACHINE_VIEW.AddAgentDefinition,
```
madushajg, line 253: These views have no case in `getDisplayLabel`, so the default branch returns the raw enum value and `buildBreadcrumbItem`s still renders a skipped view when it's the active tail (`!isLast && skippedViews.has(view)`). With the `SKIP` suffix added in this PR, the breadcrumb on the Add Agent Definition page now literally reads `"Add Agent Definition SKIP"`. Please add case `MACHINE_VIEW.AddAgent:` / case `MACHINE_VIEW.AddAgentDefinition:` to `getDisplayLabel` returning e.g. `"Agent"` / `"Agent Definition"`, like every other `SKIP` view in this set.

packages/ballerina-visualizer/src/views/BI/AgentDefinitionEditor/AgentDefinitionDesigner.tsx
```diff
@@ -1139,7 +1140,7 @@ export function AgentDefinitionDesigner(props: AgentDefinitionDesignerProps) {
         types: (runFn.returnType?.types?.length ?? 0) > 0
             ? runFn.returnType.types.map((t, i) => (i === 0 ? { ...t, ballerinaType: outputType } : t))
             : [{ fieldType: "TYPE", selected: true, ballerinaType: outputType }],
-    }] : [];
+    }] : [], [runFn?.returnType?.value]);
```
madushajg: `runFn` goes `undefined` → defined on the first model load, but the dep only tracks `runFn?.returnType?.value`. If the loaded run has no `returnType.value` (which this code explicitly handles — `outputTypeOf` returns `""` and there's a fallback `types` entry), the dep stays undefined, the memo keeps the initial `[]`, and the `outputTypeFields.length > 0` guard on line 1552 silently prevents the Response Type panel from ever opening. Add the existence of `runFn` to the deps.

e.g. `[Boolean(runFn), runFn?.returnType?.value]`

packages/ballerina-visualizer/src/components/TopNavigationBar/index.tsx
dan-niles: Fixed in 8ea7ded756 by adding `getDisplayLabel` cases for `AddAgent`/`AddAgentDefinition`, returning "Agent"/"Agent Definition" instead of the raw enum.

packages/ballerina-visualizer/src/views/BI/AgentDefinitionEditor/AgentDefinitionDesigner.tsx
dan-niles: Fixed in 8ea7ded756 — added `Boolean(runFn)` to the memo dep array so the fields recompute once the run function loads, even if `returnType.value` is empty.

madushajg (approved):

SasinduDilshara (approved):

PR #1013, opened 2026-09-05
Handle corrupted BIR cache

thuva9872:
## Purpose

When a module's compiled BIR cache is corrupted, the project loads empty because the compiler fails with an invalid magic number error while reading the cached `.bir`. Previously the only recovery was to manually delete ~/.ballerina, which isn't practical for low-code users.

This PR fixes this as follows:
- Language server: PullModuleExecutor now detects the corrupt-BIR failure, parses the affected module's coordinates (org/name/version) from the error, and emits a new `projectService/corruptBirCache` notification (payload CorruptBirCacheParams) to the client, along with the running distribution version.
- VS Code extension: On receiving the notification, we surface a user-friendly prompt (no internal "BIR" jargon) offering a one-click Clear cache & reload:
  - "The cache for module '<org>/<name>:<version>' is corrupted, so the project may appear empty. Clear the cache for this module and reload?"
  - Falls back to a generic message when the module can't be identified.
- Clicking the action clears only the affected module's cache under the active distribution's cache-<distVersion> directory, then reloads the window. If the module can't be identified — or the targeted clear matches nothing — it falls back to clearing the full module cache for that distribution so the user still recovers.

https://github.com/user-attachments/assets/dfd56f76-5bfc-4ace-9530-d4689024c4bc

image

Fixes: https://github.com/wso2/product-integrator/issues/2251

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary

- Detects corrupted BIR cache failures in the language server.
- Sends package and distribution details through `projectService/corruptBirCache`.
- Prompts users to clear the affected cache and reload VS Code.
- Clears all matching distribution caches when targeted recovery is unavailable.
- Resolves package-level cache paths, including submodules.
- Adds validation and test coverage for recovery behavior.

### DIFF
--- a/packages/ballerina-extension/src/core/extended-language-client.ts
+++ b/packages/ballerina-extension/src/core/extended-language-client.ts
@@ -321,6 +321,7 @@ import { debug, handlePullModuleProgress } from "../utils";
 import { CMP_LS_CLIENT_COMPLETIONS, CMP_LS_CLIENT_DIAGNOSTICS, getMessageObject, sendTelemetryEvent, TM_EVENT_LANG_CLIENT } from "../features/telemetry";
 import { DefinitionParams, InitializeParams, InitializeResult, Location, LocationLink, TextDocumentPositionParams } from 'vscode-languageserver-protocol';
 import { updateProjectArtifacts } from "../utils/project-artifacts";
+import { CorruptBirCachePayload, promptClearCorruptBirCache } from "../utils/bir-cache-recovery";
 import { RPCLayer } from "../../src/RPCLayer";
 import { VisualizerWebview } from "../../src/views/visualizer/webview";
 
@@ -534,7 +535,8 @@ enum EXTENDED_APIS {
     MULE_TO_BI = 'projectService/importMule',
     MIGRATION_TOOL_STATE = 'projectService/stateCallback',
     MIGRATION_TOOL_LOG = 'projectService/logCallback',
-    PUSH_MIGRATED_PROJECT = 'projectService/pushMigratedProject'
+    PUSH_MIGRATED_PROJECT = 'projectService/pushMigratedProject',
+    CORRUPT_BIR_CACHE = 'projectService/corruptBirCache'
 }
 
 enum EXTENDED_APIS_ORG {
@@ -653,6 +655,14 @@ export class ExtendedLangClient extends LanguageClient implements ExtendedLangCl
             } catch (error) {
                 console.error("Error in PUBLISH_ARTIFACTS handler:", error);
             }
+        });
+    }
+
+    registerCorruptBirCache(): void {
+        // A corrupt/incompatible cached BIR makes the project load empty. The LS
+        // reports the affected module here; offer to clear just that module's cache and reload.
+        this.onNotification(EXTENDED_APIS.CORRUPT_BIR_CACHE, (res: CorruptBirCachePayload) => {
+            void promptClearCorruptBirCache(res);
         });
     }
 
--- a/packages/ballerina-extension/src/stateMachine.ts
+++ b/packages/ballerina-extension/src/stateMachine.ts
@@ -583,6 +583,8 @@ const stateMachine = createMachine<MachineContext>(
                 try {
                     // Register the event driven listener to get the artifact changes
                     context.langClient.registerPublishArtifacts();
+                    // Register the corrupt-BIR-cache listener to recover from corrupted BIR cache failures
+                    context.langClient.registerCorruptBirCache();
                     // IF the project info is not set, we don't need to build the project structure
                     if (context.projectInfo) {
 
--- a/packages/ballerina-extension/src/test-support/birCacheRecovery.test.ts
+++ b/packages/ballerina-extension/src/test-support/birCacheRecovery.test.ts
@@ -1,0 +1,147 @@
+/**
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com) All Rights Reserved.
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * @jest-environment node
+ *
+ *  A corrupt cached BIR makes projects load empty. The LS sends a
+ * `projectService/corruptBirCache` notification with the affected module's coordinates and the
+ * running distribution version; the client clears that module's compiled cache under
+ * cache-<distVersion>. These L1 tests pin three invariants: (1) only a well-formed module
+ * coordinate is accepted (a malformed/hostile payload can never become an fs path); (2) the clear
+ * removes ONLY the affected module's compiled cache and only under the active distribution's
+ * cache-<distVersion> — never other distributions, the pulled bala/, other modules or versions, or
+ * the distribution itself.
+ */
+
+import * as os from "os";
+import * as path from "path";
+import * as fs from "fs/promises";
+import { isValidModule, resolveModuleCacheDirs, clearModuleBirCache } from "../utils/bir-cache-recovery";
+
+describe("isValidModule", () => {
+    it("accepts a well-formed coordinate", () => {
+        expect(isValidModule({ org: "ballerina", name: "ai", version: "1.14.1" })).toBe(true);
+        expect(isValidModule({ org: "ballerinax", name: "aws.s3", version: "2.1.0" })).toBe(true);
+    });
+
+    it("rejects missing or empty segments", () => {
+        expect(isValidModule(null)).toBe(false);
+        expect(isValidModule(undefined)).toBe(false);
+        expect(isValidModule({ org: "ballerina", name: "ai" })).toBe(false);
+        expect(isValidModule({ org: "ballerina", name: "ai", version: "" })).toBe(false);
+    });
+
+    it("rejects path-traversal / unsafe segments", () => {
+        expect(isValidModule({ org: "ballerina", name: "ai", version: ".." })).toBe(false);
+        expect(isValidModule({ org: "..", name: "ai", version: "1.0.0" })).toBe(false);
+        expect(isValidModule({ org: "ballerina", name: "a/i", version: "1.0.0" })).toBe(false);
+        expect(isValidModule({ org: "ballerina", name: "ai", version: "1.0.0/../../etc" })).toBe(false);
+    });
+});
+
+describe("clearModuleBirCache (targeted, cache-only)", () => {
+    let home: string;
+    const reposFor = (h: string) => path.join(h, ".ballerina", "repositories");
+
+    // Build a realistic ~/.ballerina tree under a temp home.
+    async function seed(h: string): Promise<void> {
+        const files = [
+            // target: two distribution caches for the corrupt module+version
+            "repositories/central.ballerina.io/cache-2201.13.5/ballerina/ai/1.14.1/bir/ai.bir",
+            "repositories/central.ballerina.io/cache-2201.13.4/ballerina/ai/1.14.1/bir/ai.bir",
+            // keep: other version of same module
+            "repositories/central.ballerina.io/cache-2201.13.5/ballerina/ai/1.13.0/bir/ai.bir",
+            // keep: other module
+            "repositories/central.ballerina.io/cache-2201.13.5/ballerina/io/1.6.0/bir/io.bir",
+            // keep: pulled bala source (never cleared)
+            "repositories/central.ballerina.io/bala/ballerina/ai/1.14.1/java21/pkg.json",
+            // keep: the distribution itself
+            "distributions/ballerina-2201.13.5/bin/bal",
+        ];
+        for (const rel of files) {
+            const full = path.join(h, ".ballerina", rel);
+            await fs.mkdir(path.dirname(full), { recursive: true });
+            await fs.writeFile(full, "x");
+        }
+    }
+
+    const exists = async (rel: string): Promise<boolean> => {
+        try {
+            await fs.stat(path.join(home, ".ballerina", rel));
+            return true;
+        } catch {
+            return false;
+        }
+    };
+
+    beforeEach(async () => {
+        home = await fs.mkdtemp(path.join(os.tmpdir(), "bir-cache-test-"));
+        await seed(home);
+    });
+
+    afterEach(async () => {
+        await fs.rm(home, { recursive: true, force: true });
+    });
+
+    it("scopes to the active distribution's cache-<distVersion> only", async () => {
+        const dirs = await resolveModuleCacheDirs(
+            reposFor(home),
+            { org: "ballerina", name: "ai", version: "1.14.1" },
+            "2201.13.5"
+        );
+        const rels = dirs.map((d) => path.relative(reposFor(home), d));
+        expect(rels).toEqual([path.join("central.ballerina.io", "cache-2201.13.5", "ballerina", "ai", "1.14.1")]);
+    });
+
+    it("considers every cache-* dir when no distVersion is given", async () => {
+        const dirs = await resolveModuleCacheDirs(reposFor(home), { org: "ballerina", name: "ai", version: "1.14.1" });
+        const rels = dirs.map((d) => path.relative(reposFor(home), d)).sort();
+        expect(rels).toEqual([
+            path.join("central.ballerina.io", "cache-2201.13.4", "ballerina", "ai", "1.14.1"),
+            path.join("central.ballerina.io", "cache-2201.13.5", "ballerina", "ai", "1.14.1"),
+        ]);
+    });
+
+    it("removes the corrupt module's cache only under the active distribution, nothing else", async () => {
+        const removed = await clearModuleBirCache(
+            { org: "ballerina", name: "ai", version: "1.14.1" },
+            { distVersion: "2201.13.5", homeDir: home }
+        );
+
+        expect(removed).toHaveLength(1);
+        // target gone from the active distribution's cache
+        expect(await exists("repositories/central.ballerina.io/cache-2201.13.5/ballerina/ai/1.14.1")).toBe(false);
+        // OTHER distribution's cache for the same module is left intact
+        expect(await exists("repositories/central.ballerina.io/cache-2201.13.4/ballerina/ai/1.14.1/bir/ai.bir")).toBe(true);
+        // everything else intact
+        expect(await exists("repositories/central.ballerina.io/cache-2201.13.5/ballerina/ai/1.13.0/bir/ai.bir")).toBe(true);
+        expect(await exists("repositories/central.ballerina.io/cache-2201.13.5/ballerina/io/1.6.0/bir/io.bir")).toBe(true);
+        expect(await exists("repositories/central.ballerina.io/bala/ballerina/ai/1.14.1/java21/pkg.json")).toBe(true);
+        expect(await exists("distributions/ballerina-2201.13.5/bin/bal")).toBe(true);
+    });
+
+    it("removes nothing when the module is not cached", async () => {
+        const removed = await clearModuleBirCache(
+            { org: "ballerina", name: "nope", version: "9.9.9" },
+            { distVersion: "2201.13.5", homeDir: home }
+        );
+        expect(removed).toEqual([]);
+        expect(await exists("repositories/central.ballerina.io/cache-2201.13.5/ballerina/ai/1.14.1/bir/ai.bir")).toBe(true);
+    });
+});
--- a/packages/ballerina-extension/src/test-support/birCacheRecovery.test.ts
+++ b/packages/ballerina-extension/src/test-support/birCacheRecovery.test.ts
@@ -19,20 +19,26 @@
 /**
  * @jest-environment node
  *
- *  A corrupt cached BIR makes projects load empty. The LS sends a
+ * A corrupt cached BIR makes projects load empty. The LS sends a
  * `projectService/corruptBirCache` notification with the affected module's coordinates and the
  * running distribution version; the client clears that module's compiled cache under
- * cache-<distVersion>. These L1 tests pin three invariants: (1) only a well-formed module
- * coordinate is accepted (a malformed/hostile payload can never become an fs path); (2) the clear
- * removes ONLY the affected module's compiled cache and only under the active distribution's
- * cache-<distVersion> — never other distributions, the pulled bala/, other modules or versions, or
- * the distribution itself.
+ * cache-<distVersion>. These L1 tests pin two invariants: (1) only a well-formed module coordinate
+ * is accepted (a malformed/hostile payload can never become an fs path); (2) the clear removes ONLY
+ * the affected module's compiled cache and only under the active distribution's cache-<distVersion>
+ * — never other distributions, the pulled bala/, other modules or versions, or the distribution
+ * itself.
+ *
+ * The clear invariant is table-driven from JSON fixtures under fixtures/bir-cache/ (see
+ * docs/TEST_PLAN.md §5). Each fixture materializes a realistic ~/.ballerina tree in a temp home,
+ * runs the real clear against it, and asserts what is resolved, removed, and kept. Add a fixture
+ * (e.g. issue-2251.json) to guard a new case.
  */
 
 import * as os from "os";
 import * as path from "path";
 import * as fs from "fs/promises";
-import { isValidModule, resolveModuleCacheDirs, clearModuleBirCache } from "../utils/bir-cache-recovery";
+import { loadFixtures } from "@wso2/test-config/fixtures";
+import { CorruptModule, isValidModule, resolveModuleCacheDirs, clearModuleBirCache } from "../utils/bir-cache-recovery";
 
 describe("isValidModule", () => {
     it("accepts a well-formed coordinate", () => {
@@ -55,93 +61,66 @@ describe("isValidModule", () => {
     });
 });
 
-describe("clearModuleBirCache (targeted, cache-only)", () => {
-    let home: string;
-    const reposFor = (h: string) => path.join(h, ".ballerina", "repositories");
+// A single clear scenario: seed `tree` under a temp ~/.ballerina, clear `module` (optionally scoped
+// to `distVersion`), then assert the resolved/removed/kept paths. All paths are relative to the
+// .ballerina root and use "/" separators (normalized per-OS below).
+interface BirCacheFixture {
+    description?: string;
+    module: CorruptModule;
+    distVersion?: string;
+    tree: string[];
+    expectedResolved?: string[];
+    expectedRemoved: string[];
+    expectedKept: string[];
+}
 
-    // Build a realistic ~/.ballerina tree under a temp home.
-    async function seed(h: string): Promise<void> {
-        const files = [
-            // target: two distribution caches for the corrupt module+version
-            "repositories/central.ballerina.io/cache-2201.13.5/ballerina/ai/1.14.1/bir/ai.bir",
-            "repositories/central.ballerina.io/cache-2201.13.4/ballerina/ai/1.14.1/bir/ai.bir",
-            // keep: other version of same module
-            "repositories/central.ballerina.io/cache-2201.13.5/ballerina/ai/1.13.0/bir/ai.bir",
-            // keep: other module
-            "repositories/central.ballerina.io/cache-2201.13.5/ballerina/io/1.6.0/bir/io.bir",
-            // keep: pulled bala source (never cleared)
-            "repositories/central.ballerina.io/bala/ballerina/ai/1.14.1/java21/pkg.json",
-            // keep: the distribution itself
-            "distributions/ballerina-2201.13.5/bin/bal",
-        ];
-        for (const rel of files) {
-            const full = path.join(h, ".ballerina", rel);
-            await fs.mkdir(path.dirname(full), { recursive: true });
-            await fs.writeFile(full, "x");
-        }
-    }
+const fixtures = loadFixtures<BirCacheFixture>(__dirname, "fixtures", "bir-cache");
 
-    const exists = async (rel: string): Promise<boolean> => {
+describe("clearModuleBirCache (targeted, cache-only) — fixtures", () => {
+    const toOsPath = (rel: string): string => path.join(...rel.split("/"));
+
+    it("has fixtures to run", () => {
+        expect(fixtures.length).toBeGreaterThan(0);
+    });
+
+    it.each(fixtures.map((f) => [f.name, f.data] as [string, BirCacheFixture]))("%s", async (_name, fx) => {
+        const home = await fs.mkdtemp(path.join(os.tmpdir(), "bir-cache-test-"));
+        const ballerinaDir = path.join(home, ".ballerina");
+        const reposDir = path.join(ballerinaDir, "repositories");
+        const relToBallerina = (abs: string): string => path.relative(ballerinaDir, abs);
+        const exists = async (rel: string): Promise<boolean> => {
+            try {
+                await fs.stat(path.join(ballerinaDir, toOsPath(rel)));
+                return true;
+            } catch {
+                return false;
+            }
+        };
+
         try {
-            await fs.stat(path.join(home, ".ballerina", rel));
-            return true;
-        } catch {
-            return false;
+            // Seed the realistic ~/.ballerina tree.
+            for (const rel of fx.tree) {
+                const full = path.join(ballerinaDir, toOsPath(rel));
+                await fs.mkdir(path.dirname(full), { recursive: true });
+                await fs.writeFile(full, "x");
+            }
+
+            if (fx.expectedResolved) {
+                const resolved = await resolveModuleCacheDirs(reposDir, fx.module, fx.distVersion);
+                expect(resolved.map(relToBallerina).sort()).toEqual(fx.expectedResolved.map(toOsPath).sort());
+            }
+
+            const removed = await clearModuleBirCache(fx.module, { distVersion: fx.distVersion, homeDir: home });
+            expect(removed.map(relToBallerina).sort()).toEqual(fx.expectedRemoved.map(toOsPath).sort());
+
+            for (const rel of fx.expectedRemoved) {
+                expect(await exists(rel)).toBe(false);
+            }
+            for (const rel of fx.expectedKept) {
+                expect(await exists(rel)).toBe(true);
+            }
+        } finally {
+            await fs.rm(home, { recursive: true, force: true });
         }
-    };
-
-    beforeEach(async () => {
-        home = await fs.mkdtemp(path.join(os.tmpdir(), "bir-cache-test-"));
-        await seed(home);
-    });
-
-    afterEach(async () => {
-        await fs.rm(home, { recursive: true, force: true });
-    });
-
-    it("scopes to the active distribution's cache-<distVersion> only", async () => {
-        const dirs = await resolveModuleCacheDirs(
-            reposFor(home),
-            { org: "ballerina", name: "ai", version: "1.14.1" },
-            "2201.13.5"
-        );
-        const rels = dirs.map((d) => path.relative(reposFor(home), d));
-        expect(rels).toEqual([path.join("central.ballerina.io", "cache-2201.13.5", "ballerina", "ai", "1.14.1")]);
-    });
-
-    it("considers every cache-* dir when no distVersion is given", async () => {
-        const dirs = await resolveModuleCacheDirs(reposFor(home), { org: "ballerina", name: "ai", version: "1.14.1" });
-        const rels = dirs.map((d) => path.relative(reposFor(home), d)).sort();
-        expect(rels).toEqual([
-            path.join("central.ballerina.io", "cache-2201.13.4", "ballerina", "ai", "1.14.1"),
-            path.join("central.ballerina.io", "cache-2201.13.5", "ballerina", "ai", "1.14.1"),
-        ]);
-    });
-
-    it("removes the corrupt module's cache only under the active distribution, nothing else", async () => {
-        const removed = await clearModuleBirCache(
-            { org: "ballerina", name: "ai", version: "1.14.1" },
-            { distVersion: "2201.13.5", homeDir: home }
-        );
-
-        expect(removed).toHaveLength(1);
-        // target gone from the active distribution's cache
-        expect(await exists("repositories/central.ballerina.io/cache-2201.13.5/ballerina/ai/1.14.1")).toBe(false);
-        // OTHER distribution's cache for the same module is left intact
-        expect(await exists("repositories/central.ballerina.io/cache-2201.13.4/ballerina/ai/1.14.1/bir/ai.bir")).toBe(true);
-        // everything else intact
-        expect(await exists("repositories/central.ballerina.io/cache-2201.13.5/ballerina/ai/1.13.0/bir/ai.bir")).toBe(true);
-        expect(await exists("repositories/central.ballerina.io/cache-2201.13.5/ballerina/io/1.6.0/bir/io.bir")).toBe(true);
-        expect(await exists("repositories/central.ballerina.io/bala/ballerina/ai/1.14.1/java21/pkg.json")).toBe(true);
-        expect(await exists("distributions/ballerina-2201.13.5/bin/bal")).toBe(true);
-    });
-
-    it("removes nothing when the module is not cached", async () => {
-        const removed = await clearModuleBirCache(
-            { org: "ballerina", name: "nope", version: "9.9.9" },
-            { distVersion: "2201.13.5", homeDir: home }
-        );
-        expect(removed).toEqual([]);
-        expect(await exists("repositories/central.ballerina.io/cache-2201.13.5/ballerina/ai/1.14.1/bir/ai.bir")).toBe(true);
     });
 });
--- a/packages/ballerina-extension/src/test-support/birCacheRecovery.test.ts
+++ b/packages/ballerina-extension/src/test-support/birCacheRecovery.test.ts
@@ -20,13 +20,14 @@
  * @jest-environment node
  *
  * A corrupt cached BIR makes projects load empty. The LS sends a
- * `projectService/corruptBirCache` notification with the affected module's coordinates and the
- * running distribution version; the client clears that module's compiled cache under
- * cache-<distVersion>. These L1 tests pin two invariants: (1) only a well-formed module coordinate
- * is accepted (a malformed/hostile payload can never become an fs path); (2) the clear removes ONLY
- * the affected module's compiled cache and only under the active distribution's cache-<distVersion>
- * — never other distributions, the pulled bala/, other modules or versions, or the distribution
- * itself.
+ * `projectService/corruptBirCache` notification with the affected *package's* coordinates and the
+ * running distribution version; the client clears that package's compiled cache under
+ * cache-<distVersion>. (The cache is keyed by package, not module: a package's submodule BIRs all
+ * live under `cache-<dist>/<org>/<packageName>/<version>/bir/` — see the submodule fixture.) These
+ * L1 tests pin two invariants: (1) only a well-formed package coordinate is accepted (a
+ * malformed/hostile payload can never become an fs path); (2) the clear removes ONLY the affected
+ * package's compiled cache and only under the active distribution's cache-<distVersion> — never
+ * other distributions, the pulled bala/, other packages or versions, or the distribution itself.
  *
  * The clear invariant is table-driven from JSON fixtures under fixtures/bir-cache/ (see
  * docs/TEST_PLAN.md §5). Each fixture materializes a realistic ~/.ballerina tree in a temp home,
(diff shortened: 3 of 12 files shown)


packages/ballerina-extension/src/core/extended-language-client.ts
```diff
@@ -657,6 +659,14 @@ export class ExtendedLangClient extends LanguageClient implements ExtendedLangCl
         });
     }
 
+    registerCorruptBirCache(): void {
+        // A corrupt/incompatible cached BIR makes the project load empty. The LS
+        // reports the affected module here; offer to clear just that module's cache and reload.
+        this.onNotification(EXTENDED_APIS.CORRUPT_BIR_CACHE, (res: CorruptBirCachePayload) => {
```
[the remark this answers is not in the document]
thuva9872: The handler here is a thin pass-through with no logic — payload validation, path-scoping, and cache clearing are all covered by the L1 tests in birCacheRecovery.test.ts. A unit test mocking onNotification would only assert the TS side calls the helper with the same object, which the type system already guarantees, and it wouldn't verify the actual cross-process contract (that the server's @JsonNotification string matches the client constant) — that's only meaningful in an E2E test.

packages/ballerina-language-server/langserver-commons/src/main/java/org/ballerinalang/langserver/commons/client/ExtendedLanguageClient.java
```diff
@@ -44,4 +44,7 @@ public interface ExtendedLanguageClient extends LanguageClient {
 
     @JsonNotification("projectService/pushMigratedProject")
     void pushMigratedProject(Object notification);
+
+    @JsonNotification("projectService/corruptBirCache")
+    void corruptBirCache(Object notification);
```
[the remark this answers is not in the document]
thuva9872: ExtendedLanguageClient is an lsp4j client proxy — the server side is a runtime-generated proxy that always reflects the current interface, and the actual client (the VS Code extension) consumes the notification over JSON-RPC without implementing this Java interface. There's no separately-compiled hand-written implementor that could hit AbstractMethodError.

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
```diff
@@ -0,0 +1,192 @@
...
+function isWithin(parent: string, child: string): boolean {
+    const rel = path.relative(parent, child);
+    return !!rel && !rel.startsWith("..") && !path.isAbsolute(rel);
+}
+
+function cacheDirMatches(cacheDir: string, distVersion?: string): boolean {
+    if (!cacheDir.startsWith("cache-")) {
+        return false;
+    }
+    return !distVersion || cacheDir === `cache-${distVersion}`;
```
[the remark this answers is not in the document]
thuva9872: distVersion is always populated server-side — PullModuleExecutor unconditionally calls params.setDistVersion(RepoUtils.getBallerinaShortVersion()) before sending the notification, so the "absent/invalid" branch is only a defensive fallback, not a live path. Requiring a valid distVersion before removal would mean that in the (unexpected) case it's missing, we clear nothing and leave the user with the corrupt cache and empty project — i.e. we'd fail to recover, which is the exact bug this fixes. The BIR cache is regenerable build output (clearing extra cache-* dirs just forces a recompile, no data loss), and even without distVersion the clear stays scoped to the specific module and within repositories/*/cache-*, never touching bala/ sources or distributions/. I'd keep the current behavior.

packages/ballerina-extension/src/test-support/birCacheRecovery.test.ts
```diff
@@ -0,0 +1,147 @@
...
+        expect(isValidModule({ org: "ballerina", name: "ai", version: "1.0.0/../../etc" })).toBe(false);
+    });
+});
+
+describe("clearModuleBirCache (targeted, cache-only)", () => {
+    let home: string;
+    const reposFor = (h: string) => path.join(h, ".ballerina", "repositories");
+
+    // Build a realistic ~/.ballerina tree under a temp home.
+    async function seed(h: string): Promise<void> {
```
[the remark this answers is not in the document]
thuva9872: addressed in [caf5878](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/caf58780c672739e35ca8b586c9ac8197c649380)

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
```diff
@@ -288,6 +290,13 @@ public static CompletableFuture<Void> resolveModules(String fileUri, ExtendedLan
                         clientLogger.logError(LSContextOperation.WS_EXEC_CMD,
                                 "Pull modules failed for project: " + project.sourceRoot().toString(),
                                 t, null, (Position) null);
+                        Optional<CorruptBirCacheParams> corruptBir = detectCorruptBirCache(t);
+                        if (corruptBir.isPresent()) {
+                            CorruptBirCacheParams params = corruptBir.get();
+                            params.setProjectUri(project.sourceRoot().toUri().toString());
+                            params.setDistVersion(RepoUtils.getBallerinaShortVersion());
+                            languageClient.corruptBirCache(params);
+                        }
```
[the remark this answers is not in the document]
thuva9872: Resolved in [b78e466](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/b78e466a275661518a83f1bbf7be2578fab8306c)

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
```diff
@@ -0,0 +1,192 @@
...
+    distVersion?: string
+): Promise<string[]> {
+    const targets: string[] = [];
+    for (const repo of await listSubDirs(reposDir)) {
+        const repoDir = path.join(reposDir, repo);
+        for (const cacheDir of await listSubDirs(repoDir)) {
+            if (!cacheDirMatches(cacheDir, distVersion)) {
+                continue;
+            }
+            const target = path.join(repoDir, cacheDir, module.org, module.name, module.version);
```
pasindufernando1: `module.name` is the **module** name here, but this path segment needs the **package** name.

The payload's `name` is parsed out of the LS diagnostic, which is built from `PackageID.toString()` — that renders `<org>/<module>:<version>`. The cache directory, though, is keyed by the package: `FileSystemCache.getBirPath()` resolves `packageCacheDirPath()` as `org / PackageManifest.name() / version`, and the BIR itself lands at `<version>/bir/<module>.bir`.

For a default module the two names coincide, so `ballerina/http:2.16.6` works. For any submodule they diverge — verified against a real `~/.ballerina`:

```
actual    repositories/central.ballerina.io/cache-2201.13.4/ballerina/http/2.16.6/bir/http.httpscerr.bir
resolved  repositories/central.ballerina.io/cache-2201.13.4/ballerina/http.httpscerr/2.16.6      <- does not exist
```

`removeIfExists` then returns `false`, so `removed.length === 0` and `promptClearCorruptBirCache` falls through to `clearAllBirCaches` — the submodule case silently wipes the whole `cache-<distVersion>` tree instead of the one module, which is the outcome this PR is meant to replace.

Worth noting that a client-side `name.split(".")[0]` won't fix it: package names may themselves contain dots (`ballerina/data.jsondata`, `ballerina/ai.eval` are both real package dirs), so the package/module boundary can't be recovered lexically from the module name alone.

Two ways out, in preference order:

1. Send the package coordinate alongside the module name in `CorruptBirCacheParams` — the LS can resolve the owning package from the compilation's dependency graph, where both names are known — and use the package name for this segment.
2. Failing that, have the client pick the `<org>/<pkg>/<version>` dir whose `bir/` actually contains `<module>.bir`, rather than inferring the `<pkg>` segment.

Either way it'd be worth a fixture covering a dotted submodule, since the current tests only exercise coordinates where package and module names match.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
```diff
@@ -324,6 +337,71 @@ public static CompletableFuture<Void> resolveModules(String fileUri, ExtendedLan
...
+     * Detects a corrupt/incompatible cached-BIR failure anywhere in the throwable's cause chain and,
+     * when found, returns the coordinates of the affected package (best-effort; coordinates may be
+     * {@code null} if they cannot be parsed from the message).
+     *
+     * @param throwable the completion throwable to inspect
+     * @param project   the project whose resolution maps the failing module to its package
+     * @return the corrupt-BIR parameters, or empty if this is not a corrupt-BIR failure
+     */
+    static Optional<CorruptBirCacheParams> detectCorruptBirCache(Throwable throwable, Project project) {
+        for (Throwable cause = throwable; cause != null && cause != cause.getCause(); cause = cause.getCause()) {
```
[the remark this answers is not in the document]
thuva9872: The current loop already inspects the terminal cause for every realistic exception chain, and that's proven by the passing test (PullModuleExecutorCorruptBirTest 7/7, including the case where the corrupt message sits on the terminal cause). The only scenario the finding's premise would apply to is a nonstandard getCause() that returns this — which the JDK contract prevents and the compiler's exceptions don't do — and the suggested while (cause != null) rewrite would actually introduce an infinite-loop risk on that exact input.

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
```diff
@@ -0,0 +1,196 @@
...
+    distVersion?: string
+): Promise<string[]> {
+    const targets: string[] = [];
+    for (const repo of await listSubDirs(reposDir)) {
+        const repoDir = path.join(reposDir, repo);
+        for (const cacheDir of await listSubDirs(repoDir)) {
+            if (!cacheDirMatches(cacheDir, distVersion)) {
+                continue;
+            }
+            const target = path.join(repoDir, cacheDir, pkg.org, pkg.packageName, pkg.version);
```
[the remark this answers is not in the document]
thuva9872: fixed in [20fa170](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/20fa170e54c1aa87c9b7226c5a1a754cc752ed05)

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
thuva9872: resolved in [0c58000](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/0c58000a61e6b3a9f961f2145337f48f95ecca07)

pasindufernando1 (commented):
Re-review: three blocking issues, all in `bir-cache-recovery.ts`. Remaining findings (BIR detection missing the `unsupported program file version` / truncated-BIR variants, reload-on-zero-removals, symlinked repo dirs, cause-cycle guard) are non-blocking and can follow up.

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
```diff
@@ -0,0 +1,199 @@
...
+    try {
+        realReposDir = await fs.realpath(reposDir);
+        realDir = await fs.realpath(dir);
+    } catch {
+        return false; // missing target or unresolvable path — nothing safe to remove
+    }
+    if (!isWithin(realReposDir, realDir)) {
+        return false; // target escapes the repositories root via a symlinked component
+    }
+    await fs.rm(realDir, { recursive: true, force: true });
```
pasindufernando1: **Blocker — unhandled `fs.rm` rejection.**

On Windows the JVM holds handles on `cache-<v>/<org>/<pkg>/<ver>/java21/*.jar` and `bir/*.bir`, so `rm` rejects with EPERM/EBUSY (`force: true` only suppresses ENOENT). Nothing catches it, and `promptClearCorruptBirCache` is invoked as `void promptClearCorruptBirCache(res)` in `extended-language-client.ts` — so the rejection is unhandled: no message to the user, no reload, and a half-deleted cache dir that is worse than the original corruption.

Wrap the clear + reload in try/catch with a user-visible failure, and pass `{ maxRetries, retryDelay }` to `fs.rm`.

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
```diff
@@ -0,0 +1,199 @@
...
+        if (choice !== action) {
+            return;
+        }
+
+        const removed = target
+            ? await clearPackageBirCache(target, { distVersion })
+            : await clearAllBirCaches({ distVersion });
+        // If the targeted clear matched nothing (unexpected layout), fall back to clearing the whole
+        // distribution cache so the user still recovers rather than reloading into the same state.
+        if (target && removed.length === 0) {
```
pasindufernando1: **Blocker — this fallback escalates the user's consent.**

The user agreed to clear one module; this deletes every `repositories/*/cache-<dist>` — all BIR *and* platform jars for the active distribution, forcing a full recompile of every dependency of every project.

It is not a rare path: it fires whenever the package name is wrong, e.g. the submodule case where `resolvePackageName` fails and the LS falls back to the module name `ai.observe` (pinned by the new `testUnresolvedModuleFallsBackToModuleName`). The client then probes `cache-<v>/ballerina/ai.observe/1.14.1`, finds nothing, and wipes the whole cache.

Either state this in the prompt or confirm separately before the broad wipe.

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
```diff
@@ -0,0 +1,199 @@
...
+function isSafeSegment(segment: unknown): segment is string {
+    return typeof segment === "string" && segment !== "." && segment !== ".." && SAFE_SEGMENT.test(segment);
+}
+
+/** True when the notification payload names a package safe to turn into cache paths. */
+export function isValidPackage(pkg: Partial<CorruptPackage> | null | undefined): pkg is CorruptPackage {
+    return !!pkg && isSafeSegment(pkg.org) && isSafeSegment(pkg.packageName) && isSafeSegment(pkg.version);
+}
+
+function reposDirFor(homeDir: string): string {
```
pasindufernando1: **Blocker — `BALLERINA_HOME_DIR` is ignored.**

Ballerina resolves its home via `RepoUtils.createAndGetHomeReposPath()`, which prefers `$BALLERINA_HOME_DIR` (`ProjectDirConstants.HOME_REPO_ENV_KEY`) and only then `user.home` (confirmed in the shipped 2201.13.0 bytecode). With that env var set — containerized BI runtime, CI, relocated home — the targeted clear finds nothing, the fallback finds nothing, and the window reloads into the same empty project after the user was told the cache was cleared.

The LS already knows the real path; send it in `CorruptBirCacheParams` rather than guessing client-side. Worth doing before the message shape ships.

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
thuva9872: addressed in [0ec6ced](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/0ec6ced278a1bf389a9a0d230e649518df008e99)

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
thuva9872: addressed in [0ec6ced](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/0ec6ced278a1bf389a9a0d230e649518df008e99)

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
thuva9872: addressed in [0ec6ced](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/0ec6ced278a1bf389a9a0d230e649518df008e99)

pasindufernando1 (dismissed):
LGTM — all three blockers are addressed in 0ec6ced, and I verified each:

- **`fs.rm` rejection**: `{ maxRetries: 3, retryDelay: 100 }` plus a try/catch that surfaces a modal error and returns instead of reloading, so the rejection can no longer escape the `void promptClearCorruptBirCache(...)` call.
- **Consent escalation**: the prompt now states the broader scope up front — "If the module's cache can't be located, all module caches for this distribution are cleared."
- **`BALLERINA_HOME_DIR`**: `reposPath` is now sent from the LS. Confirmed against the 2201.13.4 bytecode that `RepoUtils.createAndGetHomeReposPath()` returns the *home* dir (`$BALLERINA_HOME_DIR`, else `user.home/.ballerina`), so `.resolve(ProjectConstants.REPOSITORIES_DIR)` is correct rather than a doubled segment, and the client's env fallback matches those semantics.

Two non-blocking follow-ups, at your discretion:

1. No tests cover the new `reposDir` precedence, the `BALLERINA_HOME_DIR` fallback, or the failure path.
2. When the targeted clear *and* the fallback both remove nothing, the window still reloads silently into the same broken state — worth an error naming the path searched instead.

madushajg (dismissed):

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
```diff
@@ -0,0 +1,234 @@
...
+            const target = path.join(repoDir, cacheDir);
+            if (await removeIfExists(target, reposDir)) {
+                removed.push(target);
+            }
+        }
+    }
+    return removed;
+}
+
+let promptShown = false; // don't stack a prompt per repeated notification
```
samithkavishke, line 165: **Severity: HIGH**

`promptShown` only dedupes concurrent notifications within the current extension host's lifetime, and is reset in the `finally` block after every attempt — success or failure. Since the fix action itself reloads the window, this flag can never actually observe "we already tried this and it came back." If the corrupt-BIR condition recurs shortly after a clear (e.g. via the `.tmp`-staging race in ballerina-lang's `FileSystemCache.cacheBir` — a single hardcoded temp file shared across concurrent BIR writes, unrelated to any stale/incompatible Central package — or a persistently hostile local environment: AV interference, near-full disk), the user gets the exact same "Clear cache & reload" prompt indefinitely, with nothing ever surfacing that the fix isn't working.

Suggest persisting a small marker via `ballerinaExtInstance.context.globalState` (already used this way elsewhere in this codebase, e.g. `file-utils.ts`, for state that must survive a reload) keyed by `org/packageName/version`, storing the timestamp of the last clear. If the same coordinate is reported again within some window after a recorded clear, skip the "clear & reload" action and show a message that this looks like a recurring issue (possible disk/concurrency problem, not a stale cache) with a suggestion to file an issue — rather than repeating a fix that evidently isn't fixing anything.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
```diff
@@ -288,6 +294,15 @@ public static CompletableFuture<Void> resolveModules(String fileUri, ExtendedLan
                         clientLogger.logError(LSContextOperation.WS_EXEC_CMD,
                                 "Pull modules failed for project: " + project.sourceRoot().toString(),
                                 t, null, (Position) null);
+                        Optional<CorruptBirCacheParams> corruptBir = detectCorruptBirCache(t, project);
```
samithkavishke, line 303: **Severity: MEDIUM**

This existing file already fixed one version of this problem: `PULL_IN_PROGRESS_PROJECTS` (declared a few lines above, guard checked at the top of `resolveModules`) exists specifically because "the PROJECT_UPDATE event published by a running pull re-triggers another pull... causing an endless pull storm when pulls fail." That guard only protects re-entrancy *within a single LS process* though. `promptClearCorruptBirCache` (client side) calls `workbench.action.reloadWindow`, which restarts this very process, so it gives no protection against: clear cache → reload → recompile → corrupt again → `detectCorruptBirCache` fires again → same prompt. Also worth noting as a root-cause concern: ballerina-lang's `FileSystemCache.cacheBir()` stages every cached BIR through one shared, non-per-module `.tmp` file — concurrent compilation (parallel modules, or two windows sharing `$BALLERINA_HOME/repositories`) can race on it and produce a torn `.bir` that trips this exact detector with no relation to a stale/bad Central package. Re-pulling does nothing for that class of cause. See the companion comment on `bir-cache-recovery.ts` for a suggested cross-reload guard.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
```diff
@@ -324,6 +339,71 @@ public static CompletableFuture<Void> resolveModules(String fileUri, ExtendedLan
...
+                continue;
+            }
+            Matcher matcher = CORRUPT_BIR_MODULE_PATTERN.matcher(message);
+            if (!matcher.find()) {
+                return Optional.of(new CorruptBirCacheParams(null, null, null));
+            }
+            String org = matcher.group(1);
+            String moduleName = matcher.group(2);
+            String version = matcher.group(3);
+            String packageName = resolvePackageName(project, org, moduleName, version).orElse(moduleName);
```
samithkavishke, line 376: **Severity: MEDIUM**

`resolvePackageName` only searches `project.currentPackage().getResolution().dependencyGraph()` — i.e. the root/first member package. In a multi-package workspace, `resolveModules` (this same file) explicitly builds a separate `memberProjects` list because `project.currentPackage()` only returns the first member. If the corrupt BIR belongs to a dependency reachable only from the 2nd+ member package, this lookup never finds it, `resolvePackageName` returns empty, and the caller falls back to the raw module name as the package name — wrong for a submodule (e.g. sends `ai.observe` instead of `ai`). The client's targeted `clearPackageBirCache` then won't match anything real and falls back to a full distribution-cache wipe instead of the intended single-module fix.

Question: should this iterate over all member projects, not just `project.currentPackage()`?

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
```diff
@@ -324,6 +339,71 @@ public static CompletableFuture<Void> resolveModules(String fileUri, ExtendedLan
                 });
     }
 
+    // The compiler's BIR reader throws with this signature when a cached BIR is corrupt/incompatible,
+    // e.g. "failed to load the module 'ballerina/ai:1.14.1' from its BIR due to: invalid magic number [...]".
+    private static final Pattern CORRUPT_BIR_MODULE_PATTERN = Pattern.compile(
```
samithkavishke, line 351: **Severity: LOW**

This pattern requires both an org segment (`org/`) and a version segment (`:version`). But `PackageID#toString()` (ballerina-lang) omits the org entirely for an anonymous-org package and omits the version entirely when it's empty — verified in `PackageID.java:259-274`. In that case the compiler's message would read like `failed to load the module 'ai' from its BIR...` with no `/` or `:`, `matcher.find()` returns false, and detection degrades to the coordinate-less fallback (generic message, full-cache wipe) instead of the targeted single-module recovery. Likely a rare edge case in practice — flagging for awareness rather than blocking.

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
```diff
@@ -0,0 +1,234 @@
...
+
+        try {
+            const removed = target
+                ? await clearPackageBirCache(target, { distVersion, reposDir })
+                : await clearAllBirCaches({ distVersion, reposDir });
+            // The targeted clear matched nothing — the corrupt cache is on disk (that is why the LS
+            // reported it), but the coordinates did not resolve to it, e.g. a submodule where the LS
+            // fell back to the module name instead of the package name. Clear the whole distribution
+            // cache so the user still recovers; the prompt above states this broader scope up front.
+            if (target && removed.length === 0) {
```
samithkavishke, line 249: **Severity: LOW**

Worth combining with the `promptShown` comment above: if this fallback path (full distribution cache wipe) still doesn't resolve the issue and the same corrupt-BIR notification comes back, that's an even stronger signal this isn't a cache-staleness problem at all. The same repeat-occurrence tracking should cover this path too, not just the targeted-clear path above.

samithkavishke (changes requested):

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
thuva9872: Clearing removes the corrupt BIR and the next build regenerates it deterministically from the bala, so there's no recurring corruption to loop on — after clear+reload the project builds a fresh, valid cache (and works cacheless until then). So a cross-reload guard isn't needed here.

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
thuva9872: refer https://github.com/ballerina-platform/ballerina-vscode/pull/1013#discussion_r3977188677

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
thuva9872: For the fallback specifically: it wipes the whole cache-<distVersion> tree, so everything recompiles fresh from balas on reload.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
thuva9872: This detector only fires for corrupt BIRs in the central repository cache (`cache-<distVersion>/<org>/<pkg>/<version>`), whose dependencies always carry an org and version — anonymous-org/empty-version packages are local code that never lands in that cache. And even if such a message reached here, without org+version there's no cache path to target, so the coordinate-less fallback (full-dist cache wipe) is the only possible recovery anyway.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
thuva9872: resolved in [953f2cd](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/953f2cdc78aafcba2de8e847e71f3a65f1082675)

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
```diff
@@ -288,6 +294,15 @@ public static CompletableFuture<Void> resolveModules(String fileUri, ExtendedLan
...
                                 "Pull modules failed for project: " + project.sourceRoot().toString(),
                                 t, null, (Position) null);
+                        Optional<CorruptBirCacheParams> corruptBir = detectCorruptBirCache(t, project);
+                        if (corruptBir.isPresent()) {
+                            CorruptBirCacheParams params = corruptBir.get();
+                            params.setProjectUri(project.sourceRoot().toUri().toString());
+                            params.setDistVersion(RepoUtils.getBallerinaShortVersion());
+                            params.setReposPath(RepoUtils.createAndGetHomeReposPath()
+                                    .resolve(ProjectConstants.REPOSITORIES_DIR).toString());
+                            languageClient.corruptBirCache(params);
```
samithkavishke, line 312: **[MEDIUM] Duplicate/competing user notifications on the same failure**

When `detectCorruptBirCache` finds a match, we call `languageClient.corruptBirCache(params)` (triggers a blocking modal client-side) but then fall through unconditionally into the existing `if (t.getCause() instanceof UserErrorException) {...} else { CommandUtil.notifyClient(..., "Failed to pull modules!") }` block, firing a second, non-modal toast for the same failure. A user hitting a corrupt BIR cache will see both the "Clear cache & reload" modal and a generic "Failed to pull modules!" toast pop up together. Consider `return`ing after the corrupt-BIR notification so the generic path is skipped.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
```diff
@@ -324,6 +339,77 @@ public static CompletableFuture<Void> resolveModules(String fileUri, ExtendedLan
...
+     * {@code null} if they cannot be parsed from the message).
+     *
+     * @param throwable the completion throwable to inspect
+     * @param project   the project whose resolution maps the failing module to its package
+     * @return the corrupt-BIR parameters, or empty if this is not a corrupt-BIR failure
+     */
+    static Optional<CorruptBirCacheParams> detectCorruptBirCache(Throwable throwable, Project project) {
+        for (Throwable cause = throwable; cause != null && cause != cause.getCause(); cause = cause.getCause()) {
+            String message = cause.getMessage();
+            if (message == null || !message.contains("invalid magic number") || !message.contains("BIR")) {
```
samithkavishke, line 366: **[LOW] Case-sensitive substring check next to a case-insensitive regex for the same message**

`!message.contains("invalid magic number") || !message.contains("BIR")` is case-sensitive, while the regex used a couple lines below for the same message is compiled with `Pattern.CASE_INSENSITIVE`. If the compiler's error text casing ever changes, detection silently falls back to the generic error path — low severity since it degrades gracefully, but worth aligning for consistency.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
```diff
@@ -324,6 +339,77 @@ public static CompletableFuture<Void> resolveModules(String fileUri, ExtendedLan
...
+                        continue;
+                    }
+                    for (Module module : pkg.modules()) {
+                        if (module.moduleName().toString().equals(moduleName)) {
+                            return Optional.of(pkg.packageName().value());
+                        }
+                    }
+                }
+            }
+        } catch (Throwable e) {
```
samithkavishke, line 433: **[LOW] Broad `catch (Throwable e)` in `resolvePackageName(...)` masks resolution failures**

On any resolution failure it falls back to using the raw module name as the package name. For a submodule (e.g. `ai.observe`) that fallback won't match any real cache directory, so the client's targeted `clearPackageBirCache` matches nothing and silently escalates to wiping the entire distribution cache — even for genuine bugs in dependency-graph traversal, not just "package not found". Mitigated by the client-side full-wipe fallback, but there's no logging when this happens, and the "clear only the affected module" promise degrades more often than necessary.

packages/ballerina-extension/src/stateMachine.ts
```diff
@@ -583,6 +583,8 @@ const stateMachine = createMachine<MachineContext>(
                 try {
                     // Register the event driven listener to get the artifact changes
                     context.langClient.registerPublishArtifacts();
+                    // Register the corrupt-BIR-cache listener to recover from corrupted BIR cache failures
+                    context.langClient.registerCorruptBirCache();
```
samithkavishke, line 587: **[LOW] `registerCorruptBirCache()` can be registered more than once per session**

This sits in `registerProjectArtifactsStructure`, reached via `activateLS`, which several transitions in this same state machine loop back into. Each re-entry stacks another `onNotification` handler on the same long-lived language client. The module-level `promptShown` guard prevents multiple modals from stacking, but doesn't prevent redundant handler invocations (extra `clearPackageBirCache`/`clearAllBirCaches`/`reloadWindow` calls) on mid-session re-entry. This mirrors the pre-existing `registerPublishArtifacts()` pattern on the line above, so it's not a new regression introduced by this PR — flagging for awareness rather than as a blocker.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
thuva9872: The matched text is a fixed compiler message (... from its BIR due to: invalid magic number) with stable casing — lowercase prose plus the BIR acronym — so the case-sensitive contains checks match it faithfully and there's no realistic casing drift to guard against. I'd rather keep the checks matching the literal message than add case-insensitivity that only obscures what's actually being matched. Leaving as-is.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
thuva9872: The broad catch (Throwable) is intentional: resolvePackageName runs inside the whenComplete recovery path, and its only job is a best-effort module→package mapping so the corrupt-BIR notification can still be sent. Letting an exception escape would break the recovery this feature provides, so catching broadly and falling back to the module name is the deliberate exception-safe posture.
On the "promise degrades silently" point — it isn't silent from the user's side: the confirmation modal explicitly states "If the module's cache can't be located, all module caches for this distribution are cleared," so the broader scope is disclosed upfront and accepted before anything is wiped. The user still recovers either way. Leaving as-is.

packages/ballerina-extension/src/stateMachine.ts
thuva9872: onNotification doesn't stack handlers — it's keyed by method name in a Map, so re-registering the corruptBirCache handler on each activateLS re-entry just overwrites the previous one rather than adding another. The notification fires a single handler, so promptClearCorruptBirCache runs exactly once per notification regardless of how many times it's registered. This mirrors the existing registerPublishArtifacts call right above it, which relies on the same behavior. No change needed.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
thuva9872: fixed in [f0b448c](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/f0b448c6bb6c196d9c54e2be04283954c1bfe91e)

samithkavishke (dismissed):
LGTM. Better to add regression test for  resolvePackageName missing dependencies issue

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
```diff
@@ -0,0 +1,269 @@
...
+                `Failed to clear the corrupted module cache: ${reason}. ` +
+                    `Close any running Ballerina processes and try again, or delete the cache manually.`
+            );
+            return;
+        }
+
+        await commands.executeCommand("workbench.action.reloadWindow");
+    } finally {
+        // Reset so a later occurrence can prompt again if the window was not reloaded (dismissed/failed).
+        promptShown = false;
```
samithkavishke, line 266: **HIGH**: `promptShown` is an in-memory flag reset here in the `finally` block, so it can't detect that a clear+reload already happened and the corruption recurred. If the root cause isn't a stale/incompatible cache (e.g. a concurrent write race on the BIR cache, or a disk-full/AV-locked environment), the user will see the identical "Clear cache & reload" modal indefinitely with no escalation path (e.g. to "Report an Issue") after N repeats. This was raised in review as HIGH; the response linked to a discussion but no code change followed.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
```diff
@@ -103,6 +112,9 @@ public class PullModuleExecutor implements LSCommandExecutor {
     // published in stage 4 below is also consumed by ResolveCompilationErrorsSubscriber, which
     // starts another pull for the same project, creating an endless pull loop on failures.
     private static final Set<String> PULL_IN_PROGRESS_PROJECTS = ConcurrentHashMap.newKeySet();
```
samithkavishke, line 114: **MEDIUM**: `PULL_IN_PROGRESS_PROJECTS` only guards re-entrant pulls within one LS process lifetime, but the client's recovery action restarts that very process (`workbench.action.reloadWindow`), so it provides no protection against a clear→reload→corrupt→clear loop. The author's rebuttal ("clearing regenerates deterministically from bala") covers the stale-cache case but not the concurrent-write-race / hostile-environment cases raised in the original comment.

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
```diff
@@ -0,0 +1,269 @@
...
+
+export async function resolvePackageCacheDirs(
+    reposDir: string,
+    pkg: CorruptPackage,
+    distVersion?: string
+): Promise<string[]> {
+    const targets: string[] = [];
+    for (const repo of await listSubDirs(reposDir)) {
+        const repoDir = path.join(reposDir, repo);
+        for (const cacheDir of await listSubDirs(repoDir)) {
```
samithkavishke: **LOW** (cleanup, non-blocking): `resolvePackageCacheDirs` and `clearAllBirCaches` duplicate the same repo/cache-dir directory-walk loop (`listSubDirs(reposDir)` → `listSubDirs(repoDir)` → `cacheDirMatches`). Consider extracting a shared helper/generator to avoid maintaining the traversal logic in two places.

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
thuva9872: We have changed the modal to notification based on the offline discussion. Now we show "Clear cache & Reload" and "Report an Issue". If the issue is recurring, the user can report it using the given option.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
thuva9872: The BIR is not regenerated on reload. So there will be no concurrent write race.

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
thuva9872: addressed in [be5d8c4](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/be5d8c4e26b16c72ebf8b94a3c13d176caf7c0ea)

samithkavishke (dismissed):
LGTM

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
```diff
@@ -0,0 +1,268 @@
...
+        const clearAction = "Clear cache & reload";
+        const reportAction = "Report an Issue";
+        const message = coordinate
+            ? `The cache for module '${coordinate}' is corrupted.\n` +
+              `The project will stay unresponsive until this is resolved.\n` +
+              `Clearing removes the module's cache (or all module caches for this distribution if it ` +
+              `can't be located) and reloads the window to recover.`
+            : `A module cache is corrupted.\n` +
+              `The project will stay unresponsive until this is resolved.\n` +
+              `Clearing removes the module cache and reloads the window to recover.`;
```
NipunaRanasinghe, line 227: Better to improve the error message to make the suggested recovery steps more direct. 
```suggestion
        const message = coordinate
    ? `The cache for module '${coordinate}' is corrupted.\n` +
      `The project will stay unresponsive until this is resolved.\n` +
      `Try clearing the module's cache and reloading the window, or report an issue if the problem persists.`
    : `A module cache is corrupted.\n` +
      `The project will stay unresponsive until this is resolved.\n` +
      `Try clearing the module's cache and reloading the window, or report an issue if the problem persists.`;
```

packages/ballerina-extension/src/core/extended-language-client.ts
```diff
@@ -656,6 +658,14 @@ export class ExtendedLangClient extends LanguageClient implements ExtendedLangCl
...
+            void promptClearCorruptBirCache(res);
```
dulajdilshan, line 665: **[Medium] Error handling diverges from the neighbours.** `void promptClearCorruptBirCache(res)` discards the promise, while every sibling handler in this class wraps its body in `try`/`catch` and logs — `registerPublishArtifacts` :650-657, and the three migration handlers at :671, :684, :697. `promptClearCorruptBirCache` only guards the clear step (`bir-cache-recovery.ts:252`); a rejection from `window.showErrorMessage` (:231) or `executeCommand` (:263) escapes as an unhandled rejection with nothing logged. Match the house style?

```suggestion
            promptClearCorruptBirCache(res).catch((error) =>
                console.error("CORRUPT_BIR_CACHE handler failed:", error));
```

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
```diff
@@ -0,0 +1,268 @@
...
+            const removed = target
+                ? await clearPackageBirCache(target, { distVersion, reposDir })
+                : await clearAllBirCaches({ distVersion, reposDir });
+            // The targeted clear matched nothing — the corrupt cache is on disk (that is why the LS
+            // reported it), but the coordinates did not resolve to it, e.g. a submodule where the LS
+            // fell back to the module name instead of the package name. Clear the whole distribution
+            // cache so the user still recovers; the prompt above states this broader scope up front.
+            if (target && removed.length === 0) {
+                await clearAllBirCaches({ distVersion, reposDir });
+            }
```
dulajdilshan: **[Medium] Test asserts nothing — the destructive path is the unpinned one.** I read the whole suite: it pins `isValidPackage`, `resolvePackageCacheDirs`, `clearPackageBirCache` and `buildCorruptBirIssueUrl`, but `clearAllBirCaches` and this `removed.length === 0` escalation have zero assertions anywhere. This is the branch that deletes *every* package's cache under `cache-<dist>` — the widest-blast-radius line in the PR, and the one @pasindufernando1 asked you to narrow the consent for. `AGENTS.md` requires a test at the level the behaviour belongs to, and this is plain fs logic with no `vscode` dependency. Would an `expectAllCleared` flag on the existing fixture table cover it without new harness?

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
```diff
@@ -0,0 +1,268 @@
...
+    const query = `title=${encodeURIComponent(title)}&body=${encodeURIComponent(body)}`;
+    return `${PRODUCT_INTEGRATOR_ISSUES_URL}/new?${query}`;
+}
+
+/**
+ * Surfaces the corrupt-BIR condition and, on confirmation, clears the affected package's compiled
+ * cache under the active distribution's cache directory (or all packages in that cache when the
+ * package can't be identified) and reloads the window.
+ */
+export async function promptClearCorruptBirCache(payload: CorruptBirCachePayload | null | undefined): Promise<void> {
```
dulajdilshan, line 201: **[Minor] Single Responsibility.** This one function validates the payload, formats a display coordinate, authors two message variants, shows the notification, dispatches the report action, clears with a fallback, reports clear failures, and reloads the window — the honest name needs four "and"s. That is also exactly why the region above is untestable: the decision *targeted vs whole-cache* can't be asserted without a `vscode` mock. What do you think about extracting the pure part — `payload -> {target, coordinate, message}` — so the decision becomes an L1 assertion and this function is left with just the I/O?

packages/ballerina-language-server/langserver-core/src/test/java/org/ballerinalang/langserver/command/executors/PullModuleExecutorCorruptBirTest.java
```diff
@@ -0,0 +1,203 @@
...
+                    + "invalid magic number [99, 111, 114, 114]";
+
+    @Test(description = "A nested corrupt-BIR cause is detected and its module coordinates extracted")
+    public void testNestedCorruptBirWithCoordinates() {
+        Throwable throwable = new RuntimeException("Pull modules failed",
+                new IllegalStateException(CORRUPT_WITH_COORDINATES));
+
+        // No dependency graph resolves the module here, so the package name falls back to the module
+        // name — which is correct for a default module (package name == module name).
+        Optional<CorruptBirCacheParams> result = PullModuleExecutor.detectCorruptBirCache(throwable, null);
```
dulajdilshan, line 55: **[Medium] Stated invariant is false.** The comment says "No dependency graph resolves the module here", but `project` is `null`, so `resolvePackageName` never reaches a graph at all — `List.of(project)` throws NPE (`PullModuleExecutor.java:370`) and the blanket `catch (Throwable)` returns empty. The assertion passes for a different reason than the comment claims, and the next reader will trust the comment instead of tracing it. The genuine empty-graph path is already covered by `testUnresolvedModuleFallsBackToModuleName`. Reword to say the fallback is reached because there is no project, or pass a mock with an empty graph?

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
```diff
@@ -0,0 +1,268 @@
...
+              `The project will stay unresponsive until this is resolved.\n` +
+              `Clearing removes the module cache and reloads the window to recover.`;
+
+        // VS Code dismisses a notification whenever an action button is clicked; there is no way to
+        // keep it open. So "Report an Issue" just opens the prefilled issue and lets it close.
+        const choice = await window.showErrorMessage(message, clearAction, reportAction);
+        if (choice === reportAction) {
+            openExternalUrl(buildCorruptBirIssueUrl(payload, coordinate));
+            return;
+        }
```
dulajdilshan, line 235: **[Spec · Medium] Report can consume the only prompt.** Choosing "Report an Issue" opens the URL and returns; as your comment at :229 notes, VS Code has already dismissed the notification, and `promptShown` resets in `finally` without re-showing it — so the next prompt only arrives on another pull failure. The issue asks for "a notification prompting the user to clear the cache, with an option to do so directly via a button click", and this path can leave the user in the state the message itself warns about ("The project will stay unresponsive until this is resolved"). Re-show the prompt after the browser opens, or clear-and-reload as well as reporting?

packages/ballerina-extension/src/stateMachine.ts
dulajdilshan, line 587: **[Question] Registration ordering.** This registers inside `registerProjectArtifactsStructure`, reached via `activateLS`, while the LS fires the notification from the pull's failure arm. `vscode-languageclient` drops an unhandled notification rather than queuing it — what guarantees the handler is installed before the first corrupt-BIR notification can arrive? (Different concern from @samithkavishke's double-registration point, which you answered.)

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
```diff
@@ -324,6 +346,96 @@ public static CompletableFuture<Void> resolveModules(String fileUri, ExtendedLan
...
+     * {@code null} if they cannot be parsed from the message).
+     *
+     * @param throwable the completion throwable to inspect
+     * @param project   the project whose resolution maps the failing module to its package
+     * @return the corrupt-BIR parameters, or empty if this is not a corrupt-BIR failure
+     */
+    static Optional<CorruptBirCacheParams> detectCorruptBirCache(Throwable throwable, Project project) {
+        for (Throwable cause = throwable; cause != null && cause != cause.getCause(); cause = cause.getCause()) {
+            String message = cause.getMessage();
+            if (message == null || !message.contains("invalid magic number") || !message.contains("BIR")) {
```
rosensilva, line 366: Since we are going with the notification(Not the full blocker screen), as we discussed offline, wondering if keying on the from its BIR wrapper would be simpler and cover all of them — ResolveCompilationErrorsSubscriber:63 already detects the same condition that way.  That way, we might have a temporary fix for other BIR issues, as we discussed in https://github.com/wso2/product-integrator/issues/2166

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
```diff
@@ -0,0 +1,268 @@
...
+    return path.join(ballerinaHome, "repositories");
+}
+
+function isWithin(parent: string, child: string): boolean {
+    const rel = path.relative(parent, child);
+    return !!rel && !rel.startsWith("..") && !path.isAbsolute(rel);
+}
+
+function cacheDirMatches(cacheDir: string, distVersion?: string): boolean {
+    if (!cacheDir.startsWith("cache-")) {
```
rosensilva, line 83: Does this reach the distribution's own cache? BallerinaDistributionRepository looks like it uses the unversioned FileSystemRepository constructor, which would put it at <ballerina.home>/repo/cache — no -<version> suffix, and outside repositories/. If a bundled stdlib BIR goes bad, would we clear the user's central cache while the real corrupt file survives? Could the LS send that path along too?

thuva9872: addressed in [acc4b18](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/acc4b185ab2b07fae2260201e1a9cb6e8575fe62).

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/PullModuleExecutor.java
thuva9872: Handled in [acc4b18](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/acc4b185ab2b07fae2260201e1a9cb6e8575fe62)

packages/ballerina-extension/src/stateMachine.ts
thuva9872: This is a valid concern. Addressed in [acc4b18](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/acc4b185ab2b07fae2260201e1a9cb6e8575fe62)

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
thuva9872: Addressed in [acc4b18](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/acc4b185ab2b07fae2260201e1a9cb6e8575fe62)

packages/ballerina-language-server/langserver-core/src/test/java/org/ballerinalang/langserver/command/executors/PullModuleExecutorCorruptBirTest.java
thuva9872: addressed in [acc4b18](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/acc4b185ab2b07fae2260201e1a9cb6e8575fe62)

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
thuva9872: Fixed in [acc4b18](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/acc4b185ab2b07fae2260201e1a9cb6e8575fe62)

packages/ballerina-extension/src/core/extended-language-client.ts
thuva9872: Fixed in [acc4b18](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/acc4b185ab2b07fae2260201e1a9cb6e8575fe62)

packages/ballerina-extension/src/utils/bir-cache-recovery.ts
thuva9872: Fixed in [acc4b18](https://github.com/ballerina-platform/ballerina-vscode/pull/1013/commits/acc4b185ab2b07fae2260201e1a9cb6e8575fe62)

rosensilva (approved):

PR #1015, opened 2026-09-05
Fix max token issue in AI ehancement in migration wizard

KavinduZoysa:
## Purpose
Addresses https://github.com/wso2/product-integrator/issues/2348

## Summary

- Fixes maximum-token handling in the migration wizard’s AI enhancement flow.
- Uses a consistent chat storage key for package-scoped executions.
- Distinguishes user aborts from migration failures.
- Records failed stages and allows them to retry during resumed runs.

### DIFF
--- a/packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
@@ -252,6 +252,22 @@ export class AgentExecutor extends AICommandExecutor<GenerateAgentCodeRequest> {
     /** A turn can reach both the finish and abort paths; suggestions must be scheduled once. */
     private _followupsScheduled = false;
 
+    /**
+     * Store key for every `chatStateStorage` call, mirroring the precedence in
+     * `AICommandExecutor.run()`. Prefers `chatStorage.projectRootPath` — the key
+     * `addGeneration` and `getChatHistoryForLLM` use — because a caller may scope
+     * `executionContext` to a sub-package (the migration wizard passes the package path
+     * with `workspacePath: undefined`). Deriving the key from `executionContext` instead
+     * would address a different store than the one holding the generation, and every
+     * write would silently no-op on a missing generation ID.
+     */
+    private get chatStoreKey(): string {
+        return this.config.chatStorage?.projectRootPath
+            || this.config.executionContext.workspacePath
+            || this.config.executionContext.projectPath
+            || '';
+    }
+
     constructor(config: AICommandConfig<GenerateAgentCodeRequest>) {
         super(config);
     }
@@ -339,9 +355,9 @@ export class AgentExecutor extends AICommandExecutor<GenerateAgentCodeRequest> {
             const projectRootPath = this.config.executionContext.workspacePath || this.config.executionContext.projectPath || '';
             const agentsMd = await prepareAgentsMdForTurn(workspaceId || '', threadId);
             if (agentsMd.hashToPersist !== undefined) {
-                const generation = chatStateStorage.getGeneration(projectRootPath, threadId, this.config.generationId);
+                const generation = chatStateStorage.getGeneration(this.chatStoreKey, threadId, this.config.generationId);
                 if (generation) {
-                    chatStateStorage.updateGeneration(projectRootPath, threadId, this.config.generationId, {
+                    chatStateStorage.updateGeneration(this.chatStoreKey, threadId, this.config.generationId, {
                         metadata: {
                             ...generation.metadata,
                             agentsMdLastReadHash: agentsMd.hashToPersist,
@@ -479,7 +495,7 @@ export class AgentExecutor extends AICommandExecutor<GenerateAgentCodeRequest> {
                     const stepMessages = step.response?.messages ?? [];
                     if (stepMessages.length > 0) {
                         console.log(`[AgentExecutor] Step ${step.stepNumber} saving ${stepMessages.length} message(s) to chat storage`);
-                        chatStateStorage.updateGeneration(workspaceId, threadId, this.config.generationId, {
+                        chatStateStorage.updateGeneration(this.chatStoreKey, threadId, this.config.generationId, {
                             modelMessages: [
                                 { role: "user", content: userMessageContent },
                                 ...stepMessages,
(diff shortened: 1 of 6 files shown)


packages/ballerina-extension/src/features/ai/migration/orchestrator.ts
```diff
@@ -713,7 +713,34 @@ async function runStagesForPackage(opts: StageRunnerOpts): Promise<void> {
...
+            const reason = getErrorMessage(result.error);
+            if (debugLogger) {
+                debugLogger.logError(stage.name, result.error);
+                debugLogger.logMilestone(`${stage.name} — failed: ${reason}`);
+            }
+            console.error(`[MigrationEnhancement] ${stage.name} failed:`, result.error);
+
+            recordingHandler({
+                type: "content_block",
+                content: `\n\n**${stage.name} — Failed** ❌\n\n${reason}\n\n`,
```
RNViththagan: Shall we escape `reason` here the same way the package-loop and workspace-validation catches do (`&` → `&amp;`, `<` → `&lt;`, `>` → `&gt;`)? This block goes through the same markdown renderer, which runs `rehypeRaw`, so a reason containing `<…>` would be parsed as an HTML tag and vanish. The `prompt is too long: … tokens > … maximum` message happens to render fine (a bare `>` is just text), but other failure messages won't necessarily.

Since that makes it the fifth copy of the same three `.replace()`s in this file, a tiny shared helper might be worth it too — no strong feelings.

packages/ballerina-extension/src/test-support/migrationPackageLoop.test.ts
```diff
@@ -0,0 +1,203 @@
...
+    it("reports a failed package and keeps going through the rest of the workspace", async () => {
+        // First stage of the first package fails; every later stage succeeds.
+        mockStagePlan = (_pkg, callIndex) => (callIndex === 0 ? "fail" : "ok");
+
+        await run();
+
+        expect(packagesRun()).toEqual(PACKAGES);
+        expect(emittedText()).toContain("Package `pkgA` failed");
+        expect(emittedText()).toContain("**2** of **3** packages enhanced successfully.");
+        expect(readEnhanceToml(mockProjectRoot)?.fullyEnhanced).toBe(true);
```
RNViththagan, line 177: A design question rather than a bug. After a package's stage fails, the loop now continues — which is the right behaviour — and finishes with `markEnhancementComplete()`, so `fullyEnhanced` ends up `true`, which this assertion pins. But `checkAndRunPendingEnhancement` treats `fullyEnhanced = true` as "nothing to do", so the user never gets the resume prompt — meaning `failStage`'s "a resumed run re-attempts this stage" only really kicks in for the single-package flow.

Is that intended? Quite possibly yes — the final report already shows the failure inline. If we do want the failed package to stay retryable, shall we keep `fullyEnhanced = false` whenever any package failed? Either way, a line in the `failStage` doc about the multi-package case would save the next reader the same question.

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
```diff
@@ -251,6 +251,22 @@ export class AgentExecutor extends AICommandExecutor<GenerateAgentCodeRequest> {
...
+    /**
+     * Store key for every `chatStateStorage` call, mirroring the precedence in
+     * `AICommandExecutor.run()`. Prefers `chatStorage.projectRootPath` — the key
+     * `addGeneration` and `getChatHistoryForLLM` use — because a caller may scope
+     * `executionContext` to a sub-package (the migration wizard passes the package path
+     * with `workspacePath: undefined`). Deriving the key from `executionContext` instead
+     * would address a different store than the one holding the generation, and every
+     * write would silently no-op on a missing generation ID.
+     */
+    private get chatStoreKey(): string {
```
RNViththagan, line 257: Nice fix, and the doc comment explains the trap well. One ask: the new suite pins the abort-vs-failure separation, but nothing pins this root cause itself — divergent keys → empty `modelMessages` → `uiResponse` replay → overflow. Shall we add a small contract test? Build a config where `chatStorage.projectRootPath` ≠ `executionContext.projectPath`, run one step, and assert `modelMessages` landed in the `projectRootPath` store while the package-path store stayed empty. That's the exact invariant that broke, and it would close out CodeRabbit's earlier note too.

packages/ballerina-extension/src/features/ai/migration/orchestrator.ts
```diff
@@ -904,11 +976,11 @@ export async function runMigrationAgent(): Promise<void> {
                 markEnhancementComplete();
                 console.log("[MigrationEnhancement] Migration agent completed all stages successfully.");
             } else {
-                debugLogger.logMilestone("Run aborted by user");
+                debugLogger.logMilestone(_userAbortedMigration ? "Run aborted by user" : "Run stopped early — a stage did not complete");
             }
         }
     } catch (error) {
-        if (_migrationAbortController.signal.aborted) {
+        if (_userAbortedMigration) {
```
RNViththagan, line 1010: Since stage failures now reach the `else` of this `if` (before, they were misread as a user abort and took the branch above), shall we mirror the wizard flow here? Its equivalent branch emits `eventHandler({ type: "error", … })` so the panel sees a terminal state; this one only logs.

Related: the abort branch above writes `summary.md`, but the failure path doesn't — and the old mis-routed path used to, so a single-package resume now loses its preamble. Both are small; happy for the summary one to be a follow-up.

packages/ballerina-extension/src/test-support/migrationPackageLoop.test.ts
KavinduZoysa: Yes, the final report shows the failure inline, and the run is finished as far as the workspace flow is concerned, so it stays fullyEnhanced = true and no resume prompt appears. You're right that this makes failStage's "a resumed run re-attempts this stage" effectively single-package-only; leaving the behaviour as is for now rather than widening the scope of this PR.

packages/ballerina-extension/src/features/ai/migration/orchestrator.ts
KavinduZoysa: Fixed

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
KavinduZoysa: Fixed

packages/ballerina-extension/src/features/ai/migration/orchestrator.ts
KavinduZoysa: Fixed

RNViththagan (approved):
LGTM

lochana-chathura (approved):

PR #1016, opened 2026-09-06
Fix single-select in forms for enums

samithkavishke:
## Purpose
Fix: https://github.com/wso2/product-integrator/issues/2035

Related issue case
https://github.com/user-attachments/assets/fd7581b8-3927-4d32-a683-8738f5b25bf3

Mixed types
https://github.com/user-attachments/assets/a559dd12-4da9-4a37-b978-018d88ba50e4

Boolean
https://github.com/user-attachments/assets/78950366-7f7d-4432-9ea9-2b8b01a8f686

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary

- Fix single-select handling for enum and union types in forms.
- Use declared defaults to set dropdown placeholders.
- Support enum member values in low-code inputs and generate matching pro-code values.
- Improve Boolean and enum editors with explicit default and empty-selection handling.
- Add tests for enum and Boolean dropdown behavior.

### DIFF
--- a/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
+++ b/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
@@ -826,11 +826,8 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
                         // The singleton members (e.g. the members of an enum) become single-select options, even
                         // when the union holds other member types as well.
                         if (!options.isEmpty()) {
-                            // Reorder options so that the default value appears first
-                            List<Option> orderedOptions = defaultValue == null || defaultValue.isEmpty() ? options
-                                    : reorderOptionsByDefaultValue(options, defaultValue);
-                            builder.type().fieldType(ValueType.SINGLE_SELECT).options(orderedOptions).stepOut();
-                            alignPlaceholderWithDefault(builder, orderedOptions, defaultValue);
+                            builder.type().fieldType(ValueType.SINGLE_SELECT).options(options).stepOut();
+                            alignPlaceholderWithDefault(builder, options, defaultValue);
                         }
 
                         if (!otherTypes.isEmpty()) {
@@ -1445,24 +1442,6 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
                     .filter(option -> value.equals(CommonUtils.removeQuotes(option.value()))
                             || memberName.equals(option.label()))
                     .findFirst();
-        }
-
-        /**
-         * Reorders enum options so that the option matching the defaultValue appears first in the list.
-         * This improves user experience by showing the default option at the top of dropdown lists.
-         * Returns a new list without modifying the input list.
-         *
-         * @param options      The list of Option objects to reorder
-         * @param defaultValue The default value to prioritize (may contain quotes)
-         * @return A new list with the default option first, or the original list if no match found
-         */
-        private static List<Option> reorderOptionsByDefaultValue(List<Option> options, String defaultValue) {
-            List<Option> reorderedOptions = new ArrayList<>(options != null ? options : List.of());
-            findMatchingOption(reorderedOptions, defaultValue).ifPresent(defaultOption -> {
-                reorderedOptions.remove(defaultOption);
-                reorderedOptions.addFirst(defaultOption);
-            });
-            return reorderedOptions;
         }
 
         /**
--- a/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
+++ b/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
@@ -808,17 +808,18 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
                                     addEnumOptions(enumSymbol, options, enumMemberTypes));
                         }
 
-                        // Each member is turned into an input type as it is declared, so that the modes of the
-                        // field are offered in the order of the union. The singletons share one single select
-                        // rather than getting one each, which is added where the first of them is declared and
-                        // holds the very list the remaining ones keep being collected into.
-                        boolean addedSingleSelect = false;
+                        // The members that are not singletons keep an input type each, gathered here rather than
+                        // turned into one on sight, so that the single select standing for the singletons can be
+                        // added once its options are complete rather than while they are still arriving.
+                        List<TypeSymbol> otherTypes = new ArrayList<>();
+                        // Where the union declares the first singleton, which is the place the select belongs in
+                        // among those members, so that the modes are offered in the order the union declares.
+                        int selectPosition = -1;
                         for (TypeSymbol symbol : unionTypeSymbol.memberTypeDescriptors()) {
                             TypeDescKind memberTypeKind = CommonUtil.getRawType(symbol).typeKind();
                             if (memberTypeKind == TypeDescKind.SINGLETON) {
-                                if (!addedSingleSelect) {
-                                    builder.type().fieldType(ValueType.SINGLE_SELECT).options(options).stepOut();
-                                    addedSingleSelect = true;
+                                if (selectPosition < 0) {
+                                    selectPosition = otherTypes.size();
                                 }
                                 // Skip the singletons that are already covered by the options of an enum
                                 if (!enumMemberTypes.contains(symbol.signature())) {
@@ -826,13 +827,22 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
                                     options.add(new Option(label, symbol.signature()));
                                 }
                             } else if (memberTypeKind != TypeDescKind.NIL) {
-                                // The nil member is conveyed by the `optional` flag of the property
-                                handlePrimitiveType(symbol, CommonUtils.getTypeSignature(symbol, moduleInfo),
-                                        semanticModel, moduleInfo, builder);
+                                // A nil member stands for no value, which no entry of a dropdown can offer and
+                                // no input mode of its own can express, hence it is left to the expression mode.
+                                otherTypes.add(symbol);
                             }
                         }
-                        if (addedSingleSelect) {
-                            alignPlaceholderWithDefault(builder, options, defaultValue);
+
+                        for (int i = 0; i <= otherTypes.size(); i++) {
+                            if (i == selectPosition) {
+                                builder.type().fieldType(ValueType.SINGLE_SELECT).options(options).stepOut();
+                                alignPlaceholderWithDefault(builder, options, defaultValue);
+                            }
+                            if (i < otherTypes.size()) {
+                                TypeSymbol ts = otherTypes.get(i);
+                                handlePrimitiveType(ts, CommonUtils.getTypeSignature(ts, moduleInfo), semanticModel,
+                                        moduleInfo, builder);
+                            }
                         }
 
                         // group by the fieldType
@@ -1403,6 +1413,27 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
         }
 
         /**
+         * Adds an option standing for the given enum member or constant, labelled by its name and holding the
+         * singleton it stands for as the generated value.
+         *
+         * <p>One that carries no name is left out entirely, registering nothing: the singleton it stands for is
+         * then still collected by the walk over the members of the union, which offers it labelled by its value
+         * rather than dropping it.
+         *
+         * @param member          the enum member or constant to offer
+         * @param options         the options to append to
+         * @param enumMemberTypes collects the signatures of the singleton types covered by the added options
+         */
+        private static void addMemberOption(ConstantSymbol member, List<Option> options,
+                                            Set<String> enumMemberTypes) {
+            String memberValue = member.typeDescriptor().signature();
+            member.getName().ifPresent(name -> {
+                enumMemberTypes.add(memberValue);
+                options.add(new Option(name, memberValue));
+            });
+        }
+
+        /**
          * Adds an option for each member of the given enum. A member is labelled with its name (e.g. `HIGH`), while
          * the value it holds (e.g. `"10"`) stays the generated value, so that the source keeps referring to the
          * member the way it did before the members were surfaced by name, and hence needs no module prefix.
@@ -1414,9 +1445,7 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
         private static void addEnumOptions(EnumSymbol enumSymbol, List<Option> options, Set<String> enumMemberTypes) {
             // The members are returned in the reverse order of their declaration
             for (ConstantSymbol enumMember : enumSymbol.members().reversed()) {
-                String memberValue = enumMember.typeDescriptor().signature();
-                enumMemberTypes.add(memberValue);
-                enumMember.getName().ifPresent(name -> options.add(new Option(name, memberValue)));
+                addMemberOption(enumMember, options, enumMemberTypes);
             }
         }
 
@@ -1435,10 +1464,17 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
                 return Optional.empty();
             }
             String value = CommonUtils.removeQuotes(defaultValue);
+            // The value is looked for across every option before any label is, so that a default holding the
+            // value of one member and the name of another is read as the value it plainly is.
+            Optional<Option> byValue = options.stream()
+                    .filter(option -> value.equals(CommonUtils.removeQuotes(option.value())))
+                    .findFirst();
+            if (byValue.isPresent()) {
+                return byValue;
+            }
             String memberName = removeModulePrefix(value);
             return options.stream()
-                    .filter(option -> value.equals(CommonUtils.removeQuotes(option.value()))
-                            || memberName.equals(option.label()))
+                    .filter(option -> memberName.equals(option.label()))
                     .findFirst();
         }
 
@@ -1448,11 +1484,10 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
          *
          * <p>On a single select the placeholder names the member the field defaults to, never a sample value of
          * the type: the type of a union yields an arbitrary member of it, which says nothing about the default.
-         * The declared default is therefore the only source, and it is not always resolvable to an option - the
-         * parameter may declare none, or the union may refer to constants rather than enum members (e.g.
-         * {@code http:Compression} is COMPRESSION_AUTO|COMPRESSION_ALWAYS|...), whose names the type does not
-         * carry. Both leave the placeholder empty, which presents the empty selection rather than claiming an
-         * arbitrary member is the default.
+         * The declared default is therefore the only source, and it does not always name one of the options -
+         * the parameter may declare no default at all, or declare one the union does not offer. Both leave the
+         * placeholder empty, which presents the empty selection rather than claiming an arbitrary member is the
+         * default.
          *
          * <p>The placeholder also decides the generated source: an argument matching it is omitted, and one is
          * written for every other selection. Holding the declared default is what makes that omission correct.
@@ -1467,7 +1502,10 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
             // when it does not, so that a declared default is not mistaken for an absent one.
             String declaredDefault = defaultValue == null || defaultValue.isEmpty() ? builder.defaultValue
                     : defaultValue;
-            builder.placeholder(findMatchingOption(options, declaredDefault).map(Option::value).orElse(null));
+            // A default that names none of the options leaves the field with none to present, which is said
+            // with an empty placeholder rather than none at all: the placeholder of a single select is read as
+            // a value elsewhere, and absence there is not a state those readers are prepared for.
+            builder.placeholder(findMatchingOption(options, declaredDefault).map(Option::value).orElse(""));
         }
 
         private static String removeModulePrefix(String value) {
--- a/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
+++ b/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
@@ -1445,14 +1445,19 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
         }
 
         /**
-         * Aligns the placeholder of the property with the declared default of the parameter, when the placeholder
-         * holds one of the options.
+         * Sets the placeholder of a single select to the option the parameter defaults to, and to nothing when
+         * there is no such option.
          *
-         * <p>The placeholder is derived from the type of the parameter, which for a union yields an arbitrary
-         * member of it rather than the default. That member is then presented as the default of the field, and the
-         * generated source omits an argument matching it, so selecting it produces no code at all. The declared
-         * default replaces it here, and the placeholder is dropped when the default does not resolve to an option
-         * (or the parameter declares none), which leaves no option presented as the default.
+         * <p>On a single select the placeholder names the member the field defaults to, never a sample value of
+         * the type: the type of a union yields an arbitrary member of it, which says nothing about the default.
+         * The declared default is therefore the only source, and it is not always resolvable to an option - the
+         * parameter may declare none, or the union may refer to constants rather than enum members (e.g.
+         * {@code http:Compression} is COMPRESSION_AUTO|COMPRESSION_ALWAYS|...), whose names the type does not
+         * carry. Both leave the placeholder empty, which presents the empty selection rather than claiming an
+         * arbitrary member is the default.
+         *
+         * <p>The placeholder also decides the generated source: an argument matching it is omitted, and one is
+         * written for every other selection. Holding the declared default is what makes that omission correct.
          *
          * @param builder      the builder of the property being built
          * @param options      the options of the single select
@@ -1460,26 +1465,10 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
          */
         private static void alignPlaceholderWithDefault(Builder<?> builder, List<Option> options,
                                                         String defaultValue) {
-            boolean placeholderIsAnOption = options.stream()
-                    .anyMatch(option -> option.value().equals(builder.placeholder));
-            if (!placeholderIsAnOption) {
-                return;
-            }
             // Not every caller passes the default down to the type, hence the one held by the property is used
             // when it does not, so that a declared default is not mistaken for an absent one.
             String declaredDefault = defaultValue == null || defaultValue.isEmpty() ? builder.defaultValue
                     : defaultValue;
-            if (declaredDefault == null || declaredDefault.isEmpty()) {
-                // The parameter declares no default, hence none of the options is the default of the field
-                builder.placeholder(null);
-                return;
-            }
-            // A default that does not resolve to an option is dropped rather than guessed at, as the union may
-            // refer to constants instead of enum members (e.g. `http:Compression` is
-            // COMPRESSION_AUTO|COMPRESSION_ALWAYS|...). Keeping the member derived from the type would present
-            // an arbitrary one as the default, and an argument matching it is omitted from the generated
-            // source, so selecting that member would generate nothing. Without a placeholder every selection
-            // is generated, which states what was selected even when it is what the parameter defaults to.
             builder.placeholder(findMatchingOption(options, declaredDefault).map(Option::value).orElse(null));
         }
 
--- a/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
+++ b/packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
@@ -795,28 +795,31 @@ public record Property(Metadata metadata, List<PropertyType> types, Object value
                     try {
                         List<TypeSymbol> typeSymbols = unionTypeSymbol.memberTypeDescriptors();
                         List<Option> options = new ArrayList<>();
-                        boolean allSingletons = true;
+                        List<TypeSymbol> otherTypes = new ArrayList<>();
                         for (TypeSymbol symbol : typeSymbols) {
-                            if (CommonUtil.getRawType(symbol).typeKind() == TypeDescKind.SINGLETON) {
+                            TypeDescKind memberTypeKind = CommonUtil.getRawType(symbol).typeKind();
+                            if (memberTypeKind == TypeDescKind.SINGLETON) {
                                 String label = CommonUtils.removeQuotes(symbol.signature());
-                                Option option = new Option(label, symbol.signature());
-                                options.add(option);
-                            } else {
-                                allSingletons = false;
-                                break;
+                                options.add(new Option(label, symbol.signature()));
+                            } else if (memberTypeKind != TypeDescKind.NIL) {
+                                // The nil member is conveyed by the `optional` flag of the property
+                                otherTypes.add(symbol);
                             }
                         }
 
-                        // If all the member types are singletons, treat it as a single-select option
-                        if (allSingletons) {
+                        // The singleton members (e.g. the members of an enum) become single-select options, even
+                        // when the union holds other member types as well.
+                        if (!options.isEmpty()) {
                             // Reorder options so that the default value appears first
                             if (defaultValue != null && !defaultValue.isEmpty()) {
                                 options = reorderOptionsByDefaultValue(options, defaultValue);
                             }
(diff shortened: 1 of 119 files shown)


packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
```diff
@@ -785,38 +788,55 @@ public Builder<T> typeWithExpression(TypeSymbol typeSymbol, ModuleInfo moduleInf
...
-                            if (defaultValue != null && !defaultValue.isEmpty()) {
-                                options = reorderOptionsByDefaultValue(options, defaultValue);
-                            }
-                            builder.type().fieldType(ValueType.SINGLE_SELECT).options(options).stepOut();
-                        } else {
-                            // Handle union of primitive types by defining an input type for each primitive type
-                            for (TypeSymbol ts : typeSymbols) {
+                            List<Option> orderedOptions = defaultValue == null || defaultValue.isEmpty() ? options
+                                    : reorderOptionsByDefaultValue(options, defaultValue);
+                            builder.type().fieldType(ValueType.SINGLE_SELECT).options(orderedOptions).stepOut();
```
pasindufernando1: Mixed unions now emit a three-entry `types` list (`[SINGLE_SELECT, NUMBER, EXPRESSION]` in `node_template/config/enum_int_union.json`, `agents_manager/config/agent_call_flow_node_7.json`), but `ballerina-side-panel/src/components/editors/FieldFactory.tsx:186` only ever renders `types[0]` and `types[types.length - 1]`.

Two consequences:
- The middle NUMBER/TEXT editor is unreachable — for `ai:Agent`'s `maxIter` (`int|INFER_TOOL_COUNT`) the mode switcher offers only a one-option dropdown plus raw expression, and `getDefaultExpressionMode` makes SELECT the *primary* mode for an int parameter.
- Worse, when the LS marks the middle type `selected` (`typeWithExpression` sets `selected(true)` on NUMBER for a numeric-literal argument, `Property.java:906-915`), `FieldFactory` computes `initialInputMode = NUMBER` while `editorElements` filters over `[SINGLE_SELECT, EXPRESSION]` — both entries return `null` and **the field renders no editor at all**.

Repro: `function f(Day|int day)` called as `f(5)` → blank form field.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/model/PropertyType.java
```diff
@@ -119,38 +124,55 @@ public static void typeWithExpression(Value.ValueBuilder valueBuilder, TypeSymbo
...
+                    // The nil member is conveyed by the `optional` flag of the property
+                    otherTypes.add(symbol);
                 }
             }
 
-            // If all the member types are singletons, treat it as a single-select option
-            if (allSingletons) {
+            // The singleton members (e.g. the members of an enum) become single-select options, even when the
+            // union holds other member types as well.
+            if (!options.isEmpty()) {
```
pasindufernando1: The new front-end contract stated in `EnumEditor.test.tsx` — "the backend puts the declared default of the parameter in `placeholder`" — is implemented only in the flow-model generator (`Property.java:825`, `alignPlaceholderWithDefault`). This file got the option/label change but no placeholder alignment, so the service model's placeholder stays the arbitrary type-derived member (`ListenerUtil.java:380` → `paramResult.placeholder()`).

Concretely, `listener_from_source/config/listener_from_source_1.json` has `httpVersion` with `value: ""`, `defaultValue: null`, `placeholder: "\"2.0\""` and options `HTTP_1_0/HTTP_1_1/HTTP_2_0`. `EnumEditor.tsx:40` matches `"2.0"`, so the HTTP/GraphQL listener form now shows **HTTP_2_0** as the current selection for an unset field whose real default is HTTP_1_1. (`ListenerConfigForm.tsx:233` maps `placeholder` through and never maps `defaultValue`.)

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
samithkavishke: Addressed

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/model/PropertyType.java
samithkavishke: Addressed

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
```diff
@@ -779,38 +782,52 @@ public Builder<T> typeWithExpression(TypeSymbol typeSymbol, ModuleInfo moduleInf
...
+                            } else if (memberTypeKind != TypeDescKind.NIL) {
+                                // The nil member is conveyed by the `optional` flag of the property
+                                otherTypes.add(symbol);
                             }
+                        }
+
+                        // The singleton members (e.g. the members of an enum) become single-select options, even
+                        // when the union holds other member types as well.
+                        if (!options.isEmpty()) {
                             builder.type().fieldType(ValueType.SINGLE_SELECT).options(options).stepOut();
```
NipunaRanasinghe: **High — putting the single select first makes the middle input mode unreachable.**

`FieldFactory` renders only the first and the last type (`FieldFactory.tsx:189-191`), and for a field with no value it starts on `types[0]` (`FieldFactory.tsx:156`). `maxIter` (`int|INFER_TOOL_COUNT`) goes from `[NUMBER, EXPRESSION]` to `[SINGLE_SELECT, NUMBER, EXPRESSION]` in `agent_call_flow_node_7.json`, so a new agent node now opens on a dropdown whose only entry is `INFER_TOOL_COUNT` and the number editor is dropped from the mode switcher entirely — a number can only be entered by switching to the raw expression editor. `enum_int_union.json` and `enum_string_union.json` have the same three-type shape.

Appending the single select after the `otherTypes` inputs instead of before them keeps `types[0]` as the natural mode and leaves the select as the last-but-one, which `FieldFactory` also drops — so this probably needs `FieldFactory` to render more than first-and-last. Same ordering in `PropertyType.java:164-168`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
```diff
@@ -1360,36 +1377,90 @@ private ValueType findMatchingValueType(Node node) {
...
+         * the value it holds (e.g. `"10"`) stays the generated value, so that the source keeps referring to the
+         * member the way it did before the members were surfaced by name, and hence needs no module prefix.
+         *
+         * @param enumSymbol      the enum definition
+         * @param options         the options to append to
+         * @param enumMemberTypes collects the signatures of the singleton types covered by the added options
+         */
+        private static void addEnumOptions(EnumSymbol enumSymbol, List<Option> options, Set<String> enumMemberTypes) {
+            // The members are returned in the reverse order of their declaration
+            for (ConstantSymbol enumMember : enumSymbol.members().reversed()) {
```
NipunaRanasinghe, line 1447: **Low — the option order rests on an undocumented ordering of `EnumSymbol.members()`.**

`reversed()` is correct only while `members()` keeps returning members in reverse declaration order, which is not a guarantee the symbol API states. If that ever changes, the dropdowns silently reorder; the fixtures here (`http_listener_model.json` expects 1.0, 1.1, 2.0) would catch it, so it fails loudly rather than quietly.

Sorting by each member's `getLocation()` offset would make the intent independent of the compiler's iteration order. Same line in `PropertyType.java:343`.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/BooleanEditor/BooleanEditor.tsx
```diff
@@ -64,18 +68,24 @@ export const BooleanEditor: React.FC<BooleanEditorProps> = ({ value, onChange, f
         if (value === DEFAULT_NONE_SELECTED_VALUE) {
             onChange("", 0);
         } else {
-            const bool = parseBoolean(value);
-            onChange(String(bool), String(bool).length);
+            const bool = toBooleanValue(value) ?? String(false);
```
NipunaRanasinghe: **Nit.** The `?? String(false)` fallback cannot fire.

`e.target.value` only ever comes from `dropdownItems`, so it is `"true"`, `"false"` or `DEFAULT_NONE_SELECTED_VALUE`, and the last is handled by the branch above. `toBooleanValue` therefore always returns a value here.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
samithkavishke: Agreed the ordering was arbitrary, and that's fixed — the select is no longer put first. Each union member is now turned into its input type as it's declared, and the  singletons share one select placed where the union declares the first of them, so the mode order follows the type. int|Day now yields [NUMBER, SINGLE_SELECT, EXPRESSION]; added as a test case. For "INFER_TOOL_COUNT"|int, Day|int and Day|string the singleton is declared first, so the select legitimately leads and those fields are unchanged.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
samithkavishke: Tried this. Sorting by getLocation() offset doesn't hold for enums from compiled modules — locations there don't track source position, and it reorders 8 fixtures. ai's chunk strategy goes from HTML_HEADER, HTML_PARAGRAPH, HTML_LINE, SENTENCE, WORD, CHARACTER to WORD before SENTENCE, and kafka's serializer type comes out alphabetical (SER_AVRO before SER_BYTE_ARRAY). members().reversed() is correct for both source and compiled enums as things stand, so I've kept it with the comment stating the assumption, and the fixtures as the guard you noted. Happy to revisit if the symbol API ever documents an ordering.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/BooleanEditor/BooleanEditor.tsx
samithkavishke: addressed

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/SourceBuilder.java
```diff
@@ -843,7 +844,10 @@ public SourceBuilder functionParameters(FlowNode nodeTemplate, Set<String> ignor
                     missedDefaultValue = true;
                     continue;
                 }
-                if (prop.placeholder().equals(prop.value())) {
+                // A parameter holding what it defaults to is left out of the call. The placeholder is empty when
+                // the default is not known, e.g. a single select whose default names a constant the type does
+                // not carry, and the argument is then written out rather than assumed to be the default.
+                if (Objects.equals(prop.placeholder(), prop.value())) {
```
dulajdilshan, line 850: **[Major] Required single select now writes an empty argument.**

A `REQUIRED` param declares no default -> `alignPlaceholderWithDefault` stores `null` -> `toSourceCode()` returns `""`, and the REQUIRED branch has no empty guard:

```java
// Property.java:287
if (value == null || value.toString().isEmpty()) {
    return placeholder == null ? "" : placeholder;   // "" now, was "\"gpt-4o\"" before
}
// SourceBuilder.java:829 - no isPropValueEmpty() here, unlike DEFAULTABLE below
if (!optional && kind.equals(ParameterData.Kind.REQUIRED.name())) {
    tokenBuilder.param(prop);
}
```

Untouched required enum field -> `openai:new(, apiKey)` / missing arg -> source does not compile.

7 fixtures already pin this state (`kind: REQUIRED`, `optional: false`, no `placeholder`, no `value`):

| fixture | property |
|---|---|
| `get_openai_model_provider_template.json` | `modelType` |
| `get_anthropic_...` / `get_mistral_...` / embedding templates | `modelType` |
| `node_template/config/int_enum_union.json` (new) | `day` |

Reachable from the UI: `ExpressionField.tsx:193-217` registers no required rule for the SELECT mode (`TextAreaEditor.tsx:87` and `IdentifierField.tsx:109` do for theirs).

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
```diff
@@ -1374,36 +1389,90 @@ private ValueType findMatchingValueType(Node node) {
...
-         * @param options      The list of Option objects to reorder
-         * @param defaultValue The default value to prioritize (may contain quotes)
-         * @return A new list with the default option first, or the original list if no match found
+         * @param typeSymbol the type to resolve
+         * @return the enum definition, or empty if the type does not refer to an enum
          */
-        private static List<Option> reorderOptionsByDefaultValue(List<Option> options, String defaultValue) {
-            if (options == null || options.isEmpty() || defaultValue == null || defaultValue.isEmpty()) {
-                return new ArrayList<>(options != null ? options : List.of());
+        private static Optional<EnumSymbol> getEnumSymbol(TypeSymbol typeSymbol) {
```
dulajdilshan, line 1407: **[Medium] Duplicate implementation: 4 helpers added byte-identical to both generators.**

`getEnumSymbol`, `addEnumOptions`, `findMatchingOption`, `removeModulePrefix` -- `diff` of `Property.java:1397-1476` vs `PropertyType.java:278-350` is empty apart from indentation. The rewritten union walk (`:804-828` vs `:137-166`) too.

Shared home already exists and both already use it:
- both `build.gradle` -> `implementation project(":model-generator-commons")`
- both already `import io.ballerina.modelgenerator.commons.CommonUtils;`

Only blocker: `Option` is itself duplicated -- the two `Option.java` differ **only** in the `package` line.

Already drifting: the 5th helper is not identical --

```java
// Property.java:1468 - falls back to builder state
String declaredDefault = defaultValue == null || defaultValue.isEmpty()
        ? builder.defaultValue : defaultValue;
// PropertyType.java:307 - no fallback
valueBuilder.setPlaceholder(findMatchingOption(options, defaultValue)...);
```

Same name, different answer for the same input. And the FMG fallback reads mutable builder state, so it is only right where the call site sets `.defaultValue()` *before* `.typeWithExpression()` (`McpToolKitBuilder.java:216`/`:246` does; nothing enforces it).

Suggest: `Option` + the 4 helpers -> `model-generator-commons`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
```diff
@@ -1374,36 +1389,90 @@ private ValueType findMatchingValueType(Node node) {
...
+         */
+        private static Optional<Option> findMatchingOption(List<Option> options, String defaultValue) {
+            if (options == null || options.isEmpty() || defaultValue == null || defaultValue.isEmpty()) {
+                return Optional.empty();
+            }
+            String value = CommonUtils.removeQuotes(defaultValue);
+            String memberName = removeModulePrefix(value);
+            return options.stream()
+                    .filter(option -> value.equals(CommonUtils.removeQuotes(option.value()))
+                            || memberName.equals(option.label()))
```
dulajdilshan: **[Medium] Constant unions never match, so the default gets written out redundantly.**

`getEnumSymbol` resolves `EnumSymbol` only. `grpc:Compression` is declared **constants**, not an enum:

| | |
|---|---|
| declared default | `COMPRESSION_AUTO` |
| option value | `"AUTO"` -> `removeQuotes` -> `AUTO` |
| option label | `AUTO` |
| `removeModulePrefix("COMPRESSION_AUTO")` | `COMPRESSION_AUTO` (no colon) |

Neither arm of the filter matches, placeholder cleared. 7 fixtures lose `"placeholder": "\"AUTO\""` while keeping `"defaultValue": "COMPRESSION_AUTO"` (`search_nodes/config/config4.json`, `config1`, `config2`, `config7`, `config9`, `config11`, ...).

Consequence via `SourceBuilder.java:850`: picking AUTO now writes `compression = "AUTO"` explicitly where it used to be omitted.

`ConstantSymbol` is the sibling branch of the `instanceof` that is already there.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
```diff
@@ -793,68 +796,80 @@ public Builder<T> typeWithExpression(TypeSymbol typeSymbol, ModuleInfo moduleInf
...
+                        // Each member is turned into an input type as it is declared, so that the modes of the
+                        // field are offered in the order of the union. The singletons share one single select
+                        // rather than getting one each, which is added where the first of them is declared and
+                        // holds the very list the remaining ones keep being collected into.
+                        boolean addedSingleSelect = false;
+                        for (TypeSymbol symbol : unionTypeSymbol.memberTypeDescriptors()) {
+                            TypeDescKind memberTypeKind = CommonUtil.getRawType(symbol).typeKind();
+                            if (memberTypeKind == TypeDescKind.SINGLETON) {
+                                if (!addedSingleSelect) {
+                                    builder.type().fieldType(ValueType.SINGLE_SELECT).options(options).stepOut();
```
dulajdilshan: **[Medium] The select is built before its options exist, then mutated through the record.**

```java
:820  builder.type().fieldType(ValueType.SINGLE_SELECT).options(options).stepOut();
:824  options.add(new Option(label, symbol.signature()));   // list already inside the record
```

`stepOut()` constructs the `PropertyType` immediately (`:705`), and `TypeBuilder.options()` stores the reference (`:669`) -- no copy anywhere in that chain. So every option after the first arrives by mutating a list a built record holds.

Add `List.copyOf` to `options()` or `stepOut()` at any point later and **every enum dropdown silently comes back empty** -- no unit test fails, only the JSON fixtures.

Alternative that needs no aliasing: remember the index where the first singleton was seen, add the select after the loop with the finished list.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
```diff
@@ -793,68 +796,80 @@ public Builder<T> typeWithExpression(TypeSymbol typeSymbol, ModuleInfo moduleInf
...
+                                    addedSingleSelect = true;
+                                }
+                                // Skip the singletons that are already covered by the options of an enum
+                                if (!enumMemberTypes.contains(symbol.signature())) {
+                                    String label = CommonUtils.removeQuotes(symbol.signature());
+                                    options.add(new Option(label, symbol.signature()));
+                                }
+                            } else if (memberTypeKind != TypeDescKind.NIL) {
```
dulajdilshan, line 829: **[Medium] Nil dropped on the strength of a flag that means something else.**

```java
} else if (memberTypeKind != TypeDescKind.NIL) {
    // The nil member is conveyed by the `optional` flag of the property
```

But `optional` is defaultability, not nilability:

```java
// ParameterData.java:40
 * @param optional  whether the parameter can be left blank by the user
```

A required `Day? day` has `optional: false`, so nothing conveys the nil member and its input mode is gone -- `()` only typeable via the trailing EXPRESSION mode. Same at `PropertyType.java:166`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
```diff
@@ -1374,36 +1389,90 @@ private ValueType findMatchingValueType(Node node) {
...
+         *
+         * @param enumSymbol      the enum definition
+         * @param options         the options to append to
+         * @param enumMemberTypes collects the signatures of the singleton types covered by the added options
+         */
+        private static void addEnumOptions(EnumSymbol enumSymbol, List<Option> options, Set<String> enumMemberTypes) {
+            // The members are returned in the reverse order of their declaration
+            for (ConstantSymbol enumMember : enumSymbol.members().reversed()) {
+                String memberValue = enumMember.typeDescriptor().signature();
+                enumMemberTypes.add(memberValue);
```
dulajdilshan: **[Minor] A member with no name suppresses its own singleton and disappears.**

```java
enumMemberTypes.add(memberValue);                                    // registers as "covered"
enumMember.getName().ifPresent(name -> options.add(new Option(...)));  // may not add
```

`enumMemberTypes` is the skip set at `:824`, so on an empty `getName()` the singleton is skipped there *and* no option was added -> the member vanishes from the dropdown instead of falling back to a signature-labelled option.

Move the `add` inside `ifPresent`. Same at `PropertyType.java:342`.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/EnumEditor/EnumEditor.tsx
```diff
@@ -30,34 +30,66 @@ interface EnumEditorProps {
...
-            ...baseItems,
+    const options = useMemo(
+        () => (props.items.length > 0 ? props.items : (props.field.itemOptions ?? [])),
+        [props.items, props.field.itemOptions]
+    );
+
+    // The member the parameter defaults to, which the placeholder holds as the value of that option. An
+    // optional enum defaults to nil, which refers to no member, hence there is not always one.
+    const defaultOption = useMemo(
+        () => options.find(item => item.value === props.field.placeholder),
```
dulajdilshan, line 42: **[Medium] `placeholder` now carries two incompatible meanings.**

Here it is the default option's *value*:

```tsx
options.find(item => item.value === props.field.placeholder)
```

Everywhere else in the same render path it is hint text -- `ExpressionField.tsx` forwards `placeholder={field.placeholder}` at `:236 :259 :284 :309 :333 :356 :380`. Mixed unions this PR introduces hit both consumers off one string: `enum_string_union.json` -> `types: [SINGLE_SELECT, TEXT, EXPRESSION]`, `placeholder: "\"Monday\""`.

And the sibling editor reads a different field for the same concept:

| editor | source of "the declared default" |
|---|---|
| `EnumEditor.tsx:41` | `field.placeholder` |
| `BooleanEditor.tsx:102` | `field.defaultValue` |

`field.defaultValue` is present in both cases (`"defaultValue": "COMPRESSION_AUTO"` etc. in the fixtures). Reading it in both, or adding an explicit `defaultOption` field, removes the overload.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/EnumEditor/EnumEditor.tsx
```diff
@@ -30,34 +30,66 @@ interface EnumEditorProps {
...
 
-    const selectedValue = useMemo(() => {
-        if (props.value === undefined || props.value === null || props.value === "") {
-            return DEFAULT_NONE_SELECTED_VALUE;
+    // Emptying the field and never having touched it are the same state of the parameter, as both leave the
+    // argument out of the call and let the declared default apply. They are not the same to the reader
+    // though: one is a choice just made, and presenting the default in its place would drop the selection in
+    // front of the person who made it. Which of the two it is, is therefore remembered here rather than read
+    // back from the value.
+    const [emptiedByUser, setEmptiedByUser] = useState(false);
```
dulajdilshan: **[Medium] Duplicated Code: the `emptiedByUser` mechanism written twice.**

Same non-obvious mechanism -- state + reset effect + "show default until emptied" branch -- in two sibling editors under `MultiModeExpressionEditor`:

| | `EnumEditor.tsx` | `BooleanEditor.tsx` |
|---|---|---|
| state | `:68` | `:71` |
| reset effect | `:69-71` | `:72-76` |
| branch | `:83-84` | `:101-102` |

Reset conditions have already drifted: `if (props.value)` vs `if (toBooleanValue(value) !== undefined)`.

One `useDefaultUntilEmptied(value, defaultValue)` hook covers both and keeps them from drifting further.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/model/PropertyType.java
```diff
@@ -240,6 +266,84 @@ public static void typeWithExpression(Value.ValueBuilder valueBuilder, TypeSymbo
...
+    }
+
+    /**
+     * Sets the placeholder of a single select to the option the parameter defaults to, and to nothing when there
+     * is no such option.
+     *
+     * <p>On a single select the placeholder names the member the field defaults to, never a sample value of the
+     * type: the type of a union yields an arbitrary member of it, which says nothing about the default. The
+     * declared default is therefore the only source, and it is not always resolvable to an option - the parameter
+     * may declare none, or the union may refer to constants rather than enum members (e.g. {@code ftp:Protocol}
```
dulajdilshan: **[Nitpick] Javadoc does not parse -- copy artifact.**

> the union may refer to constants rather than enum members (e.g. `ftp:Protocol` **against** `http:Compression`)

`Property.java:1452` has the intended wording: "e.g. `http:Compression` is COMPRESSION_AUTO|COMPRESSION_ALWAYS|...".

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/EnumEditor/EnumEditor.tsx
samithkavishke: The defaultValue not present in almost most of the cases in the language server ( serviceModel, etc..) So for now this PR does not aim to refactor the that behavior and we are keeping the same behavior which we had.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/SourceBuilder.java
samithkavishke: Currently, this is not present in almost all the cases and we are sticking to the same behavior for this also, Current contract is frontend should give all the required parameters, so if placeholder is not present UI should not enable the Save button.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/EnumEditor/EnumEditor.tsx
samithkavishke: Addressed

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/model/PropertyType.java
samithkavishke: Addressed

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
samithkavishke: This is not by this PR, which is already duplicated across service model and the flow model will be addressed through separate PR

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
samithkavishke: Addressed

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
samithkavishke: Comment has been changed, and the NIL isn't dropped as discussed with @rosensilva we are only giving support to add NIL through the expression mode rather than the SINGLE_SELECT option.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
samithkavishke: This has been addressed algorithm has been altered for this case

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
samithkavishke: Addressed

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
```diff
@@ -793,68 +796,90 @@ public Builder<T> typeWithExpression(TypeSymbol typeSymbol, ModuleInfo moduleInf
...
                         }
 
-                        // If all the member types are singletons, treat it as a single-select option
-                        if (allSingletons) {
-                            // Reorder options so that the default value appears first
-                            if (defaultValue != null && !defaultValue.isEmpty()) {
-                                options = reorderOptionsByDefaultValue(options, defaultValue);
+                        for (int i = 0; i <= otherTypes.size(); i++) {
+                            if (i == selectPosition) {
+                                builder.type().fieldType(ValueType.SINGLE_SELECT).options(options).stepOut();
```
NipunaRanasinghe, line 838: **High — a union with three input types still hides one of them, whichever end the select lands on.**

`FieldFactory` keeps only `types[0]` and `types[last]` (`FieldFactory.tsx:199-201`) and `ModeSwitcher` offers exactly those two (`ModeSwitcher/index.tsx:48,55`), so a middle input type is unreachable. Your new `int_enum_union.json` is `[NUMBER, SINGLE_SELECT, EXPRESSION]`, so `int|Day` now renders no `Day` dropdown at all — the feature this PR adds is invisible for that signature; `maxIter` (`"INFER_TOOL_COUNT"|int`, still `[SINGLE_SELECT, NUMBER, EXPRESSION]`) still has no number editor. Following the declaration order is right and I agree it settles the arbitrariness, but it moves which mode is lost rather than stopping the loss.

This needs `FieldFactory` to render every type instead of first-and-last; no ordering here can fix it.

samithkavishke: This is the current behavior in the fieldFactory which only supports two types at once, only the first one and the last one.

NipunaRanasinghe: @samithkavishke in that case, +1 to create an issue for that as a future improvement and merge with the current state

NipunaRanasinghe (approved):
LGTM

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/Property.java
samithkavishke: Tracked here https://github.com/wso2/product-integrator/issues/2404

PR #1019, opened 2026-09-06
Fix error when switching from boolean to expression

thuva9872:
## Purpose
Fixes: https://github.com/wso2/product-integrator/issues/2307

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

- Fixed the AWS SQS **Auto Delete Messages** field when switching from boolean input to Expression mode.
- Normalized boolean and numeric values before passing them to the chip expression editor.
- Added regression tests for inline and expanded expression editors.

### DIFF
--- a/packages/ballerina-side-panel/src/components/editors/ExpressionField.tsx
+++ b/packages/ballerina-side-panel/src/components/editors/ExpressionField.tsx
@@ -367,7 +367,7 @@ export const ExpressionField: React.FC<ExpressionFieldProps> = (props: Expressio
             completions={completions}
             onChange={onChange}
             onBlur={onBlur}
-            value={value}
+            value={typeof value === "string" || value == null ? value : String(value)}
             sanitizedExpression={sanitizedExpression}
             rawExpression={rawExpression}
             fileName={fileName}
--- a/packages/ballerina-side-panel/src/components/editors/ExpressionField.tsx
+++ b/packages/ballerina-side-panel/src/components/editors/ExpressionField.tsx
@@ -367,7 +367,7 @@ export const ExpressionField: React.FC<ExpressionFieldProps> = (props: Expressio
             completions={completions}
             onChange={onChange}
             onBlur={onBlur}
-            value={typeof value === "string" || value == null ? value : String(value)}
+            value={value}
             sanitizedExpression={sanitizedExpression}
             rawExpression={rawExpression}
             fileName={fileName}
--- a/packages/ballerina-side-panel/src/test/ExpressionField.test.tsx
+++ b/packages/ballerina-side-panel/src/test/ExpressionField.test.tsx
@@ -1,0 +1,120 @@
+/**
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com) All Rights Reserved.
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// L1 regression test for issue #2307: switching a FLAG (boolean) field to
+// Expression mode crashed because the raw JS boolean form value was handed to
+// the CodeMirror-backed chip editor, whose doc/insert APIs require a string
+// (EditorState.create({ doc: <boolean> }) throws). jsdom cannot instantiate the
+// CodeMirror editor (see docs/TEST_GUIDE "jsdom limits"), so instead of driving
+// the live editor we stub ChipExpressionEditorComponent and assert the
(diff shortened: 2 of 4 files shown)


NipunaRanasinghe (commented):
@thuva9872 can we add a regression test at the unit/component level to ensure this doesn’t happen again?

pasindufernando1 (commented):
Noticed that once we switch to the expression mode, the listener configurations section becomes focussed which would confuse the user. 

https://github.com/user-attachments/assets/f259ad99-ccde-47fb-98cd-406df15a6fa4

packages/ballerina-side-panel/src/components/editors/ExpressionField.tsx
```diff
@@ -367,7 +367,7 @@ export const ExpressionField: React.FC<ExpressionFieldProps> = (props: Expressio
             completions={completions}
             onChange={onChange}
             onBlur={onBlur}
-            value={value}
+            value={typeof value === "string" || value == null ? value : String(value)}
```
NipunaRanasinghe: **Medium — expanded mode still hands the raw boolean to the same editor.**

`InputMode.EXP` is in `EXPANDABLE_MODES`, so the expand icon renders on this field and `ExpandedEditor` (`ExpressionEditor.tsx:931`) passes `watch(key)` straight through to `ExpressionMode`, which mounts the same `ChipExpressionEditorComponent` with `value={value}` (`ExpandedEditor/modes/ExpressionMode.tsx:51`). This coercion is display-only and nothing writes back, so the form value is still boolean `false` when the user clicks expand on the AWS SQS field, and the original error reappears there.

Normalizing inside `ChipExpressionEditorComponent` — its prop is already declared `value: string | undefined` — would cover both call sites instead of one.

packages/ballerina-side-panel/src/test/ExpressionField.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+        renderExpressionField(value);
+        expect(chipValues.at(-1)).toBe(value);
+    });
+
+    it.each([
+        ["null", null],
+        ["undefined", undefined],
+    ])("leaves %s as-is (never coerced to a string)", (_desc, value) => {
+        renderExpressionField(value);
+        expect(chipValues.at(-1)).toBe(value);
```
NipunaRanasinghe: **Low — the `undefined` row passes even if the chip editor never renders.**

`chipValues.at(-1)` returns `undefined` for an empty array, so this assertion holds for the `undefined` case whether `ExpressionField` rendered the stub or bailed out at one of its early guards. The `null` row and the two boolean rows don't have that problem. `expect(chipValues).toHaveLength(1)` before the assertion pins it down.

thuva9872: resolved in [19b3369](https://github.com/ballerina-platform/ballerina-vscode/pull/1019/commits/19b336973733defb212a02e88d6b7c502c1be2df)

packages/ballerina-side-panel/src/components/editors/ExpressionField.tsx
thuva9872: resolved in [19b3369](https://github.com/ballerina-platform/ballerina-vscode/pull/1019/commits/19b336973733defb212a02e88d6b7c502c1be2df)

packages/ballerina-side-panel/src/test/ExpressionField.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+    // both must delegate the value to the single ChipExpressionEditorComponent
+    // that applies coerceChipEditorValue. The earlier inline-only fix left the
+    // expanded site (below) unguarded.
+    it.each([
+        ["inline editor (ExpressionField)", renderInlineEditor],
+        ["expanded editor (ExpressionMode)", renderExpandedEditor],
+    ])("%s hands a raw boolean to the chip editor without crashing", (_desc, renderFn) => {
+        expect(() => renderFn(false)).not.toThrow();
+        expect(chipValues).toHaveLength(1);
+        expect(chipValues.at(-1)).toBe(false);
```
NipunaRanasinghe, line 136: **Low — nothing covers the helper actually being wired into the component.**

The first describe tests `coerceChipEditorValue` standalone; this one stubs `ChipExpressionEditorComponent` out entirely and asserts it receives the raw `false`. Delete the `coerceChipEditorValue(props.value)` call at `ChipExpressionEditor.tsx:114` and both describes still pass, even though that line is the fix.

The seam is reasonable given jsdom can't instantiate CodeMirror, so this is fine to leave — but if you want it covered, asserting that `configuration.serializeValue` is handed a string would catch it.

NipunaRanasinghe (approved):
Both findings from the previous round are fixed. 

Approving with One minor suggestion:

`handleModeChange` (`FieldFactory.tsx:238`) still converts only when `typeof currentFieldValue === 'string'`, so after switching a FLAG field to Expression the form state holds boolean `false` while the editor shows `"false"`. Nothing writes back — `normalizeEditorValue(deserializeValue) !== normalizeEditorValue(normalizedValue)` is false once both are `"false"`. Not a crash and not a regression from this PR.

packages/ballerina-side-panel/src/test/ExpressionField.test.tsx
thuva9872: The standalone coerceChipEditorValue tests plus the "both mount sites render without throwing" coverage give us reasonable protection against the regression, so I'll leave the wiring assertion out for now.

PR #1043, opened 2026-09-07
Keep chat scroll position stable during streaming, add jump-to-bottom control

VellummyilumVinoth:
## Summary
- Auto-scroll previously fired unconditionally on every streamed update, snapping the chat back to the bottom even while the user was scrolled up reading earlier content.
- Auto-scroll now only runs while the user is pinned to the bottom; scrolling up disables it instead of fighting the user.
- Added a floating "Jump to bottom" button (styled like Claude Code's own control) that appears when scrolled up, dynamically tracks the height of the footer/composer stack via a ResizeObserver so it always floats just above it, and re-pins to the bottom on click.

It resolves: https://github.com/wso2/product-integrator/issues/2359

For reference Sample demo:

https://github.com/user-attachments/assets/4c284ea8-1a36-4ec3-8c4a-aa360ca1a9f5

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary

- Preserve manual upward scrolling during streamed chat responses.
- Continue auto-scrolling while the user remains at the bottom.
- Add a floating “Jump to bottom” button with footer-aware positioning.
- Restore bottom positioning and auto-scroll when the button is selected.
- Reset scroll state for new chats, thread changes, deletions, and user turns.

### DIFF
--- a/packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/index.tsx
+++ b/packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/index.tsx
@@ -58,7 +58,7 @@ import FollowupSuggestions from "../FollowupSuggestions";
 import { Attachment, AttachmentStatus, SkillEnableStage, SkillEntry, TaskApprovalRequest } from "@wso2/ballerina-core";
 import type { ClarifyEvent, ConfigurationCollectionEvent, ConnectorGenerationNotification } from "@wso2/ballerina-core";
 
-import { AIChatView, Header, HeaderButtons, ChatMessage, TurnGroup, AuthProviderChip, UsageBadge, UsageRefreshButton, ApprovalOverlay, OverlayMessage, OverlayCloseButton } from "../../styles";
+import { AIChatView, Header, HeaderButtons, ChatMessage, TurnGroup, AuthProviderChip, UsageBadge, UsageRefreshButton, ApprovalOverlay, OverlayMessage, OverlayCloseButton, JumpToBottomButton } from "../../styles";
 import { SessionHistoryDropdown } from "../SessionHistory";
 import ReferenceDropdown from "../ReferenceDropdown";
 import { VSCodeButton } from "@vscode/webview-ui-toolkit/react";
@@ -112,6 +112,87 @@ const DRIFT_CHECK_ERROR = "Failed to check drift between the code and the docume
 
 const USAGE_EXCEEDED_THRESHOLD_PERCENT = 3;
 const DISCORD_INVITE_URL = "https://discord.com/invite/wso2";
+
+// Distance (px) from the bottom still considered "pinned" — absorbs late layout growth during streaming.
+const BOTTOM_THRESHOLD_PX = 80;
+
+interface ScrollMetrics {
+    scrollHeight: number;
+    scrollTop: number;
+    clientHeight: number;
+}
+
+function isPinnedToBottom(el: ScrollMetrics, threshold: number): boolean {
+    return el.scrollHeight - el.scrollTop - el.clientHeight <= threshold;
+}
+
+// Tracks whether a scroll was caused by our own programmatic scrollIntoView (ignore) or the
+// user (don't). markUserIntent always wins over begin, even mid-guard-window.
+interface AutoScrollGuard {
+    isAutoScrolling(): boolean;
+    markUserIntent(): void;
+    begin(onFallbackSettle: () => void, fallbackMs: number): void;
+    settle(): void;
+    dispose(): void;
+}
+
+function createAutoScrollGuard(): AutoScrollGuard {
+    let autoScrolling = false;
+    let timeoutId: ReturnType<typeof setTimeout> | undefined;
+
+    function clearPendingTimeout(): void {
+        if (timeoutId !== undefined) {
+            clearTimeout(timeoutId);
+            timeoutId = undefined;
+        }
+    }
+
+    return {
+        isAutoScrolling: () => autoScrolling,
+        markUserIntent: () => {
+            autoScrolling = false;
+        },
+        begin: (onFallbackSettle, fallbackMs) => {
+            clearPendingTimeout();
+            autoScrolling = true;
+            timeoutId = setTimeout(() => {
+                timeoutId = undefined;
+                autoScrolling = false;
+                onFallbackSettle();
+            }, fallbackMs);
+        },
+        settle: () => {
+            clearPendingTimeout();
+            autoScrolling = false;
+        },
+        dispose: () => {
+            clearPendingTimeout();
+        },
+    };
+}
+
+// Callback ref (not a mount-once effect) so it reattaches whenever the node changes, including a
+// remount after unmount.
+function useResizeObserverRef<T extends Element>(
+    onResize: (rect: DOMRectReadOnly) => void
+): (node: T | null) => void {
+    const observerRef = useRef<ResizeObserver>();
+
+    return useCallback((node: T | null) => {
+        observerRef.current?.disconnect();
+        observerRef.current = undefined;
+        if (!node) {
+            return;
+        }
+        const observer = new ResizeObserver((entries) => {
+            for (const entry of entries) {
+                onResize(entry.contentRect);
+            }
+        });
+        observer.observe(node);
+        observerRef.current = observer;
+    }, [onResize]);
+}
 
 //TODO: Add better error handling from backend. stream error type and non 200 status codes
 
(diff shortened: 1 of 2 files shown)


packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/index.tsx
```diff
@@ -1744,8 +1774,16 @@ const AIChat: React.FC = () => {
...
+        // User has scrolled up to read earlier content — don't fight them.
+        if (!isPinnedToBottomRef.current) {
+            return;
+        }
         const scrollToEnd = (behavior: ScrollBehavior) => {
+            isAutoScrollingRef.current = true;
             messagesEndRef.current?.scrollIntoView({ behavior, block: "end" });
+            window.setTimeout(() => {
+                isAutoScrollingRef.current = false;
+            }, behavior === "smooth" ? 300 : 50);
```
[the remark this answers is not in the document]
VellummyilumVinoth: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/1043/commits/7af406cd9d0ab6de3c60149521de3631582c12f9

packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/index.tsx
```diff
@@ -1757,6 +1795,32 @@ const AIChat: React.FC = () => {
...
+
+    const handleJumpToBottom = useCallback(() => {
+        isPinnedToBottomRef.current = true;
+        setShowJumpToBottom(false);
+        isAutoScrollingRef.current = true;
+        messagesEndRef.current?.scrollIntoView({ behavior: "smooth", block: "end" });
+        window.setTimeout(() => {
+            isAutoScrollingRef.current = false;
+        }, 300);
+    }, []);
```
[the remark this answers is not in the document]
VellummyilumVinoth: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/1043/commits/7af406cd9d0ab6de3c60149521de3631582c12f9

submodules/wso2-vscode-extensions
```diff
@@ -1 +1 @@
-Subproject commit d3dcf45ae01e916f33f5ab8920848493802edc26
+Subproject commit 09b37c86c93420e45f7d5b5ffb2b3c200c4c70d3
```
RNViththagan: This moves the submodule back about 110 commits (`d3dcf45ae0` to `09b37c86c9`). Staging hasn't touched this pointer since your branch was cut, so merging as-is would drop the icon/HITL updates that landed on staging in the meantime (`b359337d87`, "Bump the submodule for the missing artifact icons and resync codepoints"). Looks like a stale local checkout got picked up by a `git add -A`. Shall we drop this hunk and re-point it at `d3dcf45ae0` before merging?

packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/index.tsx
```diff
@@ -1743,19 +1773,65 @@ const AIChat: React.FC = () => {
...
     useEffect(() => {
-        const scrollToEnd = (behavior: ScrollBehavior) => {
-            messagesEndRef.current?.scrollIntoView({ behavior, block: "end" });
-        };
-        scrollToEnd("smooth");
+        // User has scrolled up to read earlier content — don't fight them.
+        if (!isPinnedToBottomRef.current) {
+            return;
+        }
+        runProgrammaticScroll("smooth");
```
RNViththagan, line 1900: I think the auto-scroll-fighting-the-user case can still happen during continuous streaming: this effect re-runs on every `messages` update, and each run calls `runProgrammaticScroll` unconditionally while `isPinnedToBottomRef.current` is true. `runProgrammaticScroll` resets the 300ms guard on every call, so while chunks keep arriving faster than 300ms apart, `isAutoScrollingRef` never actually drops back to false - `handleScroll` keeps ignoring every scroll event, including a genuine user wheel-up, and the next chunk pulls them back down. It only breaks out once there's a real pause (a tool call, model thinking time). Shall we detect real user intent from `wheel`/`touchmove`/`keydown` on the scroller instead of inferring it from `scroll` events plus a timer?

packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/index.tsx
```diff
@@ -477,6 +482,31 @@ const AIChat: React.FC = () => {
...
+            return;
+        }
+        const observer = new ResizeObserver((entries) => {
+            for (const entry of entries) {
+                setFooterHeight(entry.contentRect.height);
+            }
+        });
+        observer.observe(el);
+        return () => observer.disconnect();
+    }, []);
```
RNViththagan, line 586: This effect has empty deps, so it only ever observes the footer div that existed at first mount. When Settings/MCP/Skills gets pushed, `AIChatView` - and this `footerRef` div - unmounts; on return, a new div is created but this effect never re-runs to observe it, so `footerHeight` freezes at its last value and the jump-to-bottom button can end up misplaced for the rest of the session. Shall we switch to a callback ref that (re)attaches the observer whenever the node changes, rather than a one-shot effect?

packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/index.tsx
```diff
@@ -2199,6 +2275,8 @@ const AIChat: React.FC = () => {
 
     async function handleClearChat(): Promise<void> {
         setMessages([]);
+        isPinnedToBottomRef.current = true;
```
RNViththagan: Noticed `handleSendQuery` doesn't re-pin (only clear/switch-thread/delete-thread and the jump button do, same as here). So if the user is scrolled up reading earlier messages and sends a new prompt, their own message and the streamed reply both append off-screen, with the floating button as the only hint. Shall we re-pin in `handleSendQuery` too, the same way this does?

packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/index.tsx
```diff
@@ -1743,19 +1773,65 @@ const AIChat: React.FC = () => {
...
+        if (autoScrollTimeoutRef.current !== undefined) {
+            clearTimeout(autoScrollTimeoutRef.current);
+        }
+        isAutoScrollingRef.current = true;
+        messagesEndRef.current?.scrollIntoView({ behavior, block: "end" });
+        autoScrollTimeoutRef.current = setTimeout(() => {
+            isAutoScrollingRef.current = false;
+            autoScrollTimeoutRef.current = undefined;
+            checkPinnedState();
+        }, behavior === "smooth" ? 300 : 50);
```
RNViththagan: The guard window is a fixed 300ms regardless of how long the actual smooth-scroll animation takes - for a big jump (a large tool-result card, a code block reflowing) the browser's scroll animation can run past that. If `checkPinnedState` fires while it's still mid-flight, it can read the mid-animation position as "not at bottom" and, if a chunk lands right after, the effect sees pinned=false and skips auto-scrolling, silently dropping auto-follow with no user action. Shall we key off something like a `scrollend` listener (or an explicit "still animating" flag) instead of a fixed timer?

packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/utils/scrollPinning.test.ts
```diff
@@ -0,0 +1,56 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { distanceFromBottom, isPinnedToBottom } from "./scrollPinning";
+
+describe("isPinnedToBottom", () => {
```
RNViththagan: These cover the threshold arithmetic well, but the trickier behavior - the guard re-arming on rapid updates, the ResizeObserver re-attach case, the footer-remount case - isn't under test yet. Given how easy this class of bug is to reintroduce, would a couple of component-level tests (simulating rapid `messages` updates plus a user scroll event) be worth adding here?

packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/index.tsx
```diff
@@ -113,6 +114,10 @@ const DRIFT_CHECK_ERROR = "Failed to check drift between the code and the docume
 const USAGE_EXCEEDED_THRESHOLD_PERCENT = 3;
 const QUOTA_CONTACT_EMAIL = "support@wso2.com";
 
+// Distance (px) from the bottom still considered "pinned" — absorbs late layout growth
```
RNViththagan: A few of the new comments run two lines where the why fits in one - this one, the footerRef one around line 492, and the JSDoc blocks in scrollPinning.ts. Not a big deal, just flagging since the convention here is one line unless it really needs more.

packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/utils/scrollPinning.test.ts
VellummyilumVinoth: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/1043/commits/7e7ffeea1b1a5780c5792e7e162554d6d17eb90f

packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/index.tsx
VellummyilumVinoth: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/1043/commits/7e7ffeea1b1a5780c5792e7e162554d6d17eb90f

packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/index.tsx
VellummyilumVinoth: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/1043/commits/7e7ffeea1b1a5780c5792e7e162554d6d17eb90f

packages/ballerina-visualizer/src/views/AIPanel/components/AIChat/index.tsx
VellummyilumVinoth: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/1043/commits/7e7ffeea1b1a5780c5792e7e162554d6d17eb90f

submodules/wso2-vscode-extensions
RNViththagan: Just checking on this one - the submodule pointer at head `a11e6eee07` is still `09b37c86c9`, staging is still at `d3dcf45ae0`, so this would still regress the submodule if merged as-is. Everything else here looks fixed. Is this one still pending, or did it get missed?

RNViththagan (approved):
LGTM

SasinduDilshara (approved):

PR #1044, opened 2026-09-07
Add agent building skill

damsith0:
## Summary

- Add the default `agent-builder` skill for generating valid Ballerina `ai:Agent` code.
- Document agent tools, approval predicates, subagents, knowledge bases, and session isolation.
- Register the optional skill with frontmatter validation and default enablement.

### DIFF
--- a/packages/ballerina-extension/src/features/ai/agent/skills/agent-builder/SKILL.md
+++ b/packages/ballerina-extension/src/features/ai/agent/skills/agent-builder/SKILL.md
@@ -252,7 +252,8 @@ final ai:VectorKnowledgeBase <agent>KnowledgeBase =
 
 `ingest` accepts `Document`, `Document[]` or `Chunk[]` — pass the loader's result straight through
 without unwrapping it. `retrieve(query, <limit>)` returns `ai:QueryMatch[]`, each carrying its text
-at `chunk.content`.
+at `chunk.content`. That field is typed `anydata` on the generic `ai:Chunk` — convert it
+(`chunk.content.toString()`) or narrow to `ai:TextChunk` before treating it as a `string`.
 
 The retrieval tool returns those excerpts as text and stops there. Do not have it answer the
 question itself — the agent's own instructions decide how the excerpts are used.
@@ -372,6 +373,11 @@ An event source is not a webhook: there is no acknowledgement, so blocking slows
 past the poll deadline, rebalances the group and reprocesses the events. Keep the `start` and the
 error handling, drop the session id (events are not conversations), and end the reply method with a
 `// TODO:` comment above a `log:printInfo` of the result so the unfinished step shows in the diagram.
+
+Dropping the session id is only safe when the agent is constructed with `memory = ()`. Omitting
+`sessionId` from `run` does not disable memory — it defaults to a fixed id, so every event this
+listener processes would otherwise append to and read from the same shared history. If the agent
+does have memory, generate a per-event id instead of dropping it.
 
 ### HTTP endpoints are the exception
 
--- a/packages/ballerina-extension/src/features/ai/agent/skills/agent-builder/SKILL.md
+++ b/packages/ballerina-extension/src/features/ai/agent/skills/agent-builder/SKILL.md
@@ -61,6 +61,11 @@ the agent's `tools` array. The function MUST be `isolated`, every parameter MUST
 `anydata`, and the return type MUST be a subtype of `anydata`, `stream<anydata>` or
 `http:Response`. A non-isolated function is rejected as a tool.
 
+The one exception to the `anydata` rule is `ai:Context`: a tool may take a single `ai:Context ctx`
+parameter in the first position — the runtime injects it, and the compiler plugin strips it before
+checking anything else about the signature. Name it exactly `ctx`; that is the name the tool
+generator itself reserves. See Human approval and Subagents below for how it interacts with each.
+
 The doc comment is the tool description the model sees at runtime, so always write it — a `#`
 description line, a `# + <param> - ...` line per parameter, and `# + return - ...`.
 
@@ -125,6 +130,10 @@ leave the annotation bare on everything else.
 **the same parameter list as the tool**, and return `boolean` — the compiler rejects any other
 signature. Return `true` to pause.
 
+If the tool has a leading `ai:Context ctx` parameter, omit it from the predicate — the predicate is
+compared only against the tool's own `anydata` parameters, so a predicate that also declares `ctx`
+looks mismatched to the compiler, not more specific.
+
 ```ballerina
 @ai:AgentTool {
     requiresApproval: <predicateName>
@@ -154,18 +163,18 @@ list that tool in the parent's `tools` array:
 
 ```ballerina
 # <what delegating to this subagent accomplishes, one line>
-# + context - Context injected by the runtime; forwarded so the subagent shares the caller's context
+# + ctx - Context injected by the runtime; forwarded so the subagent shares the caller's context
 # + <param> - <the request or payload to send to the subagent>
 # + sessionId - Conversation handle. Generate a unique id to start a new conversation with the agent, and reuse the same id to continue it across turns.
 # + return - <what the subagent returns>
 @ai:AgentTool
-isolated function <subAgent>AgentTool(ai:Context context, string <param>, string sessionId) returns string|error {
-    string response = check <subAgent>Agent.run(<param>, sessionId, context);
+isolated function <subAgent>AgentTool(ai:Context ctx, string <param>, string sessionId) returns string|error {
+    string response = check <subAgent>Agent.run(<param>, sessionId, ctx);
     return response;
 }
 ```
 
-Keep both extra parameters: `ai:Context` carries the parent's context down to the subagent, and
+Keep both extra parameters: `ctx` carries the parent's context down to the subagent, and
 `sessionId` is what lets the subagent hold a multi-turn conversation. Reuse the `sessionId` doc
 line above verbatim — the model reads it to decide when to generate a fresh id and when to reuse
 one.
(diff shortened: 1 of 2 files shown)


packages/ballerina-extension/src/features/ai/agent/skills/agent-builder/SKILL.md
```diff
@@ -0,0 +1,382 @@
...
+
+```ballerina
+import ballerina/ai;
+import ballerina/http;
+
+listener ai:Listener <agent>Listener = new (listenOn = check http:getDefaultListener());
+
+service /<agent\-name> on <agent>Listener {
+    resource function post chat(@http:Payload ai:ChatReqMessage request) returns ai:ChatRespMessage|error {
+        string stringResult = check <agent>Agent.run(request.message, request.sessionId);
```
RNViththagan, line 290: This takes `sessionId` straight from the client-supplied request body and hands it to `Agent.run` with no ownership check. Since sessionId is the key into the agent's persistent memory, any caller who has or guesses another session's id can read and continue a stranger's conversation - and since the doc says "Do not change the chat resource signature," every agent generated with a chat trigger gets this by default. Worth a line here treating sessionId as a bearer credential (unguessable, generated server-side or tied to an authenticated principal) rather than trusting whatever the client sends, or noting that auth belongs at the http:Listener/gateway level.

packages/ballerina-extension/src/features/ai/agent/skills/agent-builder/SKILL.md
```diff
@@ -0,0 +1,382 @@
...
+An HTTP caller is waiting for the answer, so an HTTP resource **does not** offload and has no reply
+client — it returns the agent's result as the response:
+
+```ballerina
+resource function post <path>(@http:Payload <RequestType> request) returns <ResponseType>|error {
+    string prompt = string `<what the agent should do>
+
+Request payload:
+${request.toJsonString()}`;
+    <ResponseType>|error response = <agent>Agent.run(prompt);
```
RNViththagan, line 393: This section's own text at line 173 says: "Give sessionId no default. A default makes every call that leaves it out share one memory bucket, so unrelated requests from different callers see each other's history." This HTTP-endpoint example calls run(prompt) with no sessionId at all, which is exactly that default - every caller of the generated endpoint shares one conversation history by default, with no mention of setting memory = () to opt out. The event-source guidance a few lines up ("drop the session id") has the identical issue for Kafka/event triggers. Since this section is titled "the exception" to the sessionId-handling rules above, it reads like an intentional carve-out rather than an oversight - was that the intent, or should both examples either generate a per-request sessionId or set memory = ()?

packages/ballerina-extension/src/features/ai/agent/skills/agent-builder/SKILL.md
```diff
@@ -0,0 +1,382 @@
...
+```
+
+`ai:Wso2ModelProvider` not `ai:ModelProvider`; `ai:Wso2EmbeddingProvider` not
+`ai:EmbeddingProvider`; `ai:InMemoryVectorStore` not `ai:VectorStore`. When the user names a
+provider (OpenAI, Azure, Anthropic, Ollama, …) declare that provider's own class, same rule.
+
+## Agent tools
+
+Every tool is a module-level `isolated function` annotated with `@ai:AgentTool`, listed by name in
+the agent's `tools` array. The function MUST be `isolated`, every parameter MUST be a subtype of
```
RNViththagan, line 60: This says every tool parameter MUST be a subtype of anydata, and the approval-predicate rule at line 125 says the predicate MUST take the same parameter list as the tool - but the subagent example right below (line 162) declares ai:Context context as a tool parameter, and ai:Context isn't anydata. I compiled both variants against ai 1.14.0: a predicate mirroring the tool's full signature (Context included) fails with "the requiresApproval predicate ... must have the same parameter signature as the tool"; dropping Context from the predicate compiles. So the two stated rules are wrong for exactly the case this doc's own example demonstrates - a model following them literally on a gated subagent tool gets a compile error whose message says the opposite of the actual fix. Worth stating the ai:Context exception explicitly in both rules (first, framework-injected parameter, excluded from anydata, and excluded from the predicate's required parameter list).

packages/ballerina-extension/src/features/ai/agent/skills/agent-builder/SKILL.md
```diff
@@ -0,0 +1,382 @@
...
+
+## Knowledge bases
+
+A retrieval-augmented agent needs three declarations, each with its concrete type, plus a tool that
+queries it:
+
+```ballerina
+final ai:Wso2EmbeddingProvider <agent>EmbeddingProvider = check ai:getDefaultEmbeddingProvider();
+final ai:InMemoryVectorStore <agent>VectorStore = check new ai:InMemoryVectorStore();
+final ai:KnowledgeBase <agent>KnowledgeBase =
```
RNViththagan: This declares the abstract ai:KnowledgeBase type while constructing the concrete ai:VectorKnowledgeBase, which contradicts the concrete-type rule stated earlier in the doc ("always declare the concrete class"). It compiles fine, and the flow/design diagram views are tolerant since they key off the constructor - but the artifacts/connections list keys off the declared type and silently drops anything that isn't a recognized class, so this exact example produces a knowledge base that works but never shows up as a connection. Should be `final ai:VectorKnowledgeBase <agent>KnowledgeBase = new (...)` to match every other example in this section.

packages/ballerina-extension/src/features/ai/agent/skills/registry.ts
```diff
@@ -83,7 +84,22 @@ export const skillCreatorSkill: Skill = {
     default: false,
 };
 
+// agent-builder skill
```
RNViththagan, line 87: This throw happens at module evaluation time, and tool-registry.ts imports REGISTERED_SKILLS statically with no try/catch anywhere in the chain - so a frontmatter typo here wouldn't just disable this skill, it would throw during the extension's own module load and could take the whole extension activation down with it. Not new to this PR (data-map already does this above), but this adds a second copy of the same fragile pattern while skill-creator right above stays unguarded (a broken skill-creator frontmatter would silently register a skill with an empty name instead). Might be worth logging-and-skipping the bad skill instead of throwing, applied consistently to all three, rather than adding more copies of a pattern that can crash the whole extension over one skill's markdown.

packages/ballerina-extension/src/features/ai/agent/skills/registry.ts
```diff
@@ -83,7 +84,22 @@ export const skillCreatorSkill: Skill = {
...
+const agentBuilder = parseSkillMd(agentBuilderMd);
+if (!agentBuilder.name || !agentBuilder.description) {
+    throw new Error(`[agent-builder] SKILL.md is missing required frontmatter fields (name="${agentBuilder.name}", description="${agentBuilder.description}")`);
+}
+
+export const agentBuilderSkill: Skill = {
+    name: agentBuilder.name,
+    trigger: agentBuilder.description,
+    content: agentBuilder.body,
+    optional: false,
```
RNViththagan: Curious about optional: false here - data-map has the same combination but it's backing a slash command with its own templates, so users need it always available. This skill doesn't have that coupling. Was making it permanently non-disableable (no toggle in settings) intentional, or would optional: true, default: true (on for everyone by default, but with an escape hatch) fit better?

packages/ballerina-extension/src/features/ai/agent/skills/agent-builder/SKILL.md
```diff
@@ -0,0 +1,382 @@
...
+        new ai:VectorKnowledgeBase(<agent>VectorStore, <agent>EmbeddingProvider);
+```
+
+`ingest` accepts `Document`, `Document[]` or `Chunk[]` — pass the loader's result straight through
+without unwrapping it. `retrieve(query, <limit>)` returns `ai:QueryMatch[]`, each carrying its text
+at `chunk.content`.
```
RNViththagan: Minor: this says retrieve() results carry their text at chunk.content, but doesn't mention that Document.content is typed anydata - a model implementing the retrieval tool from this description alone could reasonably write `string text = match.chunk.content;`, which won't compile. Worth a word on converting it (.toString(), or narrowing to TextChunk).

SasinduDilshara (dismissed):

NipunaRanasinghe (changes requested):
@damaya0 please address and resolve all the review suggestions

packages/ballerina-extension/src/features/ai/agent/skills/agent-builder/SKILL.md
damaya0: fixed

packages/ballerina-extension/src/features/ai/agent/skills/agent-builder/SKILL.md
damaya0: We will note that "drop the session id" is only safe when the agent's memory is ()

packages/ballerina-extension/src/features/ai/agent/skills/registry.ts
damaya0: That wasn't intentional, will make optional: true

packages/ballerina-extension/src/features/ai/agent/skills/registry.ts
damaya0: Discussed to keep it same for now

dan-niles (approved):
LGTM

SasinduDilshara (approved):

samithkavishke (approved):

NipunaRanasinghe (approved):
LGTM. 

@damaya0 Please use imperative mood for the PR title and commit messages (e.g. "Add agent building skill")

PR #1046, opened 2026-09-07
Let user navigation through the generation remount guard

RNViththagan:
Related to https://github.com/wso2/product-integrator/issues/2370

- Add a shared `shouldSuppressDisruptiveTransition` predicate so both suppression guards use the same rule instead of two separately-maintained checks
- Let `updateView` carry a `userInitiated` flag on the `VIEW_UPDATE` event, so real navigation bypasses the guard while the agent's own per-edit remount stream stays suppressed
- Flag `goBack`, `goSelected`, `addToHistory`, and `closeReviewModeIfOpen` as user-initiated
- Add a test for the suppression predicate

## Summary

- Centralized generation-time transition suppression logic.
- Added `userInitiated` metadata to view updates.
- Allowed user navigation during generation while suppressing agent-driven remounts.
- Marked history navigation and review-mode closure as user initiated.
- Added recovery when view lookup fails during generation.
- Added tests for suppression behavior.

### DIFF
--- a/packages/ballerina-extension/src/stateMachine.ts
+++ b/packages/ballerina-extension/src/stateMachine.ts
@@ -428,6 +428,15 @@ const stateMachine = createMachine<MachineContext>(
                                     evalsetData: (context, event) => event.data.evalsetData,
                                     isViewUpdateTransition: false
                                 })
+                            },
+                            // Without this, a throw here (e.g. a project-structure lookup racing a
+                            // Copilot generation's edits) leaves the machine stuck in this state
+                            // forever — falling back to the still-current view beats a wedged panel.
+                            onError: {
+                                target: "viewReady",
+                                actions: (context, event) => {
+                                    console.error('Failed to resolve the view to show', event.data);
+                                }
                             }
                         }
                     },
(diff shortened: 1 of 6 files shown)


packages/ballerina-extension/src/stateMachine.ts
```diff
@@ -1154,10 +1155,10 @@ export function updateView(refreshTreeView?: boolean, updatedIdentifier?: string
     }
 
 
-    // Skip the disruptive remount while a Copilot generation is active; notifyCurrentWebview()
-    // below still fires so the diagram's own content refresh keeps flowing.
-    if (!chatStateStorage.hasAnyActiveExecution()) {
-        stateService.send({ type: "VIEW_UPDATE", viewLocation: lastView ? newLocation : { view: "Overview" } });
+    // Only navigation the user asked for may remount mid-generation; the agent's edits replay this on every write.
+    const viewUpdate = { type: "VIEW_UPDATE", viewLocation: lastView ? newLocation : { view: "Overview" }, userInitiated: options?.userInitiated };
```
madushajg, line 1162: Letting `VIEW_UPDATE` through mid-generation can wedge the panel permanently. The `webViewLoaded` invoke (`showView`) has no `onError`, and it's built as n`ew Promise(async (resolve, reject) => {...})` — a throw inside an async executor never rejects, it just never settles. Reaching Back mid-generation now triggers `getView` → `getViewByArtifacts`, which does `currentProjectArtifacts.projects.find(...)` with no null check (unlike its sibling `getClassMemberArtifactType`, which does guard). Mid-generation is exactly when project structure is being torn down/rebuilt, so find returning undefined or the LS call throwing is realistic — and before this PR, that path was simply unreachable. Suggest adding an `onError` on the `webViewLoaded` invoke, or wrapping the `getView` call in `try/catch` with a fallback resolve.

RNViththagan: Good catch, thanks — confirmed both parts: `getViewByArtifacts` was indeed missing the null guard its sibling `getClassMemberArtifactType` has, and `showView`'s manual `new Promise(async (resolve, reject) => {...})` wrapper meant a throw never called `reject`, so it really would hang forever rather than error out.

Fixed in f956126b13:
- Removed the `new Promise` wrapper — `showView` is now a plain `async` function, so a thrown error correctly rejects the returned promise.
- Added an `onError` on the `webViewLoaded` invoke, falling back to `viewReady` (keeping whatever the last-good view was) instead of wedging.
- Added the missing `if (!project) { ... }` guard in `getViewByArtifacts`, returning the overview (matching the function's own existing "no match" fallback) instead of dereferencing `undefined`.

Verified: compile clean, full jest suite still green (only a pre-existing, unrelated font-codepoint drift test fails).

madushajg (approved):

SasinduDilshara (approved):

SasinduDilshara (approved):

PR #1049, opened 2026-09-07
Fix WSO2 Integrator Copilot server-side compaction (context management + UI)

IsuruMaduranga:
## Summary

Fixes how the WSO2 Integrator Copilot handles Anthropic server-side compaction, in two parts.

### 1. Compaction notice leaked into the chat as a raw tag (primary fix)

When compaction fired, the notice was emitted as a plain `content_block`, so it folded
into the agent-stream JSON blob and rendered as the **literal string**
`<compaction>Context compacted — …</compaction>` in the chat. Separately, the summary
extracted from the compaction block was computed and then **discarded** — never used.

- Emit compaction as a `chat_component` (`componentType: "compaction"`), folded through the
  shared `upsertComponent` path so both chat surfaces persist it consistently, and render it
  as a fixed notice row positioned before the continuing response.
- Flush the compaction block on its own `text-end` rather than deferring to the next
  `text-start`. The notice now lands immediately after the summary block ends — before any
  following tool calls — so it is genuinely before the continuing response, and a second
  back-to-back compaction can no longer reset an unflushed first block and drop its notice.
  A stream-end fallback still covers a block left open at the end of the stream.
- Keep the model-authored summary **internal** (see #2371): the `chat_component` carries no
  summary and nothing renders one. The summary is still used server-side to strip
  `<analysis>` blocks from compaction entries. The manually-triggered `compactConversation`
  action is unaffected.
- Defensive fallback: render any `<compaction>` tag folded into an agent-stream text item
  as a notice row rather than raw text, so transcripts persisted before this change (and
  any stray raw emission) no longer leak the literal tag on reload. That path renders the
  tag's inner text, which only ever held a fixed developer notice — never the summary — and
  is also what shows the compaction-disabled warning.

### 2. Context-management config: compaction only

Switch from two context-management edits (`clear_tool_uses_20250919` + `compact_20260112`)
to **compaction only**, triggering at **500K input tokens**.

Rationale: BI re-sends the entire project source in every turn's user message, so its
context is dominated by that non-clearable dump. Compaction (which summarizes the whole
history) is the load-bearing mechanism; `clear_tool_uses` (which only trims tool results)
does comparatively little for BI and adds prompt-cache churn.

The 500K trigger is **higher than the MI Copilot's 200K**, deliberately — BI's per-turn
baseline (system prompt + whole-codebase dump) is much larger, so a 200K trigger would fire
too eagerly. 500K leaves ample working headroom while staying well within Claude Sonnet 5's
1M-token context window (1M is the default for `claude-sonnet-5`; no beta header is needed).
The floor guard still disables compaction when the fixed per-turn overhead already exceeds
the trigger.

Anthropic does recommend using both edits for tool-heavy agents, so `clear_tool_uses` may be
reintroduced as a **measured follow-up** once its cache impact is verified with `appliedEdits`
logging — it is not removed because it caused a bug.

## Scope note

An earlier report of the Copilot emitting a "CRITICAL CONTEXT TO REMEMBER" hand-off and
stopping was investigated and traced to a **stuck-tool session** (the model looping on a
failing `LibraryGetTool`), **not** compaction — the compact edit has always carried our
own summarization instructions, whose output format is different. No change here targets
that behavior; the verified compaction issues are the two above.

## Testing

- Compaction fires and renders as a notice card (no raw tag, no summary shown) on a long
  tool-heavy Ballerina session (Anthropic own-key).
- Transcripts persisted before the change render the notice cleanly on reload.
- `tsc --noEmit` clean for both `ballerina-extension` and `ballerina-visualizer`; host unit
  tests (174) and webview AI unit tests (56) pass.

## Related

Addresses the compaction-notice leak described in wso2/product-integrator#2371.

## Summary

- Configure server-side compaction with `compact_20260112` at 500K input tokens.
- Emit fixed compaction notices before the continuing response.
- Keep model-authored summaries internal.
- Render persisted and streamed `<compaction>` tags as notices.
- Validate long tool sessions, persisted transcripts, and TypeScript compilation.

### DIFF
--- a/packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
@@ -109,12 +109,10 @@ function supportsCompaction(loginMethod: LoginMethod): boolean {
  * Server-side compaction trigger, in input tokens.
  *
  * Unlike MI, BI re-sends the entire project source in every turn's user message, so its
- * per-turn baseline is much larger — MI's 200K trigger would fire almost immediately and
- * leave too little working headroom. Production target: 350_000.
- *
- * NOTE: set to 100_000 for testing / initial rollout so compaction is easy to exercise.
+ * per-turn baseline is much larger — MI's 200K trigger would fire too eagerly. 500K leaves
+ * ample working headroom while staying well within the 1M context window (Claude Sonnet).
  */
-const COMPACT_TRIGGER_TOKENS = 100_000;
+const COMPACT_TRIGGER_TOKENS = 500_000;
 
 /**
  * Builds providerOptions.anthropic.contextManagement, mirroring MI's approach: compaction
@@ -484,6 +482,18 @@ export class AgentExecutor extends AICommandExecutor<GenerateAgentCodeRequest> {
             let isCompactionBlock = false;
             let compactionContent = '';
             let cleanedCompactionSummary: string | undefined;
+            // Counts compactions in this turn so each renders as its own card (upsertComponent
+            // keys by id). Mirrors MI: a notice + the extracted summary, rendered in-stream —
+            // never emitted as a raw <compaction> text block (which would render as literal text).
+            let compactionCount = 0;
+            const emitCompactionNotice = (summary?: string) => {
+                this.config.eventHandler({
+                    type: 'chat_component',
+                    componentType: 'compaction',
+                    id: `compaction-${this.config.generationId}-${compactionCount++}`,
+                    data: { summary: summary ?? '' },
+                });
+            };
 
             // Send start event to frontend
             this.config.eventHandler({ type: "start" });
--- a/packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
@@ -106,19 +106,14 @@ function supportsCompaction(loginMethod: LoginMethod): boolean {
 }
 
 /**
- * Server-side compaction trigger, in input tokens.
- *
- * Unlike MI, BI re-sends the entire project source in every turn's user message, so its
- * per-turn baseline is much larger — MI's 200K trigger would fire too eagerly. 500K leaves
- * ample working headroom while staying well within the 1M context window (Claude Sonnet).
+ * Server-side compaction trigger, in input tokens. Higher than MI's 200K because BI re-sends
+ * the whole project source each turn; 500K sits well within Claude Sonnet's 1M window.
  */
 const COMPACT_TRIGGER_TOKENS = 500_000;
 
 /**
- * Builds providerOptions.anthropic.contextManagement, mirroring MI's approach: compaction
- * only, no `clear_tool_uses`. `clear_tool_uses` *deletes* older tool results with no summary
- * to replace them, degrading the agent mid-task; `compact_20260112` *summarizes* instead,
- * preserving the context the agent needs to keep working.
+ * Builds providerOptions.anthropic.contextManagement: compaction only, no `clear_tool_uses`
+ * (which deletes tool results without summarizing). Mirrors MI.
  */
 function buildCompactionProviderOptions(loginMethod: LoginMethod, floorTokens: number) {
     if (!supportsCompaction(loginMethod)) { return undefined; }
@@ -483,15 +478,16 @@ export class AgentExecutor extends AICommandExecutor<GenerateAgentCodeRequest> {
             let compactionContent = '';
             let cleanedCompactionSummary: string | undefined;
             // Counts compactions in this turn so each renders as its own card (upsertComponent
-            // keys by id). Mirrors MI: a notice + the extracted summary, rendered in-stream —
-            // never emitted as a raw <compaction> text block (which would render as literal text).
+            // keys by id), instead of a raw <compaction> text block that would show as literal text.
             let compactionCount = 0;
-            const emitCompactionNotice = (summary?: string) => {
+            const emitCompactionNotice = () => {
+                // Notice only — the model-authored summary is kept internal (#2371), never
+                // forwarded to the webview or persisted in the transcript.
                 this.config.eventHandler({
                     type: 'chat_component',
                     componentType: 'compaction',
                     id: `compaction-${this.config.generationId}-${compactionCount++}`,
-                    data: { summary: summary ?? '' },
+                    data: {},
                 });
             };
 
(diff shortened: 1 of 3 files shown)


packages/ballerina-visualizer/src/views/AIPanel/components/AgentStreamView/StreamEntry.tsx
```diff
@@ -147,6 +193,27 @@ function renderItem(item: StreamItem, idx: number, streamActive: boolean, rpcCli
...
+                return (
+                    <React.Fragment key={idx}>
+                        <ItemRow>
+                            <span className="codicon codicon-fold" aria-hidden="true" />
+                            <ItemLabel loading={false}>Context compacted — conversation continues below</ItemLabel>
+                        </ItemRow>
+                        {summary && (
+                            <ItemDetail>
+                                <details>
+                                    <summary style={{ cursor: "pointer" }}>View summary</summary>
```
RNViththagan: This is the spot where the actual model-authored summary becomes visible to the user (click "View summary"). Per the note in the review body - matches MI's UI, but conflicts with what #2371 explicitly asks for. Wanted to anchor the discussion here since it's the exact line.

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
```diff
@@ -90,10 +90,39 @@ function supportsCompaction(loginMethod: LoginMethod): boolean {
         || loginMethod === LoginMethod.ANTHROPIC_AWS;
 }
 
+/**
+ * Server-side compaction trigger, in input tokens.
+ *
+ * Unlike MI, BI re-sends the entire project source in every turn's user message, so its
+ * per-turn baseline is much larger — MI's 200K trigger would fire too eagerly. 500K leaves
+ * ample working headroom while staying well within the 1M context window (Claude Sonnet).
```
RNViththagan: Can we confirm this against Anthropic's actual API behavior? I couldn't find a `context-1m` (or similarly named) beta header anywhere in the extension, and Claude models generally need an explicit opt-in beta to get the 1M window - without it the default is smaller. If claude-sonnet-5's actual window here is the default rather than 1M, a 500K trigger could be unreachable: BI's per-turn floor plus growth would hit `model_context_window_exceeded` well before compaction ever fires, and with `clear_tool_uses@200K` now removed there'd be no context management at all under that window. Also, the PR description says this mirrors MI's 500K, but this same comment says MI's trigger is 200K and BI deliberately differs - worth fixing whichever one is wrong.

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
```diff
@@ -559,11 +600,8 @@ export class AgentExecutor extends AICommandExecutor<GenerateAgentCodeRequest> {
                                     type: 'usage_metrics',
                                     usage: { inputTokens: 0, cacheCreationInputTokens: 0, cacheReadInputTokens: 0, outputTokens: 0 },
                                 });
-                                // Inline notice before the continuing response
-                                this.config.eventHandler({
-                                    type: 'content_block',
-                                    content: '<compaction>Context compacted — key context preserved, conversation continues below.</compaction>',
-                                });
+                                // Compaction notice + summary card, positioned before the continuing response.
```
RNViththagan: This only emits once the *next* `text-start` arrives (or at stream end) - there's no `text-end` handling, so the block closes on whatever comes next rather than on its own end. Two consequences worth knowing about: if the model does tool calls right after compacting (the common case), the notice ends up rendered after those tool-call rows, not "before the continuing response" as this comment says. And if two compactions ever fire back to back with only tool calls between them (no text in between), the second compaction's text-start resets `compactionContent` without ever flushing the first one, so the first summary is silently dropped along with its notice. Probably rare enough not to block on, but wanted to flag it since it's exactly the mechanism this PR is changing.

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
```diff
@@ -90,10 +90,39 @@ function supportsCompaction(loginMethod: LoginMethod): boolean {
...
 }
 
+/**
+ * Server-side compaction trigger, in input tokens.
+ *
+ * Unlike MI, BI re-sends the entire project source in every turn's user message, so its
+ * per-turn baseline is much larger — MI's 200K trigger would fire too eagerly. 500K leaves
+ * ample working headroom while staying well within the 1M context window (Claude Sonnet).
+ */
+const COMPACT_TRIGGER_TOKENS = 500_000;
```
RNViththagan, line 112: Small one: a few of the new comments here and in StreamEntry.tsx run 5-7 lines where the why fits in one or two - this one and the one below it, plus the JSDoc at the top of StreamEntry.tsx. Not blocking, just flagging since the convention here is to keep it short.

packages/ballerina-visualizer/src/views/AIPanel/components/AgentStreamView/StreamEntry.tsx
IsuruMaduranga: You're right — the summary was reachable via the expandable "View summary", so I've made it internal per the issue. `emitCompactionNotice()` no longer forwards the summary (the compaction `chat_component` now carries `data: {}`), and the expandable render at L197-212 is removed, so the card shows only the fixed "Context compacted — conversation continues below" notice. The summary is still computed server-side (for `stripAnalysisFromCompactionBlocks`) but never reaches the webview or the persisted transcript.

One correction on the third suggested spot (the legacy `<compaction>` text-tag path, StreamEntry.tsx L89-94): I left that rendering the tag's inner text. Checking the pre-PR git history, that emission only ever put a fixed developer notice inside the tag (`<compaction>Context compacted — key context preserved…</compaction>`), never the model summary — the summary only ever leaked through `data.summary`. That same path also renders the `compaction_disabled` warning, which should stay visible, so replacing it with a fixed string would have dropped that warning's text.

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
IsuruMaduranga: Confirmed against Anthropic's model reference: `claude-sonnet-5` has a 1M-token context window by default — no `context-1m` (or similar) beta header is required. That header belonged to the older Sonnet 4; Sonnet 5 ships 1M natively. So the 500K trigger is reachable and there's no risk of `model_context_window_exceeded` before compaction fires.

On the MI-vs-BI discrepancy: the code comment is the accurate one — BI deliberately uses 500K, higher than MI's 200K, because BI re-sends the whole project source in every turn's user message. The PR description's "mirrors MI's 500K" wording is wrong (MI triggers at 200K); I'm fixing the description.

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
IsuruMaduranga: Done in this PR rather than deferring — tightened the two JSDocs here (`COMPACT_TRIGGER_TOKENS` and `buildCompactionProviderOptions`) to one or two lines each, along with the `renderTextItem` JSDoc in StreamEntry.tsx, and kept the new `text-end` comments short.

packages/ballerina-extension/src/features/ai/agent/AgentExecutor.ts
IsuruMaduranga: Fixed — you're right on both counts. The compaction block now flushes on its own `text-end` (via a `flushCompactionBlock()` helper) instead of deferring to the next `text-start`. So the notice is emitted the moment the summary block ends — before any following tool-call rows — which makes "before the continuing response" actually true. And because each block self-closes on its `text-end`, a second back-to-back compaction can no longer reset `compactionContent` on an unflushed first block, so the first summary can't be silently dropped. The stream-end fallback is kept for a block still open when the stream ends.

RNViththagan (approved):
LGTM

SasinduDilshara (approved):

PR #1050, opened 2026-09-07
Allow clearing value for optional fields in AutoCompleteEditor

thuva9872:
## Purpose
Fixes: https://github.com/wso2/product-integrator/issues/2270

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary

- Fixed `AutoCompleteEditor` so users can clear optional autocomplete fields.
- Preserved existing values for required fields when users clear the input.
- Added tests and a fixture for the Agent tool’s optional “Approval Function” field.

### DIFF
--- a/packages/ballerina-side-panel/src/components/editors/AutoCompleteEditor.tsx
+++ b/packages/ballerina-side-panel/src/components/editors/AutoCompleteEditor.tsx
@@ -71,7 +71,7 @@ export function AutoCompleteEditor(props: AutoCompleteEditorProps) {
                 errorMsg={errorMsg || undefined}
                 {...register(field.key, {
                     required: buildRequiredRule({ isRequired: !field.optional, label: field.label }),
-                    value: getValueForDropdown(field),
+                    value: field.optional ? (field.value ?? "") : getValueForDropdown(field),
                     validate: buildValidate(field)
                 })}
                 label={capitalize(field.label)}
@@ -89,7 +89,7 @@ export function AutoCompleteEditor(props: AutoCompleteEditorProps) {
                         : field.optional
                             ? ""
                             : currentValue ?? "";
-                    setValue(field.key, newVal);
+                    setValue(field.key, newVal, { shouldDirty: true });
                     field.onValueChange?.(newVal);
                     liveDiagnostics.onValueChange(newVal);
                 }}
--- a/packages/ballerina-side-panel/src/components/editors/AutoCompleteEditor.tsx
+++ b/packages/ballerina-side-panel/src/components/editors/AutoCompleteEditor.tsx
@@ -41,6 +41,8 @@ export function AutoCompleteEditor(props: AutoCompleteEditorProps) {
 
     const value = watch(field.key);
 
+    const comboboxValue = (field.optional ? ((value as string) || null) : value) as string;
+
     // Live diagnostics: client rules (e.g. the identifier check on a free-typed value) run on
     // every change, same as TextEditor — react-hook-form's default `onSubmit` mode otherwise
     // leaves `errors` empty until submit is attempted, so a `validations[]` failure would go
@@ -65,7 +67,7 @@ export function AutoCompleteEditor(props: AutoCompleteEditorProps) {
             <AutoComplete
                 id={field.key}
                 description={field.documentation}
-                value={value as string}
+                value={comboboxValue}
                 errorMsg={errorMsg || undefined}
                 {...register(field.key, {
                     required: buildRequiredRule({ isRequired: !field.optional, label: field.label }),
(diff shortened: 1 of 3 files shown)


packages/ballerina-side-panel/src/components/editors/AutoCompleteEditor.tsx
```diff
@@ -76,13 +78,17 @@ export function AutoCompleteEditor(props: AutoCompleteEditorProps) {
...
+                    const isEmpty = val === "" || val === undefined || val === null;
                     const currentValue = value ?? getValueForDropdown(field) ?? field.value;
-                    const newVal = (val === "" || val === undefined || val === null) && currentValue
-                        ? currentValue
-                        : val;
+                    const newVal = !isEmpty
+                        ? val
+                        : field.optional
+                            ? ""
+                            : currentValue ?? "";
```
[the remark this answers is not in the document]
thuva9872: - Not a regression / same as before. The previous code used the same value ?? getValueForDropdown(field) ?? field.value derivation, and its guard (isEmpty && currentValue) ? currentValue : val also fell through to val (i.e. "") whenever currentValue was "". So the empty-required-field behavior is unchanged by this PR.
- The proposed || changes behavior in an undesirable way. For an empty field, getValueForDropdown(field) resolves to items[0], so a required field would get its first option silently injected just from a focus→blur, with no user selection. That masks the required-empty state that validation is meant to surface. Leaving it as "" is the intended behavior.
- The precondition is largely unreachable anyway. The field is registered with value: getValueForDropdown(field), so a required field with no explicit value is already defaulted to items[0]; watch(field.key) is normally items[0], not "".

packages/ballerina-side-panel/src/components/editors/AutoCompleteEditor.tsx
madushajg: `getValueForDropdown` falls back to `field.items[0]` when `field.value` is empty, regardless of optional. 

eg: an optional `AUTOCOMPLETE` with value: `""` gets seeded to `items[0]` on mount and again on remount (toggling a parent switch off/on, reopening the panel). A field the user deliberately cleared and saved comes back pre-filled with the first candidate, the same bug class this PR fixes.

Suggestion:
```ballerina
    value: field.optional ? (field.value ?? "") : getValueForDropdown(field),
```

packages/ballerina-side-panel/src/components/editors/AutoCompleteEditor.tsx
```diff
@@ -76,13 +78,17 @@ export function AutoCompleteEditor(props: AutoCompleteEditorProps) {
...
-                        : val;
+                    // A nullable Combobox reports the cleared state as null/undefined on blur.
+                    const isEmpty = val === "" || val === undefined || val === null;
+                    const currentValue = ((value as string) || undefined) ?? getValueForDropdown(field) ?? field.value;
+                    const newVal = !isEmpty
+                        ? val
+                        : field.optional
+                            ? ""
+                            : currentValue ?? "";
                     setValue(field.key, newVal);
```
madushajg: The clear is written with a bare `setValue`, so the field never becomes dirty and `AutoComplete` doesn't consume `register()`'s `onChange` either (`Combobox`/`Combobox.Input` use their own handlers, so the spread-in `onChange` is dropped), so nothing else dirties it.
That makes the clear vulnerable to `Form/index.tsx`'s reset effect: L683 refuses to preserve `""`, `L690`/`L698` then re-seed the stale `field.value`, and `L766`'s `reset(defaultValues, { keepDirtyValues: true })` can't protect a non-dirty field. Replaying that exact sequence against RHF 7.56.4: with the current code the cleared value comes back as `"approveFn"`; with `shouldDirty: true` it stays `""`. So any parent re-render that hands Form a new `formFields` array (LS round-trip, diagnostics refresh, editing a sibling field) reintroduces the #2270 snap-back.

Suggestion:

```ballerina
setValue(field.key, newVal, { shouldDirty: true });
```

This also makes `expression.modified` (`node-property-utils.ts:404`) report the clear as a real change, relevant for LS-backed `AUTOCOMPLETE` NodeProperty fields like `mediaType` in `ResponseEditor.tsx`.

thuva9872: fixed in  [1610592](https://github.com/ballerina-platform/ballerina-vscode/pull/1050/commits/1610592861898ea1487d605c3ad2992c6ec22a9f)

packages/ballerina-side-panel/src/components/editors/AutoCompleteEditor.tsx
thuva9872: fixed in [1610592](https://github.com/ballerina-platform/ballerina-vscode/pull/1050/commits/1610592861898ea1487d605c3ad2992c6ec22a9f)

madushajg (approved):

packages/ballerina-side-panel/src/test/fixtures/fields/issue-2270.json
```diff
@@ -0,0 +1,17 @@
+{
+    "description": "#2270: optional AUTOCOMPLETE (Approval Function ref) still selects AutoCompleteEditor",
```
NipunaRanasinghe, line 2: **Name the fixture for what it exercises, not the issue it came from**

`loadFixtures` derives the test title from the filename (`packages/test-config/fixtures.js`), so this case currently reports as `issue-2270 -> selects the expected editor`, which doesn't say what is being guarded. Suggest renaming to `autocomplete-optional.json` and dropping the `#2270:` prefix from `description`, leaving just `optional AUTOCOMPLETE (Approval Function ref) still selects AutoCompleteEditor` — the linked issue in the PR body already carries the traceability. Every other fixture in this directory (`readonly`, `slider`, `enum`, `repeatable-list`, ...) is named this way. The existing `issue-1491.json` and the example in the `EditorFactory.test.tsx` header comment have the same problem; I'll clean those up separately so the convention doesn't stay split.

PR #1052, opened 2026-09-08
Generate trigger creation forms from connector-owned metadata

LakshanWeerasinghe:
## Overview
This PR migrates trigger creation metadata from the Integrator/VS Code repositories into the connector package itself.

Connector packages can now ship their trigger metadata as JSON inside the .bala, allowing the Integrator to dynamically discover connectors and generate the trigger creation form without requiring a corresponding language-server service model or VS Code-specific implementation for every connector.

This addresses the onboarding and release overheads.

## High-level approach

  The implementation introduces a layered metadata model:

  1. Connector-owned metadata

     Connectors can provide:
      - metadata/trigger-metadata.json — structural and semantic trigger information.
      - metadata/trigger-ui-metadata.json — sparse UI-specific overrides such as labels, descriptions, form fields, listener mappings, and artifact presentation details.

  2. Language-server metadata resolution

     The language server resolves these files directly from the connector’s .bala or local Ballerina repository. Metadata is validated by version, parsed into shared models, cached, and handled gracefully when
     unavailable or invalid.

  3. Dynamic model synthesis

     The language server combines:
      - Connector-owned L1 trigger metadata.
      - Sparse L2 UI metadata.
      - Semantic information extracted from the connector package.
      - Existing service/listener model information.

     These inputs are compiled into the unified trigger UI schema consumed by the service designer and trigger creation flows.

  4. VS Code integration

     The VS Code extension consumes the generated model and uses connector-provided artifact metadata and icons when rendering trigger entries in the Integrator artifact tree and related views.

  5. Backward compatibility

     Existing shipped schemas and synthesized models remain supported. The generated metadata path is introduced behind controlled resolution and onboarding logic so existing connectors and specialized service
     builders continue to behave unchanged.

  ## Key changes

  - Added connector-owned trigger metadata models and JSON parsing.
  - Added support for trigger listeners, service types, annotations, rules, imports, and platform dependencies.
  - Added sparse UI metadata overlays for connector-specific form customization.
  - Added semantic introspection of connector packages to derive missing model details.
  - Added trigger model synthesis and UI schema compilation in the language server.
  - Added connector-owned artifact metadata and theme-aware icon handling.
  - Added local repository and .bala metadata resolution with caching.
  - Preserved compatibility with legacy trigger schemas and existing service-model implementations.
  - Added validation and unit coverage for metadata reading, synthesis, schema compilation, listener handling, and icon/artifact metadata.

  ## Result

  Adding a new metadata-driven trigger now primarily requires:

  1. Implementing and publishing the Ballerina connector.
  2. Adding the trigger metadata JSON files to the connector package.
  3. Providing connector-owned icon and artifact metadata where required.

  The language server and VS Code extension can then discover the connector and dynamically generate the appropriate trigger creation experience.


packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerModelReader.java
```diff
@@ -70,98 +65,56 @@ public class TriggerModelReader {
...
-                    if (!variants.isEmpty()) {
-                        registry.put(moduleName, variants);
-                    }
-                });
-                return Map.copyOf(registry);
-            }
-        } catch (IOException | JsonParseException e) {
-            return Map.of();
-        }
+    private static boolean generationEnabled(String moduleName) {
```
dulajdilshan: **[Major] The generated tier is dead in production.** `generationEnabled` requires `-Dballerina.trigger.models=generated`, and nothing sets it — not `build.gradle`, not either `testng.xml`, not the LS launch in `server.ts:242`/`:407`. So `TriggerUIMetadataCompiler` (1546 lines), `TriggerUIAuthoringParser`, `TriggerUIMetadataModel` and the `TriggerModelSynthesizer` rework never run in a shipped VSIX, while the bundled metadata they replace is already deleted. All 14 test classes call package-private `getGeneratedTriggerModel` directly, so the flag is never exercised either.

packages/ballerina-extension/src/stateMachine.ts
```diff
@@ -394,13 +401,26 @@ const stateMachine = createMachine<MachineContext>(
...
                             onDone: {
-                                target: "webViewLoading",
+                                target: "checkConnectorUpgrades",
                                 actions: assign({
                                     dependenciesResolved: true
                                 })
                             }
                         }
                     },
+                    checkConnectorUpgrades: {
```
dulajdilshan, line 412: **[Major] Webview load blocks on a user-dismissed notification.** This state sits on the path to `webViewLoading`, and its service awaits `checkAndPromptConnectorUpgrades`, which awaits `showInformationMessage(..., 'Update All', 'Not Now')` (`connector-upgrade.ts:62`) then one `showWarningMessage` per breaking item (`:73-75`). Notifications with buttons don't auto-dismiss. Open a project on a pre-schema-driven connector and the view stays loading until the user clicks. Healthy projects still pay an LS round trip before first render. Move the prompt off the render path.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/adapter/TriggerServiceAdapter.java
```diff
@@ -97,6 +98,14 @@ public static Service toServiceTemplate(TriggerUISchemaModel model, String servi
         return service;
     }
 
+    private static String effectiveTriggerKind(TriggerUISchemaModel model) {
```
dulajdilshan, line 102: **[Medium] Duplicate implementation — this rule exists 6x, its allow-list 4x.** Character-identical to `ServiceModelGeneratorService.java:1450`. Same six-element allow-list again at `TriggerUIMetadataCompiler.java:55` and `LibraryMetadataReader.java:78` — across two Gradle modules. Coalesce repeated at `TriggerProperty.java:56`, `TriggerUIMetadataModel.java:88`, `LibraryMetadataReader.readTriggerKind():309` and `triggerKind.ts:22`. A 7th kind means editing four allow-lists; miss one and `EventIntegrationPanel.tsx:47` silently drops the card. One enum in `model-generator-commons` should survive.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerUIMetadataCompiler.java
```diff
@@ -0,0 +1,1546 @@
...
+    }
+
+    static TriggerUISchemaModel apply(TriggerUISchemaModel derived, TriggerMetadataModel l1,
+                                      TriggerLibraryFacts facts,
+                                      Map<String, TriggerLibraryFacts> crossModuleFacts,
+                                      SemanticModel semanticModel, TriggerUIMetadataModel l2) {
+        if (derived == null || l2 == null) {
+            return derived;
+        }
+        JsonObject root = GSON.toJsonTree(derived).getAsJsonObject();
```
dulajdilshan, line 75: **[Medium] Wrong representation — diverges from the file beside it.** `apply()` takes a typed model, serializes to `JsonObject`, runs ~80 string-keyed tree helpers, deserializes back at `:81` (with more round trips at `:104`/`:110`). Its neighbour `TriggerModelSynthesizer` builds the same output type over 1232 lines with typed records. Same package, same stage, opposite idiom — renames in `TriggerUISchemaModel` become silent no-ops here instead of compile errors.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerUIMetadataCompiler.java
```diff
@@ -0,0 +1,1546 @@
...
+            if (l2.trigger().listenerKind() != null) {
+                if (l2.trigger().listenerKind().isEmpty()) {
+                    root.remove("listenerKind");
+                } else {
+                    root.addProperty("listenerKind", l2.trigger().listenerKind());
+                }
+            }
+        }
+        if (l2.metadata() != null) {
+            String triggerKind = l2.metadata().effectiveTriggerKind();
```
dulajdilshan, line 203: **[Medium] Property keys hardcoded where `Constants` is their home.** The file imports five names from `Constants`, then inlines the rest: `"properties"` x27, `"initProperties"` x24, `"codedata"` x24, `"listeners"` x16, `"metadata"` x14, plus `"serviceProperties"`, `"listenerForm"`, `"schemaFunctions"`, `"triggerKind"`. Combined with the JSON-tree approach, a typo compiles and silently does nothing.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReader.java
```diff
@@ -218,6 +218,251 @@ Optional<TriggerMetadataModel> readTriggerMetadataModel(Path packageRoot) {
...
+        }
+        reader.endObject();
+        if (model == null) {
+            return new ArtifactDocument(minVersion, null, null, null, null);
+        }
+        return new ArtifactDocument(minVersion, model.version(), model.triggerKind(), model.artifactInfo(),
+                model.variants());
+    }
+
+    private ArtifactDocument selectArtifactVariant(List<ArtifactDocument> variants, String version) {
```
dulajdilshan: **[Medium] Two variant selectors for one file format.** This and `selectUIMetadataVariant:431` implement the same rule — newest-to-oldest, first entry whose `minVersion` is absent or `<= version` — over two representations of the same `metadata/trigger-ui-metadata.json`. Only `:411-429` documents the envelope, and the fallbacks already differ (`:352` returns the last variant regardless of validity, `:454` returns `root`). One should project from the other.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReader.java
```diff
@@ -218,6 +218,251 @@ Optional<TriggerMetadataModel> readTriggerMetadataModel(Path packageRoot) {
...
+            fallback = variant;
+            if (variant.minVersion() == null || variant.minVersion().isBlank()
+                    || version == null || version.isBlank() || versionAtLeast(version, variant.minVersion())) {
+                return variant;
+            }
+        }
+        return fallback;
+    }
+
+    private Optional<String> readRelativeAsset(Path root, String relative) {
```
dulajdilshan: **[Medium] Three copies of the same path-escape guard.** `resolve -> normalize -> startsWith(root) -> isRegularFile -> readString` is written here, again at `readResourceFile:504`, and inline a third time at `:232`. This is the boundary stopping connector-authored relative paths escaping the `.bala` root, so three drifting copies is the wrong number — this one has an `isSafeRelativePath` pre-check the other two lack.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
```diff
@@ -510,6 +510,48 @@ private static Optional<Path> resolveOfflineModuleBala(String org, String name)
...
+     * authored for one pinned version, or a connector resolved at the version its source declares --
+     * has to say which one it means.
+     *
+     * @param buildProject the project whose environment resolves the package
+     * @param org          the package's organization
+     * @param name         the package name
+     * @param version      the exact version to resolve; {@code null} or blank resolves the newest local
+     * @return the resolved package, or {@code Optional.empty()} when it isn't available offline
+     */
+    public static Optional<Package> getModulePackageOffline(BuildProject buildProject, String org, String name,
```
dulajdilshan: **[Medium] Duplicate implementation, and it drops the memoization it copied.** Near-verbatim copy of `resolveOfflineModuleBala:499-518` twenty lines up — same descriptor build, same offline call, same `UNRESOLVED` guard, same tail, same copied comment. But that one is reached through `memoizedSampleBala:495` and this caches nothing, while both production call sites pass `getSampleProject()` (`TriggerModelReader.java:371`, `:391`) — the exact environment the memoized path already serves. No caller varies the `BuildProject`, and the 3-arg overload at `:521` has zero callers.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReader.java
```diff
@@ -89,14 +93,26 @@ public static LibraryMetadataReader getInstance() {
...
+    }
+
+    /** Reads the artifact-tree projection without materializing or caching the complete L2 document. */
+    public Optional<ArtifactMetadata> getArtifactMetadata(ModuleInfo moduleInfo) {
+        return packageRoot(moduleInfo)
+                .flatMap(root -> readArtifactMetadata(root, moduleInfo == null ? null : moduleInfo.version()));
+    }
+
+    /** Compatibility accessor for callers interested only in presentation metadata. */
+    public Optional<ArtifactInfo.Resolved> getArtifactInfo(ModuleInfo moduleInfo) {
```
dulajdilshan, line 126: **[Medium] Four near-identical accessor pairs, one dead.** `getTriggerMetadataModel`/`getTriggerUIMetadataModel`/`getArtifactMetadata`/`getArtifactInfo` each have a `…FromLocalRepository` twin differing only in `packageRoot(…)` vs `localPackageRoot(…)`. `getArtifactInfoFromLocalRepository:144` has zero callers anywhere including tests; `readArtifactInfo:227` is called only from `LibraryMetadataReaderTest.java:281`. Both are labelled "Compatibility accessor" on a surface with no prior callers. One parameterised read collapses eight methods to four.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/model/response/ConnectorUpgradeAdviceResponse.java
```diff
@@ -0,0 +1,36 @@
...
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package io.ballerina.servicemodelgenerator.extension.model.response;
+
+import io.ballerina.servicemodelgenerator.extension.connector.ConnectorUpgradeAdvisor.ConnectorUpgradeAdvice;
```
dulajdilshan: **[Medium] Layering — `model` now depends on `connector`.** This inverts the direction every other file in `model/response/` follows. The wire record is declared nested inside the analyser (`ConnectorUpgradeAdvisor.java:73`) rather than beside the response it serializes in, though this PR does add two other types to `model/response/`. Move `ConnectorUpgradeAdvice` here and let the advisor depend on it.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/models/TriggerUIMetadataModel.java
```diff
@@ -0,0 +1,546 @@
...
+        Metadata metadata,
+        Trigger trigger,
+        List<ReadOnlyMetadata> readOnlyMetadata,
+        InitForm initForm,
+        List<TargetedNode> listeners,
+        List<TargetedNode> serviceTypes,
+        String importPrefix,
+        ArtifactInfo artifactInfo) {
+
+    /** Source-compatible constructor for callers that do not author artifact metadata. */
```
dulajdilshan: **[Medium] Speculative generality — "compatibility" constructors on a brand-new file.** Six overloads documented as "Source-compatible…", "…before `triggerKind` was introduced" (`:79`), "…predating `minItems`/`defaultItems`" (`:395`), plus `Codedata:299`, `Field:354`, `ListenerForm:482`. Nothing predates a new file. Production deserializes via Gson and never calls a constructor — the only callers are `TriggerUIMetadataCompilerTest`. A test factory belongs in the test file.

packages/ballerina-side-panel/src/components/Form/validationRules.ts
```diff
@@ -335,6 +308,33 @@ export const CLIENT_VALIDATION_RULES: Record<string, ClientRule> = {
     // `vscode.validate.file.exists` and `vscode.validate.unique.in.form` need host RPC / sibling
     // scope plumbing and are intentionally not registered yet; they degrade to "unknown rule".
 
+    "vscode.validate.regex": (value, args) => {
```
dulajdilshan, line 311: **[Medium] Published rule id renamed with no alias.** `common.validate.regex` becomes `vscode.validate.regex` and its server-side implementation is deleted (`CommonRuleValidators.java`). Rule ids now arrive from connector-authored `metadata/trigger-metadata.json`, so any connector already shipping the old id loses validation on both sides — `ValidationEngine.java:43` logs and skips, the client falls through to "unknown rule". Keep the old id as an alias for a release, or state the move in the spec.

packages/ballerina-visualizer/src/components/ArtifactIcon/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+    // L2 icons are expected to be monochrome. Older bundled SVGs use explicit
+    // black/white fills instead of `currentColor`, so normalize those values
+    // before applying the metadata color as well.
+    const value = color ? svg.replace(/((?:fill|stroke)\s*=\s*["'])(currentColor|black|white|#000(?:000)?|#fff(?:fff)?)(["'])/gi,
+        `$1${color}$3`) : svg;
+    return `data:image/svg+xml;charset=utf-8,${encodeURIComponent(value)}`;
+}
+
+/** Renders L2-owned SVG text, then the Central URL, then the generic kind icon. */
+export function ArtifactIcon({ icon, kind, size = 38 }: ArtifactIconProps) {
```
dulajdilshan: **[Medium] Dead component, and a third copy of the theme probe.** 69 lines for rendering the new L2 light/dark SVG pair, with no importer anywhere. Meanwhile the panel that should use it discards the SVGs: `EventIntegrationPanel.tsx:112` reads `toIconDescriptor(item.icon)?.url ?? ""`, so connectors shipping `trigger-ui-metadata` icons still render the Central URL. `isLightTheme():29` also re-implements the body-class check at `MarkdownRenderer.tsx:154` and, differently, at `NodeIcon/index.tsx:156,232`.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/ConnectorUpgradeAdvisor.java
```diff
@@ -0,0 +1,217 @@
...
+                }
+                String currentVersion = ConnectorVersionResolver.resolve(project, orgName, moduleName, null);
+                if (currentVersion == null) {
+                    return;
+                }
+                if (alreadySupported(metadataReader, orgName, moduleName, currentVersion, minSupportedVersion)) {
+                    return;
+                }
+                boolean breaking = crossesMinorBoundary(currentVersion, minSupportedVersion);
+                boolean explicitlyPinned = isExplicitlyPinned(project, orgName, moduleName);
```
dulajdilshan: **[Medium] `explicitlyPinned` is threaded through three packages and read by nothing.** Declared on the wire record `:75`, published non-optional at `extended-lang-client.ts:1466`, hardcoded `true` at `rpc-manager.ts:498`. No consumer — and its own javadoc (a bump must edit `Ballerina.toml`) describes behaviour `pullAndBumpConnectors` explicitly does not implement (`connector-upgrade.ts:91`). Either the toml edit lands or the field goes; a false value on a public wire type is worse than neither.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerModelReader.java
```diff
@@ -284,22 +209,58 @@ public Optional<TriggerUISchemaModel> getSchemaDrivenTriggerModel(String orgName
...
+        Resolution resolution = resolveSchemaDrivenTriggerModel(orgName, moduleName, version);
         if (resolution.cacheable()) {
             schemaDrivenTriggerCache.put(key, resolution.model());
         }
         return resolution.model();
     }
 
+    /**
+     * The L1 + semantic facts + L2 generated model, cached by {@code org/module:version}. This is the
+     * default resolution tier as of the L1+L2 cutover -- {@link #getGeneratedTriggerModel} itself stays
```
dulajdilshan: **[Minor] Comment contradicts the code and the comment 160 lines above.** This says the generated tier "is the default resolution tier as of the L1+L2 cutover"; `:78-80` says generation is opt-in and the default keeps the established path. `generationEnabled():83` makes the second one right.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/models/ArtifactInfo.java
```diff
@@ -0,0 +1,85 @@
...
+                return displayLabel;
+            }
+            String unqualified = serviceType.contains(":")
+                    ? serviceType.substring(serviceType.lastIndexOf(':') + 1)
+                    : serviceType;
+            return displayLabelOverrides.getOrDefault(unqualified, displayLabel);
+        }
+    }
+
+    /** Raw, trusted SVG text returned on the LS wire, plus the optional brand tint. */
```
dulajdilshan: **[Minor] Two consecutive javadoc blocks on `ResolvedIcon`.** Only the second attaches to the record, so the sentence explaining the trust assumption is dropped from the generated docs. Merge them.

packages/ballerina-extension/src/webview-communication/DefaultServer.ts
```diff
@@ -293,6 +293,13 @@ export class DefaultServer {
             await waitForLangClientReady();
             return serviceDesigner.getServiceInitModel(p);
         });
+        this.register("getConnectorUpgradeAdvice", async (p) => {
+            await waitForLangClientReady();
+            return serviceDesigner.getConnectorUpgradeAdvice(p);
+        });
+        this.register("pullConnectorUpgrade", async (p) => {
```
dulajdilshan, line 300: **[Minor] Diverges from the registration above it.** `getConnectorUpgradeAdvice:296` awaits `waitForLangClientReady()`; this doesn't, though it reaches the language client via `executeCommand`. The handler is also `async` with no `await`.

```suggestion
        this.register("pullConnectorUpgrade", async (p) => {
            await waitForLangClientReady();
            return serviceDesigner.pullConnectorUpgrade(p);
        });
```

packages/ballerina-extension/src/stateMachine.ts
```diff
@@ -67,6 +68,7 @@ interface MachineContext extends VisualizerLocation {
     isBISupported: boolean;
     errorCode: string | null;
     dependenciesResolved?: boolean;
+    connectorUpgradesCheckedPaths?: string[];
```
dulajdilshan: **[Minor] Machine-context array that only ever grows.** Appended at `:417-419`, never trimmed or reset. Bounded by distinct projects per session so it stays small, but a `Set` on long-lived machine context would make the choice deliberate.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReader.java
```diff
@@ -57,28 +70,19 @@ public final class LibraryMetadataReader {
 
     private static final Logger LOGGER = Logger.getLogger(LibraryMetadataReader.class.getName());
 
-    private static final String TRIGGER_METADATA_RESOURCE_PATH = "resources/trigger-metadata.json";
-    private static final String TRIGGER_UI_SCHEMA_RESOURCE_PATH = "resources/trigger-ui-schema.json";
-    private static final String PACKAGED_TRIGGER_METADATA_ROOT = "trigger-metadata-models";
-    private static final String PACKAGED_TRIGGER_METADATA_FILE = "trigger-metadata.json";
+    private static final String TRIGGER_METADATA_RESOURCE_PATH = "metadata/trigger-metadata.json";
+    private static final String TRIGGER_UI_METADATA_RESOURCE_PATH = "metadata/trigger-ui-metadata.json";
```
dulajdilshan, line 70: **[Minor] Cache sized for one caller, used by another.** `MAX_CACHE_SIZE = 2` is "sized for the designer, which resolves one connector at a time", but `getArtifactMetadata` is called per module from `Artifact.java:296` and `TriggerSearchUtil.java:126` iterates every local-repository module — both thrash it. It also sits on top of `PackageUtil.memoizedSampleBala` for the same lookup: two caches for one resolution.

packages/ballerina-extension/src/features/project/connector-upgrade.ts
```diff
@@ -0,0 +1,121 @@
...
+    const fileUri = Uri.file(path.isAbsolute(targetFile) ? targetFile : path.join(projectPath, targetFile))
+        .toString();
+
+    try {
+        await StateMachine.langClient().executeCommand({
+            command: PULL_MODULE_COMMAND,
+            arguments: [{ key: ARG_KEY_DOC_URI, value: fileUri }]
+        });
+        notifyCurrentWebview();
+        return { succeeded: advice, failed: [] };
```
dulajdilshan: **[Minor] Reports every item as succeeded on one project-wide pull.** `PULL_MODULE` re-resolves the whole project, so a partial outcome returns as `{succeeded: advice, failed: []}`. `ServiceCreationView.tsx:267` treats `result.success` as proof and re-fetches, showing the unsupported-version card again with no explanation. Derive success per item, or narrow the return to a boolean.

packages/ballerina-core/src/icons/brand-icon-registry.ts
```diff
@@ -49,6 +44,7 @@ export const BRAND_ICON_REGISTRY: Record<string, BrandIcon> = {
     http: { glyph: "bi-globe" },
     mcp: { glyph: "bi-mcp" },
     solace: { glyph: "bi-solace", color: "#00C895" },
+    "solace.jms": { glyph: "bi-solace", color: "#00C895" },
```
dulajdilshan: **[Nitpick]** `"solace.jms"` is deleted from its old position and re-added two lines under `solace` with identical content — no behaviour change, two extra hunks in a 212-file PR.

dulajdilshan (commented):
**GENERAL [Medium] — Regression tests deleted while the feature they guard survives.** `DriverDependencyFieldTest.java` (184 lines) and two `ServiceModelAPITests` cases explicitly labelled *"Regression for … (product-integrator#2020)"* are removed, but `PlatformDependencyEditUtil.populateDriverDependencyFields:147` / `findDriverDependencyFields:65` are still live (`SchemaDrivenServiceBuilder.java:97`) and `TriggerUIMetadataCompiler.java:658` still emits `driverDependency` codedata. The surviving `PlatformDependencyEditUtilTest` covers only the `isDeclared`/`findDeclaredPath`/`addIfMissing` primitives, never the field-hiding or the `[[platform.java21.dependency]]` edit wiring. The coverage was not re-homed: `src/test/resources/platform_dependency/both_declared` is now referenced by no test in the tree. CONTRIBUTING's regression policy makes this a documented-standard violation.

**GENERAL [Medium] — `getConnectorUpgradeAdvice` is plumbed through the whole webview RPC stack with no webview caller.** Added to `rpc-types/service-designer/index.ts:50`, `rpc-type.ts:52`, `rpc-client.ts:176`, `rpc-handler.ts:93`, `DefaultServer.ts:296` and `rpc-manager.ts:472` — six files — but the only consumer is the extension host, which calls `StateMachine.langClient().getConnectorUpgradeAdvice` directly (`connector-upgrade.ts:43`). Only `pullConnectorUpgrade` has a real webview caller. Drop the unused half until something needs it.

**GENERAL [Nitpick] — `@since` applied three ways in one PR.** New files carry `1.3.0` (`ConnectorUpgradeAdvisor`, `ModelResolutionIssue`, `TriggerPropertiesRegistry`, `ServiceListenerUsageScanner`), `1.10.0` (`ArtifactInfo`, `TriggerUIMetadataModel`, `ListenerChoiceDeriver`), or nothing (`ArtifactIcon`, `ArtifactMetadata`, `TriggerUIAuthoringParser`, `TriggerUIMetadataCompiler`, `ConnectorUpgradeAdviceResponse`). Pick one convention or drop the tag.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerModelReader.java
LakshanWeerasinghe: Fixed — removed the flag/cache/exclusion-set entirely; the L1-required, L2-optional path is now the only resolution path.

packages/ballerina-extension/src/stateMachine.ts
LakshanWeerasinghe: Fixed — the check now runs fire-and-forget; the transition to `webViewLoading` no longer waits on it.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/adapter/TriggerServiceAdapter.java
LakshanWeerasinghe: Fixed — added a canonical `TriggerKind` enum in `model-generator-commons`; all Java call sites now use it. Left `triggerKind.ts` separate (cross-language) with a sync-comment.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerUIMetadataCompiler.java
LakshanWeerasinghe: Agreed, this is real — didn't do the full rewrite in this pass, too large/risky to fold in here. Tracking as follow-up.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerUIMetadataCompiler.java
LakshanWeerasinghe: Fixed — added named constants, replaced the hardcoded literals.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReader.java
LakshanWeerasinghe: Already fixed by an earlier commit on this branch (removed the variants envelope entirely).

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReader.java
LakshanWeerasinghe: Fixed — extracted `resolveSafeRelativePath` as the one guarded resolve; all three call sites use it now.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/PackageUtil.java
LakshanWeerasinghe: Fixed — removed the duplicate/unmemoized overload; the version-aware lookup goes through the memoized path now.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReader.java
LakshanWeerasinghe: Fixed — removed the dead accessor.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/model/response/ConnectorUpgradeAdviceResponse.java
LakshanWeerasinghe: Fixed — moved `ConnectorUpgradeAdvice` into `model/response`.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/models/TriggerUIMetadataModel.java
LakshanWeerasinghe: Fixed — removed the compat constructors from production; added test-only factories in `TriggerUIMetadataCompilerTest`.

packages/ballerina-side-panel/src/components/Form/validationRules.ts
LakshanWeerasinghe: Left client-only for now — a server-side alias would reintroduce a real `Pattern.matches()` vs `RegExp.test()` mismatch on unanchored patterns. No alias added in this pass.

packages/ballerina-visualizer/src/components/ArtifactIcon/index.tsx
LakshanWeerasinghe: Fixed — removed the unused component; consolidated the theme-probe duplication into one shared `isLightTheme`.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/ConnectorUpgradeAdvisor.java
LakshanWeerasinghe: Fixed — removed `explicitlyPinned` everywhere (Java record, TS interface, rpc-manager).

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerModelReader.java
LakshanWeerasinghe: Fixed.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/models/ArtifactInfo.java
LakshanWeerasinghe: Fixed — merged into one javadoc block.

packages/ballerina-extension/src/webview-communication/DefaultServer.ts
LakshanWeerasinghe: Fixed.

packages/ballerina-extension/src/stateMachine.ts
LakshanWeerasinghe: Fixed — now a `Set`.

packages/ballerina-language-server/model-generator-commons/src/main/java/io/ballerina/modelgenerator/commons/trigger/LibraryMetadataReader.java
LakshanWeerasinghe: Removed the stale comment; didn't resize the cache in this pass.

packages/ballerina-extension/src/features/project/connector-upgrade.ts
LakshanWeerasinghe: Fixed — added `partitionByPostPullResolution`, checked per-connector after the pull/build.

packages/ballerina-core/src/icons/brand-icon-registry.ts
LakshanWeerasinghe: Fixed — moved back to its original position.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerModelReader.java
```diff
@@ -284,16 +169,15 @@ public Optional<TriggerUISchemaModel> getSchemaDrivenTriggerModel(String orgName
...
-        if (bundled.isPresent() || orgName == null || moduleName == null) {
-            return bundled;
+        if (orgName == null || moduleName == null) {
+            return Optional.empty();
         }
-        String key = orgName + "/" + moduleName;
+        String key = orgName + "/" + moduleName + ":" + (version == null ? "" : version);
         Optional<TriggerUISchemaModel> cached = schemaDrivenTriggerCache.getIfPresent(key);
-        if (cached != null) {
+        if (cached != null && cached.isPresent()) {
```
NipunaRanasinghe: `Resolution.ABSENT` is cacheable and stores `Optional.empty()`, but this guard only returns on a cache *hit that's present*. A cached empty result is written but never read back, so every request for a connector with no trigger metadata re-runs full resolution (offline package load + semantic-model compile) instead of hitting cache. Was this intentional, or should it just be `cached != null`?

packages/ballerina-extension/src/features/project/connector-upgrade.ts
```diff
@@ -0,0 +1,354 @@
...
+}
+
+/** Runs a real `bal build` (a fresh process, unlike PULL_MODULE's in-process resolve) to pick up an
+ * on-disk Ballerina.toml edit and regenerate Dependencies.toml from it. */
+async function rebuildProject(projectPath: string): Promise<boolean> {
+    let buildCommand = 'bal build';
+    const config = workspace.getConfiguration('ballerina');
+    const ballerinaHome = config.get<string>('home');
+    if (ballerinaHome) {
+        buildCommand = path.join(ballerinaHome, 'bin', buildCommand);
```
NipunaRanasinghe: `path.join(ballerinaHome, 'bin', 'bal build')` produces one string like `<home>/bin/bal build`, which `runCommandWithOutput` spawns with `shell: true` and no args array. If `ballerinaHome` contains a space (e.g. Windows `C:\Program Files\...`), the shell splits at the first space and fails to find the executable. The rest of the codebase already has `ballerinaExtInstance.getBallerinaCmd()` + `quoteShellPath` for exactly this — worth reusing instead of rebuilding the path here.

packages/ballerina-extension/src/features/project/connector-upgrade.ts
```diff
@@ -0,0 +1,354 @@
...
+            versionRange: new Range(document.positionAt(valueStart), document.positionAt(valueEnd))
+        };
+    }
+    return undefined;
+}
+
+/** The latest version Ballerina Central publishes for `orgName/packageName`, or `undefined` on failure. */
+function getLatestPackageVersion(orgName: string, packageName: string): Promise<string | undefined> {
+    return new Promise((resolve) => {
+        const req = https.request({
```
NipunaRanasinghe, line 318: No timeout on this request. If Central hangs or stalls mid-response, this promise never settles — and since `bumpPinnedDependencies` awaits it per-connector in a loop, the whole "Update" flow hangs with no feedback or way to cancel. Consider `req.setTimeout(...)` + resolving `undefined` on timeout.

packages/ballerina-extension/src/features/project/connector-upgrade.ts
```diff
@@ -0,0 +1,354 @@
...
+            });
+        });
+        req.on('error', () => resolve(undefined));
+        req.end();
+    });
+}
+
+/** Numeric (not lexical) semver comparison, so "1.10.0" correctly outranks "1.9.0". */
+function compareSemver(a: string, b: string): number {
+    const partsOf = (v: string) => v.split('.').map((part) => Number.parseInt(part, 10) || 0);
```
NipunaRanasinghe, line 359: Minor: `"2.0.0-beta".split('.')` → `parseInt("0-beta")` → `0`, so a pre-release compares equal to its base release and could be picked as "latest" if Central ever lists one. Low risk in practice, but worth a guard if pre-releases are possible in the response.

packages/ballerina-extension/src/features/project/connector-upgrade.ts
```diff
@@ -0,0 +1,354 @@
...
+    let match: RegExpExecArray | null;
+    while ((match = tableRegex.exec(text)) !== null) {
+        const block = match[0];
+        const orgMatch = block.match(/^\s*org\s*=\s*"([^"]+)"/m);
+        const nameMatch = block.match(/^\s*name\s*=\s*"([^"]+)"/m);
+        if (orgMatch?.[1] !== orgName || nameMatch?.[1] !== packageName) {
+            continue;
+        }
+        const versionMatch = block.match(/^\s*version\s*=\s*"([^"]+)"/m);
+        if (!versionMatch || versionMatch.index === undefined) {
```
NipunaRanasinghe, line 292: Minor: a matching org/name block with no `version =` field returns `undefined` here, i.e. treated as "not pinned" rather than "pinned but no version to bump." Probably moot since the LS side already filters out `repository = "local"` deps before advice reaches this code, but flagging in case there's another version-less pin shape.

packages/ballerina-language-server/service-model-generator/modules/service-model-generator-ls-extension/src/main/java/io/ballerina/servicemodelgenerator/extension/connector/TriggerModelReader.java
LakshanWeerasinghe: Fixed — the guard was `cached != null && cached.isPresent()`, so a cached absent result was written but never read back. Now just `cached != null`.

packages/ballerina-extension/src/features/project/connector-upgrade.ts
LakshanWeerasinghe: Fixed — now uses `extension.ballerinaExtInstance.getBallerinaCmd()` + `quoteShellPath` instead of manually joining `ballerina.home`.

packages/ballerina-extension/src/features/project/connector-upgrade.ts
LakshanWeerasinghe: Fixed — added a 10s `req.setTimeout` that resolves `undefined` on timeout.

packages/ballerina-extension/src/features/project/connector-upgrade.ts
LakshanWeerasinghe: Fixed — `compareSemver` now splits off the `-suffix` and ranks a pre-release below its own base version.

packages/ballerina-extension/src/features/project/connector-upgrade.ts
LakshanWeerasinghe: Left as-is — confirmed the LS side already filters out `repository = "local"` deps before advice reaches this code, and a real explicit pin conventionally always carries a version, so no concrete failure path found.

NipunaRanasinghe (dismissed):
LGTM

NipunaRanasinghe (dismissed):

NipunaRanasinghe (approved):

PR #1056, opened 2026-09-08
Fix array-typed and unsupported-type configurable handling in Config.toml

VellummyilumVinoth:
## Description
- Array-typed configurables (e.g. `string[] & readonly` OAuth scopes) were serialized as a quoted TOML string instead of a real array — `writeConfigValuesToConfig` never branched on array types, so `@iarna/toml` faithfully quoted whatever raw string it was given. Added array-aware parsing (bracket literal or bare comma/whitespace-separated input), per-element type coercion, and a fix for `@iarna/toml`'s numeric underscore-grouping inside arrays.
- Added a guard rejecting langlib-qualified int subtypes (`int:Signed8/16/32/64`, `int:Unsigned8/16/32`, scalar or array) at config-collection time — these compile but fail at runtime with "configurable variable '...' is not supported" (confirmed empirically against Ballerina 2201.13.4; the array form actually crashes with a raw Java stack trace rather than a clean error). Strengthened the agent's own prompt guidance to avoid generating these types in the first place.
- Added a second guard rejecting arrays of any other unsupported element type (record/map/json/etc.) before they're silently mis-written.
- Fixed `getAllConfigStatus` incorrectly treating any array (including `[]`) as "filled", and `readExistingConfigValues` dropping array values entirely instead of pre-filling them.
- Added client-side validation for numeric array fields in the Configuration Collector UI, so invalid input (e.g. `"8080, abc"` for an `int[]`) is caught before submit instead of

It resolves: https://github.com/wso2/product-integrator/issues/2322

**For reference:**

https://github.com/user-attachments/assets/03b9f884-214b-4fd1-8a92-34704ced5f11

**Config.toml**

Screenshot 2026-09-14 at 09 52 10

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary

- Add end-to-end support for array-typed configurables in `Config.toml`.
- Parse and serialize array values with per-element type conversion.
- Preserve populated arrays during status checks and value pre-fill.
- Reject unsupported array element types and langlib-qualified integer subtypes.
- Add array input validation to the Configuration Collector UI.
- Update agent guidance to support primitive arrays and avoid unsupported integer subtypes.

### DIFF
--- a/packages/ballerina-extension/src/features/ai/agent/prompts.ts
+++ b/packages/ballerina-extension/src/features/ai/agent/prompts.ts
@@ -191,7 +191,7 @@ ${getLanglibInstructions()}
 
 ## Code Structure
 - In WSO2 Integrator, Automation is simply an app with a main method unless user specifically mentions a service. Cron Job kind of requirements are handled in the deployment level for Kubernetes or Integration platform level.
-- Define required configurables for the query. Use only string, int, decimal, boolean types in configurable variables. Never assign hardcoded default values to configurables.
+- Define required configurables for the query. Use only string, int, byte, float, decimal, boolean types (or arrays of them) in configurable variables — never a langlib-qualified numeric subtype, even if a connector's client method parameter uses one; those are not supported as configurable types and fail at runtime. Widen to the base type and cast when calling the client. Never assign hardcoded default values to configurables.
 - Initialize any necessary clients with the correct configuration based on the retrieved libraries at the module level (before any function or service declarations).
 - Implement the main function OR service to address the query requirements.
 
--- a/packages/ballerina-extension/src/utils/toml-utils.ts
+++ b/packages/ballerina-extension/src/utils/toml-utils.ts
@@ -29,6 +29,241 @@ export function isPlaceholderValue(value: string | undefined | null): boolean {
     return typeof value === "string" && /^\$\{[^}]+\}$/.test(value);
 }
 
+// Strips '& readonly' and a trailing '?' so e.g. "string[] & readonly" reduces to "string[]".
+export function stripTypeDecorations(type: string): string {
+    return type.replace(/&\s*readonly/g, "").replace(/\?\s*$/, "").trim();
+}
+
+// Matches an array type suffix, fixed-length or not: "[]", "[2]", "[10]", ... (Ballerina supports
+// fixed-length array configurables, e.g. `configurable string[2] items = ?;`).
+const ARRAY_TYPE_SUFFIX = /\[\d*\]\s*$/;
+
+export function isArrayType(type: string): boolean {
+    return ARRAY_TYPE_SUFFIX.test(type);
+}
+
+// Declared size of a fixed-length array type (e.g. "string[2]" -> 2); undefined for "string[]".
+export function fixedArrayLength(type: string): number | undefined {
+    const match = type.match(/\[(\d+)\]\s*$/);
+    return match ? parseInt(match[1], 10) : undefined;
+}
+
+export function arrayElementType(type: string): string {
+    return type.replace(ARRAY_TYPE_SUFFIX, "").trim();
+}
+
+// True if a comma exists at the top level (outside quotes, outside nested brackets/braces). A
+// quote only opens at an element boundary, so a mid-word apostrophe (e.g. "Bob's") can't swallow
+// the rest of the string; depth is clamped at 0 so an unbalanced closing bracket can't either.
+function hasTopLevelComma(value: string): boolean {
+    let depth = 0;
+    let quote: string | null = null;
+    let isEscaped = false;
+    let atBoundary = true;
+    for (const char of value) {
+        if (quote) {
+            if (isEscaped) {
+                isEscaped = false;
+            } else if (char === "\\") {
+                isEscaped = true;
+            } else if (char === quote) {
+                quote = null;
+            }
+            continue;
+        }
+        if ((char === "\"" || char === "'") && atBoundary) {
+            quote = char;
+            continue;
+        }
+        if (char === "[" || char === "{") {
+            depth++;
+        } else if (char === "]" || char === "}") {
+            depth = Math.max(0, depth - 1);
+        } else if (depth === 0 && char === ",") {
+            return true;
+        }
+        atBoundary = /\s/.test(char);
+    }
+    return false;
+}
+
+// Splits on top-level occurrences of a separator character, never inside a quoted element or a
+// nested array/inline-table. Same boundary-gated quoting and depth clamp as hasTopLevelComma.
+function tokenizeTopLevel(value: string, isSeparator: (char: string) => boolean): string[] {
+    const values: string[] = [];
+    let current = "";
+    let depth = 0;
+    let quote: string | null = null;
+    let isEscaped = false;
+    let atBoundary = true;
+
+    const pushCurrent = () => {
+        const trimmed = current.trim();
+        if (trimmed.length > 0) {
+            values.push(trimmed);
+        }
+        current = "";
+        atBoundary = true;
+    };
+
+    for (const char of value) {
+        if (quote) {
+            current += char;
+            if (isEscaped) {
+                isEscaped = false;
+            } else if (char === "\\") {
+                isEscaped = true;
+            } else if (char === quote) {
+                quote = null;
+            }
+            continue;
+        }
+        if ((char === "\"" || char === "'") && atBoundary) {
+            quote = char;
+            current += char;
+            continue;
+        }
+        if (char === "[" || char === "{") {
+            depth++;
+        } else if (char === "]" || char === "}") {
+            depth = Math.max(0, depth - 1);
+        }
+        if (depth === 0 && isSeparator(char)) {
+            pushCurrent();
+        } else {
+            current += char;
+            if (!/\s/.test(char)) {
+                atBoundary = false;
+            }
+        }
+    }
+    pushCurrent();
+    return values;
+}
+
+// Splits an array literal's elements: on commas if any exist at the top level, else on
+// whitespace. Either way, a quoted element (e.g. "New York") is never split on its own internal
+// characters — quote an element to protect an internal space from being treated as a separator.
+function splitArrayElements(value: string): string[] {
+    const isSeparator = hasTopLevelComma(value) ? (c: string) => c === "," : (c: string) => /\s/.test(c);
+    return tokenizeTopLevel(value, isSeparator);
+}
+
+// Decodes TOML basic-string escapes (\" \\ \n \t \r \b \f). Only meaningful for a value that was
+// actually double-quote-delimited — a bare token or a single-quoted TOML literal string has no
+// escape syntax to decode.
+function decodeTomlEscapes(value: string): string {
+    return value.replace(/\\(["\\bfnrt])/g, (_match, escaped: string) => {
+        switch (escaped) {
+            case "n": return "\n";
+            case "t": return "\t";
+            case "r": return "\r";
+            case "b": return "\b";
+            case "f": return "\f";
+            default: return escaped; // \" or \\
+        }
+    });
+}
+
+// Strips a single matching pair of surrounding quotes, decoding TOML escapes for a double-quoted
+// (but not single-quoted, which TOML treats as a literal string) element.
+function unquote(value: string): string {
+    const trimmed = value.trim();
+    if (trimmed.length >= 2) {
+        const first = trimmed[0];
+        const last = trimmed[trimmed.length - 1];
+        if ((first === "\"" || first === "'") && first === last) {
+            const inner = trimmed.slice(1, -1);
+            return first === "\"" ? decodeTomlEscapes(inner) : inner;
+        }
+    }
+    return trimmed;
+}
+
+const STRICT_INT_PATTERN = /^[+-]?\d+$/;
+// Whole-token match only — parseFloat accepts a numeric prefix of a longer string (e.g. "12ms" -> 12),
+// which would silently write a different value than what was typed.
+const STRICT_DECIMAL_PATTERN = /^[+-]?(?:\d+(?:\.\d+)?|\.\d+)(?:[eE][+-]?\d+)?$/;
+const BYTE_MIN = 0;
+const BYTE_MAX = 255;
+
+// float/decimal needs a decimal point or exponent to stay a float in TOML — @iarna/toml drops it
+// for an integral value (e.g. 1.0 -> 1), which Ballerina's config provider then rejects.
+function formatNumericValue(value: number, isFloatLike: boolean): string {
+    const str = String(value);
+    if (!isFloatLike || str.includes(".") || /e/i.test(str)) {
+        return str;
+    }
+    return `${str}.0`;
+}
+
+// `context` distinguishes an array-element error ("in array") from a scalar one ("").
+function parseIntegerToken(variableName: string, value: string, isByte: boolean, context: string): number {
+    if (!STRICT_INT_PATTERN.test(value)) {
+        throw new Error(`Invalid integer value${context} for ${variableName}`);
+    }
+    const n = parseInt(value, 10);
+    if (isByte && (n < BYTE_MIN || n > BYTE_MAX)) {
+        throw new Error(`Byte value${context} for ${variableName} must be between ${BYTE_MIN} and ${BYTE_MAX}`);
+    }
+    return n;
+}
+
+function parseDecimalToken(variableName: string, value: string, context: string): number {
+    if (!STRICT_DECIMAL_PATTERN.test(value)) {
+        throw new Error(`Invalid decimal value${context} for ${variableName}`);
+    }
+    return parseFloat(value);
+}
+
+function coerceArrayElement(variableName: string, raw: string, elementType: string): unknown {
+    const value = unquote(raw);
+    switch (elementType) {
+        case "int":
+        case "byte":
+            return parseIntegerToken(variableName, value, elementType === "byte", " in array");
+        case "decimal":
+        case "float":
+            return parseDecimalToken(variableName, value, " in array");
+        case "boolean":
+            if (value !== "true" && value !== "false") {
+                throw new Error(`Invalid boolean value in array for ${variableName}`);
+            }
+            return value === "true";
+        default:
+            return value;
+    }
+}
+
+// Parses a submitted array value (a bracket literal or a bare comma/whitespace-separated list)
+// into a real JS array so @iarna/toml emits a TOML array instead of a quoted string.
+function parseArrayConfigValue(variableName: string, value: string, varType: string): unknown[] {
+    const trimmed = value.trim();
+    const elementType = arrayElementType(stripTypeDecorations(varType));
+
+    const inner = trimmed.startsWith("[") && trimmed.endsWith("]") ? trimmed.slice(1, -1).trim() : trimmed;
+    const elements = inner ? splitArrayElements(inner) : [];
+
+    const expectedLength = fixedArrayLength(varType);
+    if (expectedLength !== undefined && elements.length !== expectedLength) {
+        throw new Error(`Expected ${expectedLength} value(s) for ${variableName}, got ${elements.length}`);
+    }
+
+    return elements.map(el => coerceArrayElement(variableName, el, elementType));
+}
+
+// Renders a JS array back into a bracket-literal string for display/pre-fill in the UI.
+function stringifyArrayConfigValue(value: unknown[]): string {
+    const renderElement = (v: unknown) => {
+        if (typeof v !== "string") {
+            return String(v);
+        }
+        const escaped = v.replace(/\\/g, "\\\\").replace(/"/g, "\\\"");
+        return `"${escaped}"`;
+    };
+    return `[${value.map(renderElement).join(", ")}]`;
+}
+
 function readTomlSection(
     configPath: string,
     orgName: string,
@@ -58,7 +293,10 @@ export function getAllConfigStatus(
         return status;
     }
     for (const [key, value] of Object.entries(section)) {
-        if (value !== null && typeof value !== "object") {
+        // An explicitly-submitted empty array is a deliberate, meaningful value, not "unset" —
+        // treated as filled here to agree with computeCollectStatus/classifySubmission, which
+        // already treat the pre-filled "[]" display string as provided.
+        if (value !== null && (typeof value !== "object" || Array.isArray(value))) {
             status[key] = "filled";
         }
     }
@@ -94,24 +332,18 @@ export function writeConfigValuesToConfig(
         }
     }
 
-    const numericKeys = new Set<string>();
     for (const [variableName, value] of Object.entries(configValues)) {
-        const varType = typeMap.get(variableName) || "string";
-        if (varType === "int" || varType === "byte") {
-            const intValue = parseInt(value, 10);
-            if (isNaN(intValue)) {
-                throw new Error(`Invalid integer value for ${variableName}`);
-            }
-            section[variableName] = intValue;
-            numericKeys.add(variableName);
+        const varType = stripTypeDecorations(typeMap.get(variableName) || "string");
+        if (isArrayType(varType)) {
+            section[variableName] = parseArrayConfigValue(variableName, value, varType);
+        } else if (varType === "int" || varType === "byte") {
+            section[variableName] = parseIntegerToken(variableName, value, varType === "byte", "");
         } else if (varType === "decimal" || varType === "float") {
-            const decimalValue = parseFloat(value);
-            if (isNaN(decimalValue)) {
-                throw new Error(`Invalid decimal value for ${variableName}`);
-            }
-            section[variableName] = decimalValue;
-            numericKeys.add(variableName);
+            section[variableName] = parseDecimalToken(variableName, value, "");
         } else if (varType === "boolean") {
+            if (value !== "true" && value !== "false") {
+                throw new Error(`Invalid boolean value for ${variableName}`);
+            }
             section[variableName] = value === "true";
         } else {
             section[variableName] = value;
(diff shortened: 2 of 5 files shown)


packages/ballerina-extension/src/utils/toml-utils.ts
```diff
@@ -29,6 +29,122 @@ export function isPlaceholderValue(value: string | undefined | null): boolean {
...
+// Parses a submitted array value (a bracket literal or a bare comma/whitespace-separated list)
+// into a real JS array so @iarna/toml emits a TOML array instead of a quoted string.
+function parseArrayConfigValue(variableName: string, value: string, varType: string): unknown[] {
+    const trimmed = value.trim();
+    const elementType = stripTypeDecorations(varType).replace(/\[\]\s*$/, "").trim();
+
+    let elements: string[];
+    if (trimmed.startsWith("[") && trimmed.endsWith("]")) {
+        const inner = trimmed.slice(1, -1).trim();
+        elements = inner ? splitTopLevelValues(inner) : [];
```
RNViththagan: Inside a bracket literal this always comma-splits, even when the input has no commas. `[8080 9090]` for an int[] becomes one element `"8080 9090"`, and `coerceArrayElement`'s `parseInt` on that reads `8080` and silently stops - 9090 is gone rather than rejected. Same for decimal[]. The UI's own validator (`splitArrayElementsForValidation` in the visualizer file) does whitespace-split inside brackets, so the form says it's valid and the backend then drops half the values. Shall we fall back to whitespace-splitting here too when there's no comma inside the brackets, same as the bare-value branch below?

packages/ballerina-extension/src/utils/toml-utils.ts
```diff
@@ -29,6 +29,122 @@ export function isPlaceholderValue(value: string | undefined | null): boolean {
...
+    } else if (trimmed.includes(",")) {
+        elements = splitTopLevelValues(trimmed);
+    } else {
+        elements = trimmed.split(/\s+/).filter(e => e.length > 0);
```
RNViththagan: This whitespace split isn't quote-aware like `splitTopLevelValues` above, so a string element containing a space can't be entered without brackets: `New York` becomes `["New", "York"]`, and quoting it makes it worse - `"New York"` becomes `["\"New", "York\""]`. Only the bracketed form `[New York]` gives the right single element. Could this reuse `splitTopLevelValues` (which already handles quotes) instead of a plain `split(/\s+/)`?

packages/ballerina-extension/src/utils/toml-utils.ts
```diff
@@ -29,6 +29,122 @@ export function isPlaceholderValue(value: string | undefined | null): boolean {
...
+            if (isNaN(n)) {
+                throw new Error(`Invalid integer value in array for ${variableName}`);
+            }
+            return n;
+        }
+        case "decimal":
+        case "float": {
+            const n = parseFloat(value);
+            if (isNaN(n)) {
+                throw new Error(`Invalid decimal value in array for ${variableName}`);
```
RNViththagan: Ran this through `@iarna/toml` directly and it looks like exponent-notation floats break the file: `stringify({a: {b: {rate: 1e-7}}})` emits `rate = 1e-7.0`, which is invalid TOML and fails to re-parse. A `decimal[]` element like `1e-7` (or a small value JS renders in exponent form, e.g. `0.0000001`) would hit this on write. Once it does, `readTomlSection`'s `parse()` starts throwing and every later read sees "no config", and `writeConfigValuesToConfig` itself throws at its own `parse()` call on the next write - so it's not just a bad value, the whole Config.toml becomes unusable until someone manually fixes it. This looks like it predates this PR for scalar decimals too, but the new array path adds a second, more natural way to hit it (a rates/thresholds array is exactly where scientific notation shows up). Might be worth reformatting exponent output before it's written, for both paths.

packages/ballerina-extension/src/utils/toml-utils.ts
```diff
@@ -95,9 +215,17 @@ export function writeConfigValuesToConfig(
     }
 
     const numericKeys = new Set<string>();
+    const numericArrayKeys = new Set<string>();
+    const NUMERIC_ELEMENT_TYPES = new Set(["int", "byte", "decimal", "float"]);
     for (const [variableName, value] of Object.entries(configValues)) {
-        const varType = typeMap.get(variableName) || "string";
-        if (varType === "int" || varType === "byte") {
+        const varType = stripTypeDecorations(typeMap.get(variableName) || "string");
+        if (varType.endsWith("[]")) {
```
RNViththagan: This only catches `[]`-suffixed types. A fixed-length array type like `string[2]` (if Ballerina supports fixed-length configurables - not sure it does) wouldn't match here, wouldn't match the int-subtype/unsupported-element guards in config-collector.ts either, and would fall through to the plain string branch - reproducing the original bug this PR fixes. Worth a quick check on whether that type is reachable, and if so, rejecting it the same way as the other unsupported cases rather than silently mis-writing it?

packages/ballerina-visualizer/src/views/AIPanel/components/ConfigurationCollector/index.tsx
```diff
@@ -301,7 +312,32 @@ function getFieldConfig(type: string | undefined): FieldConfig {
...
+        const elementType = type.replace(/&\s*readonly/g, "").replace(/\?\s*$/, "").replace(/\[\]\s*$/, "").trim();
+        const isNumericInt = NUMERIC_INT_TYPES.has(elementType);
+        const isNumericFloat = NUMERIC_FLOAT_TYPES.has(elementType);
+        return {
+            inputKind: "text",
+            placeholder: "Enter values separated by commas, e.g. read, write",
+            validate: (v) => {
+                if (!v.trim() || (!isNumericInt && !isNumericFloat)) return null;
+                for (const el of splitArrayElementsForValidation(v)) {
+                    if (isNumericInt && (isNaN(parseInt(el, 10)) || !Number.isInteger(parseFloat(el)))) {
```
RNViththagan: Similar shape to CodeRabbit's boolean-array comment: `isNumericInt` here accepts anything `Number.isInteger(parseFloat(el))` allows, which is looser than what actually gets written server-side (`parseInt(value, 10)` in `coerceArrayElement`). `1e3` passes validation as 1000 but writes as `1`; `8_080` passes but writes as `8`. Since the whole point of this validator is to catch what the backend can't handle, would it be worth matching the backend's actual coercion here (or tightening both to a strict `^-?\d+$`) so the two agree?

packages/ballerina-extension/src/utils/toml-utils.ts
```diff
@@ -58,7 +174,11 @@ export function getAllConfigStatus(
         return status;
     }
     for (const [key, value] of Object.entries(section)) {
-        if (value !== null && typeof value !== "object") {
+        // An array counts as filled only when non-empty — "[]" is indistinguishable from "not set".
+        const isFilled = Array.isArray(value)
```
RNViththagan: With this fix, `getAllConfigStatus` reports an empty array as "missing", but elsewhere (pre-fill / the collect-mode status the agent tool reads) an empty array read back from disk is treated as a set value, not absent. So submitting `[]` for an array field: status flips to "missing" again, the agent re-collects, the form pre-fills with `[]` since that's what's on disk, user submits `[]` again - and it's stuck deciding the field is both filled and not. Is that expected, or should `[]` count as filled everywhere once a user has explicitly submitted it?

packages/ballerina-extension/src/utils/toml-utils.ts
```diff
@@ -29,6 +29,122 @@ export function isPlaceholderValue(value: string | undefined | null): boolean {
...
+        case "decimal":
+        case "float": {
+            const n = parseFloat(value);
+            if (isNaN(n)) {
+                throw new Error(`Invalid decimal value in array for ${variableName}`);
+            }
+            return n;
+        }
+        case "boolean":
+            return value === "true";
```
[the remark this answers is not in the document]
VellummyilumVinoth: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/1056/commits/558592f32e4d1015255405b7627d4c9e28ee445a

packages/ballerina-extension/src/utils/toml-utils.ts
VellummyilumVinoth: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/1056/commits/558592f32e4d1015255405b7627d4c9e28ee445a

RNViththagan (approved):
LGTM

SasinduDilshara (dismissed):

packages/ballerina-extension/src/utils/toml-utils.ts
```diff
@@ -142,10 +366,21 @@ export function writeConfigValuesToConfig(
...
-                        const pattern = new RegExp(`^(\\s*${escapedKey}\\s*=\\s*)[0-9][0-9_]*(?:\\.[0-9]+)?`, "gm");
+                        const pattern = new RegExp(
+                            `^(\\s*${escapedKey}\\s*=\\s*)[+-]?[0-9][0-9_]*(?:\\.[0-9_]+)?(?:[eE][+-]?[0-9]+)?(?:\\.0)?`,
+                            "gm"
+                        );
                         sectionSlice = sectionSlice.replace(pattern, `$1${numValue}`);
                     }
                 }
+                // Same fixups, applied per-element within a numeric array's '[ ... ]' span.
+                for (const key of numericArrayKeys) {
```
pasindufernando1: **Blocker: the numeric repair pass is keyed to the submission, not to the file — so numeric keys already in the section get corrupted on every write.**

`numericKeys` / `numericArrayKeys` (lines 310-311) are populated only while looping over `configValues`, i.e. the keys the user just submitted. But `stringify(config)` at line 354 re-serialises the **whole** file, so any numeric key already on disk and not resubmitted is transformed by `@iarna` and never repaired.

This is reachable without any unusual input, because `classifySubmission` (`config-collector.ts:151`) only forwards fields the user actually filled — the ones it describes as *"preserved silently on disk"* (line 800) are exactly the ones that get damaged. It also happens whenever the agent calls collect with a subset of variables, which it is instructed to do.

Verified against this branch:

```
--- round 1: tiny + port both submitted (correct) ---
[myorg.mypkg]
tiny = 1e-7
port = 8080

--- round 2: only an unrelated 'other' submitted ---
[myorg.mypkg]
tiny = 1e-7.0
port = 8_080
other = 5
```

`8_080` is valid TOML but rejected by Ballerina. `1e-7.0` is not valid TOML at all, which makes this unrecoverable rather than merely wrong:

- `readTomlSection` catches the parse error and returns `null` (line 254), so `getAllConfigStatus` returns `{}` — every value already on disk reports **missing**.
- `writeConfigValuesToConfig` re-throws it (line 296), so the next collect fails outright. The tool can no longer repair the file it just broke; only a hand-edit of `Config.toml` recovers it.

And `1e-7` is a value the collector itself writes, so two ordinary collect calls are enough.

Suggested fix — drive the repair from the section rather than from the submission, which also lets both sets go away:

```ts
for (const [key, numValue] of Object.entries(section)) {
    if (typeof numValue !== "number") { continue; }
    // ...existing escapedKey + pattern + replace...
}
for (const [key, arrValue] of Object.entries(section)) {
    if (!Array.isArray(arrValue) || !arrValue.every(e => typeof e === "number")) { continue; }
    // ...existing '[ ... ]' span fixup...
}
```

Worth noting the deeper cause: round-tripping the entire file through `parse` -> `stringify` and reconstructing the lost detail with regexes is what creates this class of bug. `parse` also collapses `rate = 1.0` to `1`, so an untouched `float` key degrades to an int even with the above fix. Editing only the lines the tool owns (splice the changed keys, leave the rest byte-identical) would remove the underscore, exponent and `.0` problems for every key you aren't writing.

packages/ballerina-extension/src/utils/toml-utils.ts
```diff
@@ -29,6 +29,216 @@ export function isPlaceholderValue(value: string | undefined | null): boolean {
...
+        case "int":
+        case "byte": {
+            if (!STRICT_INT_PATTERN.test(value)) {
+                throw new Error(`Invalid integer value in array for ${variableName}`);
+            }
+            return parseInt(value, 10);
+        }
+        case "decimal":
+        case "float": {
+            const n = parseFloat(value);
```
pasindufernando1: **Blocker: `float[]` / `decimal[]` (and scalar `float` / `decimal`) with whole-number values are written as TOML integers.**

`parseFloat("1.0") === 1` — JS has one number type — and `@iarna` serialises integral numbers with no decimal point. Verified on this branch:

```
input: rates = "1.0, 2.0", rate = "1.0"   (configurable float[] / float)
output:
[myorg.mypkg]
rates = [ 1, 2 ]
rate = 1
```

Ballerina's config provider is strict about int-vs-float, so the run fails with a type mismatch on both keys — the value looks accepted in the form and breaks at runtime, which is the class of silent mis-write this PR is fixing elsewhere.

`fixNumericToken` can't rescue it: it only *removes* characters (strips `_`, strips a `.0` that follows an exponent) and has no type information at that point.

The scalar path at line 327 has the same hole, and the repair pass makes it worse rather than better — line 373 rewrites the token as `` `$1${numValue}` ``, and `String(1)` is `"1"`, so even if `@iarna` had emitted something usable it would be flattened back to an integer.

Suggested fix: keep the declared type available to the repair step and format integral float/decimal values explicitly, e.g. `Number.isInteger(numValue) ? `${numValue}.0` : String(numValue)` for keys declared `float`/`decimal`, and likewise when joining a numeric array's elements.

packages/ballerina-extension/src/utils/toml-utils.ts
```diff
@@ -180,6 +415,8 @@ export function readExistingConfigValues(
                 existingValues[name] = value;
             } else if (typeof value === "number") {
                 existingValues[name] = value.toString();
+            } else if (Array.isArray(value)) {
```
pasindufernando1, line 438: Nice catch fixing the array drop here — but the `boolean` case is still unhandled, and the omission is now load-bearing in three places.

A `configurable boolean x` already set to `false` on disk is not returned by this function, so:

1. `buildInitialValues` (`ConfigurationCollector/index.tsx:470`) fills the gap from `getFieldConfig`'s `defaultValue: "true"`, the select renders `true`, and an unmodified save rewrites `x = true` — silently inverting the user's `false`.
2. `computeCollectStatus` reports the variable **missing** when the user skips it, while `getAllConfigStatus` (line 274) reports it **filled** — the two disagree on the same file.

One branch alongside the array one clears all of it:

```ts
} else if (typeof value === "boolean") {
    existingValues[name] = value ? "true" : "false";
}
```

Pre-existing rather than introduced here, but it's the other half of the bug this hunk fixes, so it seems worth folding in.

pasindufernando1 (commented):
Reviewed this against the branch and reproduced the numeric behaviour by exercising `toml-utils.ts` directly (it only imports `fs`, `path` and `@iarna/toml`, so it runs standalone). The direction looks right and the array support is a real improvement — but two of the findings below will corrupt `Config.toml` in normal use and I think should be fixed before this merges:

1. The numeric repair pass only covers keys in the current submission, so numeric values already in the section are corrupted on every write — and one of the corrupted forms (`1e-7.0`) is invalid TOML, which leaves the file unreadable *and* unwritable by the tool.
2. Integral `float` / `decimal` values (scalar and array) are written as TOML integers and rejected at runtime.

The third is a one-line addition to a hunk this PR already touches.

Separately: there's no unit coverage for any of the new parsing logic, and `writeConfigValuesToConfig` / `getAllConfigStatus` / `readExistingConfigValues` have no callers outside `config-collector.ts` — so tests would be cheap to add and would have caught both blockers.

packages/ballerina-extension/src/utils/toml-utils.ts
VellummyilumVinoth: Addressed via https://github.com/ballerina-platform/ballerina-vscode/pull/1056/commits/0bf99754dfc31ddb97d3f23bd8939e40b9397924

packages/ballerina-extension/src/utils/toml-utils.ts
VellummyilumVinoth: Addressed via https://github.com/ballerina-platform/ballerina-vscode/pull/1056/commits/0bf99754dfc31ddb97d3f23bd8939e40b9397924

packages/ballerina-extension/src/utils/toml-utils.ts
VellummyilumVinoth: Addressed via https://github.com/ballerina-platform/ballerina-vscode/pull/1056/commits/0bf99754dfc31ddb97d3f23bd8939e40b9397924

packages/ballerina-extension/src/utils/toml-utils.ts
```diff
@@ -29,6 +29,217 @@ export function isPlaceholderValue(value: string | undefined | null): boolean {
...
+// Renders a JS array back into a bracket-literal string for display/pre-fill in the UI.
+function stringifyArrayConfigValue(value: unknown[]): string {
+    const renderElement = (v: unknown) => {
+        if (typeof v !== "string") {
+            return String(v);
+        }
+        const escaped = v.replace(/\\/g, "\\\\").replace(/"/g, "\\\"");
+        return `"${escaped}"`;
+    };
+    return `[${value.map(renderElement).join(", ")}]`;
```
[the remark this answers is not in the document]
VellummyilumVinoth: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/1056/commits/0d978e8cfe01e24116d40e75804a46db47571ddc

packages/ballerina-extension/src/utils/toml-utils.ts
```diff
@@ -29,6 +29,217 @@ export function isPlaceholderValue(value: string | undefined | null): boolean {
...
+            }
+            return parseInt(value, 10);
+        }
+        case "decimal":
+        case "float": {
+            const n = parseFloat(value);
+            if (isNaN(n)) {
+                throw new Error(`Invalid decimal value in array for ${variableName}`);
+            }
+            return n;
```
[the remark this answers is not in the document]
VellummyilumVinoth: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/1056/commits/0d978e8cfe01e24116d40e75804a46db47571ddc

packages/ballerina-extension/src/utils/toml-utils.ts
```diff
@@ -29,6 +29,217 @@ export function isPlaceholderValue(value: string | undefined | null): boolean {
...
+
+function coerceArrayElement(variableName: string, raw: string, elementType: string): unknown {
+    const value = unquote(raw);
+    switch (elementType) {
+        case "int":
+        case "byte": {
+            if (!STRICT_INT_PATTERN.test(value)) {
+                throw new Error(`Invalid integer value in array for ${variableName}`);
+            }
+            return parseInt(value, 10);
```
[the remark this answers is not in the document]
VellummyilumVinoth: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/1056/commits/0d978e8cfe01e24116d40e75804a46db47571ddc

packages/ballerina-extension/src/features/ai/agent/prompts.ts
```diff
@@ -191,7 +191,7 @@ ${getLanglibInstructions()}
 
 ## Code Structure
 - In WSO2 Integrator, Automation is simply an app with a main method unless user specifically mentions a service. Cron Job kind of requirements are handled in the deployment level for Kubernetes or Integration platform level.
-- Define required configurables for the query. Use only string, int, decimal, boolean types in configurable variables. Never assign hardcoded default values to configurables.
+- Define required configurables for the query. Use only string, int, decimal, boolean types in configurable variables — never a langlib-qualified numeric subtype (e.g. \`int:Signed32\`, \`int:Unsigned16\`) even if a connector's client method parameter uses one; those are not supported as configurable types and fail at runtime. Widen to the base type (\`int\`) and cast when calling the client. Never assign hardcoded default values to configurables.
```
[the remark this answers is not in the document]
VellummyilumVinoth: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/1056/commits/0d978e8cfe01e24116d40e75804a46db47571ddc

pasindufernando1 (dismissed):
LGTM. All three findings are addressed, and I verified the fixes against `0d978e8c` by exercising `toml-utils.ts` directly:

- **Repair pass scope** — now driven from `Object.entries(section)`, so a numeric key already on disk survives a write that doesn't resubmit it. `tiny = 1e-7` / `port = 8080` stay intact, and check mode reports `filled` instead of `{}`.
- **Integral `float`/`decimal`** — `formatNumericValue` keeps the point: `rates = [ 1.0, 2.0 ]`, `rate = 1.0`.
- **Boolean read** — the `typeof value === "boolean"` branch fixes the pre-fill, the silent `false` -> `true` flip, and the status disagreement.

Nice additions beyond what was raised: `STRICT_DECIMAL_PATTERN` (so `2x` no longer truncates to `2`), the byte range checks, and passing `allSourceVariables` to the writer so the repair knows the type of a key this round didn't resubmit. The regression tests in `tomlUtils.test.ts` cover the right cases.

packages/ballerina-extension/src/utils/toml-utils.ts
```diff
@@ -126,28 +349,46 @@ export function writeConfigValuesToConfig(
...
+            }
+            // Same idea for a numeric array: rebuild its '[ ... ]' span from the real values.
+            for (const [key, arrValue] of Object.entries(section)) {
+                if (!Array.isArray(arrValue) || arrValue.length === 0 || !arrValue.every(e => typeof e === "number")) {
+                    continue;
                 }
-                tomlContent = tomlContent.slice(0, sectionStart) + sectionSlice + tomlContent.slice(sectionEnd);
+                const declaredType = stripTypeDecorations(typeMap.get(key) || "");
+                const elementType = isArrayType(declaredType) ? arrayElementType(declaredType) : declaredType;
+                const isFloatLike = elementType === "decimal" || elementType === "float";
```
madushajg: This writes invalid TOML when the declared type isn't float/decimal but the values are mixed integral/fractional. Repro: `rates = [1.0, 2.5]` on disk + a write of an unrelated key (or a key whose LS type fell back to `"string"` at `config-collector.ts:444`) emits `rates = [ 1, 2.5 ]`, which `@iarna/toml` rejects with `"Inline lists must be a single type"`. From then on `getAllConfigStatus` silently returns `{}` and every subsequent write/remove/rename throws, so the user's `Config.toml` needs a manual fix.

packages/ballerina-extension/src/utils/toml-utils.ts
```diff
@@ -29,6 +29,232 @@ export function isPlaceholderValue(value: string | undefined | null): boolean {
...
+    }
+    pushCurrent();
+    return values;
+}
+
+// Splits an array literal's elements: on commas if any exist at the top level, else on
+// whitespace. Either way, a quoted element (e.g. "New York") is never split on its own internal
+// characters — quote an element to protect an internal space from being treated as a separator.
+function splitArrayElements(value: string): string[] {
+    const isSeparator = hasTopLevelComma(value) ? (c: string) => c === "," : (c: string) => /\s/.test(c);
```
madushajg, line 148: `'` is treated as a quote opener here, so any unmatched apostrophe in unquoted input swallows the following separators.

eg: `"Bob's`, `Alice's"` for a string[] yields one element, `Bob's`, `Alice's`, with no error and no client-side validation to catch it. 

Consider only honoring `'` as a delimiter when it's balanced (or only at an element boundary), and clamping depth at 0 so unbalanced brackets don't disable separator detection.

packages/ballerina-extension/src/utils/toml-utils.ts
VellummyilumVinoth: Address Via https://github.com/ballerina-platform/ballerina-vscode/pull/1056/commits/2247f3c2707ff01bd039ca653052e18277e60cc0

packages/ballerina-extension/src/utils/toml-utils.ts
VellummyilumVinoth: Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/1056/commits/2247f3c2707ff01bd039ca653052e18277e60cc0

pasindufernando1 (dismissed):
LGTM

madushajg (dismissed):

packages/ballerina-extension/src/features/ai/agent/prompts.ts
```diff
@@ -191,7 +191,7 @@ ${getLanglibInstructions()}
 
 ## Code Structure
 - In WSO2 Integrator, Automation is simply an app with a main method unless user specifically mentions a service. Cron Job kind of requirements are handled in the deployment level for Kubernetes or Integration platform level.
-- Define required configurables for the query. Use only string, int, decimal, boolean types in configurable variables. Never assign hardcoded default values to configurables.
+- Define required configurables for the query. Use only string, int, byte, float, decimal, boolean types, scalar or array, in configurable variables — never a langlib-qualified numeric subtype, even if a connector's client method parameter uses one; those are not supported as configurable types and fail at runtime. Widen to the base type and cast when calling the client. Never assign hardcoded default values to configurables.
```
SasinduDilshara: Does byte allows?

SasinduDilshara: We can remove the scalar part ryt?

VellummyilumVinoth: Yes, `byte` is allowed. It is an officially supported configurable type in Ballerina, with values in the 0–255 range. https://ballerina.io/learn/provide-values-to-configurable-variables/

VellummyilumVinoth: Yes we can.
Addressed in https://github.com/ballerina-platform/ballerina-vscode/pull/1056/commits/533f932e88fa5f952ac4ee4ce0f5a4d817fdc813

SasinduDilshara (approved):

samithkavishke (approved):
LGTM

PR #1062, opened 2026-09-08
Fix Send Email activity field types and keyword escaping in generated activities

rdulmina:
Fixes wso2/product-integrator#2390 and wso2/product-integrator#2391 in the prebuilt activities.

## 1. Send Email fields only accepted an expression (wso2/product-integrator#2390)

`to`, `subject`, `body` and `from` declared a single `EXPRESSION` type, so entering a literal address meant typing it with Ballerina quotes — `user@example.com` produced invalid source. Every other builtin activity field of this shape (REST `path`, SOAP `action`/`path`) already offers both modes, so this was an inconsistency in the Email strategy rather than a missing capability.

They are now `TEXT` (selected by default) + `EXPRESSION`, and source generation goes through the existing `BuiltinActivityStrategy.addQuotedArg`, so text input is emitted as a quoted string literal and expression input verbatim.

For the read-back path, `addDualTypePathProperty` is generalised into `addDualTypeProperty` (adds an `expressionType` and a `required` flag) so the four required email fields round-trip the mode the same way REST/SOAP already do: a double-quoted string literal in source reopens the field in Text mode, anything else in Expression mode. `addRequiredExpressionProperty` became unused and is removed.

## 2. Generated activities dropped the `'` escape on keyword parameter names (wso2/product-integrator#2391)

Creating an activity from a connection action derived its parameters from `ParameterSymbol.getName()`, which returns the bare name, so an action such as

```ballerina
remote function convert(string 'from, string to) returns string|error;
```

produced an invalid signature:

```ballerina
isolated function convertActivity(string from, string to) returns string|error {
//                                       ^ `from` is a keyword
```

Inheriting the escape from the action parameter itself does not work on this path — only the connector index stores escaped names (`central-index.sqlite` has 2411 parameters starting with `'`), and the analyzer reads the semantic model, where no escaped variant exists. The fix uses the keyword syntax API instead: `CommonUtil.escapeReservedKeyword` (backed by `SyntaxInfo.isKeyword`), which `CodeAnalyzer` already uses for exactly this and which is idempotent, so an already-escaped index name passes through untouched. Resource path parameters and expanded included-record fields get the same treatment.

Two related points surfaced while fixing it:

- `Documentation.parameterMap()` may be keyed by the name as written in the doc comment, so the parameter description is now looked up under both spellings — otherwise escaping the name would silently drop the doc line.
- The bug was worse than a bad signature. The wizard matches the derived parameter against the action node template by name:

  ```ts
  const property = newProperties[param.name];   // key is "'from" -> undefined
  if (!property) { continue; }
  ```

  so with an index-resolved connector the argument was silently never wired into the action call. The lookup now accepts either spelling, since a project-local connector is keyed by the bare name.

`# + 'from - ...` alongside `string 'from` compiles clean on 2201.13.4, so the generated doc line needs no special casing.

## Testing

150 tests, no failures:

| Suite | Tests |
| --- | --- |
| `NodeTemplateTest` | 120 |
| `SourceGeneratorTest` | passed |
| `AnalyzeActivityActionTest` | 7 (incl. new `testKeywordParamNameIsEscaped`) |
| `GenActivityTest` | 11 |
| `WorkflowNodeTemplateTest` | 10 |
| `WorkflowManagerTest` | 2 |

`AnalyzeActivityActionTest` gains a case covering the escaping, backed by a `convert(string 'from, string to)` action added to the `analyze_action` test project. `builtin_activity_email.json` is updated for the new field types. Visualizer `tsc` is clean for the edited file (the 11 pre-existing `Type 'Timeout'` errors elsewhere are untouched).

## Not addressed

The quote-stripping in `addDualTypeProperty` is naive (`value.substring(1, len - 1)`), so a text value containing an escaped quote degrades slightly on each round trip — `"he said \"hi\""` reads back as `he said \"hi\"` and re-emits double-escaped. That predates this change and affects REST `path` and SOAP `action` equally, so it is left out of scope.

- Added default Text and Expression modes for Send Email `to`, `subject`, `body`, and `from` fields.
- Preserved quoted Text values and unquoted Expression values during source generation and read-back.
- Escaped reserved keyword names in generated parameters, resource paths, and record fields.
- Updated documentation lookup, action-template property lookup, and tests for escaped names.


packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -2548,33 +2551,45 @@ private void addOptionalAdvancedExpression(Map<String, String> opts, String optK
...
-        nodeBuilder.properties().custom()
+        Property.Builder<FormBuilder<NodeBuilder>> builder = nodeBuilder.properties().custom()
                 .metadata().label(label).description(description).stepOut()
                 .type().fieldType(Property.ValueType.TEXT).ballerinaType("string")
-                    .selected(isStringLit).stepOut()
-                .type().fieldType(Property.ValueType.EXPRESSION).ballerinaType("string")
-                    .selected(!isStringLit).stepOut()
+                    .selected(textSelected).stepOut()
+                .type().fieldType(Property.ValueType.EXPRESSION).ballerinaType(expressionType)
+                    .selected(!textSelected).stepOut()
```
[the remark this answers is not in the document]
rdulmina: Valid — added in 97dd2a4788.

The gap was real: `node_template/config/builtin_activity_email.json` pins only the fresh template, `workflow_builtin_activity_empty_path` covers the absent-argument case for the REST `path` field rather than email, and `to_source/config/builtin_activity_email_manual_retry_to_source.json` predates this change — its values are already quoted under an `EXPRESSION`-only type, so the new `addQuotedArg` TEXT branch and the `'from` keyword argument were never exercised end to end.

New L3 test `BuiltinEmailFieldModeTest` reads a workflow that calls the activity three times, one per entry mode:

```ballerina
() _ = check ctx->callActivity(activity:sendEmail, {connection: smtpClient, to: "user@example.com", subject: "Order update", 'from: "no-reply@example.com", body: "Your order has shipped."});
() _ = check ctx->callActivity(activity:sendEmail, {connection: smtpClient, to: recipient, subject: subjectLine, 'from: recipient, body: subjectLine});
() _ = check ctx->callActivity(activity:sendEmail, {connection: smtpClient});
```

It asserts, per field, that both modes are offered with the right one selected, the value display form, the `ballerinaType` on each side (including `string|string[]` for `to`'s expression mode — the `expressionType` parameter added here), and `codedata.kind == REQUIRED`. Then it feeds each node back through `getSourceCode` and asserts the regenerated `args` mapping:

| entered as | regenerated |
| --- | --- |
| string literals | `{connection: smtpClient, to: "user@example.com", subject: "Order update", 'from: "no-reply@example.com", body: "Your order has shipped."}` |
| expressions | `{connection: smtpClient, to: recipient, subject: subjectLine, 'from: recipient, body: subjectLine}` |
| absent | `{connection: smtpClient}` |

7/7 pass, and they are not vacuous: reverting the fix (`textSelected = false` in `CodeAnalyzer.addDualTypeProperty`) fails `testStringLiteralsOpenInTextMode`, `testAbsentArgumentsOpenInTextMode` and `testStringLiteralsRegenerateQuoted`, while the two expression-mode tests correctly stay green.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/EmailActivityStrategy.java
```diff
@@ -259,11 +264,12 @@ public String connectionBallerinaType() {
...
-        addArg(args, TO_KEY, "to", props);
-        addArg(args, SUBJECT_KEY, "subject", props);
-        addArg(args, FROM_KEY, FROM_PARAM, props);
-        addArg(args, BODY_KEY, "body", props);
+        // Required params (order matches sendEmail signature). Each is TEXT/EXPRESSION dual-typed:
+        // TEXT input is emitted as a quoted string literal, EXPRESSION input verbatim.
+        BuiltinActivityStrategy.addQuotedArg(args, "to", props, TO_KEY);
+        BuiltinActivityStrategy.addQuotedArg(args, "subject", props, SUBJECT_KEY);
+        BuiltinActivityStrategy.addQuotedArg(args, FROM_PARAM, props, FROM_KEY);
+        BuiltinActivityStrategy.addQuotedArg(args, "body", props, BODY_KEY);
```
dulajdilshan: **[Standards · Medium] Duplicate implementation — and the two String params are transposed.**

`addArg` (`:288`) is `addQuotedArg` minus the quoting. Both end `args.add(paramName + ": " + value)`. They now sit 3 lines apart with the middle two args **swapped**:

```java
:269  BuiltinActivityStrategy.addQuotedArg(args,      "to",  props, TO_KEY);   // paramName, props, propKey
:275  addArg(                             optFields, CC_KEY, "cc",  props);   // propKey,   paramName, props
```

All `String` -> a transposition compiles and silently emits `to: <value of prop "to">` under the wrong name.

The optional fields are EXPRESSION-only, so `isTextSelected` -> `false` and `addQuotedArg` behaves identically for them. Delete `addArg`, call `addQuotedArg` for all 11.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/EmailActivityStrategy.java
```diff
@@ -72,53 +72,58 @@ public class EmailActivityStrategy implements BuiltinActivityStrategy {
...
 
-        // Subject — required string
+        // Subject — required; plain text by default, switchable to an expression
         nodeBuilder.properties().custom()
                 .metadata()
                     .label("Subject")
                     .description("Email subject line")
                     .stepOut()
-                .type().fieldType(Property.ValueType.EXPRESSION).ballerinaType(STRING_TYPE).selected(true).stepOut()
+                .type().fieldType(Property.ValueType.TEXT).ballerinaType(STRING_TYPE).selected(true).stepOut()
```
dulajdilshan: **[Standards · Medium] Duplicated Code — same 13-line chain 4x.**

`:75-131` repeats one property chain for To/Subject/Body/From. Subject (`:90-102`) and Body (`:104-116`) are byte-identical bar two strings and the key:

```java
.metadata().label(LABEL).description(DESC).stepOut()
.type().fieldType(Property.ValueType.TEXT).ballerinaType(STRING_TYPE).selected(true).stepOut()
.type().fieldType(Property.ValueType.EXPRESSION).ballerinaType(EXPR_TYPE).selected(false).stepOut()
.codedata().kind(REQUIRED.name()).stepOut()
.value("").editable(true).stepOut().addProperty(KEY);
```

This PR had to edit all four to insert the one TEXT line — the cost is already paid. Only `label`, `description`, `expressionType` and `key` vary; a `addDualTypeFormProperty(nodeBuilder, key, label, desc, exprType)` collapses it to 4 calls.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -2548,33 +2551,45 @@ private void addOptionalAdvancedExpression(Map<String, String> opts, String optK
...
+     * @param advanced       {@code true} to mark the property as advanced (for SOAP path/action)
+     * @param required       {@code true} for a REQUIRED, non-optional parameter (the email fields)
+     */
+    private void addDualTypeProperty(Map<String, String> src, String key, String label,
+                                     String description, String placeholder, String expressionType,
+                                     boolean advanced, boolean required) {
         String value = src.getOrDefault(key, "");
         boolean isStringLit = value.length() >= 2 && value.startsWith("\"") && value.endsWith("\"");
         String displayValue = isStringLit ? value.substring(1, value.length() - 1) : value;
+        // An absent argument carries no evidence of the entry mode, so fall back to the strategies'
```
dulajdilshan: **[Standards · Medium] The "is this a quoted literal" test exists twice, in two classes.**

```java
// CodeAnalyzer:2573                       - read side (strip quotes, pick TEXT)
value.length() >= 2 && value.startsWith("\"") && value.endsWith("\"")
// BuiltinActivityStrategy:219 isBallerinaStringExpression - write side (skip re-quoting)
value.startsWith("\"") && value.endsWith("\"") && value.length() > 1
```

Same predicate, same module, two names, and they are the two halves of one round-trip that the javadoc at `:2560` already admits must stay in sync ("Mirrors the node template built by the builtin activity strategies"). Nothing enforces it.

Call `BuiltinActivityStrategy.isBallerinaStringExpression(value)` here so read and write cannot drift.

(Separate from CodeRabbit's escape-decoding point on the same pair — this is about the duplicated predicate, not the escaping.)

packages/ballerina-visualizer/src/views/BI/FlowDiagram/NewActivityFromConnection.tsx
```diff
@@ -553,7 +553,12 @@ export function NewActivityFromConnection(props: NewActivityFromConnectionProps)
                     parameterDescription: param.description,
                 });
             }
-            const property = newProperties[param.name];
+            // The LS escapes a keyword parameter name ('from); the action node template keys the
+            // matching property the same way when it comes from the connector index, but by the bare
+            // symbol name when it is resolved from the project's own semantic model. Try both so the
+            // argument is always wired up.
+            const propertyKey = param.name in newProperties ? param.name : param.name.replace(/^'/, "");
```
dulajdilshan: **[Standards · Medium] Escape in the LS, un-escape in the UI — and the primary branch never fires.**

The comment says the template keys the property escaped "when it comes from the connector index". I could not find that branch. Every parameter property key goes through `ParamUtils.removeLeadingSingleQuote(...)`, unconditionally:

```java
CallBuilder.java:407   return FlowNodeUtil.getPropertyKey(ParamUtils.removeLeadingSingleQuote(paramData.name()));
CallBuilder.java:176   String unescapedParamName = ParamUtils.removeLeadingSingleQuote(paramData.name());
CallBuilder.java:338   String unescapedParamName = ParamUtils.removeLeadingSingleQuote(paramData.name());
```

`RemoteActionCallBuilder` / `ResourceActionCallBuilder` extend `CallBuilder` and no subclass overrides it. So keys are always bare -> `param.name in newProperties` is always `false` for a keyword param and only the regex fallback ever runs.

Two consequences:
- `ActionSignatureAnalyzer:69` and `:211` assert the opposite ("matches the action node template's property key"). Wrong invariant to leave in a javadoc.
- the strip is hand-rolled in TS while `ParamUtils.removeLeadingSingleQuote` already does it in Java.

Escape only where source text is emitted (the signature, and the argument at `:566`), keep property keys bare on both sides, and this branch goes away.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/ActionSignatureAnalyzer.java
```diff
@@ -198,17 +206,26 @@ private static Analysis analyze(FunctionTypeSymbol functionTypeSymbol, SemanticM
...
+     * Adds a derived parameter. {@code name} is the raw symbol/field name; it is escaped with
+     * {@link CommonUtil#escapeReservedKeyword} so a keyword name (e.g. {@code from}) is carried into
+     * the generated activity signature — and matched against the action node template's property
+     * key — in its quoted form ({@code 'from}). Escaping is idempotent, so an already-escaped name
+     * coming from the connector index is left alone.
+     */
     private static void addParam(List<DerivedParam> params, List<String> reasons, String name,
                                  TypeSymbol type, boolean required, TypeSymbol anydata,
                                  SemanticModel semanticModel, String description) {
+        String escapedName = CommonUtil.escapeReservedKeyword(name);
```
dulajdilshan: **[Standards · Minor] The escaped name leaks into UI labels.**

`escapedName` is now what the frontend receives as `param.name`, and the form uses it as display text:

```tsx
NewActivityFromConnection.tsx:340   label: param.name,                        // renders 'from
NewActivityFromConnection.tsx:359   label: `Value for '${param.name}'`,       // renders Value for ''from'
NewActivityFromConnection.tsx:364   documentation: `The fixed value used for '${param.name}' ...`
```

Source-level escaping in user-facing text. Either keep the raw name alongside the escaped one on `DerivedParam`, or strip the quote at the label sites.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -2548,33 +2551,45 @@ private void addOptionalAdvancedExpression(Map<String, String> opts, String optK
...
+     * argument is absent altogether; EXPRESSION otherwise. Mirrors the node template built by the
+     * builtin activity strategies, so a saved node reopens in the mode it was entered in and an
+     * untouched field opens in the same mode a fresh node would.
+     *
+     * @param expressionType the ballerinaType advertised by the EXPRESSION type (the TEXT type is
+     *                       always {@code string})
+     * @param advanced       {@code true} to mark the property as advanced (for SOAP path/action)
+     * @param required       {@code true} for a REQUIRED, non-optional parameter (the email fields)
+     */
+    private void addDualTypeProperty(Map<String, String> src, String key, String label,
```
dulajdilshan: **[Standards · Minor] Flag arguments, and `addDualTypePathProperty` is now a Middle Man.**

8 positional params ending in two bare booleans — every email call site reads `..., "", "string", false, true)` (`:2500-2508`), where `false, true` means "not advanced, required". Unreadable at the call site.

And `addDualTypePathProperty` (`:2551`) is now a single delegating line, so there are two entry points for one behaviour.

Suggest one signature with the two booleans replaced by an explicit shape (a small record, or `ValueType`-style constants), and drop the path wrapper.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/java/io/ballerina/flowmodelgenerator/extension/BuiltinEmailFieldModeTest.java
```diff
@@ -0,0 +1,265 @@
...
+ * Tests the TEXT/EXPRESSION dual-typed fields of the Send Email builtin activity over a full
+ * read-save cycle: the flow model is generated from source, the mode each field reopens in is
+ * asserted, and the same node is fed back to the source generator to check the arguments are
+ * regenerated as written.
+ *
+ * <p>The three {@code sendEmail} calls in the test source cover the three ways a field can appear
+ * in source: as a quoted string literal (TEXT), as any other expression (EXPRESSION), and as an
+ * argument that is absent altogether (TEXT, matching the fresh node template).
+ *
+ * @since 1.5.0
```
dulajdilshan: **[Standards · Nitpick] `@since 1.5.0` on a brand-new class.**

New test classes in this package tag the current version — `ConnectorIndexKeywordTest` is `1.8.0`, the Copilot suites are `1.7.0`. `1.5.0` reads as backdated.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/EmailActivityStrategy.java
dulajdilshan, line 226: **[Standards · Medium] Duplicate implementation — and the two String params are transposed.**

`addArg` (`:288`) is `addQuotedArg` minus the quoting. Both end `args.add(paramName + ": " + value)`. They now sit 3 lines apart with the middle two args **swapped**:

```java
:269  BuiltinActivityStrategy.addQuotedArg(args,      "to",  props, TO_KEY);   // paramName, props, propKey
:275  addArg(                             optFields, CC_KEY, "cc",  props);   // propKey,   paramName, props
```

All `String` -> a transposition compiles and silently emits `to: <value of prop "to">` under the wrong name.

The optional fields are EXPRESSION-only, so `isTextSelected` -> `false` and `addQuotedArg` behaves identically for them. Delete `addArg`, call `addQuotedArg` for all 11.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/EmailActivityStrategy.java
dulajdilshan: **[Standards · Medium] Duplicated Code — same 13-line chain 4x.**

`:75-131` repeats one property chain for To/Subject/Body/From. Subject (`:90-102`) and Body (`:104-116`) are byte-identical bar two strings and the key:

```java
.metadata().label(LABEL).description(DESC).stepOut()
.type().fieldType(Property.ValueType.TEXT).ballerinaType(STRING_TYPE).selected(true).stepOut()
.type().fieldType(Property.ValueType.EXPRESSION).ballerinaType(EXPR_TYPE).selected(false).stepOut()
.codedata().kind(REQUIRED.name()).stepOut()
.value("").editable(true).stepOut().addProperty(KEY);
```

This PR had to edit all four to insert the one TEXT line — the cost is already paid. Only `label`, `description`, `expressionType` and `key` vary; a `addDualTypeFormProperty(nodeBuilder, key, label, desc, exprType)` collapses it to 4 calls.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
dulajdilshan, line 2669: **[Standards · Medium] The "is this a quoted literal" test exists twice, in two classes.**

```java
// CodeAnalyzer:2573                       - read side (strip quotes, pick TEXT)
value.length() >= 2 && value.startsWith("\"") && value.endsWith("\"")
// BuiltinActivityStrategy:219 isBallerinaStringExpression - write side (skip re-quoting)
value.startsWith("\"") && value.endsWith("\"") && value.length() > 1
```

Same predicate, same module, two names, and they are the two halves of one round-trip that the javadoc at `:2560` already admits must stay in sync ("Mirrors the node template built by the builtin activity strategies"). Nothing enforces it.

Call `BuiltinActivityStrategy.isBallerinaStringExpression(value)` here so read and write cannot drift.

(Separate from CodeRabbit's escape-decoding point on the same pair — this is about the duplicated predicate, not the escaping.)

packages/ballerina-visualizer/src/views/BI/FlowDiagram/NewActivityFromConnection.tsx
dulajdilshan: **[Standards · Medium] Escape in the LS, un-escape in the UI — and the primary branch never fires.**

The comment says the template keys the property escaped "when it comes from the connector index". I could not find that branch. Every parameter property key goes through `ParamUtils.removeLeadingSingleQuote(...)`, unconditionally:

```java
CallBuilder.java:407   return FlowNodeUtil.getPropertyKey(ParamUtils.removeLeadingSingleQuote(paramData.name()));
CallBuilder.java:176   String unescapedParamName = ParamUtils.removeLeadingSingleQuote(paramData.name());
CallBuilder.java:338   String unescapedParamName = ParamUtils.removeLeadingSingleQuote(paramData.name());
```

`RemoteActionCallBuilder` / `ResourceActionCallBuilder` extend `CallBuilder` and no subclass overrides it. So keys are always bare -> `param.name in newProperties` is always `false` for a keyword param and only the regex fallback ever runs.

Two consequences:
- `ActionSignatureAnalyzer:69` and `:211` assert the opposite ("matches the action node template's property key"). Wrong invariant to leave in a javadoc.
- the strip is hand-rolled in TS while `ParamUtils.removeLeadingSingleQuote` already does it in Java.

`CodeAnalyzer:2153` states the repo's actual rule, in the same feature:

```java
// A keyword field name is keyword-escaped in source ('from) but plain everywhere else.
String key = ParamUtils.removeLeadingSingleQuote(sf.fieldName().toString().trim());
```

Escape only where source text is emitted (the signature, and the argument at `:566`); keep keys bare on both sides and this branch goes away.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/ActionSignatureAnalyzer.java
dulajdilshan: **[Standards · Minor] The escaped name leaks into UI labels.**

`escapedName` is now what the frontend receives as `param.name`, and the form uses it as display text:

```tsx
NewActivityFromConnection.tsx:340   label: param.name,                        // renders 'from
NewActivityFromConnection.tsx:359   label: `Value for '${param.name}'`,       // renders Value for ''from'
NewActivityFromConnection.tsx:364   documentation: `The fixed value used for '${param.name}' ...`
```

Source-level escaping in user-facing text. Either keep the raw name alongside the escaped one on `DerivedParam`, or strip the quote at the label sites.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
dulajdilshan, line 2661: **[Standards · Minor] Flag arguments, and `addDualTypePathProperty` is now a Middle Man.**

8 positional params ending in two bare booleans — every email call site reads `..., "", "string", false, true)` (`:2500-2508`), where `false, true` means "not advanced, required". Unreadable at the call site.

And `addDualTypePathProperty` (`:2551`) is now a single delegating line, so there are two entry points for one behaviour.

Suggest one signature with the two booleans replaced by an explicit shape (a small record, or `ValueType`-style constants), and drop the path wrapper.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/java/io/ballerina/flowmodelgenerator/extension/BuiltinEmailFieldModeTest.java
dulajdilshan: **[Standards · Nitpick] `@since 1.5.0` on a brand-new class.**

New test classes in this package tag the current version — `ConnectorIndexKeywordTest` is `1.8.0`, the Copilot suites are `1.7.0`. `1.5.0` reads as backdated.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -2495,14 +2495,17 @@ private void populateSoapProperties(Map<String, String> src) {
...
-        addRequiredExpressionProperty(src, EmailActivityStrategy.BODY_KEY,
-                "Body", "Plain-text body of the email", "string");
+        // to/subject/from/body are TEXT/EXPRESSION dual-typed: a string literal (or an absent
+        // argument) reopens the form in text mode, anything else in expression mode.
+        addDualTypeProperty(src, EmailActivityStrategy.TO_KEY,
+                "To", "Recipient email address (or list of addresses)", "", "string|string[]",
+                false, true);
+        addDualTypeProperty(src, EmailActivityStrategy.SUBJECT_KEY,
+                "Subject", "Email subject line", "", "string", false, true);
+        addDualTypeProperty(src, EmailActivityStrategy.FROM_KEY,
```
dulajdilshan, line 2600: **[Standards · Medium] Wiring: email field order flips between the fresh node and a reopened one.**

Two builders produce the same form, in different orders:

| | order |
|---|---|
| template `EmailActivityStrategy` `:88 :102 :116 :131` | to, subject, **body, from** |
| read-back `populateEmailProperties` `:2500-2508` | to, subject, **from, body** |

`FormBuilder.nodeProperties` is a `LinkedHashMap` (`:83`), so insertion order is the serialized order. The committed template fixture confirms it — `node_template/config/builtin_activity_email.json` keys read `connection, to, subject, body, from, cc, ...`.

Create an email node -> To/Subject/Body/From. Save, reopen -> To/Subject/From/Body.

Predates this PR, but the PR rewrites all 8 of these lines, so it is a swap of `:2505` and `:2507` now. `getCallActivityArgs` (`:269-272`) and the `sendEmail` signature both use to/subject/from/body, so the template is the odd one out — move `From` above `Body` there.

`BuiltinEmailFieldModeTest` can't catch this: it looks fields up by name (`EMAIL_FIELDS`), never asserts order.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -2548,33 +2551,45 @@ private void addOptionalAdvancedExpression(Map<String, String> opts, String optK
...
-                .type().fieldType(Property.ValueType.EXPRESSION)
-                    .ballerinaType(ballerinaType).selected(true).stepOut()
-                .codedata().kind(ParameterData.Kind.REQUIRED.name()).stepOut()
-                .value(value).editable(true)
-                .stepOut().addProperty(key);
+                .editable(true).optional(!required).advanced(advanced);
+        if (required) {
+            // Reassign rather than discard: stepOut() only hands back the parent builder, so
+            // dropping its result trips SpotBugs' RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT.
+            builder = builder.codedata().kind(ParameterData.Kind.REQUIRED.name()).stepOut();
```
dulajdilshan: **[Standards · Minor] The SpotBugs workaround is removable — drop `stepOut()` instead of assigning it.**

The comment is right that `stepOut()` has no side effect — `FacetedBuilder:35` is literally `return parentBuilder;`, so `RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT` is a true positive.

But the work is already done by the calls before it: `Property.Builder.codedata()` caches the child builder on the parent, and `kind()` sets the field and returns `this`. Both have side effects, so neither is flagged. Just don't call `stepOut()`:

```suggestion
            builder.codedata().kind(ParameterData.Kind.REQUIRED.name());
```

Removes the reassignment and the two-line comment explaining it.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/java/io/ballerina/flowmodelgenerator/extension/BuiltinEmailFieldModeTest.java
dulajdilshan: Actually, the new version should be something: 5.14.0. cc: @NipunaRanasinghe @samithkavishke

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/EmailActivityStrategy.java
rdulmina: Fixed in f9758f6523. `addArg` is gone; all eleven arguments go through `addQuotedArg`.

Confirmed the behaviour is unchanged for the optional seven before swapping them over: each is declared `EXPRESSION`-only both in the template (`EmailActivityStrategy` `:95-215`) and on read-back (`CodeAnalyzer.addOptionalAdvancedExpression`, single `EXPRESSION` type, `selected(true)`), so `isTextSelected` returns false and `addQuotedArg` emits `paramName + ": " + value` — byte-identical to what `addArg` did. `addArg` had exactly one caller, so nothing else was relying on it.

The transposition risk you describe is the real reason to do this rather than just reorder the parameters: with four `String`-ish positions and no type to stop you, a swap compiles and silently emits an argument under the wrong name.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/EmailActivityStrategy.java
rdulmina: Fixed in f9758f6523. The four chains are now four calls:

```java
BuiltinActivityStrategy.addRequiredDualTypeProperty(nodeBuilder, TO_KEY,
        "To", "Recipient email address (or list of addresses)", STRING_OR_STRING_ARRAY);
BuiltinActivityStrategy.addRequiredDualTypeProperty(nodeBuilder, SUBJECT_KEY,
        "Subject", "Email subject line", STRING_TYPE);
```

The helper lives on `BuiltinActivityStrategy` rather than in the email strategy, so it sits next to `addQuotedArg` and its javadoc can point at `CodeAnalyzer.addDualTypeProperty` as the read-side counterpart that has to agree with it.

Not adopted for the other strategies in this PR: Rest has 1 such chain and Soap 2, but they vary in placeholder/optional/advanced and neither file is otherwise touched here, so converting them would widen the diff into untouched code. The helper is in the shared interface and ready for them.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
rdulmina: Fixed in f9758f6523 — the read side now calls the write side's predicate:

```java
boolean isStringLit = BuiltinActivityStrategy.isBallerinaStringExpression(value);
```

Checked they were genuinely interchangeable first: `length() >= 2` and `length() > 1` are the same condition, and the short-circuit order does not change the result for `"` (both false). `isBallerinaStringExpression` is a static interface method, so it is already public.

Only the predicate is shared — the read side additionally strips the quotes for display, which the write side has no use for. If that pairing drifts again the natural next step is a helper returning both the flag and the display value, but sharing the test is what the round-trip actually depends on.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/NewActivityFromConnection.tsx
rdulmina: Confirmed and fixed in f9758f6523 — the branch was dead, exactly as you describe.

Verified the claim end to end before changing anything: `getParameterPropertyKey` is defined once (`CallBuilder:403`), called from one place (`:331`), and **no subclass overrides it** — `RemoteActionCallBuilder` declares no `addProperty` at all and `ResourceActionCallBuilder` just calls `setParameterProperties`. So every parameter key is stripped, `'from' in newProperties` is always false, and only the regex fallback ever ran. The javadoc invariant at `ActionSignatureAnalyzer:69`/`:211` was wrong and is rewritten.

One qualification on the remedy: keeping keys bare on both sides is necessary but not sufficient, because the escaped spelling is genuinely required in two places — `:552`, where it becomes the generated activity's parameter name, and `:565`, the argument passed to the action. `ActivityGenerator:339` reads that value straight into the signature, so dropping the escaping there would regress what this PR set out to fix.

So `DerivedParam` carries both, per your suggestion on the labels thread, and the TSX picks per site:

```ts
const property = newProperties[param.name];        // keys are always bare
...
value = isChecked ? param.escapedName : ...        // source text keeps the quote
```

The hand-rolled `replace(/^'/, "")` is gone, and `ParamUtils.removeLeadingSingleQuote` now does the stripping on the Java side where you pointed out it belongs.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/ActionSignatureAnalyzer.java
rdulmina: Fixed in f9758f6523, together with the property-key thread — one change closes both, as you suggested there: `DerivedParam` now carries both spellings.

```java
public record DerivedParam(String name, String escapedName, String type, boolean required,
                           String description) {
```

`name` is bare and is what the form renders (`label: param.name`) and what property lookups use; `escapedName` is used at exactly the two places the name becomes source text — `:552` (the generated activity's parameter name) and `:565` (the argument passed to the action).

One correction to my own assumption while doing this: the raw symbol name for a keyword parameter **already** carries the quote, so `escapeReservedKeyword` was a no-op and the bare spelling is the one that has to be derived, via `ParamUtils.removeLeadingSingleQuote`. The comment claiming "the symbol name does not [carry the quote]" was wrong; `AnalyzeActivityActionTest` caught it immediately.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
rdulmina: Fixed in f9758f6523. The two booleans are one explicit argument, and the wrapper is gone:

```java
private enum DualTypeUse { REQUIRED, OPTIONAL, OPTIONAL_ADVANCED }
```

Call sites now read `..., "string|string[]", DualTypeUse.REQUIRED)` for the email fields, `DualTypeUse.OPTIONAL` for the REST path and `DualTypeUse.OPTIONAL_ADVANCED` for the SOAP action/path. `addDualTypePathProperty` is deleted and its three callers moved to the single entry point.

Kept the two halves together deliberately: dropping the wrapper on its own would have pushed `"string", false, false` into those three call sites, making them worse — it only reads better once the flags are gone too.

Unrelated thing noticed while moving these, not changed here: the SOAP action advertises `string?` as its EXPRESSION type in the template (`SoapActivityStrategy:80`) but `string` on read-back. Same family as the field-order mismatch, so worth a follow-up.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/java/io/ballerina/flowmodelgenerator/extension/BuiltinEmailFieldModeTest.java
rdulmina: Changed to `1.8.0` in f9758f6523, matching the newest classes in this package (`ConnectorIndexKeywordTest`), so it is no longer backdated.

Leaving this thread open for your follow-up question rather than resolving it: `5.14.0` is a module-wide policy call, not a fix to this one file. Every `@since` in `flow-model-generator-ls-extension` is on the library series (`1.0.0`–`2.0.0`, most common `1.0.0`); `5.14.0` is the extension's version (`packages/ballerina-extension/package.json` is `5.14.0-SNAPSHOT`, matching this branch's base). Switching to the product version would want doing across the module in one go, and it is your and @NipunaRanasinghe / @samithkavishke's call — happy to make that sweep as a separate change once you have decided.

Worth noting for whichever way it goes: `AnalyzeActivityActionTest` carries `@since 1.5.0` and is touched by this PR, so if `1.5.0` is considered backdated then that file is too.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
rdulmina: Confirmed and fixed in f9758f6523 — this was the only user-visible bug in the review.

Verified the mechanism before changing anything: `FormBuilder:83` is a `LinkedHashMap`, and the committed template fixture reads `['connection','to','subject','body','from',...]` against a read-back order of to/subject/from/body.

Moved `From` above `Body` in the template, as you suggested — the signature, `getCallActivityArgs` and the read-back all use to/subject/from/body, so the template was the outlier. The fixture is regenerated to match.

You are right that the test could not catch it, and there is a sharper reason than the lookup-by-name: `JsonObject.equals` is order-insensitive, so **no** golden fixture can catch a field-order regression — the fixture would have kept passing after the swap. So the guard has to be an explicit assertion, which `testFieldOrderMatchesTheNodeTemplate` now does: it pulls the node template and the reopened node and pins both against to/subject/from/body.

Checked it actually bites — putting `Body` back before `From` fails it:

```
testFieldOrderMatchesTheNodeTemplate FAILED
  The node template's email field order changed: Lists differ at element [2]: from != body
```

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
rdulmina: Fixed in f9758f6523 — applied your suggestion:

```java
if (use == DualTypeUse.REQUIRED) {
    builder.codedata().kind(ParameterData.Kind.REQUIRED.name());
}
```

Verified the reasoning rather than taking it on faith: `Property.Builder.codedata()` caches the child builder on the parent (`:589-593`) and `build()` reads `codedataBuilder.build()` (`:1421`), so the codedata is applied without stepping back out; and `kind()` assigns the field before returning `this` (`PropertyCodedata:56-59`), so ignoring *its* return value is not a no-side-effect call. `spotbugsMain` passes on the result, so the warning the workaround existed for does not come back.

The reassignment and its two-line comment are both gone.

packages/ballerina-core/src/interfaces/extended-lang-client.ts
```diff
@@ -1074,8 +1074,12 @@ export interface ActivityActionAnalysis {
...
     /** When unsupported, the human-readable reasons. */
     reasons: string[];
-    /** The derived activity parameters. */
-    params: { name: string; type: string; required: boolean; description?: string }[];
+    /**
+     * The derived activity parameters. `name` is the bare parameter name — it matches the action
+     * node template's property key and is what the form shows; `escapedName` carries the leading
+     * quote for a Ballerina keyword (`'from`) and belongs only in text emitted as source.
+     */
+    params: { name: string; escapedName: string; type: string; required: boolean; description?: string }[];
```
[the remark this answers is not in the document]
rdulmina: Valid — the field is optional as of 61d6928189, so an older server degrades rather than breaks.

Checked how the two are actually deployed before choosing, because that decides which of your two options applies. There is no lockstep guarantee to state: the language server is not bundled with this extension, it is started out of the user's Ballerina distribution — `getServerOptionsUsingCLI` runs `bal start-language-server`, and the java path passes `-Dballerina.home=${ballerinaHome}` with a classpath from that home. The version check that exists (`extension.ts:542`) only warns, and its floor is Swan Lake Beta 3, so it does not gate this at all. A newer extension really can meet an older server.

So the compatibility handling is in the client:

```ts
// A language server older than the escapedName field sends only name, already carrying
// the keyword quote, so falling back to it yields the same source either way.
const escapedName = param.escapedName ?? param.name;
```

That fallback is exact rather than approximate: before this field existed, `name` *was* the escaped spelling, so an older server's response produces the same generated source it always did. What it cannot restore is the property lookup for a keyword parameter — an older server sends `'from` where the node template's key is `from` — but that path was already broken for keyword parameters before this PR, which is the defect the property-key thread describes. This does not regress it.

Worth noting the exposure is narrow in practice: `analyzeActivityAction` is new in this same unreleased cycle, so no released distribution has the API without the field.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-ls-extension/src/test/java/io/ballerina/flowmodelgenerator/extension/BuiltinEmailFieldModeTest.java
samithkavishke: Ideally it should `5.14.0` since language server is not packaged separately as earlier but this is not a mandatory since there are confusing parts in the code also.

samithkavishke (commented):
LGTM

samithkavishke (approved):
LGTM

hasithaa (approved):

PR #1064, opened 2026-09-08
Prevent Component Diagram Links from Overlapping Workflow Nodes

madushajg:
## Purpose
$title
Addresses: https://github.com/wso2/product-integrator/issues/2358

Screenshot 2026-09-08 at 23 57 29

## Summary

- Prevent component diagram links from crossing workflow and unrelated nodes, including curved paths.
- Add Bézier path utilities, obstruction detection, fixed port anchors, and waypoint-based rerouting.
- Refactor diagram construction into reusable utilities and improve GraphQL and AI service port positioning.
- Standardize node border colors and resting-link opacity.
- Add overlap fixtures and regression tests for routing, geometry, node height, and port anchors.


packages/component-diagram/src/utils/diagram.ts
```diff
@@ -167,9 +183,796 @@ export function autoDistribute(engine: DiagramEngine) {
...
+        if (eventIndex === -1) {
+            return center; // workflow nodes have no other row-level ports
+        }
+        return box.top + ENTRY_HEADER_HEIGHT + eventIndex * ENTRY_ROW_HEIGHT + ENTRY_ROW_HEIGHT / 2;
+    }
+
+    const service = node.node as CDService;
+    if (service?.type === "ai:Service" || service?.type === "graphql:Service") {
+        return center;
+    }
```
pasindufernando1, line 307: The premise here doesn't hold for `ai:Service`. `AIServiceWidget` renders `chatFunction`/`decisionFunction` as ordinary `FunctionBox` body rows below the header, each with its own `PortWidget` — they aren't at the node's vertical center, so this branch anchors them wrongly and the obstruction pass skips detours it should make.

Repro: AI service at y=100 with 2 resource functions (height 176, real row-1 center 240, center assumed here 188), workflow box `[200,225]` in the middle column, connection box `[158,222]`. No waypoints are added and the rendered curve penetrates the workflow node by ~12px. An otherwise identical `http:Service` node anchors at 244 and reroutes correctly — so the bug this PR fixes still reproduces on AI agent diagrams.

The `graphql:Service` half is the same class of gap. Worth falling through to the row-index math below for both instead of returning `center`.

Note this is invisible to `linkOverlap.test.ts`: `getLinkGeometry` derives both endpoints from `getLinkAnchors`, i.e. the same approximation being validated, so the suite stays green on `6-ai-agent-complex` and `7-graphql-complex`.

madushajg: Fixed with bc746b2038950536f0a80c83b35b0e9618c2725c

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -167,9 +183,794 @@ export function autoDistribute(engine: DiagramEngine) {
...
+        const eventIndex = events.findIndex((event) => node.getEventPort(event) === port);
+        if (eventIndex === -1) {
+            return center; // workflow nodes have no other row-level ports
+        }
+        return box.top + ENTRY_HEADER_HEIGHT + eventIndex * ENTRY_ROW_HEIGHT + ENTRY_ROW_HEIGHT / 2;
+    }
+
+    const service = node.node as CDService;
+    if (service?.type === "graphql:Service") {
+        return center;
```
rosensilva: The router assumes every GraphQL link starts from the middle of the service box. But each link actually starts from its own function row or group header.
Because it checks the wrong starting point, it can miss a workflow blocking the link. For example, with three visible queries and collapsed Mutation and Subscription groups, a query link still crosses a workflow even though the router thinks the path is clear.
Fix: Use the position of the specific port the link connects to, accounting for expanded and collapsed groups. Alternatively, run routing after the browser has measured the ports.
Update the test to verify that links from these ports avoid workflow nodes, instead of expecting every port to use the service box’s center.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -167,9 +183,794 @@ export function autoDistribute(engine: DiagramEngine) {
...
+        // it would silently invalidate the safety argument above, since an outer segment would
+        // then run between a source/target point and a waypoint on the *wrong* side, spanning
+        // back across the obstruction's own X-range instead of staying outside it.
+        const leftWaypoint = { x: columnLeft - detourX, y: laneY };
+        const rightWaypoint = { x: columnRight + detourX, y: laneY };
+        const [sourceSideWaypoint, targetSideWaypoint] = sourceIsLeft
+            ? [leftWaypoint, rightWaypoint]
+            : [rightWaypoint, leftWaypoint];
+        link.point(sourceSideWaypoint.x, sourceSideWaypoint.y, 1);
+        link.point(targetSideWaypoint.x, targetSideWaypoint.y, 2);
```
rosensilva, line 614: The link’s detour is calculated when the diagram is first arranged. When a user drags a node, the detour stays in its old position.
This can make the link double back and cross the connection box. Moving a workflow can also put it in the path of a previously clear link.
Fix: Recalculate affected link paths when a node moves. Check both links connected to that node and other links that the moved node might block.
let's Add a test that drags a connection or workflow and verifies that the links still avoid node boxes.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -296,20 +1102,15 @@ export const getModelId = (nodeId: string) => {
...
+        return ENTRY_HEADER_HEIGHT + numFunctions * ENTRY_ROW_HEIGHT + ROW_PADDING + ENTRY_VIEW_ALL_BUTTON_HEIGHT;
     }
 
     if (numFunctions <= 2) {
-        return BASE_HEIGHT + numFunctions * FUNCTION_HEIGHT + PADDING;
+        return ENTRY_HEADER_HEIGHT + numFunctions * ENTRY_ROW_HEIGHT + ROW_PADDING;
     }
 
-    return BASE_HEIGHT + 2 * FUNCTION_HEIGHT + PADDING + VIEW_ALL_BUTTON_HEIGHT;
+    return ENTRY_HEADER_HEIGHT + 2 * ENTRY_ROW_HEIGHT + ROW_PADDING + ENTRY_VIEW_ALL_BUTTON_HEIGHT;
```
dulajdilshan: **[Major] Node is 8px short at exactly 3 functions — and this pass now depends on that height.**

| funcs | widget renders | `calculateEntryNodeHeight` | |
|---|---|---|---|
| 1 | 1 row → 128 | 128 | ok |
| 2 | 2 rows → 176 | 176 | ok |
| **3** | 3 rows, no button → **224** | 72+2·48+8+40 = **216** | **−8** |
| 4 | 2 rows + button → 216 | 216 | ok |

At 3, `partitionRegularServiceFunctions` shows all rows (`<= SHOW_ALL_THRESHOLD`, which is 3) and `GeneralWidget:253` hides the button (needs `> 3`) — but `:1109`'s literal `<= 2` still sends it down the preview branch.

Why it bites *this* PR: `getNodeBoundingBox` reads `node.height`, so `polylineCrossesBox` is blind to the bottom 8px of every 3-function service — precisely the overlap this pass exists to remove — and `LINK_DETOUR_MARGIN` (16) is effectively halved there.

`:190` says these constants were extracted "so the two can't drift out of sync". The literal `2` is the drift they don't cover — derive the branch from `SHOW_ALL_THRESHOLD` / `PREVIEW_COUNT` instead?

packages/component-diagram/src/test/linkOverlap.test.ts
```diff
@@ -0,0 +1,66 @@
...
+import model10 from "../stories/10-workflow-function-port-overlap.json";
+import model11 from "../stories/11-stacked-workflows-overlap.json";
+
+/** The imported JSON's inferred shape is narrower than CDModel's optional fields, hence the cast. */
+const fixture = (name: string, model: unknown): [string, CDModel] => [name, model as CDModel];
+
+const fixtures = [
+    fixture("1-empty", model1),
+    fixture("2-only-automation", model2),
+    fixture("3-simple-service", model3),
```
dulajdilshan: **[Medium] No fixture has a 3-function service, so this corpus can't catch the sizing bug above.**

Function counts across all 11 fixtures: 0, 1, 2, 4 — never 3.

And the checker can't catch it independently: `linkOverlapChecker.ts:189` derives its boxes from the same `getNodeBoundingBox`, so a wrong `node.height` makes production and test wrong in step — green means "consistent", not "correct".

(Distinct from CodeRabbit's `:194-196` note — that one is about vacuous passing via skipped links.)

Cheapest fix: bump one fixture's service to 3 functions. Better: a unit test pinning `calculateEntryNodeHeight(n)` against the row count the widget renders for n = 0..5, which locks the invariant `:190` claims.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -167,9 +183,794 @@ export function autoDistribute(engine: DiagramEngine) {
...
+ * unrelated node occupying a column it skips over. This pass finds those links and adds two
+ * waypoints that route them through the nearest free vertical gap in the offending column
+ * instead - above or below whichever node(s) are in the way - rather than through it.
+ *
+ * The check is purely geometric - it doesn't know or care that the "workflow" column is the one
+ * usually in the way - and it asks the question against the curve NodeLinkModel actually paints,
+ * by sampling the very same bezier segments (see sampleBezierPath / getBezierSegments) rather
+ * than against the straight chord between the link's endpoints.
+ *
+ * Testing the chord is what this pass used to do, and it is *not* a safe approximation: with a
```
dulajdilshan: **[Minor] A few comments narrate the PR's own history rather than the code's contract.**

Here: "Testing the chord is what this pass used to do … passed 0.3px clear … entered that node by 4.4px". And `constants.ts:50`: "see the ON_SURFACE/OUTLINE_VARIANT contrast mismatch this replaced".

Excellent PR-description material. In source they describe code a future reader can't see and can't verify, so they rot silently.

The invariants sitting beside them are genuinely good — keep those, move the before/after to the commit message?

packages/component-diagram/src/utils/diagram.ts
madushajg: Fixed with https://github.com/ballerina-platform/ballerina-vscode/pull/1064/commits/830c9d43c03e49e3e88f7e8b1ea9fc1e6c93ec70

packages/component-diagram/src/utils/diagram.ts
madushajg: Fixed differently than suggested with  https://github.com/ballerina-platform/ballerina-vscode/pull/1064/commits/03232903139aae018b4ae1b18bba678fba814e02. Rather than recomputing link routing on every drag (which we tried and found introduces its own correctness issues, e.g. rerouted links crossing other workflow nodes), we removed the CSS rule that was forcing the link layer to render above the node layer. Nodes already have opaque backgrounds, so with normal stacking restored, any link segment that overlaps a node (whether from initial layout or after a drag) now simply renders behind it instead of visibly crossing over it, no rerouting needed at all. This is also how n8n (and most node-graph editors — Blender's shader editor, Unreal Blueprints) handle this: they don't dynamically reroute links around dragged nodes either, they just let links pass behind nodes.

Added `Diagram.test.tsx` snapshot coverage (already exercises the layer stacking) rather than a dedicated drag test, since there's no longer any drag-triggered recompute logic to test — the fix is a static `z-order` change that holds regardless of node position.

packages/component-diagram/src/utils/diagram.ts
madushajg: Fixed with https://github.com/ballerina-platform/ballerina-vscode/pull/1064/commits/6e8dd13f4ff0026dd92b19b8bee39bcc57eeff6f

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -167,9 +183,803 @@ export function autoDistribute(engine: DiagramEngine) {
...
+        // Only ever linked while collapsed, in which case exactly PREVIEW_COUNT rows are
+        // visible above it (see partitionRegularServiceFunctions in Diagram.tsx).
+        return box.top + ENTRY_HEADER_HEIGHT + PREVIEW_COUNT * ENTRY_ROW_HEIGHT + ENTRY_VIEW_ALL_BUTTON_HEIGHT / 2;
+    }
+
+    // A specific function's own port. Ports are added in the same order functions are shown in
+    // (see EntryNodeModel's constructor and partitionRegularServiceFunctions in Diagram.tsx),
+    // and a function is only ever linked via its own port while visible, so its position among
+    // the node's out-ports - after the leading generic "out" port - is exactly its row index.
+    const rowIndex = node.getOutPorts().indexOf(port as NodePortModel) - 1;
```
samithkavishke, line 340: **[HIGH] `getPortAnchorY`'s generic row-index fallback assumes port array order == rendered row order — false for `AIServiceWidget`.**

`rowIndex = node.getOutPorts().indexOf(port) - 1` relies on the comment's claim that "ports are added in the same order functions are shown in." Ports are added in `EntryNodeModel`'s constructor in `service.resourceFunctions` array order (raw source order), but `AIServiceWidget.tsx` (lines 180-185) unconditionally renders the chat resource row before the decision resource row regardless of that array's order:
```tsx
{chatFunction && <FunctionBox func={chatFunction} .../>}
{decisionFunction && <FunctionBox func={decisionFunction} .../>}
```
If a service's `resourceFunctions` lists the decision resource before the chat resource (e.g. that's the order the two `resource function` declarations appear in the `.bal` source), the port added first is the decision port, so this fallback computes decision → row 0 (top) and chat → row 1 (bottom) — the opposite of what's actually rendered. Any link into that port gets anchored to the wrong Y, producing a visibly kinked link or one that cuts through node content, which is exactly the defect class this PR sets out to fix. Not covered by `avoidLinkObstructions.test.ts`/`linkOverlapChecker.ts`, which only exercise generic/GraphQL services.

Suggested fix: give `AIServiceWidget` the same `rowOffsetY`-stamping treatment already used for GraphQL ports (see `computeGraphQLPortOffsets`), rather than relying on array-order matching render order.

packages/component-diagram/src/components/DiagramCanvas.tsx
```diff
@@ -43,10 +43,6 @@ const Container = styled.div<{ color: string; background: string }>`
     background-image: radial-gradient(var(--vscode-editor-inactiveSelectionBackground) 10%, transparent 0px);
     background-size: 16px 16px;
     font-family: "GilmerRegular";
```
samithkavishke, line 54: **[MEDIUM] `& svg:first-child { z-index: 1; }` was removed from `Container` with no replacement.**

This rule forced the first SVG child of the canvas container (the react-diagrams link/overlay SVG layer) to stack above sibling layers. Since `Container`'s direct children are flex items (`display: flex` + `> * { height: 100%; ... }`), they participate in stacking via `z-index` without needing an explicit `position`, so this wasn't dead CSS. Nothing else in the diff re-establishes a stacking rule for that layer. If any other canvas layer (e.g. the overlay/background layer) is also a sibling positioned/flex element, dropping this could let links render underneath other layers in some environments — silently reintroducing exactly the overlap class this PR targets. jsdom-based snapshot tests don't evaluate real CSS paint/stacking, so a regression here wouldn't be caught by the new test suite. Was this rule confirmed unnecessary (e.g. because the SVG layer is now always painted last in DOM order), or was it dropped incidentally during the refactor?

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -167,9 +183,803 @@ export function autoDistribute(engine: DiagramEngine) {
...
+ * to derive them the same way, and deriving them twice is how the two would drift apart.
+ */
+export function getLinkAnchors(link: NodeLinkModel): LinkAnchors | null {
+    const { sourceNode, targetNode } = link;
+    if (!sourceNode || !targetNode || sourceNode === targetNode) {
+        return null;
+    }
+    const sourceBox = getNodeBoundingBox(sourceNode);
+    const targetBox = getNodeBoundingBox(targetNode);
+    const sourceIsLeft = sourceBox.left <= targetBox.left;
```
samithkavishke: **[MEDIUM] `sourceIsLeft` conflates "which node is positioned further left" with "which side a port physically renders on" — the two aren't always the same thing.**

`createNodesLink`/`createPortsLink` always attach via `sourceNode.getOutPort()` and `targetNode.getInPort()`, and those ports are rendered on a *fixed* side regardless of on-canvas position (see `RightPortWidget`/`LeftPortWidget` in `ConnectionNodeWidget.tsx`, and the analogous top/bottom port widgets elsewhere) — the source's out port is always on its own right edge, the target's in port always on its own left edge. Here, though:
```ts
const sourceIsLeft = sourceBox.left <= targetBox.left;
source.x = sourceIsLeft ? sourceBox.right : sourceBox.left;
target.x = sourceIsLeft ? targetBox.left : targetBox.right;
```
if `sourceIsLeft` is ever `false` (source positioned at/right of target), this returns `sourceBox.left`/`targetBox.right` — the *opposite* edges from where the ports actually render — even though the comment above (and the one in `avoidLinkObstructions`, lines 428-431) explicitly claims this "stays correct for a link ever drawn right-to-left." It doesn't, for the reason above.

Given `autoDistribute`'s strictly increasing column layout (listener < entry < workflow < connection) and that every `createNodesLink`/`createPortsLink` call site links strictly forward through those columns, `sourceIsLeft` appears to always be `true` today, so this looks unreachable in practice rather than a live bug. But since the comments assert general correctness the code doesn't actually have, any future change that adds a link type running against that column order (e.g. a workflow-to-workflow back-edge, or a reordered column layout) would silently mis-anchor without any test catching it. Consider anchoring by fixed port side (`sourceBox.right` / `targetBox.left`, always) rather than by relative position, which would make the invariant actually hold rather than coincidentally hold today.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -167,9 +183,803 @@ export function autoDistribute(engine: DiagramEngine) {
...
+        if (eventIndex === -1) {
+            return center; // workflow nodes have no other row-level ports
+        }
+        return box.top + ENTRY_HEADER_HEIGHT + eventIndex * ENTRY_ROW_HEIGHT + ENTRY_ROW_HEIGHT / 2;
+    }
+
+    const service = node.node as CDService;
+    if (service?.type === "graphql:Service") {
+        const offset = (port as NodePortModel).rowOffsetY;
+        return offset === undefined ? center : box.top + offset;
```
samithkavishke: **[LOW] `rowOffsetY` silently falls back to node center if a GraphQL port is ever anchored without going through `buildDiagramData`/`computeGraphQLPortOffsets` first.**

`avoidLinkObstructions`/`getPortAnchorY` are only fed correct GraphQL row anchors because `buildDiagramData` stamps `NodePortModel.rowOffsetY` before `autoDistribute` runs; nothing in the type system enforces that ordering. Any future caller that builds nodes/links directly (a partial/incremental diagram update, a new test harness) would get `rowOffsetY === undefined` for every GraphQL port and silently fall back to `center` here — reintroducing the exact "link cuts through a function row" bug this PR fixes, with no runtime error to flag it. Worth a comment noting the invariant explicitly, or an assertion/dev-mode warning if `rowOffsetY` is unexpectedly missing for a GraphQL port.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -167,9 +183,803 @@ export function autoDistribute(engine: DiagramEngine) {
...
+ * Shared row metrics for the plain entry/workflow body layout: a fixed-height header block,
+ * then a uniform-height row per function/event, optionally followed by a "view all" row (see
+ * `Node`/`Box`/`FunctionBoxWrapper` in `nodes/EntryNode/components/styles.ts` and the row
+ * components in `GeneralWidget.tsx`). `calculateEntryNodeHeight`/`calculateWorkflowNodeHeight`
+ * (which size a node) and `getPortAnchorY` (which locates a specific row's port for link
+ * routing) both derive from these same numbers so the two can't drift out of sync.
+ */
+const ROW_PADDING = 8;
+const ENTRY_HEADER_HEIGHT = 64 + ROW_PADDING;
+const ENTRY_ROW_HEIGHT = 40 + ROW_PADDING;
```
samithkavishke, line 209: **[LOW] `ENTRY_ROW_HEIGHT`/`ENTRY_VIEW_ALL_BUTTON_HEIGHT`/`ROW_PADDING` duplicate the `GQL_FUNCTION_HEIGHT`/`GQL_SHOW_BUTTON_HEIGHT`/`GQL_PADDING` constants introduced later in the same file (~line 1134-1138) with identical numeric values (40+8, 40, 8 respectively).** Both sets are freshly extracted by this PR from previously-inline literals, so this is a new split rather than pre-existing design. If the shared 40px row height is ever tweaked, both `ENTRY_*` and `GQL_*` (and every function keying off them — `getPortAnchorY`, `calculateEntryNodeHeight`, `calculateGraphQLNodeHeight`, `computeGraphQLPortOffsets`) need to be updated in lockstep, or the regular-service and GraphQL layouts silently drift apart in row height while everything still compiles. Consider a single shared `ROW_HEIGHT`/`PADDING`/`SHOW_BUTTON_HEIGHT` set.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -167,9 +183,803 @@ export function autoDistribute(engine: DiagramEngine) {
...
+    const service = node.node as CDService;
+    if (service?.type === "graphql:Service") {
+        const offset = (port as NodePortModel).rowOffsetY;
+        return offset === undefined ? center : box.top + offset;
+    }
+
+    if (port === node.getViewAllResourcesPort()) {
+        // Only ever linked while collapsed, in which case exactly PREVIEW_COUNT rows are
+        // visible above it (see partitionRegularServiceFunctions in Diagram.tsx).
+        return box.top + ENTRY_HEADER_HEIGHT + PREVIEW_COUNT * ENTRY_ROW_HEIGHT + ENTRY_VIEW_ALL_BUTTON_HEIGHT / 2;
```
samithkavishke: **[LOW] The view-all-button anchor hardcodes `PREVIEW_COUNT * ENTRY_ROW_HEIGHT` to locate the row, duplicating the slicing logic that independently lives in `partitionRegularServiceFunctions`.**

The comment here acknowledges the coupling ("exactly `PREVIEW_COUNT` rows are visible above it") but nothing enforces it beyond convention — a future change to how many rows are shown before the view-all button (e.g. a per-node override, or a different threshold for AI/GraphQL variants) would silently desync this routing offset from the actual rendered layout, without a compiler or test catching the drift.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -167,9 +183,803 @@ export function autoDistribute(engine: DiagramEngine) {
...
+        // overlap rather than by "sits strictly between the two columns" is what makes the detour
+        // construction below provably safe.
+        const obstructions = allNodes
+            .filter((node) => node !== link.sourceNode && node !== link.targetNode)
+            .map(getNodeBoundingBox)
+            .filter((box) => box.right > anchorLeft.x && box.left < anchorRight.x);
+
+        // Sampled once and reused for every box below, and again for `naiveY` - it's the same
+        // curve throughout, and sampling is the expensive part of this pass.
+        const polyline = sampleBezierPath([anchorLeft, anchorRight], CURVE_SAMPLES_PER_SEGMENT);
```
samithkavishke, line 513: **[LOW] `sampleBezierPath` (64 cubic-bezier samples) runs unconditionally before checking whether `obstructions` is even non-empty.**

`obstructions` is computed above this line, but the emptiness check (`if (!obstructions.some(...)) return;`) happens *after* `sampleBezierPath` is called, so every link with zero X-overlapping nodes (the common case — most links span adjacent columns with nothing between them) still pays for a full 64-point curve sampling pass it doesn't need. Moving the `obstructions.length === 0` check above the `sampleBezierPath` call would skip this for the majority of links on every `autoDistribute()` call (initial mount + every expand/collapse re-layout).

packages/component-diagram/src/utils/diagram.ts
madushajg: Good catch.  `getPortAnchorY` now resolves AI chat/decision ports by the widget's actual render order instead of assuming it matches `resourceFunctions` array order. Added a swapped-order regression test, verified it fails without the fix. Simpler than the suggested stamping approach since AI's row order isn't UI-state-dependent like GraphQL's.

packages/component-diagram/src/components/DiagramCanvas.tsx
madushajg: This is intentional. Left without an explicit `z-index` on purpose so the node layer (added second) stays above the link layer by default DOM-order stacking: while dragging a node, a link crossing near it shouldn't flash on top (n8n-style). Added a comment recording that intent so it isn't flagged as dropped CSS again.

packages/component-diagram/src/utils/diagram.ts
madushajg: Fixed. Anchors now always use the fixed port side (source's right edge, target's left edge) instead of comparing box positions, which only coincidentally matched today. Simpler than before, and actually correct. Added a regression test, verified it fails without the fix.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -355,6 +1206,52 @@ export const calculateGraphQLNodeHeight = (
...
 };
 
+/**
+ * Computes the Y offset (from the node's top) of each visible GraphQL function row and each
+ * group's header/"show more" row - whichever row a link attached to that group's header port
+ * would actually leave from. Walks the same groups, in the same order and with the same
+ * collapsed/expanded rules `GraphQLServiceWidget` renders them with, so a row's offset here always
+ * matches where it actually draws.
+ */
+function computeGraphQLPortOffsets(
```
rosensilva, line 1273: new calculation adds a 40-pixel “Show More” row whenever more than two functions are visible. The widget shows that button only when the group contains more than three functions.
With exactly three queries, this shifts the calculated Mutation and Subscription ports 40 pixels too far down, so routing still checks the wrong path

Fix: Use visibleItems.length + hiddenItems.length > SHOW_ALL_THRESHOLD for the button condition in both computeGraphQLPortOffsets and calculateGraphQLNodeHeight. Add a three-query case; the new GraphQL tests use two queries.

packages/component-diagram/src/utils/diagram.ts
madushajg: Added a `console.warn` when a GraphQL port's `rowOffsetY` is missing, rather than just a comment ,surfaces the misuse instead of silently degrading. Confirmed it doesn't fire anywhere in the existing suite (55 tests, zero warnings) before adding a test that reproduces the gap directly and checks both the warning and the center fallback.

packages/component-diagram/src/utils/diagram.ts
madushajg: Fixed. `GraphQLServiceWidget` reuses the exact same row components as the regular entry widget, so the matching values weren't coincidental. Removed the duplicate `GQL_PADDING/GQL_FUNCTION_HEIGHT`/`GQL_SHOW_BUTTON_HEIGHT`; GraphQL sizing now reads the same `ROW_PADDING`/`ENTRY_ROW_HEIGHT`/`ENTRY_VIEW_ALL_BUTTON_HEIGHT` constants directly, so they can no longer drift apart. Kept `GQL_BASE_HEIGHT`/`GQL_HEADER_HEIGHT` separate since those are genuinely different values, not duplicates. No new test needed. Pure refactor, confirmed by all snapshots and existing GraphQL tests passing unchanged.

packages/component-diagram/src/utils/diagram.ts
madushajg: Valid. Actually a triplication, not just two places: `calculateEntryNodeHeight` hardcoded the same assumption too, alongside the anchor and the actual slicing in `partitionRegularServiceFunctions`. Extracted `visibleRowCountWhenCollapsed()` as the single source; all three now call it instead of each hardcoding `PREVIEW_COUNT`. No existing test covered this anchor, so added one and confirmed it fails if the count ever desyncs.

packages/component-diagram/src/utils/diagram.ts
madushajg: Moved the empty-obstructions check above the sampling call; identical output, just skips the wasted 64-point sample on the common (no-obstruction) path.

rosensilva (dismissed):

packages/component-diagram/src/utils/diagram.ts
madushajg: At exactly 3 total items in a group, the old formula reserved a 40px button row the widget never renders, shifting every later group's ports down. Applied the suggested fix in both functions, added the 3-query test case, verified it fails with height 447 without the fix.

rosensilva (approved):

PR #1065, opened 2026-09-09
Bump test dependency versions 

LakshanWeerasinghe:
## Purpose
$subject

Updates Ballerina dependency versions for the `staging/5.14.0` branch.

- Refreshes language-server dependency lock files.
- Adds the `ballerina/udp` dependency.
- Updates dependency lists for `graphql`, `oracledb`, and `solace`.
- Updates `stdlibGraphqlVersion` to `1.18.1`.


NipunaRanasinghe (approved):
LGTM

packages/ballerina-language-server/gradle.properties
```diff
@@ -83,7 +83,7 @@ stdlibSoapVersion=2.3.1
 stdlibWebsubVersion=2.15.0
 
 # Stdlib Level 07
-stdlibGraphqlVersion=1.17.0
+stdlibGraphqlVersion=1.18.1
```
pasindufernando1, line 86: This change is not needed. Just bumping the version in the ballerina_dependencies is required. gradle.properties values refer to the packed versions in the distribution which is 1.17.0. So lets remove this change

LakshanWeerasinghe: This was needed because of a BIR failure

pasindufernando1 (changes requested):

packages/ballerina-language-server/gradle.properties
pasindufernando1: There is no harm in bumping as already the custom home will bear 18.1 which is a newer version. Lets keep this for the moment. Will look at the BIR failure cause later

PR #1067, opened 2026-09-09
Rename WSO2 Integration Intelligence back to WSO2 Integrator Copilot

RNViththagan:
Related to https://github.com/wso2/product-integrator/issues/2397
Related to https://github.com/wso2/product-integrator/issues/2399

- Rename the AI assistant from "WSO2 Integration Intelligence" (WII/Wii) back to "WSO2 Integrator Copilot" across every user-facing string — AI panel, floating orb, mini chat, migration panel, status bar, error messages, command titles, and the agent's own system prompt (WII/Wii identity removed).
- Restore the original AI-panel welcome subtitle ("Build integrations faster with AI. Describe what you need and get working integrations instantly.").
- Hide the editor-toolbar "Open WSO2 Integrator Copilot" button while the chat panel is already open, so it isn't a redundant entry point into the same chat (#2399) — new `ballerina.aiPanelOpen` context key.

- Renamed user-facing AI assistant references to “WSO2 Integrator Copilot” across the extension, panels, status bar, prompts, errors, and visualizer.
- Restored the AI panel welcome subtitle.
- Added panel context keys to hide the toolbar command while the AI panel is open or visible.
- Added tests for `ballerina.aiPanelVisible` context updates.

### DIFF
--- a/packages/ballerina-extension/package.json
+++ b/packages/ballerina-extension/package.json
@@ -1157,7 +1157,7 @@
                 {
                     "command": "ballerina.open.ai.panel",
                     "group": "navigation@1",
-                    "when": "resourceLangId == ballerina || isBallerinaDiagram || (isBalVisualizerActive && isBIProject && !ballerina.copilotAmbientPresent)",
+                    "when": "(resourceLangId == ballerina || isBallerinaDiagram || (isBalVisualizerActive && isBIProject && !ballerina.copilotAmbientPresent)) && !ballerina.aiPanelVisible",
                     "title": "Open WSO2 Integrator Copilot"
                 },
                 {
--- a/packages/ballerina-extension/package.json
+++ b/packages/ballerina-extension/package.json
@@ -223,7 +223,7 @@
                         "Simple gradient orb without a ring."
                     ],
                     "default": "animated",
-                    "description": "WSO2 Integration Intelligence orb style."
+                    "description": "WSO2 Integrator Copilot orb style."
                 },
                 "ballerina.enableNotebookDebug": {
                     "type": "boolean",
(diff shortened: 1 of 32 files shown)


packages/ballerina-extension/package.json
```diff
@@ -1157,8 +1157,8 @@
                 {
                     "command": "ballerina.open.ai.panel",
                     "group": "navigation@1",
-                    "when": "resourceLangId == ballerina || isBallerinaDiagram || (isBalVisualizerActive && isBIProject && !ballerina.copilotAmbientPresent)",
-                    "title": "Open WSO2 Integration Intelligence"
+                    "when": "(resourceLangId == ballerina || isBallerinaDiagram || (isBalVisualizerActive && isBIProject && !ballerina.copilotAmbientPresent)) && !ballerina.aiPanelOpen",
```
madushajg: `aiPanelOpen` means the panel is alive, not on screen — the panel opens ViewColumn.Beside and is routinely buried behind another tab in the side group. When that happens this button disappears even though `ballerina.open.ai.panel` would do the useful thing (`openAIWebview` → `reveal()`), and the status bar item is hidden too while the state is idle, leaving only the command palette. Consider driving a `ballerina.aiPanelVisible` key from `setAiPanelVisible()` instead, which is already tracked.

RNViththagan: Good catch — switched to `ballerina.aiPanelVisible` driven from `setAiPanelVisible()`, so the button hides only while the panel is actually on screen and still shows (to reveal it) when it's buried behind another tab. Added a small test for the context-key transition.

madushajg (approved):

NipunaRanasinghe (approved):
@RNViththagan we need an approval from the AI team before merging

shafreenAnfar (approved):

PR #1071, opened 2026-09-10
Render the connection init form in parameter declaration order

pasindufernando1:
## Purpose

Fixes wso2/product-integrator#2402 — the connector configuration panel rendered its fields alphabetically instead of in `init` declaration order. Pinecone's **Create Vector Store** showed `API Key` above `Service URL`, and `Query Mode` below `Pinecone Configuration` / `HTTP Configuration`.

## Cause

The LS sends `getNodeTemplate` properties in declaration order, and `Form` renders `formFields` as given, so the order was lost in the conversion. `ConnectionCreator` used `convertConfig`, which sorts keys by default (`node-property-utils.ts:478`). Sorted Pinecone keys are `apiKey, config, httpConfig, queryMode, serviceUrl`; `Form`'s two passes (plain, then `advanced`) split that into exactly the reported order.

The other connector path — `AddConnectionWizard` → `FlowNodeForm` → `convertNodePropertiesToFormFields` — never sorted, so the same `init` read differently depending on which panel opened it. Not Pinecone-specific: every connection kind created through this overlay was affected.

## Changes

- `ConnectionSelector/connectionFormFields.ts` (new) — `convertConnectionConfig()`, order-preserving.
- `ConnectionCreator` uses it. Covers create and edit overlays, and every connection kind (connector, vector store, model/embedding provider, chunker, memory store).
- Dropped the sort from the unreferenced `ballerina-side-panel/src/utils/convertConfig.ts`, whose doc comment claims this overlay, so wiring it up cannot reintroduce the bug.

`convertConfig`'s alphabetical default is unchanged — `FunctionForm`, `AgentToolForm` and `AIAgentSidePanel` lay themselves out around it, and flipping it would relayout five unrelated forms.

## Checks

- New L1 invariant test over captured `getNodeTemplate` payloads (Pinecone `VectorStore`, MySQL `Client`); 4 of its 6 cases fail against the pre-fix call.
- Panel now reads `Service URL, API Key`, then advanced `Query Mode, Pinecone Configuration, HTTP Configuration`.
- `ballerina-visualizer` 561 tests, `ballerina-side-panel` 222 tests, `scripts/typecheck-tests.js` clean; `tsc --noEmit` output identical to baseline.

- Preserve connection form field order based on `init` parameter declarations.
- Apply the order-preserving conversion to create and edit overlays for all connection types.
- Keep alphabetical ordering for unrelated forms.
- Add connection fixtures and tests for declaration-order behavior.

### DIFF
--- a/packages/ballerina-side-panel/src/utils/convertConfig.ts
+++ b/packages/ballerina-side-panel/src/utils/convertConfig.ts
@@ -28,7 +28,7 @@ import { FormField } from "../components/Form/types";
  */
 export function convertConfig(properties: NodeProperties, skipKeys: string[] = []): FormField[] {
     const formFields: FormField[] = [];
-    for (const key of Object.keys(properties).sort()) {
+    for (const key of Object.keys(properties)) {
         if (skipKeys.includes(key)) {
             continue;
         }
--- a/packages/ballerina-visualizer/src/components/ConnectionSelector/ConnectionCreator.tsx
+++ b/packages/ballerina-visualizer/src/components/ConnectionSelector/ConnectionCreator.tsx
@@ -20,12 +20,12 @@ import { useCallback, useContext, useEffect, useMemo, useRef, useState } from "r
 import { FormField, FormValues, FormImports } from "@wso2/ballerina-side-panel";
 import { useRpcContext } from "@wso2/ballerina-rpc-client";
 import { PanelOverlayContext } from "../../views/BI/FlowDiagram/context/PanelOverlayContext";
-import { convertConfig } from "../../utils/bi";
 import { ArtifactForm } from "../../views/BI/Forms/ArtifactForm";
 import { RelativeLoader } from "../RelativeLoader";
 import { InfoBox } from "../InfoBox";
 import { ConnectionCreatorProps } from "./types";
 import { getConnectionSpecialConfig } from "./config";
+import { convertConnectionConfig } from "./connectionFormFields";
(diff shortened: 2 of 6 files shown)


packages/ballerina-side-panel/src/utils/convertConfig.ts
```diff
@@ -28,7 +28,7 @@ import { FormField } from "../components/Form/types";
  */
 export function convertConfig(properties: NodeProperties, skipKeys: string[] = []): FormField[] {
     const formFields: FormField[] = [];
-    for (const key of Object.keys(properties).sort()) {
+    for (const key of Object.keys(properties)) { // LS order is the `init` declaration order (#2402)
```
NipunaRanasinghe: comment is redundant IMO

pasindufernando1: Removed, along with the same comment in `connectionFormFields.ts`. The rationale is in the commit message.

pasindufernando1: Fixed

NipunaRanasinghe (approved):

packages/ballerina-side-panel/src/utils/convertConfig.ts
pasindufernando1: Fixed

madushajg (approved):

PR #1073, opened 2026-09-10
Add visible `Open WSO2 Integrator Copilot` link when panel is closed

VellummyilumVinoth:
## Summary
- The Copilot orb icon shown in the Package/Integration Overview empty state during an active run isn't perceived as clickable from the user's point of view — it just looks like a status indicator.
- Adds a clearly clickable "Open WSO2 Integrator Copilot" link under the status text ("Working on it…"), shown only after the Copilot panel has been closed, so users have an obvious way back in.
- Reuses the existing `showOpenCopilot` flag and `openCopilotPanel` helper already wired to the orb button — no new state, RPC, or extension-host changes needed.

It resolves: https://github.com/wso2/product-integrator/issues/2416

Reference: 

https://github.com/user-attachments/assets/45ceb106-04ee-4cb8-af54-a281c0de74e2

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/packages/ballerina-visualizer/src/views/BI/PackageOverview/CopilotComposer.tsx
+++ b/packages/ballerina-visualizer/src/views/BI/PackageOverview/CopilotComposer.tsx
@@ -19,6 +19,7 @@
 import { CSSProperties, useEffect, useLayoutEffect, useRef, useState } from "react";
 import { keyframes } from "@emotion/react";
 import styled from "@emotion/styled";
+import { VSCodeLink } from "@vscode/webview-ui-toolkit/react";
 import { AgentRunStatus, AttachmentStatus, SHARED_COMMANDS } from "@wso2/ballerina-core";
 import { useRpcContext } from "@wso2/ballerina-rpc-client";
 import { Button, Icon, ThemeColors } from "@wso2/ui-toolkit";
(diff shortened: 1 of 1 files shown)


packages/ballerina-visualizer/src/views/BI/PackageOverview/CopilotComposer.tsx
```diff
@@ -520,6 +521,11 @@ export function CopilotComposer({ onAddArtifactManually, hiding }: CopilotCompos
                     <Heading>{runHeading}</Heading>
                     {runDetail && <Subtitle $visible={!aiPanelOpen}>{runDetail}</Subtitle>}
                     {submittedPrompt && <PromptEcho>{submittedPrompt}</PromptEcho>}
+                    {showOpenCopilot && (
```
RNViththagan: There's a narrow window right after Send where the RPC has resolved (so the composer flips to run mode with showOpenCopilot true) but the extension side hasn't actually finished opening the panel yet (openAIWebview isn't awaited on the extension side, and there's a tab-cleanup step before the panel object exists). Clicking the link in that window fires a second OPEN_AI_PANEL while currentPanel is still undefined, so both calls pass the guard and you get two panels. Pre-existing timing gap, not introduced by this PR, but this makes it clickable where before it was just a static status. Would gating on something like `working && showOpenCopilot` (run actually reported by the host) rather than showOpenCopilot alone be enough to close that window on the UI side, or is that not worth it given how narrow it is?

packages/ballerina-visualizer/src/views/BI/PackageOverview/CopilotComposer.tsx
```diff
@@ -520,6 +521,11 @@ export function CopilotComposer({ onAddArtifactManually, hiding }: CopilotCompos
...
+                        <VSCodeLink onClick={() => openCopilotPanel(rpcClient)} style={{ marginTop: 8 }}>
```
RNViththagan: Minor: VSCodeLink with no href doesn't get a real anchor (no href means it's out of the tab order, no link role), so this is mouse-only - matches most other VSCodeLink usages in this codebase, so not a new pattern, just flagging since a couple of the recent commits on this same page were specifically about keyboard accessibility. Not blocking either way.

RNViththagan (approved):
LGTM

SasinduDilshara (approved):

PR #1076, opened 2026-09-10
Add EU region support for Copilot backend URL routing

kalaiyarasiganeshalingam:
## Purpose
Introduce dynamic backend URL switching based on the user's Choreo region. BACKEND_URL, DEVANT_TOKEN_EXCHANGE_URL, and OLD_BACKEND_URL are now mutable and updated at login time via setBackendRegion(), which resolves the correct endpoint from a region-keyed map covering US and EU prod/dev/stage environments.

Adds EU backend routing for Copilot production, development, and staging environments. The extension selects regional endpoints at login through `setBackendRegion()` and updates `BACKEND_URL`, `DEVANT_TOKEN_EXCHANGE_URL`, and `OLD_BACKEND_URL`. Authentication state persists the selected region, restores it after restart, and refreshes the client when the backend URL changes. Authentication and token refresh continue when platform-extension state lookup fails.

Fixes https://github.com/wso2/product-integrator/issues/1890

### DIFF
--- a/packages/ballerina-extension/.env.example
+++ b/packages/ballerina-extension/.env.example
@@ -1,2 +1,7 @@
 COPILOT_ROOT_URL=https://7eff1239-64bb-4663-b256-30a00d187a5c-prod.e1-us-east-azure.choreoapis.dev/copilot
 COPILOT_DEV_ROOT_URL=https://7eff1239-64bb-4663-b256-30a00d187a5c-dev.e1-us-east-azure.choreoapis.dev/copilot
+# COPILOT_STAGE_ROOT_URL=https://7eff1239-64bb-4663-b256-30a00d187a5c-staging.e1-us-east-azure.choreoapis.dev/copilot
+
+COPILOT_EU_ROOT_URL=https://366b8031-ab7f-4a98-a1bd-3369fbd00eee-prod.e1-eu-west-cdp.choreoapis.dev/copilot
+# COPILOT_EU_DEV_ROOT_URL=
+# COPILOT_EU_STAGE_ROOT_URL=
--- a/packages/ballerina-extension/.env.example
+++ b/packages/ballerina-extension/.env.example
@@ -1,6 +1,6 @@
 COPILOT_ROOT_URL=https://7eff1239-64bb-4663-b256-30a00d187a5c-prod.e1-us-east-azure.choreoapis.dev/copilot
 COPILOT_DEV_ROOT_URL=https://7eff1239-64bb-4663-b256-30a00d187a5c-dev.e1-us-east-azure.choreoapis.dev/copilot
-# COPILOT_STAGE_ROOT_URL=https://7eff1239-64bb-4663-b256-30a00d187a5c-staging.e1-us-east-azure.choreoapis.dev/copilot
+# COPILOT_STAGE_ROOT_URL=
 
 COPILOT_EU_ROOT_URL=https://366b8031-ab7f-4a98-a1bd-3369fbd00eee-prod.e1-eu-west-cdp.choreoapis.dev/copilot
 # COPILOT_EU_DEV_ROOT_URL=
--- a/packages/ballerina-extension/src/features/ai/utils.ts
+++ b/packages/ballerina-extension/src/features/ai/utils.ts
@@ -64,12 +64,20 @@ export let DEVANT_TOKEN_EXCHANGE_URL: string = _defaultBackendUrl + "/auth-api/v
 export let OLD_BACKEND_URL: string = _defaultBackendUrl + "/v2.0";
 
 export const setBackendRegion = (region: string): void => {
+    if (config.get('rootUrl')) {
+        return;
+    }
     const normalized = region?.trim().toLowerCase();
     const key = devantEnv ? `${normalized}-${devantEnv}` : normalized;
-    const regionalUrl = COPILOT_ROOT_URLS.get(key) || COPILOT_ROOT_URLS.get(normalized);
+    const regionalUrl = COPILOT_ROOT_URLS.get(key);
+    if (!regionalUrl) {
+        console.error(`No backend URL configured for region '${normalized}'`);
+        return;
+    }
     BACKEND_URL = regionalUrl;
     DEVANT_TOKEN_EXCHANGE_URL = regionalUrl + "/auth-api/v1.0/auth/token-exchange";
     OLD_BACKEND_URL = regionalUrl + "/v2.0";
+    console.log(`[Region] ${region} → BACKEND_URL: ${BACKEND_URL}`);
 };
 
 export async function closeAllBallerinaFiles(dirPath: string): Promise<void> {
--- a/packages/ballerina-extension/src/features/ai/utils.ts
+++ b/packages/ballerina-extension/src/features/ai/utils.ts
@@ -45,15 +45,35 @@ const PLATFORM_ENV_SETTING = "WSO2.WSO2-Platform.Advanced.ChoreoEnvironment";
 const devantEnv = (process.env.CHOREO_ENV || process.env.CLOUD_ENV
     || workspace.getConfiguration().get<string>(PLATFORM_ENV_SETTING) || "").trim().toLowerCase();
 const COPILOT_ROOT_URLS = new Map<string, string>([
-    ["dev", process.env.COPILOT_DEV_ROOT_URL],
-    ["stage", process.env.COPILOT_STAGE_ROOT_URL || process.env.COPILOT_DEV_ROOT_URL],
+    ["us", process.env.COPILOT_ROOT_URL],
+    ["us-dev", process.env.COPILOT_DEV_ROOT_URL],
+    ["us-stage", process.env.COPILOT_STAGE_ROOT_URL || process.env.COPILOT_DEV_ROOT_URL],
+    ["eu", process.env.COPILOT_EU_ROOT_URL],
+    ["eu-dev", process.env.COPILOT_EU_DEV_ROOT_URL || process.env.COPILOT_EU_ROOT_URL],
+    ["eu-stage", process.env.COPILOT_EU_STAGE_ROOT_URL || process.env.COPILOT_EU_DEV_ROOT_URL || process.env.COPILOT_EU_ROOT_URL],
 ]);
-export const BACKEND_URL: string = config.get('rootUrl') || COPILOT_ROOT_URLS.get(devantEnv) || process.env.COPILOT_ROOT_URL;
+const _defaultBackendUrl: string = config.get('rootUrl') || COPILOT_ROOT_URLS.get(devantEnv) || process.env.COPILOT_ROOT_URL;
 
-export const DEVANT_TOKEN_EXCHANGE_URL: string = BACKEND_URL + "/auth-api/v1.0/auth/token-exchange";
+export let BACKEND_URL: string = _defaultBackendUrl;
+
+export let DEVANT_TOKEN_EXCHANGE_URL: string = _defaultBackendUrl + "/auth-api/v1.0/auth/token-exchange";
 
 // This refers to old backend before FE Migration. We need to eventually remove this.
-export const OLD_BACKEND_URL: string = BACKEND_URL + "/v2.0";
+export let OLD_BACKEND_URL: string = _defaultBackendUrl + "/v2.0";
+
+export const setBackendRegion = (region: string): void => {
+    const normalized = region?.trim().toLowerCase();
+    const key = devantEnv ? `${normalized}-${devantEnv}` : normalized;
+    const regionalUrl = COPILOT_ROOT_URLS.get(key) || COPILOT_ROOT_URLS.get(normalized);
+    if (regionalUrl) {
+        BACKEND_URL = regionalUrl;
+        DEVANT_TOKEN_EXCHANGE_URL = regionalUrl + "/auth-api/v1.0/auth/token-exchange";
+        OLD_BACKEND_URL = regionalUrl + "/v2.0";
+        console.log(`[Region] ${region} → BACKEND_URL: ${BACKEND_URL}`);
+    } else {
+        console.log(`[Region] ${region} → no URL found, keeping current: ${BACKEND_URL}`);
+    }
+};
 
 export async function closeAllBallerinaFiles(dirPath: string): Promise<void> {
     // Check if the directory exists
--- a/packages/ballerina-extension/src/features/ai/utils.ts
+++ b/packages/ballerina-extension/src/features/ai/utils.ts
@@ -46,9 +46,11 @@ const devantEnv = (process.env.CHOREO_ENV || process.env.CLOUD_ENV
     || workspace.getConfiguration().get<string>(PLATFORM_ENV_SETTING) || "").trim().toLowerCase();
 const COPILOT_ROOT_URLS = new Map<string, string>([
     ["us", process.env.COPILOT_ROOT_URL],
+    ["us-prod", process.env.COPILOT_ROOT_URL],
     ["us-dev", process.env.COPILOT_DEV_ROOT_URL],
     ["us-stage", process.env.COPILOT_STAGE_ROOT_URL || process.env.COPILOT_DEV_ROOT_URL],
     ["eu", process.env.COPILOT_EU_ROOT_URL],
+    ["eu-prod", process.env.COPILOT_EU_ROOT_URL],
     ["eu-dev", process.env.COPILOT_EU_DEV_ROOT_URL],
     ["eu-stage", process.env.COPILOT_EU_STAGE_ROOT_URL || process.env.COPILOT_EU_DEV_ROOT_URL],
 ]);
@@ -63,21 +65,22 @@ export let DEVANT_TOKEN_EXCHANGE_URL: string = _defaultBackendUrl + "/auth-api/v
 // This refers to old backend before FE Migration. We need to eventually remove this.
 export let OLD_BACKEND_URL: string = _defaultBackendUrl + "/v2.0";
 
-export const setBackendRegion = (region: string): void => {
+export const setBackendRegion = (region: string): boolean => {
     if (config.get('rootUrl')) {
-        return;
+        return true;
     }
     const normalized = region?.trim().toLowerCase();
     const key = devantEnv ? `${normalized}-${devantEnv}` : normalized;
-    const regionalUrl = COPILOT_ROOT_URLS.get(key);
+    const regionalUrl = COPILOT_ROOT_URLS.get(key) || (devantEnv ? COPILOT_ROOT_URLS.get(normalized) : undefined);
     if (!regionalUrl) {
         console.error(`No backend URL configured for region '${normalized}'`);
-        return;
+        return false;
     }
     BACKEND_URL = regionalUrl;
     DEVANT_TOKEN_EXCHANGE_URL = regionalUrl + "/auth-api/v1.0/auth/token-exchange";
     OLD_BACKEND_URL = regionalUrl + "/v2.0";
     console.log(`[Region] ${region} → BACKEND_URL: ${BACKEND_URL}`);
+    return true;
 };
 
 export async function closeAllBallerinaFiles(dirPath: string): Promise<void> {
--- a/packages/ballerina-extension/src/features/ai/utils.ts
+++ b/packages/ballerina-extension/src/features/ai/utils.ts
@@ -49,10 +49,12 @@ const COPILOT_ROOT_URLS = new Map<string, string>([
     ["us-dev", process.env.COPILOT_DEV_ROOT_URL],
     ["us-stage", process.env.COPILOT_STAGE_ROOT_URL || process.env.COPILOT_DEV_ROOT_URL],
     ["eu", process.env.COPILOT_EU_ROOT_URL],
-    ["eu-dev", process.env.COPILOT_EU_DEV_ROOT_URL || process.env.COPILOT_EU_ROOT_URL],
-    ["eu-stage", process.env.COPILOT_EU_STAGE_ROOT_URL || process.env.COPILOT_EU_DEV_ROOT_URL || process.env.COPILOT_EU_ROOT_URL],
+    ["eu-dev", process.env.COPILOT_EU_DEV_ROOT_URL],
+    ["eu-stage", process.env.COPILOT_EU_STAGE_ROOT_URL || process.env.COPILOT_EU_DEV_ROOT_URL],
 ]);
-const _defaultBackendUrl: string = config.get('rootUrl') || COPILOT_ROOT_URLS.get(devantEnv) || process.env.COPILOT_ROOT_URL;
+
+const defaultRegionKey = devantEnv ? `us-${devantEnv}` : "us";
+const _defaultBackendUrl: string = config.get('rootUrl') || COPILOT_ROOT_URLS.get(defaultRegionKey) || process.env.COPILOT_ROOT_URL;
 
 export let BACKEND_URL: string = _defaultBackendUrl;
 
@@ -65,14 +67,9 @@ export const setBackendRegion = (region: string): void => {
     const normalized = region?.trim().toLowerCase();
     const key = devantEnv ? `${normalized}-${devantEnv}` : normalized;
     const regionalUrl = COPILOT_ROOT_URLS.get(key) || COPILOT_ROOT_URLS.get(normalized);
-    if (regionalUrl) {
-        BACKEND_URL = regionalUrl;
-        DEVANT_TOKEN_EXCHANGE_URL = regionalUrl + "/auth-api/v1.0/auth/token-exchange";
-        OLD_BACKEND_URL = regionalUrl + "/v2.0";
-        console.log(`[Region] ${region} → BACKEND_URL: ${BACKEND_URL}`);
-    } else {
-        console.log(`[Region] ${region} → no URL found, keeping current: ${BACKEND_URL}`);
-    }
+    BACKEND_URL = regionalUrl;
+    DEVANT_TOKEN_EXCHANGE_URL = regionalUrl + "/auth-api/v1.0/auth/token-exchange";
+    OLD_BACKEND_URL = regionalUrl + "/v2.0";
 };
 
 export async function closeAllBallerinaFiles(dirPath: string): Promise<void> {
--- a/packages/ballerina-extension/src/features/ai/utils/ai-client.ts
+++ b/packages/ballerina-extension/src/features/ai/utils/ai-client.ts
@@ -60,6 +60,7 @@ export function getBedrockRegionalPrefix(region: string): string {
 
 let cachedAnthropic: ReturnType<typeof createAnthropic> | null = null;
 let cachedAuthMethod: LoginMethod | null = null;
+let cachedBaseUrl: string | null = null;
 
 /**
  * Reusable fetch function that handles authentication with token refresh.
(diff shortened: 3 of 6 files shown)


packages/ballerina-extension/src/utils/ai/auth.ts
```diff
@@ -156,6 +156,10 @@ export const getPlatformStsToken = async (): Promise<string | undefined> => {
         if (!api) {
             return undefined;
         }
+        if (api.isLoggedIn()) {
+            const { region } = api.getAuthState();
+            setBackendRegion(region.toLowerCase());
```
RNViththagan: This only runs inside getPlatformStsToken, and checkToken (views/ai-panel/utils.ts) returns early with the stored credentials whenever they exist, before ever reaching this call - so on a warm restart (Reload Window, or just reopening VS Code) with a still-valid stored token, the region is never set this session; BACKEND_URL stays at the module's default (US). It only gets corrected later if something forces a token refresh (expiry, or a 401). For an EU user that means: reload, open the panel, send a message - the request goes to the US backend, either working silently cross-region (defeats the point of having an EU endpoint) or 401ing, triggering a refresh that sets EU correctly but retries against the same cached US client (see the other comment), 401s again, and logs them out.

Would persisting the region alongside the stored credentials (read back here instead of only setting it after a fresh exchange) close this gap, or is there a simpler way to make sure this runs on every activation?

packages/ballerina-extension/src/features/ai/utils.ts
```diff
@@ -45,15 +45,32 @@ const PLATFORM_ENV_SETTING = "WSO2.WSO2-Platform.Advanced.ChoreoEnvironment";
...
+
+export let BACKEND_URL: string = _defaultBackendUrl;
+
+export let DEVANT_TOKEN_EXCHANGE_URL: string = _defaultBackendUrl + "/auth-api/v1.0/auth/token-exchange";
 
 // This refers to old backend before FE Migration. We need to eventually remove this.
-export const OLD_BACKEND_URL: string = BACKEND_URL + "/v2.0";
+export let OLD_BACKEND_URL: string = _defaultBackendUrl + "/v2.0";
+
+export const setBackendRegion = (region: string): void => {
```
RNViththagan: Even when this does run, it doesn't reach the actual generation traffic. ai-client.ts's getAnthropicClient caches the Anthropic client and only rebuilds it when the login method changes (`!cachedAnthropic || cachedAuthMethod !== loginMethod`) - the baseURL is computed from BACKEND_URL and baked into createAnthropic(...) once, and that guard never checks whether the URL/region changed. So: log in as US, send one message (client cached with the US URL), then switch to EU (log out, log in as EU, or just let a later refresh update BACKEND_URL) - the cached client keeps using the US URL for the rest of the session, silently.

Shall we track the URL the cached client was built with and rebuild when it no longer matches BACKEND_URL, alongside the login-method check?

packages/ballerina-extension/src/features/ai/utils.ts
```diff
@@ -45,15 +45,32 @@ const PLATFORM_ENV_SETTING = "WSO2.WSO2-Platform.Advanced.ChoreoEnvironment";
...
+    const normalized = region?.trim().toLowerCase();
+    const key = devantEnv ? `${normalized}-${devantEnv}` : normalized;
+    const regionalUrl = COPILOT_ROOT_URLS.get(key) || COPILOT_ROOT_URLS.get(normalized);
+    BACKEND_URL = regionalUrl;
```
RNViththagan: Small one: config.get('rootUrl') is only honoured for the initial value now - once this runs (first login or any later refresh), it unconditionally overwrites BACKEND_URL/DEVANT_TOKEN_EXCHANGE_URL/OLD_BACKEND_URL with the regional URL, silently discarding the override. Before this PR the override always won. If someone's using ballerina.rootUrl to point at a local/dev backend, it'll work until the first refresh and then quietly switch to prod. Worth an early return here when the config override is set?

packages/ballerina-extension/.env.example
```diff
@@ -1,2 +1,7 @@
 COPILOT_ROOT_URL=https://7eff1239-64bb-4663-b256-30a00d187a5c-prod.e1-us-east-azure.choreoapis.dev/copilot
 COPILOT_DEV_ROOT_URL=https://7eff1239-64bb-4663-b256-30a00d187a5c-dev.e1-us-east-azure.choreoapis.dev/copilot
+# COPILOT_STAGE_ROOT_URL=https://7eff1239-64bb-4663-b256-30a00d187a5c-staging.e1-us-east-azure.choreoapis.dev/copilot
+
+COPILOT_EU_ROOT_URL=https://366b8031-ab7f-4a98-a1bd-3369fbd00eee-prod.e1-eu-west-cdp.choreoapis.dev/copilot
+# COPILOT_EU_DEV_ROOT_URL=
```
RNViththagan: Since these are unset by default, an EU login with ChoreoEnvironment=dev resolves eu-dev -> misses -> falls back to plain eu, which is EU production - so a dev-mode EU tester silently hits prod instead of dev/staging (the US side doesn't have this gap since us-dev/us-stage are both set). Given that's real traffic against a production backend during dev testing, would a warning (or failing loudly) when the env-specific key is missing be worth adding, same as how an unset one currently just falls through quietly?

packages/ballerina-extension/src/features/ai/utils.ts
```diff
@@ -45,15 +45,32 @@ const PLATFORM_ENV_SETTING = "WSO2.WSO2-Platform.Advanced.ChoreoEnvironment";
...
+export let DEVANT_TOKEN_EXCHANGE_URL: string = _defaultBackendUrl + "/auth-api/v1.0/auth/token-exchange";
 
 // This refers to old backend before FE Migration. We need to eventually remove this.
-export const OLD_BACKEND_URL: string = BACKEND_URL + "/v2.0";
+export let OLD_BACKEND_URL: string = _defaultBackendUrl + "/v2.0";
+
+export const setBackendRegion = (region: string): void => {
+    const normalized = region?.trim().toLowerCase();
+    const key = devantEnv ? `${normalized}-${devantEnv}` : normalized;
+    const regionalUrl = COPILOT_ROOT_URLS.get(key) || COPILOT_ROOT_URLS.get(normalized);
```
RNViththagan: If both lookups miss, regionalUrl is undefined, and the next two lines concatenate that into `"undefined/auth-api/v1.0/auth/token-exchange"` and `"undefined/v2.0"` - syntactically valid-looking but garbage URLs, so the failure shows up as a confusing network/auth error rather than something obviously about a missing config. The realistic way to hit this: any existing local .env that predates this PR and doesn't have COPILOT_EU_ROOT_URL set yet - an EU login on that build silently produces this. Worth a guard (fall back to a default region, or throw something clearer) when both lookups miss?

packages/ballerina-extension/src/features/ai/utils.ts
```diff
@@ -45,15 +45,32 @@ const PLATFORM_ENV_SETTING = "WSO2.WSO2-Platform.Advanced.ChoreoEnvironment";
...
+    BACKEND_URL = regionalUrl;
+    DEVANT_TOKEN_EXCHANGE_URL = regionalUrl + "/auth-api/v1.0/auth/token-exchange";
+    OLD_BACKEND_URL = regionalUrl + "/v2.0";
```
madushajg, line 79: `regionalUrl` can be undefined (missing `COPILOT_EU_*` env var, or an unrecognized region), and with `strictNullChecks: false` this compiles fine, it would silently set `BACKEND_URL` to undefined and every AI request becomes `fetch("undefined/llm/...")` with no way to recover in-session. Please bail out early: 

```tsx
if (!regionalUrl) {
    console.error(`No backend URL configured for region '${normalized}'`);
    return;
}
``` 

before mutating the three exports.

packages/ballerina-extension/src/utils/ai/auth.ts
```diff
@@ -156,6 +156,10 @@ export const getPlatformStsToken = async (): Promise<string | undefined> => {
...
+        }
```
madushajg, line 168: `getAuthState()` can throw (see the try/catch around it in `aiMachine.ts:640`) and region can be undefined on older platform-extension versions. `region.toLowerCase()` then throws a `TypeError` that this function's catch swallows, returning undefined and breaking Copilot auth for all users, including US. Please make region resolution non-fatal: 

```tsx
try {
    const region = api.getAuthState()?.region;
    if (region) {
        setBackendRegion(region.toLowerCase());
    }
} catch {
    /* keep default region */
}
```

packages/ballerina-extension/src/features/ai/utils.ts
madushajg, line 77: When `devantEnv` is `dev`/`stage` and region is `eu`, the `eu-dev`/`eu-stage` keys are unset, so this fallback silently resolves to `COPILOT_EU_ROOT_URL`, the EU production endpoint from a `dev`/`stage` build. The fallback should stay within the same environment (or refuse to switch and keep the current URL) rather than crossing into prod.

RNViththagan (commented):
Thanks for adding this - EU routing is a real need. I think there's a gap between what this fixes and what it needs to fix though: I traced the actual call paths and the region switch doesn't reliably reach the traffic that matters (the chat/generation requests), only the login/token-exchange call. I'd hold off merging until the two items below are addressed, the other three are smaller but worth a look too. Also +1 to CodeRabbit's still-open test-coverage comment.

RNViththagan (commented):
EU routing is a good addition, but looks like it only covers part of the path - the region switch reaches login/token-exchange, not the actual chat/generation requests. Left two comments on that below, would hold off merging until they're sorted. Three smaller ones on top, and +1 on CodeRabbit's still-open test-coverage comment.

packages/ballerina-extension/src/utils/ai/auth.ts
RNViththagan: This only runs inside `getPlatformStsToken`. On a warm restart (Reload Window / reopen) with a valid stored token, `checkToken` (`views/ai-panel/utils.ts`) returns early with the stored credentials and never reaches here, so the region is never set and `BACKEND_URL` stays US for the whole session (until a refresh or 401). An EU user's first request after a reload goes to the US backend.

Could we persist the region alongside the stored credentials and apply it on restart, instead of only after a fresh token exchange?

packages/ballerina-extension/src/features/ai/utils.ts
RNViththagan, line 66: Even when this runs, it doesn't reach the generation traffic. `getAnthropicClient` in `ai-client.ts` caches the client and only rebuilds on a login-method change, not a URL change — the `baseURL` is baked in from `BACKEND_URL` once. Switch region mid-session (or let a later refresh update `BACKEND_URL`) and the cached client keeps hitting the old URL.

Could we reset `cachedAnthropic` inside `setBackendRegion` (or add the resolved URL to the cache key), so it rebuilds when the region changes?

packages/ballerina-extension/src/features/ai/utils.ts
RNViththagan, line 77: Smaller one: `config.get('rootUrl')` only wins for the initial value now — this then overwrites `BACKEND_URL`/`DEVANT_TOKEN_EXCHANGE_URL`/`OLD_BACKEND_URL` with the regional URL on every run, so the override is silently dropped (it always won before this PR). A dev pointing `ballerina.rootUrl` at a local backend gets flipped to prod on the first refresh.

Could we early-return here when the `rootUrl` override is set?

packages/ballerina-extension/.env.example
```diff
@@ -1,2 +1,7 @@
 COPILOT_ROOT_URL=https://7eff1239-64bb-4663-b256-30a00d187a5c-prod.e1-us-east-azure.choreoapis.dev/copilot
 COPILOT_DEV_ROOT_URL=https://7eff1239-64bb-4663-b256-30a00d187a5c-dev.e1-us-east-azure.choreoapis.dev/copilot
+# COPILOT_STAGE_ROOT_URL=https://7eff1239-64bb-4663-b256-30a00d187a5c-staging.e1-us-east-azure.choreoapis.dev/copilot
```
RNViththagan: This commented line still carries the real staging URL, while the EU dev/stage placeholders below are left empty. Could we drop the value here too (`# COPILOT_STAGE_ROOT_URL=`), so we're not committing an internal staging hostname to the example file?

packages/ballerina-extension/src/utils/ai/auth.ts
kalaiyarasiganeshalingam: Fixed

packages/ballerina-extension/src/features/ai/utils.ts
kalaiyarasiganeshalingam: Fixed

kalaiyarasiganeshalingam: Fixed

madushajg (dismissed):

packages/ballerina-extension/src/features/ai/utils.ts
```diff
@@ -45,15 +45,43 @@ const PLATFORM_ENV_SETTING = "WSO2.WSO2-Platform.Advanced.ChoreoEnvironment";
...
-export const OLD_BACKEND_URL: string = BACKEND_URL + "/v2.0";
+export let OLD_BACKEND_URL: string = _defaultBackendUrl + "/v2.0";
+
+export const setBackendRegion = (region: string): boolean => {
+    if (config.get('rootUrl')) {
+        return true;
+    }
+    const normalized = region?.trim().toLowerCase();
+    const key = devantEnv ? `${normalized}-${devantEnv}` : normalized;
+    const regionalUrl = COPILOT_ROOT_URLS.get(key) || (devantEnv ? COPILOT_ROOT_URLS.get(normalized) : undefined);
```
RNViththagan: The `us-prod` / `eu-prod` entries above already resolve the default `prod` config, so this `|| get(normalized)` fallback isn't needed for prod any more - and because `COPILOT_EU_DEV_ROOT_URL` / `COPILOT_EU_STAGE_ROOT_URL` are unset in `.env.example`, it sends a dev/stage EU tester to EU **production** with a dev/stage token (the case flagged earlier on this line, which the previous commit had removed). Shall we drop the fallback and let a missing env-specific URL fail loud via the `return false` below? Pairs with the `.env.example` note.

packages/ballerina-extension/src/views/ai-panel/utils.ts
```diff
@@ -50,6 +52,13 @@ export const checkToken = async (): Promise<AuthCredentials | undefined> => {
             // First check if we have stored credentials
             const credentials = await getAccessToken();
```
RNViththagan: `getAccessToken()` can refresh an expired token, and that refresh exchanges against `DEVANT_TOKEN_EXCHANGE_URL` *before* the stored region is applied at line 59. On a cold start the platform extension may not be hydrated yet (`isLoggedIn()` is false inside `getPlatformStsToken` while `getStsToken()` still works), so an EU user's refresh goes to the US exchange - and a rejected exchange lands in `silentLogout`, which also signs them out of the platform. Shall we apply the stored region before calling `getAccessToken()`? e.g. read `getAuthCredentials()` raw, `setBackendRegion(secrets.region)` if BI_INTEL, then proceed. With the helper suggested in `auth.ts`, using live-then-stored here also gives legacy records without `region` a fallback.

packages/ballerina-extension/src/features/ai/utils/ai-client.ts
```diff
@@ -71,7 +72,7 @@ let cachedAuthMethod: LoginMethod | null = null;
  * @param options - Fetch options
  * @returns Promise<Response>
  */
-export async function fetchWithAuth(input: string | URL | Request, options: RequestInit = {}): Promise<Response | undefined> {
+export async function fetchWithAuth(input: string | URL | Request, options: RequestInit = {}): Promise<Response> {
     try {
         const credentials = await getAccessToken();
```
RNViththagan: `getAccessToken()` here can refresh and move `BACKEND_URL`, but `input` was built from the client's baked `baseURL` - so both the first `fetch(input)` and the 401 retry below re-use the stale URL. `cachedBaseUrl` fixes the *next* client build, not the in-flight request, and a region-mismatched request ends in `SILENT_LOGOUT` (which signs the user out of the platform too). Shall we rebase `input` onto the live `BACKEND_URL` after each refresh? e.g. capture `BACKEND_URL` before the call and, if it changed and `input` is a string, `input.replace(before, BACKEND_URL)` before fetching.

packages/ballerina-extension/src/utils/ai/auth.ts
```diff
@@ -398,10 +408,17 @@ export const getRefreshedAccessToken = async (): Promise<string> => {
                 console.log('Refreshing token via STS exchange...');
                 const newSecrets = await refreshTokenViaStsExchange();
 
-                // Update stored credentials
+                // Update stored credentials, persisting region so warm restarts restore it
+                let region: string | undefined;
```
RNViththagan: This block is repeated at `aiMachine.ts:496` and `views/ai-panel/utils.ts:76`, and the fourth BI_INTEL store in `getTokenForDefaultModel` (`features/ai/utils.ts:204-208`) was missed, so that path still persists no region. Shall we pull it into one helper here in `auth.ts` and use it at all four sites? Two things worth building in while we're at it:

- Gate on `isLoggedIn()`: `getAuthState()` returns `{ userInfo: null, region: "US" }` whenever the platform is logged out or not yet hydrated, so this unguarded read can persist `"us"` for an EU user if a refresh races startup - and every later warm restart would then restore US.
- Fall back to the previously stored region instead of dropping it.

Something like:

```ts
export type CopilotRegion = "us" | "eu";

export const getPlatformRegion = async (): Promise<CopilotRegion | undefined> => {
    try {
        const api = await getPlatformExtensionAPI();
        if (!api?.isLoggedIn()) { return undefined; } // getAuthState() reports "US" while logged out / unhydrated
        const region = api.getAuthState()?.region?.trim().toLowerCase();
        return region === "us" || region === "eu" ? region : undefined;
    } catch { return undefined; }
};

export const storeBiIntelCredentials = async (secrets: BIIntelSecrets): Promise<AuthCredentials> => {
    const region = (await getPlatformRegion()) ?? (await getStoredRegion());
    const credentials: AuthCredentials = { loginMethod: LoginMethod.BI_INTEL, secrets: { ...secrets, ...(region && { region }) } };
    await storeAuthCredentials(credentials);
    return credentials;
};
```

Then each site becomes a single `await storeBiIntelCredentials(secrets)`, and `region?: CopilotRegion` on `BIIntelSecrets` tightens the type too.

packages/ballerina-extension/.env.example
```diff
@@ -1,2 +1,7 @@
 COPILOT_ROOT_URL=https://7eff1239-64bb-4663-b256-30a00d187a5c-prod.e1-us-east-azure.choreoapis.dev/copilot
 COPILOT_DEV_ROOT_URL=https://7eff1239-64bb-4663-b256-30a00d187a5c-dev.e1-us-east-azure.choreoapis.dev/copilot
+# COPILOT_STAGE_ROOT_URL=
+
+COPILOT_EU_ROOT_URL=https://366b8031-ab7f-4a98-a1bd-3369fbd00eee-prod.e1-eu-west-cdp.choreoapis.dev/copilot
+# COPILOT_EU_DEV_ROOT_URL=
```
RNViththagan: `createEnvDefinePlugin` only defines keys that are declared in `.env`, so a commented line can't be supplied by a CI secret - the bundle keeps a literal `process.env.COPILOT_EU_DEV_ROOT_URL` that is `undefined` at runtime. Shall we declare these empty (`COPILOT_EU_DEV_ROOT_URL=`, `COPILOT_EU_STAGE_ROOT_URL=`) so an env-supplied value is picked up, and forward them in `.github/actions/build/action.yml` like `COPILOT_DEV_ROOT_URL`? Then the fallback at `features/ai/utils.ts:74` isn't needed.

packages/ballerina-extension/src/features/ai/utils.ts
```diff
@@ -45,15 +45,43 @@ const PLATFORM_ENV_SETTING = "WSO2.WSO2-Platform.Advanced.ChoreoEnvironment";
...
+    if (!regionalUrl) {
+        console.error(`No backend URL configured for region '${normalized}'`);
+        return false;
```
RNViththagan: Neither caller checks this return (`auth.ts:163`, `views/ai-panel/utils.ts:59`), so an unconfigured region only logs and the request goes on with the previous URLs. Shall we either act on it (a one-time `showWarningMessage` from the helper is more visible than `console.error`) or drop the boolean? Related: on `false` the last account's URLs stay in place - a small `resetBackendRegion()` back to the default, also called from `logout()`, keeps that predictable.

packages/ballerina-extension/src/features/ai/utils.ts
```diff
@@ -45,15 +45,43 @@ const PLATFORM_ENV_SETTING = "WSO2.WSO2-Platform.Advanced.ChoreoEnvironment";
...
+
+export let BACKEND_URL: string = _defaultBackendUrl;
+
+export let DEVANT_TOKEN_EXCHANGE_URL: string = _defaultBackendUrl + "/auth-api/v1.0/auth/token-exchange";
 
 // This refers to old backend before FE Migration. We need to eventually remove this.
-export const OLD_BACKEND_URL: string = BACKEND_URL + "/v2.0";
+export let OLD_BACKEND_URL: string = _defaultBackendUrl + "/v2.0";
+
+export const setBackendRegion = (region: string): boolean => {
```
RNViththagan: There are no tests for the region resolution yet (CodeRabbit asked too), and this module imports the state machines so a jest L1 test can't load it. Shall we move `devantEnv`, the URL map and `setBackendRegion` into a small leaf module (e.g. `features/ai/backend-url.ts`, importing only `vscode`) and add a unit test covering `us`/`eu` x `prod`/`dev`/`stage`, an unknown region, and the `rootUrl` short-circuit? It also removes the `auth.ts` <-> `utils.ts` import cycle.

packages/ballerina-extension/src/features/ai/utils.ts
RNViththagan: The `us-prod` / `eu-prod` entries above already handle the default `prod` setup, so this `|| get(normalized)` fallback is not needed for prod any more. But it still runs for dev/stage: `COPILOT_EU_DEV_ROOT_URL` and `COPILOT_EU_STAGE_ROOT_URL` are empty in `.env.example`, so an EU tester on dev/stage gets sent to EU **production** with a dev/stage token. This was flagged earlier on this line, and the previous commit had removed it. Shall we drop the fallback and let a missing URL fail with the `return false` below?

packages/ballerina-extension/src/views/ai-panel/utils.ts
RNViththagan, line 64: `getAccessToken()` can refresh an expired token. That refresh calls the token exchange URL before the saved region is set on line 59. On a cold start the platform extension may not be ready yet (`isLoggedIn()` is false inside `getPlatformStsToken`, but `getStsToken()` still works), so an EU user's refresh can go to the US exchange URL. If that fails we hit `silentLogout`, which also logs them out of the platform. Shall we set the saved region first, before calling `getAccessToken()`? For example: read `getAuthCredentials()`, call `setBackendRegion(secrets.region)` if it is BI_INTEL, then continue. With the helper suggested in `auth.ts`, this also covers old saved logins that have no `region`.

packages/ballerina-extension/src/features/ai/utils/ai-client.ts
RNViththagan, line 77: `getAccessToken()` here can refresh the token and change `BACKEND_URL`. But `input` was built from the old client URL, so both the first `fetch(input)` and the 401 retry below still use the old URL. `cachedBaseUrl` fixes the *next* client, not this request. A request sent to the wrong region ends in `SILENT_LOGOUT`, which also logs the user out of the platform. Shall we update `input` to the current `BACKEND_URL` after each refresh? For example: save `BACKEND_URL` before the call, and if it changed and `input` is a string, do `input.replace(before, BACKEND_URL)` before fetching.

packages/ballerina-extension/src/utils/ai/auth.ts
RNViththagan: This same block appears at `aiMachine.ts:496` and `views/ai-panel/utils.ts:76`. The fourth place that saves a BI_INTEL login, `getTokenForDefaultModel` (`features/ai/utils.ts:204-208`), was missed, so that path still saves no region. Shall we move this into one helper here in `auth.ts` and use it in all four places? Two things worth adding while we do that:

- Check `isLoggedIn()` first. `getAuthState()` returns `{ userInfo: null, region: "US" }` when the platform is logged out or not ready yet. So this read can save `"us"` for an EU user if a refresh runs during startup, and every restart after that would go to US.
- If the platform region is not available, keep the previously saved region instead of dropping it.

Something like:

```ts
export type CopilotRegion = "us" | "eu";

export const getPlatformRegion = async (): Promise<CopilotRegion | undefined> => {
    try {
        const api = await getPlatformExtensionAPI();
        if (!api?.isLoggedIn()) { return undefined; } // getAuthState() reports "US" when logged out / not ready
        const region = api.getAuthState()?.region?.trim().toLowerCase();
        return region === "us" || region === "eu" ? region : undefined;
    } catch { return undefined; }
};

export const storeBiIntelCredentials = async (secrets: BIIntelSecrets): Promise<AuthCredentials> => {
    const region = (await getPlatformRegion()) ?? (await getStoredRegion());
    const credentials: AuthCredentials = { loginMethod: LoginMethod.BI_INTEL, secrets: { ...secrets, ...(region && { region }) } };
    await storeAuthCredentials(credentials);
    return credentials;
};
```

Then each place becomes one call: `await storeBiIntelCredentials(secrets)`.

RNViththagan: Thanks, the helper looks good and covers the 4th site too. One thing still missing is the `isLoggedIn()` check. `getAuthState()` gives region "US" when the platform is logged out or not ready yet, so this can still save "us" for an EU user at startup. Shall we add `if (!api?.isLoggedIn()) { return undefined; }` before reading the region?

RNViththagan (approved):
LGTM

SasinduDilshara (dismissed):

packages/ballerina-extension/src/features/ai/utils.ts
```diff
@@ -48,15 +48,43 @@ const PLATFORM_ENV_SETTING = "WSO2.WSO2-Platform.Advanced.ChoreoEnvironment";
...
-    ["dev", process.env.COPILOT_DEV_ROOT_URL],
-    ["stage", process.env.COPILOT_STAGE_ROOT_URL || process.env.COPILOT_DEV_ROOT_URL],
+    ["us", process.env.COPILOT_ROOT_URL],
+    ["us-prod", process.env.COPILOT_ROOT_URL],
+    ["us-dev", process.env.COPILOT_DEV_ROOT_URL],
+    ["us-stage", process.env.COPILOT_STAGE_ROOT_URL || process.env.COPILOT_DEV_ROOT_URL],
+    ["eu", process.env.COPILOT_EU_ROOT_URL],
+    ["eu-prod", process.env.COPILOT_EU_ROOT_URL],
+    ["eu-dev", process.env.COPILOT_EU_DEV_ROOT_URL],
+    ["eu-stage", process.env.COPILOT_EU_STAGE_ROOT_URL || process.env.COPILOT_EU_DEV_ROOT_URL],
```
gigara: Fallback to Dev without a user notification may confuse

gigara (approved):

samithkavishke:
@coderabbitai review

SasinduDilshara (approved):

packages/ballerina-extension/src/features/ai/utils.ts
RNViththagan: This is intentional. We don't have a separate stage environment for Copilot, so stage uses the dev backend for both dev and stage. The fallback just reflects that, rather than a silent misconfiguration. @kalaiyarasiganeshalingam we can add a short comment on the map entry if we want to make that clear.

RNViththagan (approved):
LGTM

PR #1077, opened 2026-09-10
Fix checkpoint restore reporting failure, damaging files, and racing new generations

RNViththagan:
Related to https://github.com/wso2/product-integrator/issues/2406

A checkpoint restore rewrote the workspace correctly and then reported failure, so the chat history was never rewound. Two separate things caused that, and the same pass tightens what a restore is allowed to write and delete.

- Arm the Language Server artifact-update wait *before* applying the edits, and treat its timeout as advisory instead of rejecting — the notification handler has no replay, so a notification that arrived while the restore was still finishing was lost and the 10s timeout then always fired
- Move the post-restore data mapper refresh and webview notification out of the block that decides the outcome, so a view that fails to refresh no longer turns a completed restore into a failed one
- Never overwrite a file whose bytes on disk are not valid UTF-8: a snapshot holds decoded strings, so writing one back replaced real bytes with U+FFFD (keystores, images, binary test data). The files left as they are are named in a warning
- Send non-`.bal` deletions to the OS trash instead of unlinking them — the snapshot holds no copy of a file it never captured, so that delete was the user's only copy (`.bal` deletions already got this from `WorkspaceEdit`)
- Refuse any snapshot path that resolves outside the workspace root, for writes and deletions alike
- Skip rewriting a file that already matches, comparing against an unsaved editor when one is open; this replaces the `Ballerina.toml`-only workaround and spares the Language Server a recompile per untouched file
- Isolate each non-`.bal` write and delete, so one unwritable path applies everything else and then reports what failed instead of abandoning the restore half-applied
- Arm the artifact wait only when there is actually a `.bal` edit to apply, rather than whenever the project contains a `.bal` file

It also closes the window a restore leaves open, which is what made the restore look flaky (https://github.com/wso2/product-integrator/issues/2421):

- Refuse a new generation while a restore is running — `generateAgent` throws, as it already does for a concurrent generation, so the panel's catch resets its spinner and shows why. Without this the restore implicitly accepted the generation it was reverting, captured a torn checkpoint and baseline from disk, then sliced the running turn out of the thread while it was still streaming
- Refuse a restore or a revert while a response is still running — the mirror of the guard the thread actions already had, which also covers the moment right after Stop, while the executor is still unwinding
- Refuse a second restore while one is in flight, so the review bar's discard and a separator click cannot interleave two restores over the same files
- Extend the existing guard so new chat, thread switch and thread delete are refused during a restore too; deleting the active thread mid-restore used to have the restore recreate it empty and make it the active one

The refusal is extension-side state rather than the panel's own flag, which is webview state: it is gone after a reload and the extension never sees it, while the chat store is what a concurrent run corrupts. A reopened panel still shows an untruncated thread until it reloads — that needs a completion notification and stays in #2421.

`checkpointRestore.test.ts` adds 11 unit tests covering the restore outcome (notification early / late / absent, failed edit), the non-UTF-8 file, the already-matching file, the unsaved-editor case, the trash, the containment guard, the partial-write report, and the failing data mapper refresh. `restoreGuard.test.ts` adds 4 covering the refusal, its per-workspace scoping, and its release on a failed restore.

## Summary

- Improved checkpoint restore reliability and reporting.
- Added workspace path validation and safer file handling.
- Preserved non-UTF-8 files and moved non-`.bal` deletions to the OS trash.
- Skipped unnecessary writes and isolated per-file failures.
- Made artifact updates and post-restore refresh advisory.
- Added restore guards for generation, restore, revert, chat, and thread operations.
- Added tests for restore behavior, artifact waits, and concurrency guards.

### DIFF
--- a/packages/ballerina-extension/src/rpc-managers/ai-panel/rpc-manager.ts
+++ b/packages/ballerina-extension/src/rpc-managers/ai-panel/rpc-manager.ts
@@ -813,7 +813,7 @@ User reverted the last made changes. The files have been restored to the state b
             if (!found) {
                 if (chatStateStorage.hasCompactedHistory(projectRootPath, threadId)) {
                     window.showWarningMessage(
-                        "This conversation was compacted to manage memory. Undo points prior to compaction are unavailable."
+                        "This conversation was compacted to manage memory. Earlier undo points are no longer available."
                     );
                     throw new Error("Checkpoint unavailable due to compaction");
                 }
--- a/packages/ballerina-extension/src/rpc-managers/ai-panel/rpc-manager.ts
+++ b/packages/ballerina-extension/src/rpc-managers/ai-panel/rpc-manager.ts
@@ -160,6 +160,12 @@ import { approvalViewManager } from '../../features/ai/state/ApprovalViewManager
 import { chatStateStorage, isRevertible } from '../../views/ai-panel/chatStateStorage';
 import { runEventStore } from '../../features/ai/utils/run-event-store';
 import { restoreWorkspaceSnapshot } from '../../views/ai-panel/checkpoint/checkpointUtils';
+import {
+    assertNoRestoreInProgress,
+    beginRestore,
+    endRestore,
+    isRestoreInProgress
+} from '../../views/ai-panel/checkpoint/restore-state';
 import { runningServicesManager } from '../../features/ai/agent/tools/running-service-manager';
 import { executeRun } from "../../features/ai/agent/tools/ballerina-run";
 import { platformExtStore } from "../platform-ext/platform-store";
@@ -201,15 +207,20 @@ const CONNECTION_FAILURE_MESSAGE: Record<ConnectionSettleReason, string> = {
 
 /**
  * A run owns the active thread until it ends, so reparenting it mid-turn would strand the
- * run's writes. The panel already disables these actions; this backstops a webview reload
- * or a click that races the turn starting.
+ * run's writes; a restore is about to truncate that thread, and recreates it if it has gone.
+ * The panel already disables these actions; this backstops a webview reload or a click that
+ * races the turn starting.
  */
-function refuseWhileRunning(projectRootPath: string, action: string): boolean {
-    if (!runEventStore.hasActiveRun(projectRootPath)) {
-        return false;
+function refuseWhileBusy(projectRootPath: string, action: string): boolean {
+    if (runEventStore.hasActiveRun(projectRootPath)) {
+        console.warn(`[RPC] Refused ${action} — a response is still running for: ${projectRootPath}`);
+        return true;
     }
-    console.warn(`[RPC] Refused ${action} — a response is still running for: ${projectRootPath}`);
-    return true;
+    if (isRestoreInProgress(projectRootPath)) {
+        console.warn(`[RPC] Refused ${action} — a checkpoint restore is still running for: ${projectRootPath}`);
+        return true;
+    }
+    return false;
 }
 
 export class AiPanelRpcManager implements AIPanelAPI {
@@ -537,8 +548,10 @@ export class AiPanelRpcManager implements AIPanelAPI {
     }
 
     async revertGeneration(params: RevertGenerationRequest): Promise<void> {
+        const projectRootPath = resolveProjectRootPath();
+        assertNoRestoreInProgress(projectRootPath, 'revertGeneration');
+        beginRestore(projectRootPath);
         try {
-            const projectRootPath = resolveProjectRootPath();
             // Resolve the thread from the generation the bar names, not the active-thread pointer:
             // another thread can hold its own revertible generation.
             const located = chatStateStorage.findGenerationScope(projectRootPath, params.generationId);
@@ -599,6 +612,8 @@ User reverted the last made changes. The files have been restored to the state b
         } catch (error) {
             console.error("[Review Actions] Error reverting generation:", error);
             throw error;
+        } finally {
+            endRestore(projectRootPath);
         }
     }
 
@@ -778,47 +793,58 @@ User reverted the last made changes. The files have been restored to the state b
     async restoreCheckpoint(params: RestoreCheckpointRequest): Promise<void> {
         // Get project root path and thread identifiers
         const projectRootPath = resolveProjectRootPath();
-        const threadId = chatStateStorage.getActiveThreadId(resolveProjectRootPath());
+        assertNoRestoreInProgress(projectRootPath, 'restoreCheckpoint');
+        if (runEventStore.hasActiveRun(projectRootPath)) {
+            // The mirror of refuseWhileBusy: a run still writing files would land its edits on top
+            // of the restored ones, and its own generation is what the truncation is about to drop.
+            throw new Error('A response is still running. Please wait for it to finish before restoring.');
+        }
+        beginRestore(projectRootPath);
+        try {
+            const threadId = chatStateStorage.getActiveThreadId(projectRootPath);
 
-        // Find the checkpoint
-        const found = chatStateStorage.findCheckpoint(projectRootPath, threadId, params.checkpointId);
+            // Find the checkpoint
+            const found = chatStateStorage.findCheckpoint(projectRootPath, threadId, params.checkpointId);
 
-        if (!found) {
-            if (chatStateStorage.hasCompactedHistory(projectRootPath, threadId)) {
-                window.showWarningMessage(
-                    "This conversation was compacted to manage memory. Undo points prior to compaction are unavailable."
-                );
-                throw new Error("Checkpoint unavailable due to compaction");
+            if (!found) {
+                if (chatStateStorage.hasCompactedHistory(projectRootPath, threadId)) {
+                    window.showWarningMessage(
+                        "This conversation was compacted to manage memory. Undo points prior to compaction are unavailable."
+                    );
+                    throw new Error("Checkpoint unavailable due to compaction");
+                }
+                throw new Error(`Checkpoint ${params.checkpointId} not found`);
             }
-            throw new Error(`Checkpoint ${params.checkpointId} not found`);
-        }
 
-        const { checkpoint } = found;
+            const { checkpoint } = found;
 
-        // 1. Restore workspace files from checkpoint snapshot.
-        // restoreWorkspaceSnapshot reports its own failures to the user but does not throw, so a
-        // failed restore must not fall through to truncating the thread history — that loss is
-        // irreversible while the files would stay unchanged.
-        const workspaceRestored = await restoreWorkspaceSnapshot(checkpoint);
-        if (!workspaceRestored) {
-            throw new Error('Restoring the workspace from the checkpoint failed; the conversation was not rewound.');
-        }
+            // 1. Restore workspace files from checkpoint snapshot.
+            // restoreWorkspaceSnapshot reports its own failures to the user but does not throw, so a
+            // failed restore must not fall through to truncating the thread history — that loss is
+            // irreversible while the files would stay unchanged.
+            const workspaceRestored = await restoreWorkspaceSnapshot(checkpoint);
+            if (!workspaceRestored) {
+                throw new Error('Restoring the workspace from the checkpoint failed; the conversation was not rewound.');
+            }
 
-        // 2. Truncate thread history to this checkpoint
-        const restored = chatStateStorage.restoreThreadToCheckpoint(
-            projectRootPath,
-            threadId,
-            params.checkpointId
-        );
+            // 2. Truncate thread history to this checkpoint
+            const restored = chatStateStorage.restoreThreadToCheckpoint(
+                projectRootPath,
+                threadId,
+                params.checkpointId
+            );
 
-        if (!restored) {
-            throw new Error('Failed to restore thread to checkpoint');
+            if (!restored) {
+                throw new Error('Failed to restore thread to checkpoint');
+            }
+        } finally {
+            endRestore(projectRootPath);
         }
     }
 
     async clearChat(): Promise<void> {
         const projectRootPath = resolveProjectRootPath();
-        if (refuseWhileRunning(projectRootPath, 'clearChat')) { return; }
+        if (refuseWhileBusy(projectRootPath, 'clearChat')) { return; }
         // Create a new thread — preserves all existing history
         const newThreadId = chatStateStorage.createNewThread(projectRootPath);
         clearCompactionDisabledWarning(projectRootPath, newThreadId);
@@ -832,13 +858,13 @@ User reverted the last made changes. The files have been restored to the state b
 
     async switchThread(params: SwitchThreadRequest): Promise<void> {
         const projectRootPath = resolveProjectRootPath();
-        if (refuseWhileRunning(projectRootPath, 'switchThread')) { return; }
+        if (refuseWhileBusy(projectRootPath, 'switchThread')) { return; }
         chatStateStorage.switchToThread(projectRootPath, params.threadId);
     }
 
     async deleteThread(params: DeleteThreadRequest): Promise<void> {
         const projectRootPath = resolveProjectRootPath();
-        if (refuseWhileRunning(projectRootPath, 'deleteThread')) { return; }
+        if (refuseWhileBusy(projectRootPath, 'deleteThread')) { return; }
         await chatStateStorage.deleteThread(projectRootPath, params.threadId);
     }
 
--- a/packages/ballerina-extension/src/rpc-managers/ai-panel/rpc-manager.ts
+++ b/packages/ballerina-extension/src/rpc-managers/ai-panel/rpc-manager.ts
@@ -212,7 +212,7 @@ const CONNECTION_FAILURE_MESSAGE: Record<ConnectionSettleReason, string> = {
  * races the turn starting.
  */
 function refuseWhileBusy(projectRootPath: string, action: string): boolean {
-    if (runEventStore.hasActiveRun(projectRootPath)) {
+    if (runEventStore.hasActiveRun(projectRootPath) || chatStateStorage.hasActiveExecutionFor(projectRootPath)) {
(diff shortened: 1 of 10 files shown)


packages/ballerina-extension/src/test-support/checkpointRestore.test.ts
```diff
@@ -0,0 +1,308 @@
...
+        fs.writeFileSync(path.join(root, "extra.txt"), "added by the generation\n");
+
+        checkpoint = {
+            id: "cp-1",
+            messageId: "msg-1",
+            timestamp: 0,
+            fileList: ["main.bal", "Config.toml"],
+            workspaceSnapshot: { "main.bal": ORIGINAL_BAL, "Config.toml": ORIGINAL_CONFIG },
+            snapshotSize: ORIGINAL_BAL.length + ORIGINAL_CONFIG.length,
+        };
```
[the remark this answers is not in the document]
RNViththagan: Not applying this one, and I would like to check the reasoning with you.

The `issue-<n>.json` convention in the guidelines is backed by one corpus: `ballerina-side-panel/src/test/fixtures/fields/`, where each fixture is a form-field descriptor fed to a field-rendering invariant. There is no equivalent corpus or invariant for checkpoint restore in `ballerina-extension`, so there is nothing here for an `issue-2406.json` to feed.

On the substance of the rule, the tests do assert rules rather than the reported case. Each one pins a property of the restore path, not an instance of the bug: a restore never overwrites bytes it cannot reproduce, never deletes a file permanently, never writes outside the workspace root, and never has its outcome decided by an advisory refresh. The reported symptom (#2406) is covered by the third of those.

Building a fixture corpus for a restore would mean modelling workspace state (files on disk, unsaved editors, deletions) rather than a single JSON descriptor, which is a bigger change than this fix warrants. Happy to do it separately if you think the restore path should get one.

packages/ballerina-extension/src/views/ai-panel/checkpoint/checkpointUtils.ts
```diff
@@ -20,18 +20,40 @@ import * as vscode from 'vscode';
...
+// captured faithfully — writing one back replaces real bytes with U+FFFD.
+function isLosslessUtf8(bytes: Buffer): boolean {
+    return Buffer.from(bytes.toString('utf8'), 'utf8').equals(bytes);
+}
+
+// Snapshot keys are relative paths from a previous session, so they are only as trustworthy as the
+// file they were persisted in: resolve first, then refuse anything that lands outside the workspace.
+function resolveInsideWorkspace(workspaceRoot: vscode.Uri, filePath: string): vscode.Uri | null {
+    const target = path.resolve(workspaceRoot.fsPath, filePath);
+    return isPathInside(workspaceRoot.fsPath, target) ? vscode.Uri.file(target) : null;
```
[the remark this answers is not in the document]
RNViththagan: Agreed on the analysis, but leaving it for a follow-up rather than this PR.

`isPathInside` is lexical, so a snapshot path under an in-workspace directory symlink still resolves as inside and can reach a write outside the root. Worth noting for context that this is a pre-existing gap rather than something this PR introduces: before this change there was no containment check at all, so the lexical one is a strict improvement on the same path.

Doing it properly means `realpath` on both sides, not just the target: workspace roots are themselves frequently symlinked (macOS `/tmp` and `/var` among them), so resolving only the target would reject legitimate restores. It also needs to resolve the nearest existing ancestor, since the target may not exist yet on a restore that recreates a file. That is enough behaviour to want its own tests, alongside the related finding that a symlinked directory also lets a restore delete outside the workspace.

Both are tracked and will go in together.

RNViththagan:
Addressed the review feedback in 08f1f45f92.

Applied:

- Settle the artifact-update wait before running its callback. `stopListening()` had already cleared the timeout, so a throwing callback left the promise pending forever — and since `publish()` swallows subscriber errors, the restore would have hung with the restore-in-progress guard still held, refusing every later generation.
- Wrap `notifyCurrentWebview()` in its own catch. It was the one advisory step left outside the guarded block, and `sendNotification` can throw synchronously, which would have reported a completed restore as failed — the same defect this PR set out to fix.
- Drop the `fs.unlinkSync` fallback when the trash refuses a deletion. The file can be the only copy, so it is now left in place and reported instead.

Two further changes from the same pass, beyond the review comments:

- Decide snapshot fidelity at capture rather than at restore. A file whose bytes do not survive a UTF-8 round trip is now recorded in `fileList` but left out of `workspaceSnapshot`, which means a restore neither deletes it (it is listed) nor writes a lossy decode over it (it is not in the snapshot), with no change to the persisted schema. Judging it from the current disk bytes was wrong in both directions: an untouched image was reported as unrestorable on every restore, and an image the generation had deleted was recreated corrupted, because the check was skipped when the file was absent. Checkpoints captured before this change are detected by the replacement character in their content and left alone.
- Match open documents by scheme as well as path. An SCM diff (`git:`) document carries the same `fsPath` as the file it mirrors, so comparing against its HEAD content could skip restoring the real file and still report success.

Restore also reads through `fs` synchronously now instead of an awaited round trip per file, which the skip-if-unchanged comparison had made a per-file cost.

Tests: 293 pass. The non-text-file test previously pinned the incorrect behaviour and now covers capture-side exclusion, the untouched file producing no warning, a lossy legacy entry never being written back, the SCM-diff path, and a refused trash leaving the file in place. `node scripts/typecheck-tests.js`, `tsc` and tslint are clean.

Replied on the two threads not applied — the `issue-<n>.json` corpus and the `realpath` containment check.

packages/ballerina-extension/src/utils/project-artifacts-handler.ts
```diff
@@ -100,3 +100,56 @@ export class ArtifactNotificationHandler {
...
+        settle = resolve;
+    });
+
+    const unsubscribe = ArtifactNotificationHandler.getInstance().subscribe(
+        ArtifactsUpdated.method,
+        undefined,
+        payload => {
+            stopListening();
+            onArtifacts(payload.data);
+            settle(true);
```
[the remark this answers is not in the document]
RNViththagan: Fixed in 08f1f45f92. `settle(true)` now runs before the callback, and the callback is wrapped so a throw is logged rather than lost.

Worth recording how bad this was: `publish()` swallows a throwing subscriber and `stopListening()` had already cleared the timeout, so `notified` stayed pending forever. The restore awaits it, and this PR also makes the restore hold a restore-in-progress flag released in a `finally` — which would never have run. So a single throwing listener would have left every later generation refused for the rest of the session.

packages/ballerina-extension/src/views/ai-panel/checkpoint/checkpointUtils.ts
```diff
@@ -227,48 +281,39 @@ export async function restoreWorkspaceSnapshot(checkpoint: Checkpoint, skipArtif
...
-                return;
-            }
+        // Advisory, and deliberately outside the block above: the files are written and saved by
+        // now, so a view that fails to refresh must not turn a completed restore into a failed one.
+        try {
+            await renderDatamapper();
+        } catch (error) {
+            console.warn('[Checkpoint] Could not refresh the data mapper after the restore:', error);
+        }
+        notifyCurrentWebview();
```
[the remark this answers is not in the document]
RNViththagan: Fixed in 08f1f45f92, wrapped in its own catch.

This was the same defect the PR set out to fix, one line further down: I moved both advisory calls out of the outcome-determining block but only wrapped `renderDatamapper()`, while the comment above them claimed the property the code did not have. The comment now covers both, and the test that makes the data mapper refresh throw has a sibling for the notification.

packages/ballerina-extension/src/views/ai-panel/checkpoint/checkpointUtils.ts
```diff
@@ -174,29 +199,58 @@ export async function restoreWorkspaceSnapshot(checkpoint: Checkpoint, skipArtif
...
+                    }
+                    try {
+                        // The snapshot holds no copy of a file it never captured, so this deletion
+                        // is the user's only copy — .bal deletes already get the trash via WorkspaceEdit.
+                        await vscode.workspace.fs.delete(fileUri, { useTrash: true });
+                    } catch {
+                        try {
+                            fs.unlinkSync(fileUri.fsPath);
+                        } catch (error) {
+                            failed.push(`${path.basename(fileUri.fsPath)} (${(error as Error).message})`);
```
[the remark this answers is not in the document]
RNViththagan: Agreed, fixed in 08f1f45f92. The fallback is gone: if the trash refuses the deletion, or the user cancels it, the file is left alone and reported as not restored.

Falling back to a permanent delete contradicted the reason for the change, since the snapshot holds no copy of a file it never captured. A test now makes `workspace.fs.delete` reject and asserts the file is still there.

packages/ballerina-extension/src/rpc-managers/ai-panel/rpc-manager.ts
```diff
@@ -778,47 +796,59 @@ User reverted the last made changes. The files have been restored to the state b
...
-        if (!found) {
-            if (chatStateStorage.hasCompactedHistory(projectRootPath, threadId)) {
-                window.showWarningMessage(
-                    "This conversation was compacted to manage memory. Undo points prior to compaction are unavailable."
-                );
-                throw new Error("Checkpoint unavailable due to compaction");
+            if (!found) {
+                if (chatStateStorage.hasCompactedHistory(projectRootPath, threadId)) {
+                    window.showWarningMessage(
+                        "This conversation was compacted to manage memory. Undo points prior to compaction are unavailable."
```
madushajg:
```suggestion
                        "This conversation was compacted to manage memory. Earlier undo points are no longer available."
```

packages/ballerina-extension/src/utils/project-artifacts-handler.ts
```diff
@@ -100,3 +100,62 @@ export class ArtifactNotificationHandler {
...
+                console.error('[Artifacts] Artifact-update listener failed:', error);
+            }
+        }
+    );
+    const timeoutId = setTimeout(() => {
+        stopListening();
+        settle(false);
+    }, timeoutMs);
+
+    function stopListening(): void {
```
madushajg, line 154: `stopListening()` isn't idempotent, and `cancel()` can run it a second time after the notification already settled the wait (notification during `applyEdit`, then the progress block throws at `failed.length > 0 → artifactWait?.cancel()). ArtifactNotificationHandler`'s unsubscribe closure does `if (subscribers.size === 0) this.subscribers.delete(type)` over a captured Set, so the second call deletes the Set a later subscriber created and silently unsubscribes every other `artifactsUpdated` listener.

packages/ballerina-extension/src/views/ai-panel/checkpoint/checkpointUtils.ts
```diff
@@ -143,60 +180,95 @@ export async function restoreWorkspaceSnapshot(checkpoint: Checkpoint, skipArtif
...
-                        fs.unlinkSync(fileUri.fsPath);
+                // Each file is isolated: one unwritable path must not abandon the rest half-restored.
+                for (const { fileUri, filePath } of nonBalFilesToDelete) {
+                    if (!fs.existsSync(fileUri.fsPath)) {
+                        continue;
+                    }
+                    try {
+                        // The snapshot holds no copy of a file it never captured, so this deletion
+                        // is the user's only copy — .bal deletes already get the trash via WorkspaceEdit.
+                        await vscode.workspace.fs.delete(fileUri, { useTrash: true });
```
madushajg: `workspace.fs.delete(uri, { useTrash: true })` rejects outright when the filesystem provider lacks trash support (several remote/virtual providers) or when `files.enableTrash` is off. On those setups every non-`.bal` file the generation created is left behind on every restore, yet the restore still returns true and the caller truncates the chat history. The user is told only about the first 3 names. Can we check trash capability up front (and surface a clear, non-truncated list, or fall back explicitly) rather than having this degrade silently?

packages/ballerina-extension/src/views/ai-panel/checkpoint/checkpointUtils.ts
```diff
@@ -143,60 +180,95 @@ export async function restoreWorkspaceSnapshot(checkpoint: Checkpoint, skipArtif
...
+                }
+
+                const existing = readFileBytes(fileUri);
+
+                // An unsaved editor, not disk, is what the user sees and what saveAll writes back,
+                // so a file only counts as already restored when the buffer matches too. Skipping
+                // the rest spares the Language Server a recompile per untouched file.
+                const currentText = openDocumentText(fileUri) ?? existing?.toString('utf8');
+                if (currentText === content) {
+                    continue;
```
madushajg: `openDocumentText(fileUri) ?? existing?.toString('utf8')` means disk is never consulted when a document is open. Non-`.bal` files are written by the agent with `fs.writeFileSync`, so an open non-dirty editor can still hold pre-generation text that equals the snapshot while disk holds the generation's content, The file is then silently skipped and never restored. `saveAll()` only covers the dirty case. 

Suggestion:  trust the buffer only when `doc.isDirty` and otherwise comparing against existing, or requiring both to match before continue.

packages/ballerina-extension/src/utils/project-artifacts-handler.ts
RNViththagan: Good catch, and worse than a double unsubscribe — fixed in c0450d37d9.

Traced it through: the closure `subscribe()` hands back captures the Set and does `if (subscribers.size === 0) this.subscribers.delete(type)`. So the second call operates on the old, already-empty Set, sees size 0, and deletes whatever Set a later subscriber created under that key — every other `artifactsUpdated` listener in the extension goes silent, with nothing logged. Reachable exactly as you describe: notification during `applyEdit`, then the throw at `failed.length > 0` reaches `artifactWait?.cancel()`.

`stopListening()` is now guarded by a `stopped` flag. There is a regression test in `artifactUpdateWait.test.ts` that subscribes a second listener after the wait has settled, cancels the wait, and asserts the listener still receives the next notification — it reports 0 calls without the guard.

packages/ballerina-extension/src/views/ai-panel/checkpoint/checkpointUtils.ts
RNViththagan: Agreed, fixed in c0450d37d9 — I had the precedence backwards.

Taking both is the safer rule, so the skip now needs disk to match *and* no open buffer to disagree:

    const openText = openDocumentText(fileUri);
    const diskText = existing?.toString('utf8');
    if (diskText === content && (openText === undefined || openText === content)) { continue; }

That keeps the case the buffer check was added for (an unsaved editor differing from disk must still be reverted, or saveAll would write it back) while closing the one you found. I used disk-agreement rather than `isDirty` because a non-dirty buffer that VS Code has not yet reloaded is precisely the stale case — `isDirty` is false there too. Test added for a non-dirty editor holding pre-generation text while disk holds the generation's.

packages/ballerina-extension/src/views/ai-panel/checkpoint/checkpointUtils.ts
RNViththagan: Fair — it degraded silently, fixed in c0450d37d9.

The deletion now follows `files.enableTrash` rather than forcing `useTrash: true`. That is the same setting `WorkspaceEdit.deleteFile` already honours for the `.bal` deletions in this function, so both halves of a restore now behave the way the user configured deletions to behave, and someone who has deliberately turned the trash off no longer has every restore leave files behind.

If the trash is enabled but the provider refuses, the file is still kept rather than deleted permanently — that part was deliberate, since the snapshot holds no copy of a file it never captured, so the deletion would be the only copy.

Also fixed the message: it listed three names as if that were the whole set. It now shows up to five and says how many more, with every one logged.

I did not add an up-front capability probe: there is no API for whether a provider supports trash short of attempting a delete, so probing would mean a speculative delete. Happy to revisit if you know of one.

packages/ballerina-extension/src/rpc-managers/ai-panel/rpc-manager.ts
RNViththagan: Applied in c0450d37d9, thanks — reads better.

madushajg (dismissed):

packages/ballerina-extension/src/rpc-managers/ai-panel/rpc-manager.ts
```diff
@@ -445,6 +456,9 @@ export class AiPanelRpcManager implements AIPanelAPI {
 
     async generateContextTypes(params: ProcessContextTypeCreationRequest): Promise<void> {
         try {
+            // Writes a generation to the active thread without ever calling beginRun, so
+            // hasActiveRun cannot see it and the restore guard has to be asked directly.
+            assertNoRestoreInProgress(resolveProjectRootPath(), 'generateContextTypes');
```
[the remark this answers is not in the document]
RNViththagan: Good catch — the gap is real, fixed in the latest commit, though by a different route than holding a second flag.

`ContextTypesExecutor` extends `AICommandExecutor`, which already registers an active execution (`setActiveExecution`, stage 1) and clears it in a `finally`. What it skips is `runEventStore.beginRun`, which only happens under `trackForReconnection` — so the state exists, `restoreCheckpoint` was just looking in the wrong registry. It checked `runEventStore.hasActiveRun`, which by design only sees runs buffered for panel reconnection.

So rather than a second flag with its own lifecycle to leak, both guards now also consult `chatStateStorage.hasActiveExecutionFor(projectRootPath)` (new, per-workspace; the existing `hasAnyActiveExecution` is global). That covers every executor that registers one — the type creator, doc generation and the rest — for the whole duration, released by the `finally` that already exists, and it extends `refuseWhileBusy` too, so new chat, thread switch and thread delete are refused during those runs as well.

The other direction stays as it was: `generateContextTypes` asserts no restore is in progress before starting.

Tests added against the real `ChatStateStorage`: an execution that never begins a tracked run reports the workspace busy, clearing it frees the workspace, and another workspace is unaffected.

SasinduDilshara (approved):

madushajg (approved):

PR #1078, opened 2026-09-10
Animate MCP toolkit tools during dev-time tracing

dan-niles:
## Summary

Fixes https://github.com/wso2/product-integrator/issues/2072

Dev-time trace animation never lit up MCP toolkit tool nodes. The trace only carries the MCP server's actual tool name, never the toolkit variable name shown in the diagram, so name-based matching could never succeed for MCP tools.

This PR threads the MCP toolkit's class name through the whole pipeline (language server → trace span attribute → webview) and matches on that instead, so MCP tool nodes now pulse correctly when traced.

## Implementation

- **Language server**: `CodeAnalyzer`/`AiUtils` resolve the MCP toolkit's module-qualified class name and emit it as `className` on the tool's entry in `agentInfo.tools`
- **Runtime → trace**: relies on `module-ballerina-ai#165`, which records the toolkit's class name on `execute_tool` spans (`gen_ai.tool.toolkit.name`)
- **Extension**: `trace-animation.ts` reads that span attribute and threads it through as `activeToolKitName` on the animation event
- **Webview**: `DiagramContext` stores it per trace entry; `AgentNodeWidget` and `AgentCallNodeWidget` match MCP tools against it by toolkit class name instead of by tool name

https://github.com/user-attachments/assets/f20cc11e-6abb-4009-9450-790f3fd8fe06

## Summary

- Fixes dev-time trace animation for MCP toolkit tools in agent nodes.
- Passes the module-qualified MCP toolkit class name from the language server through trace events to the webview.
- Matches MCP toolkit traces to the correct diagram nodes.
- Updates agent and tool animation states, including active borders, connectors, dimming, and hover styles.


packages/bi-diagram/src/components/nodes/agentNodeUtils.ts
```diff
@@ -45,3 +45,31 @@ export function sanitizeAgentData(data: AgentData): AgentData {
...
+
+function mcpToolKitSimpleClassName(tool: ToolData): string | undefined {
+    return tool.type === "MCP Server" ? tool.className?.split(":").pop() : undefined;
+}
+
+// MCP tools trace by toolkit class name, not tool.name (which is the toolkit variable name, never seen in a trace).
+export function isToolTraceActive(tool: ToolData, activeToolNames: string[], activeToolKitNames: string[]): boolean {
+    const mcpClassName = mcpToolKitSimpleClassName(tool);
+    if (mcpClassName) {
+        return activeToolKitNames.some(name => name?.includes(mcpClassName));
```
samithkavishke: **[MEDIUM] MCP match is keyed on toolkit *class name*, not on the specific toolkit field/instance — two fields of the same MCP toolkit class light up together.**

`mcpToolKitSimpleClassName` (line 49-51) reduces `tool.className` down to just the simple class name, and `isToolTraceActive`/`toolEntryMatchesTools` (line 63-74) match any trace entry whose `toolKitName` *contains* that simple name. There's no per-field/instance discriminator (e.g. the toolkit variable name) folded into the match.

Concrete failure: an agent with two MCP toolkit fields of the same class (e.g. two `GitHubMcpToolKit` instances pointed at different servers/tokens — a very normal pattern for "prod" + "staging" or multi-tenant setups) will have identical `tool.className` for both `ToolData` entries. When only one of them actually executes a tool, `activeToolKitNames` contains that one class name, and `isToolTraceActive` (line 57) returns `true` for **both** `ToolData` entries because it only checks whether the name is present anywhere, not which specific tool node is executing. Both toolkit nodes will animate as active even though only one actually ran.

Separately (same mechanism, lower likelihood), the `.includes()` substring check on line 57 and line 73 also risks a false-positive across *different* MCP toolkit classes if one's simple name is a substring of another's recorded toolkit name (e.g. class `SearchToolKit` vs. a differently-named toolkit whose trace value ends in `...WebSearchToolKit`), since it's not an exact/boundary-safe match.

Suggest keying the match on something unique per field (e.g. thread the toolkit variable/field name through in addition to/instead of the class name) or at least using an exact-match / suffix-with-boundary check instead of a bare substring test, if class-name granularity is accepted as a known limitation this should probably be called out in the PR description.

packages/bi-diagram/src/components/nodes/agentNodeUtils.ts
```diff
@@ -45,3 +45,31 @@ export function sanitizeAgentData(data: AgentData): AgentData {
...
+    if (entry.toolName && tools.some(t => t.name === entry.toolName)) {
+        return true;
+    }
+    const toolKitName = entry.toolKitName;
+    if (!toolKitName) {
+        return false;
+    }
+    return tools.some(t => {
+        const mcpClassName = mcpToolKitSimpleClassName(t);
+        return Boolean(mcpClassName) && toolKitName.includes(mcpClassName);
```
samithkavishke: Same substring-match mechanism as the comment on line 57 — flagging the second call site so both are addressed together if the matching strategy changes.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
```diff
@@ -1801,26 +1801,37 @@ public static boolean readRequiresApproval(Symbol toolSymbol, Project project) {
...
+        return mcpToolKitClassName(typeSymbol);
     }
 
-    public static boolean isMcpToolKitType(TypeSymbol typeSymbol) {
-        if (typeSymbol.nameEquals("McpToolKit") && typeSymbol.getModule()
-                .map(module -> CommonUtils.isAiModule(module.id().orgName(), module.id().packageName()))
-                .orElse(false)) {
-            return true;
+    public static Optional<String> mcpToolKitClassName(TypeSymbol typeSymbol) {
+        if (!isMcpToolKitType(typeSymbol)
```
samithkavishke: **[LOW] Redundant raw-type resolution.** `isMcpToolKitType(typeSymbol)` (called on this line) already computes `CommonUtils.getRawType(typeSymbol)` internally to check for `ClassSymbol`/`isAiMcpBaseToolKit`, and then this method immediately recomputes `CommonUtils.getRawType(typeSymbol)` again on the next line to get the `ClassSymbol` for the class/module name. It's cheap per-call, but on a project with many tool references this doubles the raw-type resolution work for no reason. Consider having `isMcpToolKitType` return the resolved `ClassSymbol` (or refactor into a single helper that resolves-and-classifies once) so both branches reuse the same computation.

packages/bi-diagram/src/components/nodes/agentNodeUtils.ts
dan-niles: Fixed the substring risk in 933150f644 — both call sites now compare exact simple class names instead of `.includes()`, so `SearchToolKit` won't false-match inside `WebSearchToolKit` anymore.

On the class-vs-instance granularity: agreed this is an acceptable known limitation. The MCP toolkit code-gen always creates a distinct class per added toolkit, so two same-class instances only happen in a rare pro-code scenario (manually reusing one toolkit class for multiple instances), which isn't a pattern we need to support precisely for this dev-time visual trace. Currently we don't have a mechanism to fetch the instance name at run time and pass it in the trace, so we have gone with the class name.

packages/bi-diagram/src/components/nodes/agentNodeUtils.ts
dan-niles: Fixed alongside the other call site in 933150f644 — same exact simple-class-name comparison now, no more substring containment.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
dan-niles: Fixed in f1a78c0260 — factored the raw-type resolution into a single `asMcpToolKitClass` helper that both `isMcpToolKitType` and `mcpToolKitClassName` now share, so it's resolved once instead of twice.

dan-niles: Double-checked this is behavior-preserving: `isAiMcpBaseToolKit` checks for a `*McpBaseToolKit` type inclusion, and in `ballerina/ai`'s `toolkit.bal`, `McpToolKit` itself is `public isolated class McpToolKit { *McpBaseToolKit; ... }`. So the nameEquals("McpToolKit") branch and the isAiMcpBaseToolKit branch were never independent — whenever the first matches, the second already holds on the same resolved class. The refactor only removes the duplicate `getRawType` call, no behavior change.

madushajg (dismissed):

samithkavishke (dismissed):
LGTM

MohamedSabthar (dismissed):

madushajg (approved):

MohamedSabthar (approved):

PR #1079, opened 2026-09-10
Simplify agent run node design

dan-niles:
## Summary

Simplifies the agent and durable agent run node's design by removing the system prompt, tools and model provider circles. Adds a clear label indicating that this is an agent run node and adds an `Open Agent` button that redirects to the agent instance. 

Fixes https://github.com/wso2/product-integrator/issues/2428

Before: 

image

After:

image

## Summary

- Simplified agent and durable agent run nodes by removing embedded prompt, tool, and model-provider details.
- Added agent references with labels and an **Open Agent** action.
- Added compact reference layouts, sizing, styling, and durable-agent color support.
- Improved result-variable resolution for uncaptured calls and invalid assignment expressions.
- Added sizing coverage for durable-agent reference nodes.


MohamedSabthar (dismissed):

packages/bi-diagram/src/visitors/SizingVisitor.ts
```diff
@@ -421,6 +420,13 @@ export class SizingVisitor implements BaseVisitor {
             return;
         }
 
+        // Reference mode (a run() call site) collapses to the same simple reference row
+        // AgentCallNode uses — no side circle columns are painted, so no side space is reserved.
+        if (this.isDurableAgentReference) {
+            this.setNodeSize(node, halfNodeWidth, halfNodeWidth, NODE_HEIGHT + AGENT_CALL_REFERENCE_HEIGHT);
+            return;
+        }
+
```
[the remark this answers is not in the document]
dan-niles: Added L1 coverage for both sizing paths in `DurableAgentReferenceSizing.test.ts`, asserting `lw`/`rw`/`ch` for an `agentBox` durable-agent run under `new SizingVisitor(false)` and `new SizingVisitor(true)`.

MohamedSabthar (dismissed):

packages/bi-diagram/src/utils/node.ts
```diff
@@ -522,6 +522,18 @@ export function getNodeTitle(node: FlowNode) {
     return label;
 }
 
+// getFlowModel always requests forceAssign, so properties.variable can carry a suggested name that
+// doesn't match the source — for a discard (`_ = ...`) or a plain reassignment to an existing variable
+// (`draft = ...`, no type). The statement's own source is the ground truth for what it assigns to.
+export function getResultVariableName(node: FlowNode): string | undefined {
+    const source = node.codedata?.sourceCode?.trim();
+    const lhs = source?.match(/^([^=]*)=[^=]/)?.[1]?.trim();
```
madushajg, line 531: The `^([^=]*)=[^=]` match can't tell an assignment `=` from a named-argument `=`: for `check agent->run(query = "hi");` this returns `"agent->run(query"` and renders that as the node's variable name. Consider anchoring on the assignment by requiring the match to end before the first `(`, or bailing out when the captured LHS contains `(`.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
```diff
@@ -864,6 +887,83 @@ export function DurableAgentRunNodeWidget(props: DurableAgentRunNodeWidgetProps)
...
+                                <NodeIcon type={model.node.codedata.node} size={24} />
+                                <NodeStyles.RunBadge>
+                                    <Icon name="bi-play" iconSx={{ fontSize: "20px" }} sx={{ color: "var(--vscode-charts-green)", display: "flex", justifyContent: "center", alignItems: "center" }} />
+                                </NodeStyles.RunBadge>
+                            </NodeStyles.IconBox>
+                            <NodeStyles.Row readOnly={readOnly}>
+                                <NodeStyles.Header onClick={handleOnClick}>
+                                    <NodeStyles.Title>durable agent : run</NodeStyles.Title>
+                                    <NodeStyles.Description>
+                                        {model.node.properties?.variable?.value as ReactNode}
```
madushajg: This is the same `forceAssign` phantom-name case `getResultVariableName` was added for in `AgentCallNodeWidget` (line 499). A `_ = ...` discard here will still show a suggested name that isn't in the source. Use `getResultVariableName(model.node)` for consistency.

packages/bi-diagram/src/utils/node.ts
dan-niles: Fixed in 8e4f128d4e — bail to the suggested name when the captured LHS contains `(`, since that means the `=` found is a named-argument, not an assignment.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
dan-niles: Fixed in 8e4f128d4e — now uses `getResultVariableName(model.node)` here for consistency with AgentCallNodeWidget. (There's a second, identical read at line 1112 with the same latent issue, but it's pre-existing code untouched by this PR, so left it alone.)

packages/bi-diagram/src/utils/node.ts
```diff
@@ -522,6 +522,18 @@ export function getNodeTitle(node: FlowNode) {
...
+    if (!lhs || lhs.includes("(")) {
+        return node.properties?.variable?.value as string | undefined;
+    }
```
dulajdilshan, line 534: **[Major] Stated invariant is false — residual of the named-arg fix.** The comment at `:525` says `properties.variable` "can carry a suggested name that doesn't match the source", but both bail-outs at `:531` return exactly that: `agent->run(x = 1);` takes the `includes("(")` arm and `agent->run(a);` takes the `!lhs` arm, so an uncaptured call still renders a phantom name — `8e4f128d4e` swapped garbage (`"agent->run(x"`) for a different wrong name rather than removing it.

I verified the premise rather than assuming it: `CodeAnalyzer.java:489` returns early for a remote call under `forceAssign`, which is what makes the LS hand back a suggested variable for a call that assigns nothing.

A reader trusts the comment instead of the code, and the next caller of this shared helper inherits the same surprise. Should *both* bail-outs return `undefined` — "no assignment LHS in the source" is the honest answer for an uncaptured call — and the widget render no name? (Fixing only the `(` arm leaves `agent->run(a);` still phantom, which is why I'm not offering this as a one-click suggestion.)

packages/bi-diagram/src/utils/node.ts
```diff
@@ -522,6 +522,18 @@ export function getNodeTitle(node: FlowNode) {
     return label;
 }
 
+// getFlowModel always requests forceAssign, so properties.variable can carry a suggested name that
+// doesn't match the source — for a discard (`_ = ...`) or a plain reassignment to an existing variable
+// (`draft = ...`, no type). The statement's own source is the ground truth for what it assigns to.
+export function getResultVariableName(node: FlowNode): string | undefined {
+    const source = node.codedata?.sourceCode?.trim();
```
dulajdilshan, line 530: **[Medium] Fragile parse — the shapes the `(` guard doesn't cover.** `/^([^=]*)=[^=]/` still mis-reads three statement forms I ran it on: `x += 1;` yields `"+"`, `[a, b] = f();` yields `"b]"`, and because `[^=]` matches newlines, a source whose first line is `// set x = 1` yields that comment's token.

None is the named-argument case @madushajg raised, so the `includes("(")` bail doesn't catch them; the cost is a node labelled with something that isn't a variable. Would anchoring the match to the first line and rejecting an LHS that isn't a plain identifier be worth it here — or is the honest fix to name it `getAgentRunResultVariableName` so its narrow domain is visible to the next caller of shared `utils/node.ts`?

packages/bi-diagram/src/visitors/SizingVisitor.ts
```diff
@@ -421,6 +420,13 @@ export class SizingVisitor implements BaseVisitor {
             return;
         }
 
+        // Reference mode (a run() call site) collapses to the same simple reference row
+        // AgentCallNode uses — no side circle columns are painted, so no side space is reserved.
+        if (this.isDurableAgentReference) {
+            this.setNodeSize(node, halfNodeWidth, halfNodeWidth, NODE_HEIGHT + AGENT_CALL_REFERENCE_HEIGHT);
```
dulajdilshan: **[Medium] Duplicated Code — two homes for one height.** `NODE_HEIGHT + AGENT_CALL_REFERENCE_HEIGHT` is computed here and again at `agentNodeLayout.ts:41`, which is the file that exists to own agent-node height arithmetic.

Both boxes are meant to be the same size, so the next time the reference row grows a row or a gap, one edit moves one box and the other silently drifts (DRY). What do you think about routing the durable reference case through `getAgentNodeContainerHeight` too, so the arithmetic lives in one place?

packages/bi-diagram/src/visitors/SizingVisitor.ts
```diff
@@ -55,7 +56,9 @@ import { Branch, FlowNode } from "../utils/types";
 export class SizingVisitor implements BaseVisitor {
     private skipChildrenVisit = false;
 
-    constructor() {
+    // True when durable-agent-run boxes are being sized for a run() call site rather than the
+    // agent's own declaration — see endVisitDurableAgentRun.
+    constructor(private isDurableAgentReference: boolean = false) {
```
dulajdilshan, line 60: **[Medium] Flag argument.** A positional boolean makes the call sites unreadable — `new SizingVisitor(true)` at `Diagram.tsx:179` and in the new test says nothing about what is true — and it bakes a *view-mode* fact (`FlowDiagram/index.tsx:4233` sets `durableAgentReference: !agentOnlyView`) into a geometry visitor.

Boolean parameters that switch a unit between two behaviours are the shape that makes the third mode add a second positional boolean. Would `constructor(options: { durableAgentReference?: boolean } = {})` be acceptable, so the next mode reads as a named field at every call site?

packages/bi-diagram/src/resources/constants.ts
```diff
@@ -167,6 +167,8 @@ export const AGENT_NODE_TOOL_SECTION_GAP = 120;
 export const AGENT_CALL_TOOL_SECTION_GAP = 80;
 export const AGENT_NODE_ADD_TOOL_BUTTON_WIDTH = 20;
 export const AGENT_CALL_AGENT_ROW_HEIGHT = 38;
+export const AGENT_NODE_USAGE_GAP = 10;
+export const AGENT_CALL_REFERENCE_HEIGHT = 58;
```
dulajdilshan, line 169: **[Medium] Dead code.** `AGENT_NODE_USAGE_GAP` has zero callers anywhere under `packages/` (grepped `*.ts`/`*.tsx`) — the `gap: 10px` it names is hard-coded at `AgentCallNodeWidget.tsx:165`. The same grep shows this PR's deletions also orphaned `AGENT_CALL_TOOL_SECTION_GAP` (`:167`) and `AGENT_CALL_AGENT_ROW_HEIGHT` (`:169`), which now have only their own declarations.

An exported constant advertises a contract; three that nothing consumes send the next reader looking for one. Use the new one in both `Column`s, or drop it and the two orphans?

packages/bi-diagram/src/components/nodes/AgentCallNode/AgentCallNodeWidget.tsx
```diff
@@ -673,344 +535,15 @@ export function AgentCallNodeWidget(props: AgentCallNodeWidgetProps) {
...
-                                        appearance="icon"
-                                        onClick={handleViewAgentClick}
-                                    >
-                                        <Icon name="bi-function-flow" sx={{ width: 16, height: 16 }} iconSx={{ fontSize: 16 }} />
-                                    </NodeStyles.MenuButton>
-                                </Tooltip>
-                            )}
-                        </AgentRow>
-                    )}
+                    <AgentReferenceRow
```
dulajdilshan, line 539: **[Spec · Medium] Partial implementation — the dead click.** `clickable` is true whenever the callback exists, but `goToAgentFromRunNode` (`AIChatAgent/utils.ts:475-476`) and `handleOnGoToDurableAgent` (`FlowDiagram/index.tsx:4089-4092`) both return silently when the agent declaration can't be resolved — I read both handlers; the cross-file case works, the unresolved case does nothing.

The requirement is *"an `Open Agent` button that redirects to the agent instance"*, and a button that looks enabled and does nothing is worse than one that isn't offered. How should this behave when the declaration is unresolved — disable the button, or surface the failure?

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
```diff
@@ -687,7 +710,7 @@ export function DurableAgentRunNodeWidget(props: DurableAgentRunNodeWidgetProps)
         if (item.kind === "peer") {
             // Delegating runs another durable agent, so it is marked as the agentic workflow it is
             // rather than as a plain tool function.
-            return <Icon name="bi-ai-agent" sx={{ fontSize: "24px" }} />;
+            return <NodeIcon type="AGENT_RUN" size={24} />;
```
dulajdilshan, line 713: **[Question] Why `AGENT_RUN` for the peer glyph?** `b87a622f42` swapped the raw `Icon` for `NodeIcon type="AGENT_RUN"` to pick up the shared cyan, which reads right — but `ddc1e23b70` in this same PR adds `"DURABLE_AGENT_RUN"` to that cyan group, and that's what the node's own `codedata.node` says.

Both resolve to the same colour today, so nothing is broken; I'm asking because if the two ever diverge the peer icon would follow the wrong one. Was `AGENT_RUN` deliberate, or just the first spelling that produced the right cyan?

packages/bi-diagram/src/utils/node.ts
dan-niles: Fixed in e5ae16e0de — both bail-out arms now return `undefined` instead of `properties.variable.value`. Verified: an uncaptured call renders no name instead of a phantom one.

packages/bi-diagram/src/utils/node.ts
dan-niles: Fair, but this helper only ever sees agent-run call-statement source, so `+=`/destructuring/comment-first-line can't actually occur here — skipping the rename/anchoring for now. Happy to revisit if this ever gets reused for a broader statement shape.

packages/bi-diagram/src/visitors/SizingVisitor.ts
dan-niles: Agreed this is DRY debt, but leaving it as-is for this PR — not routing it through getAgentNodeContainerHeight to keep this change scoped. Will follow up separately if the reference row size changes again.

packages/bi-diagram/src/visitors/SizingVisitor.ts
dan-niles: Good point, but skipping the options-object refactor here to keep this PR scoped to the sizing fix — will pick it up if/when a third mode is added.

packages/bi-diagram/src/resources/constants.ts
dan-niles: Confirmed AGENT_NODE_USAGE_GAP plus the two orphaned constants have no callers — leaving the cleanup out of this PR to keep it scoped, but agree it should happen. Will remove/consume in a follow-up.

packages/bi-diagram/src/components/nodes/AgentCallNode/AgentCallNodeWidget.tsx
dan-niles: Confirmed this is a real gap — canViewAgent is true whenever the callback exists, but resolution can still fail silently at click time. Leaving the UX call (disable vs. surface an error) for a follow-up since it needs a product decision, not folding it into this fix.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
dan-niles: AGENT_RUN was just the first spelling that produced the right cyan, not deliberate. Leaving as-is since both resolve identically today, but agree DURABLE_AGENT_RUN would be the more correct choice if they ever diverge.

dulajdilshan: Ack!

packages/bi-diagram/src/resources/constants.ts
dan-niles: Fixed in 1edb0453f3 — wired AGENT_NODE_USAGE_GAP into AgentCallNodeWidget's Column in place of the hard-coded `gap: 10px`, and removed the two orphaned constants outright since nothing consumed them (left DurableAgentRunNodeWidget's Column alone — its gap is 8px, not 10, so reusing the same constant there would've been a visual change, not a cleanup).

packages/bi-diagram/src/visitors/SizingVisitor.ts
dan-niles: Fixed in fec3edf48f — routed this through getAgentNodeContainerHeight(node, NodeTypes.AGENT_CALL_NODE), the same helper already used two lines above for the typed-agent case. Returns the identical value, no behavior change.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
```diff
@@ -910,7 +1010,7 @@ export function DurableAgentRunNodeWidget(props: DurableAgentRunNodeWidgetProps)
                                     <div className="connector-icon">{renderCapabilityIcon(item)}</div>
                                 </foreignObject>
 
-                                {!isAgentReference && <g
+                                <g
```
NipunaRanasinghe, line 1013: Dropping `!isAgentReference` here is right — the new early return at L894 makes this block unreachable in reference mode. But the identical guard survives on the right-column delete `<g>` (L1325) and on the affordance block (L1178). Both are now dead; worth removing them in the same pass so the cleanup isn't half-applied.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
```diff
@@ -633,15 +659,12 @@ export function DurableAgentRunNodeWidget(props: DurableAgentRunNodeWidgetProps)
         },
         { id: "goToSource", label: "Source", onClick: () => onGoToSource() },
         { id: "delete", label: "Delete", onClick: () => deleteNode() },
+        ...(isAgentBox && isAgentReference && agentNode?.onGoToAgent ? [{
```
NipunaRanasinghe, line 662: `isAgentBox &&` is redundant — `if (!isAgentBox) return` at L830 already guarantees it. Minor, but it means the "Open Agent" menu entry and the reference box at L894 are gated by two separate conditions that have to stay in sync.

packages/bi-diagram/src/components/Diagram.tsx
```diff
@@ -176,7 +176,7 @@ export function Diagram(props: DiagramProps) {
 
         const initVisitor = new InitVisitor(flowModel, currentExpandedErrorHandler);
         traverseFlow(flowModel, initVisitor);
-        const sizingVisitor = new SizingVisitor();
+        const sizingVisitor = new SizingVisitor(agentNode?.durableAgentReference === true);
```
NipunaRanasinghe, line 179: `agentNode` isn't in the deps of the effect that runs the visitors (`[model, showErrorFlow, expandedErrorHandler]`, L155). The widget re-renders when `agentNode` changes, but the layout doesn't re-run — so if `durableAgentReference` flips without `model` changing, the reserved box height and the rendered content disagree.

Cleaner fix: derive the flag inside `endVisitDurableAgentRun` from node metadata, where the sibling decision (`nodeMetadata.agentBox`) already lives, instead of threading a diagram-wide boolean through the constructor.

packages/bi-diagram/src/utils/node.ts
```diff
@@ -522,6 +522,19 @@ export function getNodeTitle(node: FlowNode) {
     return label;
 }
 
+// getFlowModel always requests forceAssign, so properties.variable can carry a suggested name that
+// doesn't match the source — for an uncaptured call, a discard (`_ = ...`), or a plain reassignment
+// to an existing variable (`draft = ...`, no type). The statement's own source is the ground truth
+// for what it assigns to, so an uncaptured call renders no name rather than a suggested one.
+export function getResultVariableName(node: FlowNode): string | undefined {
+    const source = node.codedata?.sourceCode?.trim();
+    const lhs = source?.match(/^([^=]*)=[^=]/)?.[1]?.trim();
```
NipunaRanasinghe, line 531: This is the only non-visual logic in the PR and it's untested — the sizing change got a test, this didn't. The `(` guard for uncaptured calls is neat but fragile enough to need one.

Cases worth pinning:
- `_ = agent->run(...)` returns `"_"`, but the comment above lists discard as a case that "renders no name". Comment and behavior disagree.
- `x += 1` → `"+"` (`[^=]*` swallows the `+`). Not reachable for agent nodes today, but this is exported from general utils under a general name.
- `[a, b] = ...` → `"b]"`.

packages/bi-diagram/src/components/nodes/AgentWidget/agentNodeLayout.ts
```diff
@@ -38,8 +38,7 @@ const layoutStrategies = {
         const descriptionHeight = hasPrompt ? 115 : agentInfo?.description ? 95 : 0;
         return Math.max(NODE_HEIGHT + memoryHeight + descriptionHeight, NODE_HEIGHT + AGENT_NODE_TOOL_SECTION_GAP + toolHeight);
     },
-    [NodeTypes.AGENT_CALL_NODE]: (toolHeight: number) => NODE_HEIGHT + AGENT_CALL_TOOL_SECTION_GAP
-        + AGENT_NODE_TOOL_GAP * 2 + 38 + toolHeight,
+    [NodeTypes.AGENT_CALL_NODE]: () => NODE_HEIGHT + AGENT_CALL_REFERENCE_HEIGHT,
```
NipunaRanasinghe, line 41: This always adds `AGENT_CALL_REFERENCE_HEIGHT`, but `AgentReferenceRow` returns `null` when there's no label and no `goToAgent` (AgentReferenceRow.tsx L95). In that case the box reserves 58px for a row that never renders, leaving an empty band at the bottom.

packages/bi-diagram/src/components/nodes/AgentWidget/AgentReferenceRow.tsx
```diff
@@ -0,0 +1,126 @@
...
+    onButtonHoverChange: (hovered: boolean) => void;
+};
+
+// Read-only agent metadata (model/tools/memory) lives in the property panel now; this row only opens the agent.
+export function AgentReferenceRow({ label, buttonLabel = "Open Agent", clickable, onOpen, onButtonHoverChange }: AgentReferenceRowProps) {
+    if (!label && !clickable) {
+        return null;
+    }
+    const handleKeyDown = (event: React.KeyboardEvent) => {
+        if (event.key === "Enter" || event.key === " ") {
```
NipunaRanasinghe, line 99: Nit: Space is handled without `event.preventDefault()`, so pressing it also scrolls the diagram container.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
```diff
@@ -864,6 +887,83 @@ export function DurableAgentRunNodeWidget(props: DurableAgentRunNodeWidgetProps)
...
+                                <NodeIcon type={model.node.codedata.node} size={24} />
+                                <NodeStyles.RunBadge>
+                                    <Icon name="bi-play" iconSx={{ fontSize: "20px" }} sx={{ color: "var(--vscode-charts-green)", display: "flex", justifyContent: "center", alignItems: "center" }} />
+                                </NodeStyles.RunBadge>
+                            </NodeStyles.IconBox>
+                            <NodeStyles.Row readOnly={readOnly}>
+                                <NodeStyles.Header onClick={handleOnClick}>
+                                    <NodeStyles.Title>durable agent : run</NodeStyles.Title>
+                                    <NodeStyles.Description>
+                                        {getResultVariableName(model.node) as ReactNode}
```
NipunaRanasinghe, line 936: Nit: `as ReactNode` is unnecessary — `getResultVariableName` returns `string | undefined`, already a valid `ReactNode`. `AgentCallNodeWidget` uses it uncast.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
```diff
@@ -864,6 +887,83 @@ export function DurableAgentRunNodeWidget(props: DurableAgentRunNodeWidgetProps)
...
+                    disabled={disabled}
+                    hovered={isBoxHovered && !isOpenAgentHovered}
+                    hasError={hasError}
+                    readOnly={readOnly}
+                    isActiveBreakpoint={isActiveBreakpoint}
+                    isSelected={isSelected}
+                    onMouseEnter={() => setIsBoxHovered(true)}
+                    onMouseLeave={() => setIsBoxHovered(false)}
+                    onContextMenu={!readOnly ? handleOnContextMenu : undefined}
+                    title="Configure Run"
```
NipunaRanasinghe, line 907: `title="Configure Run"` covers the whole box, but there's no `onClick` on it — only the icon and header open the form. `AgentCallNodeWidget` puts `onClick={!readOnly ? handleOnClick : undefined}` on its box (L436). Inconsistent hit target for the same affordance.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
dan-niles: Removed the dead `!isAgentReference` guard at both spots (the affordances block and the right-column delete `<g>`), same as the L1013 cleanup.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
dan-niles: Leaving this one as-is for now — I can't confirm `isAgentBox` is false only when `isAgentReference` is also false, so dropping the check risks surfacing "Open Agent" on the compact buildAndRun node. Will revisit once that's confirmed.

packages/bi-diagram/src/components/Diagram.tsx
dan-niles: Added `agentNode?.durableAgentReference` to the effect's deps for now in  in https://github.com/ballerina-platform/ballerina-vscode/pull/1079/commits/429de9aa409207ff73850ffd702227adfa5efdcd.

packages/bi-diagram/src/utils/node.ts
dan-niles: Fixed (in https://github.com/ballerina-platform/ballerina-vscode/pull/1079/commits/429de9aa409207ff73850ffd702227adfa5efdcd) the discard mismatch — `_ = ...` now returns undefined like the comment says, and added a test file (getResultVariableName.test.ts) covering captured/uncaptured/discard/reassignment. Left the `+=` and destructuring cases alone since they're unreachable for agent nodes today, as you noted.

packages/bi-diagram/src/components/nodes/AgentWidget/agentNodeLayout.ts
dan-niles: Confirmed this is reachable for an AGENT_CALL node with no connection set yet. Fixed by only reserving AGENT_CALL_REFERENCE_HEIGHT when AgentReferenceRow will actually render (mirrors the widget's own label/clickable check) in https://github.com/ballerina-platform/ballerina-vscode/pull/1079/commits/429de9aa409207ff73850ffd702227adfa5efdcd.

packages/bi-diagram/src/components/nodes/AgentWidget/AgentReferenceRow.tsx
dan-niles: Fixed in https://github.com/ballerina-platform/ballerina-vscode/pull/1079/commits/429de9aa409207ff73850ffd702227adfa5efdcd, added preventDefault() on Space.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
dan-niles: Removed the unnecessary cast in https://github.com/ballerina-platform/ballerina-vscode/pull/1079/commits/429de9aa409207ff73850ffd702227adfa5efdcd.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
dan-niles: Fixed in https://github.com/ballerina-platform/ballerina-vscode/pull/1079/commits/429de9aa409207ff73850ffd702227adfa5efdcd, added onClick={!readOnly ? handleOnClick : undefined} to match AgentCallNodeWidget's box.

packages/bi-diagram/src/components/nodes/AgentWidget/agentNodeLayout.ts
```diff
@@ -38,13 +38,17 @@ const layoutStrategies = {
         const descriptionHeight = hasPrompt ? 115 : agentInfo?.description ? 95 : 0;
         return Math.max(NODE_HEIGHT + memoryHeight + descriptionHeight, NODE_HEIGHT + AGENT_NODE_TOOL_SECTION_GAP + toolHeight);
     },
-    [NodeTypes.AGENT_CALL_NODE]: (toolHeight: number) => NODE_HEIGHT + AGENT_CALL_TOOL_SECTION_GAP
-        + AGENT_NODE_TOOL_GAP * 2 + 38 + toolHeight,
-} satisfies Record<AgentWidgetType, (toolHeight: number, agentInfo?: NodeMetadata["agentInfo"]) => number>;
+    [NodeTypes.AGENT_CALL_NODE]: (_toolHeight: number, _agentInfo: NodeMetadata["agentInfo"] | undefined, node?: FlowNode) => {
+        const hasReferenceRow =
+            node?.codedata?.node !== "AGENT_CALL" ||
```
NipunaRanasinghe, line 43: Non-blocking follow-up — this covers `AGENT_CALL` but not `AGENT_RUN`.

`AGENT_CALL_NODE_KINDS = ["AGENT_CALL", "AGENT_RUN"]` (`ballerina-core/src/flow-model/flow-model-utils.ts:27`) and `endVisitAgentRun` delegates to `endVisitAgentCall` (`SizingVisitor.ts:372`), so an `AGENT_RUN` node takes the `!== "AGENT_CALL"` branch and always reserves 58px — while `AgentCallNodeWidget` still passes an empty `label` and `clickable === false`, so `AgentReferenceRow` returns `null`. Same empty band, other node kind.

Keying off what the widget actually branches on (connection non-empty) and letting the durable-reference caller opt in explicitly would avoid having to keep this literal in sync with the kind list.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
NipunaRanasinghe, line 662: Non-blocking, carried over: `isAgentBox &&` is still redundant given the `!isAgentBox` early return at L830. Fine to leave as-is.

NipunaRanasinghe (approved):
Approving
Two non-blocking notes inline; neither needs to hold the merge.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/DurableAgentRunNodeWidget.tsx
dan-niles: Checked this properly — confirmed `isAgentBox &&` is load-bearing, not redundant. `isAgentBox === false` (the compact node) only ever renders when `agentOnlyView` is false, since `agentOnlyView` filters `displayModel` down to just `[Start, agent box]` (FlowDiagram/index.tsx L4167-4173) whenever it's true. So wherever the compact node exists, `isAgentReference` (`!agentOnlyView`) is always true, and `agentNode.onGoToAgent` is wired unconditionally — dropping `isAgentBox &&` would put a spurious "Open Agent" entry on every compact buildAndRun node's menu. Keeping the check as-is.

SasinduDilshara (approved):

PR #1083, opened 2026-09-11
Add Librarian and LibraryResearcher subagents for Copilot library resolution

IsuruMaduranga:
## Purpose
The WSO2 Integrator Copilot's three library tools (`LibraryGetTool`, `LibrarySearchTool`, `HealthcareLibraryProviderTool`) put rendered library documentation straight into the main agent's context and filtered it with a one-shot Haiku call. The alpha test tracker recorded the failures this produced: class-member APIs reported as non-existent (xlsx), the `@csv:Name` binding annotation never discovered because it lives in `ballerina/data.csv` rather than the connector (ftp, smb, xlsx, Azure Files), silent defaults accepted (AWS SNS target type), library functions hand-rolled (`time:dateValidate`, `civilAddDuration`), inverted claims about library semantics (jwt), and the same docs re-fetched every prompt.

Related issue: https://github.com/wso2/product-integrator/issues/2371

## Goals
- Library lookup runs in a subagent with its own context, so rendered docs never enter the main agent's window; the main agent receives a compact, verbatim report.
- Full library API surface on disk (no filter), including class/client/resource member docs, greppable by the subagent.
- Transitive binding modules, silent defaults and existing library functions are reported explicitly, and "does X do Y" questions are answered from the docs or declared unstated.
- Lookups for independent systems run in parallel; a running subagent shows what it is doing instead of a static row.

## Approach
**Subagents.** A generic `Subagent` tool spawns one of two types: `Librarian` (docs only) and `LibraryResearcher` (docs plus web tools for libraries with thin metadata). Each runs a real `generateText` loop (Sonnet 5, 50-step guard, 16K output) with tools `library_search`, `library_docs` (renders a library's full API to `~/.ballerina/copilot/lib-docs/<version>/<org>__<name>.md`, cached per distribution), and path-scoped `docs_grep` / `docs_read` / `docs_list`. Runs are foreground or background; `task_output` waits for (and is woken by) a background task, `kill_task` stops one, background tasks die at run end, and completions are announced through a `<system-reminder>` appended to the next tool result. History persists under `threads/<id>/subagents/<taskId>/` so a report's `Resume id` continues the conversation. The report has a fixed shape (Libraries, API surface, Usage notes, Defaults that matter, Use these instead of hand-rolling, Gaps, Doc files).

**Renderer.** `to-syntax-string.ts` now emits `# + param` / `# + return` docs for class, client and resource methods, as it did for module-level functions. The omission hid facts such as `time:Zone.utcFromCivil` erroring when `civil.timeAbbrev` is missing.

**Parallel briefs.** The main-agent prompt leaves the split to the model: one Librarian per independent system in the same message (at most 3; the AI SDK runs a step's tool calls concurrently), one brief when systems are related (a listener and its binding module, a client and its auth module). `ensureLibraryDocs` shares one in-flight fetch per library.

**Progress in the UI.** The runner reports each step that made tool calls and the Subagent tool re-emits it as a `tool_result` on the same call id with `partial: true`, `status: "running"`, the activity and the step. `ToolResult` gains the tool-agnostic `partial` field, carried onto the persisted transcript item, so both chat surfaces keep the row spinning ("Consulting the Librarian — reading ballerinax/aws.sns docs (step 3)…"), the composer indicator re-labels, and the status bar moves ("Librarian: reading ballerinax/aws.sns docs"). A background start is `status: "started"`. The prompt also asks for one sentence to the user before a foreground brief.

**Prompt rules** (Librarian): follow `// Special Agent Note` markers including on Verify questions; fetch the `ballerina/data.*` binding module when a signature binds payloads; list defaults that change behaviour; name existing functions instead of hand-rolling; answer Verify questions only from the docs, with a hard stop for absent facts and no leanings; never leave a Gap for an unread section; language rules are not docs.

## UI Component Development
- [ ] Added reusable UI components to the ui-toolkit — N/A: no new components; the change re-words existing tool rows and status labels.
- [x] Use ui-toolkit components wherever possible — existing `Codicon` and row components only.
- [x] Matches with the native VSCode look and feel.

## User stories
- As a developer, when I ask the Copilot to build an integration, it uses the real library API (member methods, binding annotations, correct defaults) instead of hand-rolling or inventing it.
- As a developer, I can see what the Copilot's library lookup is doing while it runs, and lookups for several connectors run at once.

## Release note
The WSO2 Integrator Copilot now resolves libraries through dedicated Librarian subagents that read the full library documentation, report binding annotations, defaults and existing functions verbatim, run in parallel for independent systems, and show their progress in the chat.

## Documentation
N/A — no user-facing configuration; the behaviour change is internal to the Copilot's library lookup and the wording of tool rows.

## Training
N/A

## Certification
N/A — no change to certified user-facing features.

## Marketing
N/A

## Automation tests
- Unit tests: `packages/ballerina-extension` `pnpm run test:unit` — 35 suites / 334 tests (new: `libDocsCache`, `docsFsTools`, `subagentStore`, `subagentBackground`, `subagentHelpers`, `subagentProgress`, `toSyntaxStringMethodDocs`, `agentStatusLabels`); `packages/ballerina-visualizer` AI suites — 64 tests (`streamSerialization`, `toolDisplay`, `promptHandoff`). `tsc` clean in both packages; tslint clean.
- Integration tests: verified headless against the repo's language server and Sonnet 5 with 11 briefs, one per library failure mode in the alpha tracker (xlsx member methods, CSV `@csv:Name` via data.csv, jwt verify, SNS defaults, time arithmetic, email `ImapListener`, ftp/smb header-row semantics, jwt secondary claims, `jwt:Payload` shape, GitHub pagination/serviceUrl, Service Bus listener). All 11 pass at the provider's default thinking effort; four concurrent LS renders complete in 6.2 s; two parallel Librarians finish in 68 s against 110 s sequential. The rig is local (uses a real API key) and is not part of this PR.

## Security checks
- Ran FindSecurityBugs plugin and verified report? N/A (TypeScript; no Java changes)
- Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes. `docs_read`/`docs_grep` are path-confined to the docs cache directory (symlinks resolved); `Config.toml` remains unreadable to the agent.

## Samples
N/A

## Related PRs
N/A

## Migrations (if applicable)
Library docs caches rendered by a build before this change (under `~/.ballerina/copilot/lib-docs/`) lack method parameter/return docs; the cache is keyed by extension and distribution version, so a new extension version renders fresh files. No user action.

## Test environment
macOS 15 (Darwin 24.2), JDK 21 (language server), Node 22.3 (rush build) and 26.3 (tests), Ballerina 2201.13.3.

## Learning
The MI Copilot's subagent framework (`agents/subagents/`, `tools/subagent_tool.ts`) is the reference for the tool shape, background runs and resume. One finding while porting: in the Vercel AI SDK, each step's `response.messages` is the cumulative list, so concatenating per-step lists duplicates history triangularly; this change takes `result.response.messages` instead.

## Summary

- Replaced direct library tools with resumable `Librarian` and `LibraryResearcher` subagents.
- Added cached library documentation with API details, bindings, defaults, parameters, and return values.
- Added scoped documentation search and read tools with bounded output and validated paths.
- Added background execution, parallel lookups, progress updates, task control, persistence, resume support, and run cleanup.
- Updated the AI panel and mini chat to display subagent activity, partial results, and task status.
- Improved subagent usage tracking without emitting additional usage events.
- Improved documentation pattern validation and library selection consistency.
- Added unit and UI coverage for caching, subagent workflows, task handling, progress updates, documentation rendering, and transcript serialization.


packages/ballerina-extension/src/features/ai/agent/subagents/librarian/system.ts
```diff
@@ -0,0 +1,112 @@
...
+    LIBRARY_DOCS_TOOL_NAME,
+    LIBRARY_SEARCH_TOOL_NAME,
+} from "./tools";
+
+
+export const LIBRARIAN_REPORT_FORMAT = `## Libraries
+- org/name — one line on why it was chosen (alternatives ruled out: …)
+
+## API surface
+Per library, only what the brief needs, verbatim from the docs files in \`\`\`ballerina blocks: client init and the config record fields that matter, the specific remote/resource functions, request/response record types with their load-bearing fields, error types, listener and service signatures when the brief listens for events.
```
SasinduDilshara, line 31:
```suggestion
Per library, only what the brief needs, verbatim from the docs files in \`\`\`ballerina blocks: client init and the config record fields that matter, the specific remote/resource functions, request/response record types with their load-bearing fields, error types, classes, objects, annotations listener and service signatures when the brief listens for events.
```

packages/ballerina-extension/src/features/ai/agent/subagents/runner.ts
```diff
@@ -0,0 +1,117 @@
...
+        model,
+        messages,
+        tools: definition.buildTools(params.ctx),
+        stopWhen: stepCountIs(SUBAGENT_MAX_STEPS),
+        maxOutputTokens: SUBAGENT_MAX_OUTPUT_TOKENS,
+        abortSignal: params.abortSignal,
+        providerOptions: buildReasoningProviderOptions(params.reasoning),
+        // The final step (report text, no tool calls) is followed at once by the completion result, so
+        // only steps that called tools are worth announcing.
+        onStepFinish: onProgress ? (step) => {
```
RNViththagan, line 91: Traced this one against the installed SDK (`ai@6.0.103`): when a background subagent's run ends mid-tool-call (thread switched away from, checkpoint restore, etc.), `cleanupRunningBackgroundSubagents` fires the terminal `status:"aborted"` result via `onRunEnd`, aborts, and deletes the map entry - but the SDK's own `finish-step` handling (`dist/index.js` around the transform's `finish-step` branch) calls `await notify({ event: currentStepResult, callbacks: onStepFinish })` unconditionally, with no check against the abort signal. So a tool call that was already in flight (e.g. `library_docs` rendering, which can run up to ~180s) still finishes, `onStepFinish` still fires here, and `onProgress` -> `emitProgress` (subagent-tool.ts) emits a fresh `status:"running"` partial on the *same* `toolCallId` as the terminal result already sent - with nothing checking `entry.runEnded`/the abort signal before emitting it.

If the webview doesn't independently refuse to let a partial supersede a final result on the same call id, that would flip a finished row back to "running" for good, since the subagent's real completion (or its AbortError) is discarded once `entry.runEnded` is true, so nothing ever corrects it. Concretely: switch threads (or restore a checkpoint) while a Librarian is mid-render, and that turn's tool row can be left permanently spinning even though the run is long over. Shall we gate `emitProgress`/the `onStepFinish` callback on `abortController.signal.aborted` (or `entry.runEnded`) here?

packages/ballerina-extension/src/features/ai/agent/tools/subagent-tool.ts
```diff
@@ -0,0 +1,266 @@
...
+import { markForegroundRun, validateResume } from "../subagents/resume";
+
+export const SubagentInputSchema = z.object({
+    description: z.string().describe("A short (3-5 word) description of what the subagent will do. Shown to the user."),
+    prompt: z.string().describe("The brief: what the integration must do (systems, protocols, operations, constraints, payload shapes), or the precise question. Purpose-written for the subagent, never the raw user request."),
+    subagent_type: z.enum(SUBAGENT_TYPES).describe(`The subagent to run:\n${describeSubagentTypes()}`),
+    model: z.enum(["sonnet", "haiku"]).optional().describe("Defaults to sonnet. Use haiku only for a narrow, cheap lookup (one named library, one signature)."),
+    run_in_background: z.boolean().optional().describe(`Run in the background and return a task id immediately. Only when you have independent work to do meanwhile; check the result with ${TASK_OUTPUT_TOOL_NAME} (block=true when you run out of work) or stop it with ${KILL_TASK_TOOL_NAME}.`),
+    resume: z.string().optional().describe("A task id from an earlier run (task-subagent-xxxxxxxx) to continue with a follow-up question instead of starting over."),
```
RNViththagan, line 54: `resume` is a bare string with no shape validation, and it flows straight into `getSubagentDir` (subagents/store.ts) which just `path.join(threadDir, "subagents", subagentId)`s it with no containment check - unlike `docs-fs-tools.ts`'s `resolveDocsPath`, which does realpath + relative-path checks for the same kind of boundary. A `resume` like `../../<otherThreadId>/subagents/<id>` loads (and, after the run, overwrites via `saveSubagentRun`) another thread's subagent history.

The fix already exists in this PR and just isn't wired in: `isSubagentId()` (subagents/types.ts) matches exactly the `task-subagent-[0-9a-f]{8}` shape and is unit-tested in `subagentStore.test.ts`, but has zero call sites in the actual code path. Would validating `input.resume` against it here (and/or inside `getSubagentDir`) before it touches the filesystem be enough to close this?

packages/ballerina-extension/src/features/ai/agent/subagents/resume.ts
```diff
@@ -0,0 +1,78 @@
...
+    const running = getBackgroundSubagent(resumeId);
+    if (running && !running.completed) {
+        return {
+            kind: "error",
+            error: "SUBAGENT_STILL_RUNNING",
+            message: `Subagent ${resumeId} is still running. Wait for it with ${TASK_OUTPUT_TOOL_NAME} or stop it with ${KILL_TASK_TOOL_NAME} before resuming.`,
+        };
+    }
+    const metadata = loadSubagentMetadata(threadDir, resumeId);
+    if (metadata && metadata.subagentType !== requestedType) {
```
RNViththagan, line 63: When `metadata` is `null` (missing or corrupt `metadata.json`), this check is skipped entirely and the resume proceeds as whatever `subagent_type` was requested, regardless of what actually produced the history - so a Librarian's saved conversation could get resumed as a LibraryResearcher and pick up live web tools it never had, purely because its metadata file didn't survive.

It's not even a one-time slip: after this, the same run's `saveSubagentRun` rewrites `metadata.json` with the *requested* type, so the mis-typing becomes permanent for every future resume of that id too. Would treating a null `metadata` the same as `SUBAGENT_NOT_FOUND` rather than falling through be the right call here?

packages/ballerina-extension/src/features/ai/agent/subagents/definitions.ts
```diff
@@ -0,0 +1,70 @@
...
+        buildTools: () => createLibrarianTools(),
+        followUpHint: "Re-grep the doc files you already have rather than re-fetching them.",
+    },
+    LibraryResearcher: {
+        summary: "The librarian with web research. Use when the Librarian reported a gap, when a library's docs are thin, or when you need examples or real-world behaviour (auth quirks, rate limits, per-operation service URLs). Reads the docs first, then GitHub module sources, Central and the upstream service's docs; cites sources.",
+        system: researcherSystemPrompt,
+        buildTools: (ctx) => ({
+            ...createLibrarianTools(),
+            // `true` skips the per-use approval banner: the subagent's web access is always on (owner decision).
+            web_search: createWebSearchTool(ctx.eventHandler, true),
```
RNViththagan, line 50: Passing `true` here skips `approvalManager.requestWebToolApproval` entirely - not just a banner, the actual consent gate the main agent's own `web_search`/`web_fetch` honour today, driven by the user's per-request `webSearchEnabled` toggle. So a user who's turned web access off for this chat can still have it happen invisibly: the main agent just spawns a LibraryResearcher instead, and whatever that fetches flows straight into the report with no sanitization - which `agent/prompts.ts` then tells the main agent to treat as its most-trusted source ("use them verbatim"). Since the tools this replaces were docs-only with no web reach at all, this is a new path to that behaviour, not a pre-existing one.

Would having the subagent inherit the run's actual `webSearchEnabled` flag (rather than hardcoding it on) be worth doing here, so the user's toggle still means something?

packages/ballerina-extension/src/features/ai/utils/libs/docs-fs-tools.ts
```diff
@@ -0,0 +1,163 @@
...
+    const out: string[] = [];
+    let matches = 0;
+    let truncated = false;
+    for (const file of files) {
+        if (truncated) { break; }
+        const lines = fs.readFileSync(file, "utf8").split("\n");
+        const label = path.basename(file);
+        let lastPrinted = -1;
+        for (let i = 0; i < lines.length; i++) {
+            if (!regex.test(lines[i])) { continue; }
```
RNViththagan: +1 to CodeRabbit's ReDoS flag on the pattern above - worth adding two things: `truncateLine` is only applied to lines that already matched, for display (line 118 here, and readDocs:149), so this `.test()` call itself runs against the full, untruncated line; and because it's synchronous on the extension host's own thread, a pathological pattern doesn't just slow this one grep down, it freezes the whole extension - the LS client and every open webview - until VS Code offers to restart the host. Neither `abortSignal` nor `kill_task` can interrupt a synchronous call already in progress. Given the Librarian's own prompt (librarian/system.ts) encourages alternation-heavy grep patterns, this is a realistic thing for the model to write on its own, no adversarial input needed.

packages/ballerina-extension/src/features/ai/utils/libs/lib-docs-cache.ts
```diff
@@ -0,0 +1,320 @@
...
+    return `${safe}.md`;
+}
+
+/** Version-keyed cache directory; both the extension and the distribution shape the library JSON. */
+export function getLibDocsDir(versionKey: string): string {
+    const safeKey = versionKey.replace(/[^A-Za-z0-9._-]/g, "_") || "unknown";
+    return path.join(LIB_DOCS_ROOT, safeKey);
+}
+
+export function buildVersionKey(extensionVersion: string | undefined, distributionVersion: string | undefined): string {
```
RNViththagan, line 76: Minor/non-blocking: the cache key is only extension version + distribution version, so once a library's docs are rendered they're reused as-is until one of those bumps - a connector updated on Central next week stays invisible to the Librarian until then, with no `refresh` signal in the prompt to tell the model when that might matter. The tools this replaces always fetched fresh, so this is a small step backwards on staleness specifically (everything else about the caching looks solid, including the in-flight-dedup). Worth a soft mtime TTL at some point, not blocking this PR.

packages/ballerina-extension/src/features/ai/agent/subagents/resume.ts
```diff
@@ -0,0 +1,78 @@
...
+        return {
+            kind: "error",
+            error: "SUBAGENT_TYPE_MISMATCH",
+            message: `Cannot resume: ${resumeId} is a ${metadata.subagentType} subagent, but subagent_type=${requestedType} was requested. Use subagent_type=${metadata.subagentType}.`,
+        };
+    }
+    try {
+        return { kind: "ok", messages: loadSubagentHistory(threadDir, resumeId), description: metadata?.description ?? "" };
```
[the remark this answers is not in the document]
IsuruMaduranga: Yeah, null metadata bails with not-found now instead of falling through. 7c9c324

packages/ballerina-extension/src/features/ai/agent/subagents/store.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export class SubagentNotFoundError extends Error {
+    constructor(id: string) {
+        super(`Subagent with ID ${id} not found. Cannot resume.`);
+        this.name = "SubagentNotFoundError";
+    }
+}
+
+export function getSubagentDir(threadDir: string, subagentId: string): string {
+    return path.join(threadDir, SUBAGENTS_DIR, subagentId);
```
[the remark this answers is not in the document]
IsuruMaduranga: Good catch. isSubagentId check goes in getSubagentDir now, so every read/write is covered — plus at the resume boundary.

packages/ballerina-extension/src/features/ai/agent/tools/task-tools.ts
```diff
@@ -0,0 +1,127 @@
...
+    return tool({
+        description: `Get the result of a background subagent by task id. block=true (default) waits until it completes — use this when you have no other work to do; you are woken as soon as the report is ready. block=false returns the current status immediately. Not for foreground runs (their report was returned inline).`,
+        inputSchema: z.object({
+            task_id: z.string().describe("The task id returned by Subagent with run_in_background=true."),
+            block: z.boolean().optional().describe("Wait for completion (default true)."),
+            timeout: z.number().int().optional().describe(`Maximum wait in milliseconds when block=true (default ${TASK_OUTPUT_DEFAULT_BLOCK_MS}, max ${TASK_OUTPUT_MAX_BLOCK_MS}).`),
+        }),
+        execute: async ({ task_id, block = true, timeout = TASK_OUTPUT_DEFAULT_BLOCK_MS }, options?: { toolCallId?: string; abortSignal?: AbortSignal }) => {
+            const toolCallId = options?.toolCallId ?? `fallback-${Date.now()}`;
+            const task = getBackgroundSubagent(task_id);
```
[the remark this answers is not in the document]
IsuruMaduranga: Passing runKey in now, a task from another run reads as not found.

packages/ballerina-extension/src/features/ai/agent/tools/task-tools.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export function createKillTaskTool(eventHandler: CopilotEventHandler) {
+    return tool({
+        description: "Terminate a running background subagent by task id.",
+        inputSchema: z.object({
+            task_id: z.string().describe("The task id to terminate."),
+        }),
+        execute: async ({ task_id }, options?: { toolCallId?: string }) => {
+            const toolCallId = options?.toolCallId ?? `fallback-${Date.now()}`;
+            const task = getBackgroundSubagent(task_id);
```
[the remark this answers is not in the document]
IsuruMaduranga: Same here, scoped to runKey.

packages/ballerina-extension/src/features/ai/utils/libs/docs-fs-tools.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+function truncateLine(line: string): string {
+    return line.length > GREP_MAX_LINE_CHARS ? `${line.slice(0, GREP_MAX_LINE_CHARS)}…` : line;
+}
+
+export function grepDocs(dir: string, input: DocsGrepInput): string {
+    let regex: RegExp;
+    try {
+        regex = new RegExp(input.pattern, input.case_sensitive ? "" : "i");
```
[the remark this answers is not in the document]
IsuruMaduranga: Added a star-height + backreference check that rejects the pattern before it runs. Normal patterns still go through.

packages/ballerina-extension/src/features/ai/utils/libs/to-syntax-string.ts
```diff
@@ -642,6 +642,32 @@ function renderParam(param: Parameter, externalLinks: ExternalLinkInfo[]): strin
...
+ * functions always had these lines (`renderStandaloneFunction`).
+ */
+function renderFunctionDocs(func: { description?: string; parameters: Parameter[]; return?: RemoteFunction["return"] }, indent: string): string {
+    const lines: string[] = [];
+    if (func.description) {
+        lines.push(...func.description.split("\n").map((l) => `${indent}# ${l}`));
+    }
+    for (const param of func.parameters) {
+        if (param.description) {
+            lines.push(`${indent}# + ${param.name} - ${param.description}`);
```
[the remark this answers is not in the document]
IsuruMaduranga: Done, each line gets its own # prefix now.

packages/ballerina-extension/src/test-support/libDocsCache.test.ts
```diff
@@ -0,0 +1,208 @@
...
+        const toc = libraryToc(http);
+        expect(toc).toMatch(/^- Functions \(1\): authenticateResource/m);
+        expect(toc).toMatch(/^- Types \(\d+\): CacheConfig, /m);
+        expect(toc).toMatch(/^- Services \(1\): /m);
+    });
+
+    it("caps names per category and says how many more there are", () => {
+        const many: Library = {
+            name: "x/many", description: "", clients: [], typeDefs: [],
+            functions: Array.from({ length: TOC_NAME_LIMIT + 7 }, (_, i) => ({ name: `fn${i}`, type: "function", description: "", parameters: [], return: { type: { name: "()" } } } as any)),
```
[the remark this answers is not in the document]
IsuruMaduranga: Dropped the cast, using RemoteFunction.

packages/ballerina-extension/src/test-support/subagentBackground.test.ts
```diff
@@ -0,0 +1,148 @@
...
+describe("withBackgroundNotifications", () => {
+    it("appends the reminder to string results and attaches it to object results", async () => {
+        registerBackgroundSubagent(entry("s", { completed: true, success: true }));
+        const stringTool = withBackgroundNotifications({ description: "d", inputSchema: {} as any, execute: async () => "ok" } as any, "run-a");
+        const out = await stringTool.execute!({}, {} as any);
+        expect(out).toMatch(/^ok\n\n<system-reminder>/);
+
+        registerBackgroundSubagent(entry("o", { completed: true, success: true }));
+        const objectTool = withBackgroundNotifications({ description: "d", inputSchema: {} as any, execute: async () => ({ status: "done" }) } as any, "run-a");
+        const obj = await objectTool.execute!({}, {} as any) as any;
```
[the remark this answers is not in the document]
IsuruMaduranga: Replaced these with real tool() fixtures.

packages/ballerina-extension/src/features/ai/agent/subagents/librarian/system.ts
IsuruMaduranga: Applied, thanks.

packages/ballerina-extension/src/features/ai/agent/subagents/runner.ts
IsuruMaduranga: Nice catch. Gated onStepFinish on the abort signal so the terminal row can't get flipped back to running.

packages/ballerina-extension/src/features/ai/agent/tools/subagent-tool.ts
IsuruMaduranga: Wired isSubagentId in at getSubagentDir + the resume boundary, so a ../ id can't escape the thread dir.

packages/ballerina-extension/src/features/ai/agent/subagents/resume.ts
IsuruMaduranga: Treating null metadata as not-found now.

packages/ballerina-extension/src/features/ai/agent/subagents/definitions.ts
IsuruMaduranga: This one's intentional — the researcher's web access is always on by design. Left a note in the code so it doesn't read as an oversight.

packages/ballerina-extension/src/features/ai/utils/libs/lib-docs-cache.ts
IsuruMaduranga: Agreed it's a gap on staleness. Leaving it out of this PR though, will do the mtime TTL as a follow-up.

packages/ballerina-core/src/state-machine-types.ts
```diff
@@ -516,6 +516,12 @@ export interface ToolResult {
     toolOutput?: any;
     toolCallId?: string;
     failed?: boolean;
+    /**
+     * A progress report, not the call's final result: the tool will send another `tool_result` with the
+     * same `toolCallId` later (a subagent heartbeat, a background task that just started). Renderers keep
+     * the row in its running state; the terminal result carries no `partial`.
+     */
```
samithkavishke, line 523: Do we need this comment

IsuruMaduranga: Worth keeping, I think — `partial` isn't self-explanatory. The tool sends a second `tool_result` with the same `toolCallId` later (a subagent heartbeat, or a background task that just started), and the chat row keys its "still running" state off it, so the docs save a trip through the renderers. Happy to trim it to two lines if it reads long.

packages/ballerina-extension/src/features/ai/utils/libs/docs-fs-tools.ts
```diff
@@ -0,0 +1,217 @@
...
+            continue;
+        }
+        if (c === "{") {
+            const bound = /^\{\d+(?:,\d*)?\}/.exec(pattern.slice(i));
+            if (!bound) { continue; }
+            groupHasQuantifier[groupHasQuantifier.length - 1] = true;
+            i += bound[0].length - 1;
+        }
+    }
+    return true;
```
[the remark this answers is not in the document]
IsuruMaduranga: Fair — it did slip through: star height doesn't catch ambiguous alternation, and `(a|aa)+$` really does blow up. Fixed in a2f45cf. The guard now records each alternative's first characters and rejects a quantified group whose alternatives overlap, compared case-insensitively since matching runs with `i`. `(a|b)+`, `(?:read|write)+` and `(?:https?|ftp)://[^\s]+` still pass.

SasinduDilshara (approved):

packages/ballerina-extension/src/features/ai/utils/libs/docs-fs-tools.ts
```diff
@@ -0,0 +1,341 @@
...
+            closeBranch(frame);
+            const closed = frames.pop() as GroupFrame;
+            const quantifier = pattern[i + 1] ?? "";
+            const quantified = /[*+{]/.test(quantifier);
+            const ambiguous = closed.nestedAmbiguity
+                || closed.branchFirsts.some((branch, idx) =>
+                    closed.branchFirsts.slice(idx + 1).some(other => overlaps(branch, other)));
+            if (quantified && closed.bodyHasQuantifier) { return false; }
+            if (quantified && ambiguous) { return false; }
+            const parent = frames[frames.length - 1];
```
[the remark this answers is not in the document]
IsuruMaduranga: Will fix in a separate PR

PR #1087, opened 2026-09-14
Fix Role/Instructions landing last in the agent form

dan-niles:
## Summary

Fixes https://github.com/wso2/product-integrator/issues/2440

- Role and Instructions are synthesized from the agent's `systemPrompt` field *after* every other parameter is already in the properties map, so they always rendered at the bottom of the Edit Agent form.
- This went unnoticed while those other parameters (Maximum Iterations, Tool Loading Strategy, etc.) were all tucked under Advanced Configurations; a recent change made several of them non-advanced, surfacing the ordering bug in the main form.
- Fixed by reserving Role/Instructions' slots up front (`FormBuilder.reserveProperty`) in both places that build an agent's properties from scratch: `AgentBuilder` (new agent template) and `CodeAnalyzer#checkForNewConnectionOrAgent` (inspecting an existing agent declaration).

https://github.com/user-attachments/assets/cafe238b-562d-44cb-853f-4645ed6ab294

- Fixed the agent form field order by reserving `AgentCallBuilder.ROLE` and `AgentCallBuilder.INSTRUCTIONS` properties during agent template construction, so `Role` and `Instructions` render in the intended position for new and existing agents.
- Updated `CodeAnalyzer` agent generation to consistently reserve `Role` and `Instructions` properties when inspecting agent declarations.
- Normalized `query` prompt type handling by routing `AgentCallBuilder` and `AgentRunBuilder` through a shared `fixQueryPromptType(..., true)` flow that uses `AiUtils.addPromptTypeIfUnionMember(..., defaultToPrompt)` to add/select the PROMPT option for unions that include `ai:Prompt`.
- Updated `AIChatAgentWizard.tsx` panel layout to use automatic height with visible overflow.


pasindufernando1 (dismissed):

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -3807,6 +3807,10 @@ private void checkForNewConnectionOrAgent(NewExpressionNode newExpressionNode,
             return;
         }
         startNode(kind, newExpressionNode);
+        if (kind == NodeKind.AGENT) {
+            nodeBuilder.properties().reserveProperty(AgentCallBuilder.ROLE)
+                    .reserveProperty(AgentCallBuilder.INSTRUCTIONS);
```
madushajg, line 3881: `reserveProperty(ROLE)` / `reserveProperty(INSTRUCTIONS)` runs for every `NodeKind.AGENT`, but these properties are only populated by `genAgentData` for `ImplicitNewExpressionNode`. Explicit new syntax (e.g. `check new ai:Agent(...)`) leaves them as null placeholders.

When the flow is later round-tripped through `AgentBuilder.toSource()` → `writeSystemPromptFromRoleInstructions`, those placeholders are treated as present and overwrite the existing `systemPrompt` with `{role: , instructions: }`, silently discarding the user's prompt and causing compilation errors on the next save.

Let's either populate these properties for the explicit-new path using `genAgentData` / `setAdditionalAgentProperties`, or remove them when they can't be populated.

dan-niles: Fixed by dropping the `instanceof ImplicitNewExpressionNode` guard and calling `genAgentData` for both explicit- and implicit-new (it now takes the already-resolved arg list). Added a `getModuleNodes` test with an explicit-new agent to cover it: ef88600.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
```diff
@@ -436,6 +436,62 @@ public static Property convertToSingleSelect(Property original, List<String> opt
...
+            updatedTypes.add(new PropertyType(type.fieldType(), type.ballerinaType(), type.scope(), type.options(),
+                    type.template(), type.typeMembers(), type.recordSelectorType(),
+                    isPromptValue ? false : type.selected()));
+        }
+        return new Property(
+                original.metadata(),
+                updatedTypes,
+                original.value(),
+                original.oldValue(),
+                null,
```
NipunaRanasinghe: This passes `null` for `placeholder`, so the original one is dropped. `convertToSingleSelect` just above passes `original.placeholder()`.

The goldens show the effect: `agent_call_flow_node_7/8/9.json` all lose `"placeholder": "{}"` on `query`, and in all three PROMPT ends up `"selected": false` — so someone editing an existing agent call in record/expression mode loses the hint for no reason related to this fix.

If clearing it is deliberate for PROMPT mode, could it be conditioned on `isPromptValue` and mentioned in the javadoc? Otherwise `original.placeholder()`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentRunBuilder.java
```diff
@@ -88,6 +89,7 @@ public void setConcreteConstData() {
     public void setConcreteTemplateData(TemplateContext context) {
         FlowNode callTemplate = getOrCreateCallFunctionTemplate(context);
         restoreFromTemplate(callTemplate);
+        fixQueryPromptType(this, true);
```
NipunaRanasinghe, line 92: The `defaultToPrompt = true` path has no test coverage, here or in `AgentCallBuilder`.

Every golden this PR touches (`agent_call_flow_node_7/8/9.json`) has `query` PROMPT with `"selected": false`, so only the "not selected" branch is exercised. `agent_call_template.json` and `agent_call_template_ballerina.json` still target `ballerinax/ai` 1.0.1, whose `run` takes a plain `string` query, so they no-op straight through `addPromptTypeIfUnionMember` and stayed unchanged. That also means the `isBacktickTemplateValue` selection branch and this whole AGENT_RUN fix are untested.

A template golden against `ballerina/ai` 1.15.0 plus one source case with a backtick-templated query would cover both branches.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentRunBuilder.java
```diff
@@ -113,6 +115,28 @@ void callSuperSetConcreteTemplateData(TemplateContext context) {
...
+     * Post-processes the {@code query} property on an AGENT_RUN node builder, adding a PROMPT type option
+     * when its declared type is a union containing {@code ai:Prompt} (e.g. {@code anydata|ai:Prompt|ai:Resume}
+     * for a typed agent's {@code run}). Safe to call for any node builder — exits immediately when the
+     * conditions are not met.
+     *
+     * @param nodeBuilder     the node builder to update
+     * @param defaultToPrompt selects PROMPT by default, for a blank template where the query value is
+     *                        just a generic placeholder rather than a real value read from source
+     */
+    public static void fixQueryPromptType(NodeBuilder nodeBuilder, boolean defaultToPrompt) {
```
NipunaRanasinghe, line 119: This is byte-identical to `AgentCallBuilder.fixQueryPromptType` apart from the class in the `instanceof`, and line 56 redeclares a private `QUERY = "query"` that duplicates the public `AgentCallBuilder.QUERY`. `CodeAnalyzer` then calls both back to back (4800-4801).

Both classes extend `CallBuilder` — one protected helper there, or a single static taking the expected builder class, would collapse this to one copy.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -3874,6 +3883,10 @@ private void checkForNewConnectionOrAgent(NewExpressionNode newExpressionNode,
             return;
         }
         startNode(kind, newExpressionNode);
+        if (kind == NodeKind.AGENT) {
+            nodeBuilder.properties().reserveProperty(AgentCallBuilder.ROLE)
```
NipunaRanasinghe, line 3880: One path still leaks the reservation: `genAgentData` returns early when `argumentNodes == null`, which happens for an implicit new with no arg list (`ai:Agent a = check new;` — invalid for `ai:Agent`, but reachable while typing). The reserved slots then serialize as all-null properties, and `AgentBuilder.toSource` sees `getProperty(ROLE).isPresent()` and writes an empty `systemPrompt`.

No NPE — `map(Property::value)` swallows the null and it falls back to `""` — so this is low severity, but a `removeProperty` on that early-return path would mirror the pattern already used at line 4782.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/AiUtils.java
dan-niles: Fixed. Cleared it only when PROMPT is selected, kept otherwise. Placeholder is one field on the whole property, not per-mode, so restoring it unconditionally would show a stray "{}" in the default Prompt-mode field. Goldens updated.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentRunBuilder.java
dan-niles: Added a test for the defaultToPrompt=true branch (blank template asserts PROMPT selected, others not). Didn't add coverage for the backtick-value branch or AGENT_RUN's own defaultToPrompt=true path: no existing fixture has a backtick-templated query, and a typed-agent (ai:Listener) fixture is more setup than I wanted to add in this pass. Can follow up if you'd like it covered now.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/AgentRunBuilder.java
dan-niles: Moved the shared logic into AiUtils.fixQueryPromptType(Map<String, Property>, boolean). Both builders still need their own instanceof cast to reach formBuilder, since it's protected and only reachable from within the declaring subclass's own code, but the actual get/check/put logic is now in one place.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
dan-niles: Fixed. removeProperty on the argumentNodes == null path, same pattern as the inferred-type param below.

SasinduDilshara (approved):

NipunaRanasinghe (approved):
Approving — all four comments from my earlier review are addressed in 0d2b6f9, verified against the branch:

- `placeholder` is now `isPromptValue ? null : original.placeholder()`, and the `{}` is back in `agent_call_flow_node_7/8/9.json`. Agreed that keeping it unconditionally would show stray text in Prompt mode, since `placeholder` is one field on the property rather than per-mode.
- `removeProperty(ROLE/INSTRUCTIONS)` on the `argumentNodes == null` path.
- Both `fixQueryPromptType` bodies collapsed into `AiUtils.fixQueryPromptType(Map, boolean)`; the per-builder `instanceof` casts staying put makes sense given `formBuilder` is protected.
- `testAgentCallQueryDefaultsToPromptForBlankTemplate` covers the `defaultToPrompt` path against `ballerina/ai` and would catch the union detection regressing.

The backtick-value branch and AGENT_RUN's own `defaultToPrompt` are still uncovered, but not worth blocking on. CI is green on the updated head.

PR #1089, opened 2026-09-14
Migrate the language server to Java 25 and Gradle 9.7.1

### LINKED ISSUE #2532: Migrate the Ballerina language server to Java 25
### Description

Move the language server (packages/ballerina-language-server) to Java 25, alongside the platform migration, and handle users still on a Java 21 distribution.

### Version

WI 5.1.0 alpha

pasindufernando1:
Migrates the language server to Java 25 and Gradle 9.7.1, and tells the user when their distribution cannot run it.

`ballerinaLangVersion` is now `2201.14.0-20260916-095600-2ad98357`. 2201.14.0 publishes Java 25 artifacts (`org.gradle.jvm.version=25`), so a Java 21 target fails during dependency resolution, before compilation. Gradle 8.11.1 cannot configure this build on JDK 25 at all, so it had to move too — **most of this diff is Gradle 9 collateral, not Java 25.**

For the users who are still in a earlier java 21 based distribution, they will be prompted to either update to a newer distribution or downgrade the ballerina extension version

https://github.com/user-attachments/assets/64ecb37c-b572-4373-8068-b2b17f4e2de9

## What's in it

- Gradle 8.11.1 → 9.7.1; all six CI JDK setups → 25; `sourceCompatibility = 21` → a single `options.release = 25` in the root `subprojects` block, so modules that previously declared no Java version are covered too
- Plugin migrations forced by Gradle 9: gradle-enterprise → develocity, shadow → `com.gradleup.shadow`, `Project.exec()` → injected `ExecOperations`; Mockito and JaCoCo bumped because the old ones break *running* on JVM 25
- **Two `pack` bugs from shadow 9**, both producing a jar that built and unit-tested cleanly while being unable to serve a request: `from()` stopped expanding jars, so `Main` was never on the classpath; and `duplicatesStrategy` now defaults to `EXCLUDE`, which cut `META-INF/services` for `ExtendedLanguageServerService` from 34 entries to 1, so no BI extension loaded
- **Test provisioning fix.** `bal build --sticky` only writes a package into the build-owned test home when it has to fetch it, so anything the distribution bundles at the exact locked version never lands there. Invisible until 2201.14.0 caught up with the lock, then it broke 30 tests. Those packages are now injected via `copyStdlibs`, pinned to `Dependencies.toml`
- **34 flow model goldens refreshed.** 2201.14.0 reports real declared defaults where the old ones recorded type-zero fallbacks, and picks up genuine upstream changes from ai 1.15.0 and kafka. Six langserver-core fixtures moved too: `Ballerina.toml` completions now also offer `platform.java25*` (java21 is still offered), two definition paths follow the bala's `java21` → `java25` directory, and one completion config gained `labelDetails`
- **Test-suite fixes surfaced by the move.** `langserver-commons` was missing `useTestNG()`, so Gradle ran JUnit against a TestNG test and **its 2 tests had never executed**; `flow-model-generator-ls-extension` needed its test heap raised from 2g to 4g, which the 2201.14.0 corpus now exceeds while reading bala metadata
- **setup-ballerina → v1.1.5.** v1.1.4's nightly branch downloaded `Windows Installer ZIP` / `MacOS Installer ZIP` while the install step looked for `.msi` / `.pkg`, so those jobs failed before a distribution was ever installed. Only Linux, which already asked for the deb, worked — which is why Ubuntu ran and Windows did not
- **The staging JRE** now names `jdk-25.0.3+9-jre`. Staging has no `/distributions` endpoint to query so the value is hardcoded, and it still said `jdk-21.0.5+11-jre` — plugin-dev-mode testers would have fetched a Java 21 JRE for a Java 25 distribution

## Users on an older distribution

This is the user-facing half of the change, and the reason the migration is safe to ship.

The packed jar is Java 25 bytecode, so it cannot load on the JRE a pre-2201.14.0 distribution ships. Today that failure is **completely silent**: the server dies, `extension.ts` logs to the output channel and calls `showPluginActivationError()`, whose body is entirely commented out. The user gets an editor with no completions, no diagnostics and no BI views, and nothing at all explaining why.

`checkLanguageServerJdkCompatibility()` now runs *before* the server is launched. It resolves the JDK the server would actually start with, reads its major version from `<jdkDir>/release`, and if that is below 25 it stops startup and shows a modal:

> This version of the extension requires Ballerina 2201.14.0 or later.
>
> Your Ballerina distribution (`<version>`) runs on Java `<n>`, and the language server requires Java 25.

with exactly two ways out — **Update Ballerina**, which opens the visual updater, or **Install Previous Extension Version**, which opens the extension page (VS Code has no command that opens the version picker for a specific extension, so this puts the user one gear-menu click from "Install Another Version...").

Because the modal is transient, the visualizer panel is given the same explanation and the same two actions rather than sitting on its loading frame forever.

Deliberate choices:

- **No silent fallback.** Users above the BI gate keep getting the bundled jar and are told to update. Quietly dropping them onto the distribution's own language server would take away BI capabilities with no explanation, which is worse than a clear stop
- **Checks the JRE, not the distribution version.** The JRE is the actual constraint, and reading it keeps the check correct when `findHighestVersionJdk` falls back to `JAVA_HOME`. It reads the `release` file rather than parsing a directory name
- **Skipped when `ballerina.langServerPath` is set** — that jar's Java version is the user's own concern
- **Fails open everywhere.** No JDK resolved, an undeterminable version, or a throw inside the check all let startup proceed. A bug in the check can never be what blocks a working setup

## Status

Full LS suite green locally against `2201.14.0-alpha`: **7058 tests, 0 failures** across all 12 modules.

On CI, green so far: Build repo, Fast tests, e2e groups 1-4, and LS tests on **Windows** (which the v1.1.5 bump fixed). The Ubuntu LS test job has not yet completed a run to conclusion — the last two were superseded by pushes before finishing — so that one still needs a clean result.

e2e passing is the meaningful part: it means the extension launched the Java 25 server, the pre-flight JDK check passed rather than blocking, and BI served real requests.

## Note for reviewers

The history targets Java 25, then 21, then 25 again. The Java 21 detour was deliberate — at the time the distribution still shipped a JRE 21. 2201.14.0 publishing Java 25-only libraries removed that choice.


packages/ballerina-language-server/spotbugs-exclude.xml
```diff
@@ -274,4 +274,25 @@
...
+    </Match>
+    <Match>
+        <Bug pattern="AT_STALE_THREAD_WRITE_OF_PRIMITIVE"/>
+    </Match>
+    <Match>
+        <Bug pattern="USO_UNSAFE_METHOD_SYNCHRONIZATION"/>
+    </Match>
+    <Match>
+        <Bug pattern="HSM_HIDING_METHOD"/>
+    </Match>
```
nipunayf: Do we have a separate issue to track these issues?

pasindufernando1:
Changed approach since this was drafted — it no longer targets Java 25 bytecode, so the "do not merge" reason is gone and it's safe to merge today.

The LS doesn't choose its own JVM: `findHighestVersionJdk` (`server.ts:108`) picks the highest `jdk-*-jre` from the distribution's `dependencies` dir, so it moves to Java 25 by itself once the distribution ships one. Java 21 bytecode loads on both, so compiling to 25 bought nothing and would have broken every user still on a JRE 21 distribution. What's left is the Gradle 9.7.1 upgrade, which is what actually lets the build run on a JDK 25 machine — 8.11.1 fails outright with `Type T not present`.

@dulajdilshan — you drafted this on the old premise; it needs to come out of draft for CI to run again. `ballerinaLangVersion` is deliberately not bumped here; that's a separate PR.

pasindufernando1:
Addressed both findings in 36a5453.

**`options.release = 25`** — the finding was right, and wider than reported: four places still set up JDK 21 before running Gradle against the LS (`schedule.yml` pack / build / cyclonedxBom, and the language-server publication step in `release-pre-release.yml`). The publication one is the risk, since nothing on this PR exercises it. Moved all four to JDK 25 rather than lowering the target — 2201.14.0's libraries are Java 25 artifacts, so a Java 21 target fails during dependency resolution before compilation.

**Custom language server path** — correct. `getServerOptionsUsingJava` launches a configured jar instead of the bundled one, and `usesBundledLanguageServer()` only separates the CLI path from the Java path, so a custom Java 21 jar was being blocked. The check now skips when `ballerina.langServerPath` is set.

The `ballerinaLangVersion` comment was already handled by the revert in 181a097–95a68b0.

packages/ballerina-language-server/gradle.properties
```diff
@@ -1,18 +1,19 @@
 org.gradle.caching=true
 group=io.ballerina
 name=ballerina-language-server
-ballerinaLangVersion=2201.13.4
+ballerinaLangVersion=2201.14.0-20260916-095600-2ad98357
```
dulajdilshan, line 4: **[Blocking] Scheduled nightly provisioning.** This timestamped coordinate needs the CI-artifact installer, but both scheduled pack/Windows jobs call `setup-ballerina` without `distributionVersion: nightly`; the CDN request therefore fails with the 403 documented in `build/action.yml:69-74`. Please pass `nightly` at both call sites.

packages/ballerina-extension/src/views/visualizer/webview.ts
```diff
@@ -216,6 +221,16 @@ export class VisualizerWebview {
                 retainContextWhenHidden: true,
             }
         );
+        // The blocked-startup panel posts these; the React app is never loaded in that state,
+        // so its own messaging is not available.
+        panel.webview.onDidReceiveMessage(async (message) => {
```
dulajdilshan: **[Blocking] Undisposed listener.** The new `onDidReceiveMessage` subscription's `Disposable` is discarded, so it has no panel-lifecycle disposal path. `.coderabbit.yaml` requires one for new VS Code listeners; please retain and dispose it with the panel.

.github/actions/build/action.yml
```diff
@@ -54,18 +54,24 @@ runs:
...
 
     - name: Install Ballerina
       id: ballerina
       uses: ./.github/actions/setup-ballerina
+      # TEMPORARY: 2201.14.0 has no installer on dist.ballerina.io yet (HTTP 403), so
+      # install the latest ballerina-distribution CI artifact instead. The build still
+      # compiles against the ballerinaLangVersion artifacts from GitHub Packages.
+      # Revert once 2201.14.0 is published to the CDN.
+      with:
+        distributionVersion: nightly
```
dulajdilshan, line 71: **[Medium] Misleading distribution output.** Installing `nightly` still exposes the Maven coordinate as `ballerinaVersion`, although this output is documented as the installed distribution and `reusable-build.yml:189-191` consumes it as a CDN version. Could this expose the actual installed version, or be renamed and its consumers aligned?

packages/ballerina-language-server/build.gradle
```diff
@@ -341,17 +358,28 @@ task pack(type: com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar) {
...
+
     archiveBaseName = project.name
     archiveClassifier = ''
     archiveVersion = project.version
     destinationDirectory = file("$project.buildDir")
 
     configurations = [project.configurations.externalJarDeps]
 
-    // Include the compiled classes and resources from all submodules
+    // Include the compiled classes and resources from all submodules. These must be
```
dulajdilshan: **[Medium] Unpinned fat-jar contract.** `pack` runs with tests excluded while the LS tests do not run `pack`, so a regression in flattening subproject jars or merging `META-INF/services` can ship a valid ZIP that fails at extension startup. Could we add a structural pack verification for representative classes and service entries?

.github/workflows/release-pre-release.yml
madushajg, line 90: `distributionVersion: nightly` is not set here, unlike the other three call sites (`build/action.yml`, `ls-test/action.yml`, and `reusable-build.yml`). Since those call sites were updated as part of the same JDK 21 → 25 bump in this PR, this looks like it may be an accidental omission rather than an intentional exclusion.

Could you please verify whether `distributionVersion: nightly` should also be set here? If so, leaving it out could break the pre-release publish path.

.github/workflows/schedule.yml
madushajg, line 223: ditto

madushajg, line 318: ditto

dulajdilshan (commented):
## Review summary

Standards: 5 findings (2 Blocking, 3 Medium).
Spec: no spec available.

### Done well

- `usesBundledLanguageServer()` centralizes the selection rule for both preflight and launch.
- The check fails open when it cannot resolve or identify a JDK.
- The Shadow migration explicitly preserves flat-jar and service-loader intent.

Tooling should own formatter/linter territory; none is raised here.

packages/ballerina-language-server/gradle.properties
pasindufernando1: Fixed in d2f9a58. Three call sites were missing it — `release-pre-release.yml:90` and both in `schedule.yml` — all resolving the timestamped coordinate from the working tree and 403ing on the CDN.

`e2e-scheduled.yml:140` is intentionally left: it passes `gradlePropertiesRef: nightly`, so it reads `gradle.properties` from the released tag and still gets a published version. It only breaks once a nightly tag carries a timestamped coordinate.

packages/ballerina-extension/src/views/visualizer/webview.ts
pasindufernando1: Fixed in 192ca9c. It was registered in the static `createWebview()`, where `_disposables` is not in scope — moved into the constructor and pushed onto `_disposables`, so it disposes with the panel.

.github/actions/build/action.yml
pasindufernando1: Real, but latent. The only consumer is `ExtTest_Ballerina`, gated `if: false`, so nothing reads it today — with `nightly` installed, `reusable-build.yml:190` would build a release URL from the Maven coordinate and 404.

Left as-is rather than renaming the output and its consumers inside this PR. Happy to do it here if you would rather not leave the trap for whoever re-enables that job.

packages/ballerina-language-server/build.gradle
pasindufernando1: Added `verifyPack` in d5fd371 — asserts the launcher class is at the jar root, no nested jars remain, and the extension-service file was not collapsed.

Checked it fails, not just passes: doctoring a jar to drop `Main.class`, and to cut the service file to one entry, each trips it with a message naming the cause. Those are the two regressions this PR already hit.

The service count is a floor, not the exact 34 — extensions come and go, and the regression collapsed it to 1. Not wired into a workflow yet; say the word if you want it on the pack job.

.github/workflows/release-pre-release.yml
pasindufernando1: Good catch — accidental omission, not intentional. Fixed in d2f9a58.

You were right that it would have broken the pre-release publish path: this job resolves `ballerinaLangVersion` from the working tree, so it asked the CDN for the timestamped 2201.14.0 coordinate and got a 403. Same for both `schedule.yml` sites. All three now pass `distributionVersion: nightly` with the same TEMPORARY comment, so they revert together at GA.

.github/workflows/schedule.yml
pasindufernando1: Fixed in d2f9a58 — same omission, same fix.

pasindufernando1: Fixed in d2f9a58 — same omission, same fix.

packages/ballerina-language-server/build.gradle
```diff
@@ -383,6 +411,57 @@ task pack(type: com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar) {
     }
 }
 
+// Task to verify the packed jar shape
+task verifyPack {
```
dulajdilshan, line 408: **[Major] Disconnected pack verification.** `verifyPack` only depends on `pack`; neither the package build script nor scheduled `./gradlew pack -x check` invokes it. A repeat of the flat-jar/service-entry regression can therefore publish an unloadable LS without this guard. Could `pack` finalize with `verifyPack` (or every shipping path invoke it)?

packages/ballerina-language-server/spotbugs-exclude.xml
```diff
@@ -274,4 +274,27 @@
...
 </FindBugsFilter>
```
dulajdilshan, line 302: **[Major] Global SpotBugs suppression.** These pattern-only matches suppress each detector for every current and future class, although the comment describes known legacy cases. That removes the Gradle SpotBugs gate for future concurrency/exception defects. Could we baseline the affected classes or methods instead and leave the detectors active elsewhere?

packages/ballerina-language-server/build.gradle
```diff
@@ -73,6 +75,13 @@ allprojects {
 
 subprojects {
     apply plugin: "java"
+    // Ballerina 2201.14.0 and its libraries are published as Java 25 artifacts, so the
+    // LS cannot target anything lower -- a Java 21 target fails during dependency
+    // resolution, before compilation. The LS runs on the JRE the distribution ships,
+    // which is Java 25 from 2201.14.0 onwards.
+    tasks.withType(JavaCompile).configureEach {
+        options.release = 25
```
dulajdilshan, line 80: **[Medium] pre-existing — stale JDK prerequisite exposed by this migration.** `options.release = 25` cannot compile under the JDK 21 requirement still stated in `AGENTS.md:41`, `CONTRIBUTING.md:119,245`, and `.github/workflows/README.md:368`. Could this PR update those documented build prerequisites to JDK 25?

packages/ballerina-extension/src/core/extension.ts
```diff
@@ -1678,6 +1695,85 @@ export class BallerinaExtension {
...
 
+    /**
+     * Verifies that the JRE the language server would be launched with is new enough to load
+     * it. Returns true when the server may start, false when the user has been told why it
+     * cannot and startup should be abandoned.
+     *
+     * Only applies to the bundled server: when the distribution's own server is used, our jar
+     * is never loaded and the JRE version is irrelevant.
+     */
+    private async checkLanguageServerJdkCompatibility(): Promise<boolean> {
```
dulajdilshan, line 1695: **[Medium] Unpinned compatibility gate.** No test covers JDK parsing/resolution, Java 24 blocking vs. Java 25 proceeding, custom-server bypass, fail-open behavior, warning actions, or the visualizer handoff; the only changed E2E test is unrelated and skipped. Could we add narrow L1 plus host/UI coverage for this startup boundary?

packages/ballerina-extension/e2e-test/e2e-playwright-tests/type-editor/type.spec.ts
```diff
@@ -155,7 +155,11 @@ export default function createTests() {
          * silent on invalid input. Both assertions below show nothing at all
          * before that fix.
          */
-        test('Type Editor Field Diagnostics', async ({ }, testInfo) => {
+        // SKIPPED: waitForTypeEditor() below runs before clickAddType() opens the editor, so this
+        // only passes when a sibling in this serial block happens to leave the editor open. The
+        // Playwright trace shows the visualizer loading normally and the container never being
+        // opened. Unrelated to the Java 25 migration; to be fixed properly with the team.
+        test.skip('Type Editor Field Diagnostics', async ({ }, testInfo) => {
```
dulajdilshan, line 160: **[Medium] Unrelated regression coverage removed.** This PR skips the sole Type Editor diagnostics assertions and says the failure is unrelated to the migration, leaving #961 unpinned. Could we fix the setup order (`clickAddType()` before `waitForTypeEditor()`) and retain the test, or move a linked, time-bounded disablement into a separate fix?

pasindufernando1: Keeping the skip. The failure is unrelated to this migration: the one commit it started on changes a single string on the staging update-tool path, which e2e never enters, and the Playwright trace shows the visualizer loading normally with no JDK-incompatibility panel — the editor container is simply never opened, because `waitForTypeEditor()` runs before the `clickAddType()` that opens it.

Your suggested fix is the right shape and matches that diagnosis, but I cannot run the Playwright suite locally to verify it, so I would rather not swap the order blind and spend CI cycles guessing. Being tracked for a proper fix with the team that owns these tests.

packages/ballerina-language-server/build.gradle
pasindufernando1: Fixed in d9e6c44. You were right that nothing invoked it — `schedule.yml` runs `pack -x check` and the extension build runs `build pack -x test -x check`.

`pack` now finalizes with `verifyPack`. It returns early when the archive is absent, so a `pack` failure surfaces its own error rather than a misleading missing-jar one.

packages/ballerina-language-server/spotbugs-exclude.xml
pasindufernando1: Fixed in 077aa11. Measured what they were hiding first — removed all six and ran `spotbugsMain` across all 27 modules: 133 violations, 123 of them THROWS-family across ~64 classes, 10 not.

THROWS stays global: that is the long-standing convention of LS entry points surfacing `RuntimeException`, and listing ~64 classes is an inventory rather than a baseline. The other three are now scoped to the classes that trip them — `BallerinaWorkspaceManager$ProjectContext` and `Type` (stale write), `DiagnosticsHelper` (unsafe synchronization), `AutomationBuilder`/`WorkflowBuilder` (method hiding) — so the concurrency detectors keep guarding everything else, including new code. `spotbugsMain` passes clean.

packages/ballerina-language-server/build.gradle
pasindufernando1: Fixed in 89acb2e — all four updated to JDK 25 (`AGENTS.md:41`, `CONTRIBUTING.md:119` and `:245`, `.github/workflows/README.md:368`). No Java 21 references left in the docs.

packages/ballerina-extension/src/core/extension.ts
pasindufernando1: Added in b8aa06d — 16 tests in `src/test-support/languageServerJdk.test.ts`, over real temp directories rather than an fs mock so the release-file parsing and version ordering run for real.

Covers `getJdkMajorVersion` (including the release file winning over the directory name, the `JAVA_HOME` fallback case, plus its null paths), both sides of the boundary so Java 21 and **24** block while 25 and 26 proceed, the highest-version pick, and `resolveLanguageServerJdkDir`'s truncation at `distributions` plus the null that lets the caller fail open.

Mutation-checked: flipping `REQUIRED_JDK_MAJOR_VERSION` to 21 fails exactly the three boundary assertions.

Still uncovered, and worth naming: the visualizer handoff and the modal actions need real webview scaffolding. That gap is not academic — a bug there (the React app mounting over the blocked-startup panel) was found manually today and fixed in bfc9150.

packages/ballerina-extension/e2e-test/e2e-playwright-tests/type-editor/type.spec.ts
dulajdilshan: Ack! Thanks!

.github/actions/build/action.yml
dulajdilshan: Shall we create a ticket/issue or update an existing post-release thread?

.github/actions/setup-ballerina/action.yml
```diff
@@ -43,13 +52,15 @@ runs:
           properties="${{ github.workspace }}/packages/ballerina-language-server/gradle.properties"
           version=$(sed -n 's/^ballerinaLangVersion=//p' "${properties}")
         fi
-        if [[ ! "${version}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+        # Accepts timestamped pre-GA builds; tighten back to the plain N.N.N form once
+        # 2201.14.0 is released.
```
dulajdilshan, line 53: Same as https://github.com/ballerina-platform/ballerina-vscode/pull/1089/changes#r4068609370

.github/workflows/schedule.yml
```diff
@@ -210,17 +210,23 @@ jobs:
           ref: refs/heads/${{ matrix.branch }}
           persist-credentials: false
 
-      - name: Set up JDK 21
+      - name: Set up JDK 25
         uses: actions/setup-java@c1e323688fd81a25caa38c78aa6df2d33d3e20d9 # v4
         with:
           distribution: 'temurin'
-          java-version: 21.0.3
+          java-version: 25
```
dulajdilshan: Let's check whether this sticks with a particular version. Currently, it seems that 25.0.4 is the stable production release recommended for all users

dulajdilshan (dismissed):
LGTM! with minor comments

.github/actions/build/action.yml
pasindufernando1: Created wso2-enterprise/integration-engineering#2961 for the post-GA work, linked under #2893.

It covers the three workarounds to revert together (lang version, `distributionVersion: nightly` at six sites, the version regex), plus the compiled-against vs shipped gap and the defects found along the way.

.github/workflows/schedule.yml
pasindufernando1: Good catch — it was floating, and that is a regression I introduced. This PR replaced `java-version: 21.0.3` with `java-version: 25`, so it resolves to whatever 25.x the runner tool-cache holds; the last run picked `25.0.4+1`.

Pinned all six setup-java blocks to `25.0.4` in 6cfafb3, restoring the previous convention.

.github/actions/setup-ballerina/action.yml
pasindufernando1: Will be tracked via https://github.com/wso2-enterprise/integration-engineering/issues/2961

dulajdilshan (dismissed):

packages/ballerina-extension/src/views/visualizer/webview.ts
```diff
@@ -228,6 +239,21 @@ export class VisualizerWebview {
...
     }
 
+    /** Records the failure and re-renders an open panel; the HTML is built once at creation. */
+    public static showJdkIncompatibility(info: {
+        ballerinaVersion: string;
+        jdkMajorVersion: number;
+        requiredJdkMajorVersion: number;
+        requiredBallerinaVersion: string;
+    }): void {
+        VisualizerWebview.jdkIncompatibility = info;
```
madushajg, line 255: `VisualizerWebview.jdkIncompatibility` is set once and never cleared, so once shown, every panel created for the rest of the session renders the incompatibility page including the panel opened by the "Update Ballerina" action itself (`SETUP_BALLERINA` → `SetupView`). The update UI never mounts because `loadedScript()` returns early. Clear the flag before dispatching `ballerina.update-ballerina-visually`.

pasindufernando1: Correct, and worse than cosmetic — fixed in c43023a.

Traced it: `stateMachine.ts:923` constructs a fresh `VisualizerWebview` for the setup view, `getWebviewContent()` saw the static still set, and the `loadedScript()` early-return I added for the endless-loader bug then stopped the app mounting at all. So the recovery action the modal offers did nothing.

Now cleared from both entry points, the modal button and the panel button. Deliberately **not** cleared for "Install Previous Extension Version" — that opens the extensions page rather than a `VisualizerWebview`, and the server still cannot start, so the panel should keep explaining why.

I tried to pin this with a test and dropped it: `webview.ts` reaches `@wso2/ballerina-core` plus `AiPanelWebview`, `approvalViewManager`, `agentStatusManager`, `StateMachinePopup` — ten-plus mocks to assert a two-line setter, and still not covering what broke, which is that a newly constructed panel inherits the static. That is the webview-scaffolding gap already tracked in wso2-enterprise/integration-engineering#2961; two bugs have now come out of it, both caught by review rather than CI.

madushajg (approved):

PR #1090, opened 2026-09-14
Add workflow:DurableAgent information to workflow skill in WSO2 Copilot

damsith0:
Updated `SKILL.md` with comprehensive `workflow:DurableAgent` guidance for `ballerina/workflow` 0.10.0. The documentation covers declarations, configuration, activities, tools, peers, human tasks, event handling, driver methods, validation rules, and diagnostic codes. It replaces the previous unsupported-feature note and removes obsolete 0.9.x guidance.


SasinduDilshara (dismissed):

samithkavishke:
@damsith0 address the code rabbit suggestions

packages/ballerina-extension/src/features/ai/agent/skills/workflow-builder/SKILL.md
```diff
@@ -371,11 +371,170 @@ for connection details rather than inventing keys for those modes; only `IN_MEMO
...
+string instanceId = check <agentName>.run(<query>, <input>);
+string token = check <agentName>.sendData(instanceId, "<channelName>", <data>);
+<Result> result = check <agentName>.waitForResult(instanceId);
+<Response> reply = check <agentName>.waitForDataResult(instanceId, token);
+```
+
+- **`run(string query, json input = ())` returns the new instance ID — always the ID, never the
+  result.** A durable agent may suspend for days on a human task, so no caller thread is blocked.
+  `query` is the user turn appended to the agent's system prompt; `input` is an optional structured
+  JSON payload that must match the agent's declared `inputType`. Outside a workflow this is a
```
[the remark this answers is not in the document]
damsith0: In ballerina central documentation, input is strictly given as json. Please verify, @rdulmina @hasithaa

packages/ballerina-extension/src/features/ai/agent/skills/workflow-builder/SKILL.md
```diff
@@ -371,11 +371,170 @@ for connection details rather than inventing keys for those modes; only `IN_MEMO
...
+| `model` | `ai:ModelProvider` | required |
+| `inputType` | `typedesc<json>?` | `json` |
+| `resultType` | `typedesc<anydata>?` | `()` |
+| `activities` | `(ActivityDecl\|function)[]` | `[]` |
+| `tools` | `(ToolDecl\|ai:ToolConfig\|ai:BaseToolKit\|function)[]` | `[]` |
+| `events` | `map<EventConfig>` | `{}` |
+| `humanTasks` | `map<HumanTaskDefinition>` | `{}` |
+| `peers` | `PeerDecl[]` | `[]` |
+| `maxIter` | `int` | `16` |
+| `eventTimeout` | `Duration?` | `()` |
```
hasithaa, line 425: `maxEventWaits` is missing from the table. It is the runaway backstop for a conversation nobody closes, and it matters here because `eventTimeout` defaults to "wait indefinitely" — without the cap named, a chat-like agent silently stops at the default after 50 waits and the model has no idea why.

```suggestion
| `systemPrompt` | `ai:SystemPrompt` | required |
| `model` | `ai:ModelProvider` | required |
| `inputType` | `typedesc<json>?` | `json` |
| `resultType` | `typedesc<anydata>?` | `()` |
| `activities` | `(ActivityDecl\|function)[]` | `[]` |
| `tools` | `(ToolDecl\|ai:ToolConfig\|ai:BaseToolKit\|function)[]` | `[]` |
| `events` | `map<EventConfig>` | `{}` |
| `humanTasks` | `map<HumanTaskDefinition>` | `{}` |
| `peers` | `PeerDecl[]` | `[]` |
| `maxIter` | `int` | `16` |
| `eventTimeout` | `Duration?` | `()` |
| `maxEventWaits` | `int` | `50` |
```

Worth one sentence below the table too: `eventTimeout` omitted means each event wait waits indefinitely, and `maxEventWaits` caps chat turns and events together per run — raise it for a chat-like agent whose turns are many and short.

packages/ballerina-extension/src/features/ai/agent/skills/workflow-builder/SKILL.md
```diff
@@ -371,11 +371,170 @@ for connection details rather than inventing keys for those modes; only `IN_MEMO
...
+
+| Field | Type | Default |
+|---|---|---|
+| `activity` | `function` | required — the `@workflow:Activity` function |
+| `name` | `string` | optional — the function name |
+| `description` | `string` | optional — the function's doc comment |
+| `bindings` | `map<anydata\|object {}>` | optional |
+| `requiresApproval` | `boolean` | `false` |
+| `userRoles` | `string\|string[]` | optional |
+| `retryPolicy` | `AutoRetry\|ReviewTaskDefinition\|NoAutomaticRetry` | `NoAutomaticRetry` |
```
hasithaa: These three rows are the 0.9.0 shape. 0.10.0 replaced `requiresApproval`/`userRoles` with a single `approvalPolicy` that carries its own audience, and — importantly for a skill — writing the old names is **a build error, not a silently ignored field**: the plugin reports `WORKFLOW_163` (*"'requiresApproval' is no longer a field of 'ActivityDecl': declare 'approvalPolicy'"*). `NoAutomaticRetry` is now a deprecated alias of `NoRetry`, and the retry union gained `RetryBeforeReview` (auto-retry, then a review once the attempts are spent).

```suggestion
| `approvalPolicy` | `ReviewTaskDefinition\|NoApproval` | `NoApproval` |
| `retryPolicy` | `AutoRetry\|ReviewTaskDefinition\|RetryBeforeReview\|NoRetry` | `NoRetry` |
```

So the gated form is `{activity: approveRefund, approvalPolicy: {userRoles: "manager"}}`.

packages/ballerina-extension/src/features/ai/agent/skills/workflow-builder/SKILL.md
```diff
@@ -371,11 +371,170 @@ for connection details rather than inventing keys for those modes; only `IN_MEMO
...
+
+#### `ToolDecl`
+
+An AI tool capability, with optional gating config.
+
+| Field | Type | Default |
+|---|---|---|
+| `tool` | `ai:BaseToolKit\|ai:ToolConfig\|ai:FunctionTool` | required |
+| `requiresApproval` | `boolean` | `false` |
+| `userRoles` | `string\|string[]` | optional |
```
hasithaa: Same 0.10.0 rename as on `ActivityDecl` — `requiresApproval`/`userRoles` on a `ToolDecl` fail the build with `WORKFLOW_163` (*"declare 'approvalPolicy'"*).

```suggestion
| `approvalPolicy` | `ReviewTaskDefinition\|NoApproval` | `NoApproval` |
```

Also worth adding under this table, since the skill's whole job is to stop the model generating code that will not build: a tool whose `@ai:AgentTool` declares `auth` is rejected outright (`WORKFLOW_155`) — durable agents do not run the `ai:Agent` loop, so the tool would run without token acquisition. Don't offer an authenticated tool; wrap the call in an activity instead.

packages/ballerina-extension/src/features/ai/agent/skills/workflow-builder/SKILL.md
```diff
@@ -371,11 +371,170 @@ for connection details rather than inventing keys for those modes; only `IN_MEMO
...
+| Field | Type | Default |
+|---|---|---|
+| `userRoles` | `string\|string[]` | required |
+| `title` | `string?` | `()` |
+| `description` | `string?` | `()` |
+| `timeout` | `Duration?` | `()` |
+| `taskInputType` | `typedesc<map<json>>` | `JsonObject` |
+| `resultType` | `typedesc<anydata>` | `anydata` |
+
+The first four are included from `*ReviewTaskDefinition`, and the record is open. Input supplied to
```
hasithaa: `HumanTaskDefinition` includes the whole of `*ReviewTaskDefinition`, which in 0.10.0 is nine fields, not four — the missing ones are exactly the ones a generated agent needs to express "these people, not those": `users`, `excludedUsers`, `excludedRoles`, `administratorRoles`, `administratorUsers`.

The type is also wrong in a way that will produce non-compiling code: the audience fields are `string|[string, string...]` (a non-empty **tuple**), not `string|string[]`. A literal like `["manager", "finance"]` is fine, but a `string[]` variable is not assignable, and a model told `string[]` will write one.

```suggestion
| `userRoles` | `string\|[string, string...]?` | required — `()` when only `users` may decide |
| `users` | `string\|[string, string...]` | optional |
| `excludedUsers` | `string\|[string, string...]` | optional |
| `excludedRoles` | `string\|[string, string...]` | optional |
| `administratorRoles` | `string\|[string, string...]` | optional |
| `administratorUsers` | `string\|[string, string...]` | optional |
| `title` | `string?` | `()` |
| `description` | `string?` | `()` |
| `timeout` | `Duration?` | `()` |
| `taskInputType` | `typedesc<map<json>>` | `JsonObject` |
| `resultType` | `typedesc<anydata>` | `anydata` |

The first nine are included from `*ReviewTaskDefinition`, and the record is open. `userRoles` is a
required field of nilable type: at least one of `userRoles` and `users` must name someone, and a
definition that names nobody fails the build with `WORKFLOW_164`. Input supplied to
the task is checked against `taskInputType` before the task is created. `resultType` is how an agent
declares the answer's shape; a workflow states that as `awaitHumanTask`'s `T` instead.
```

packages/ballerina-extension/src/features/ai/agent/skills/workflow-builder/SKILL.md
```diff
@@ -371,11 +371,170 @@ for connection details rather than inventing keys for those modes; only `IN_MEMO
...
+
+| Field | Type | Default |
+|---|---|---|
+| `agent` | `DurableAgent` | required — the peer agent |
+| `name` | `string` | required — tool name, unique across all capabilities |
+| `description` | `string` | optional — what the peer does, for the model |
+| `'wait` | `boolean` | `true` |
+| `callbackChannel` | `string` | optional |
+| `requiresApproval` | `boolean` | `false` |
+| `userRoles` | `string\|string[]` | optional |
```
hasithaa: `PeerDecl` was reshaped in 0.10.0 — five of these seven fields no longer exist, and each of them is a `WORKFLOW_163` build error rather than an ignored field. The record is now `{agent, description?, allowedEvents?}`, and `agent` also accepts a `@workflow:Workflow` function (started fire-and-forget, since a workflow has no events to answer on).

```suggestion
| `agent` | `DurableAgent\|function` | required — the peer agent, or a `@workflow:Workflow` function |
| `description` | `string` | optional — what the peer does, for the model |
| `allowedEvents` | `string[]` | optional — the peer events to expose; omit for all, `[]` for the run entry only |
```

packages/ballerina-extension/src/features/ai/agent/skills/workflow-builder/SKILL.md
```diff
@@ -371,11 +371,170 @@ for connection details rather than inventing keys for those modes; only `IN_MEMO
...
+`callbackChannel`, which is **required** in that case and must name a channel declared in `events`.
+
+**Do not write `'wait` in a `PeerDecl` literal.** Against `ballerina/workflow` 0.9.0 the compiler
+plugin emits invalid code for it — the build fails with `action invocation as an expression not
+allowed here` and `invalid token ':'`, reported at a line past the end of your own file because the
+fault is in generated code. Every other field of `PeerDecl`, `callbackChannel` included, is fine.
+Omit `'wait` and take its `true` default until that is fixed.
+
+On all three, `requiresApproval = true` gates every call with a `PRE_RUN` review activity, and
+`userRoles` says who may decide those reviews.
```
hasithaa: This whole passage describes the 0.9.0 model and is now actively harmful: `'wait` and `callbackChannel` are not just buggy, they are gone, and a generated literal containing either fails with `WORKFLOW_163` (*"a one-way peer event returns at once and a duplex one waits; the reply address travels with each delegation"*) — so the landmine paragraph tells the model to avoid a field while the paragraph above it still tells it the field exists. The async/callback dance it describes is no longer how delegation works.

```suggestion
A peer's identity is its own module-level variable name: that name prefixes the tool names the
model sees, and it is what must be unique across this agent's capabilities — there is no `name`
field to set. The framework advertises one tool to start the peer plus one per peer event
`allowedEvents` admits; a one-way peer event returns an acknowledgement at once, a duplex event or
the run entry waits durably for the answer. The reply address travels with each delegation, so
there is no callback channel to declare.

Gating belongs to the peer's own activities and tools, not to the delegation — `PeerDecl` carries
no approval fields.

On `ActivityDecl` and `ToolDecl`, an `approvalPolicy` such as `{userRoles: "manager"}` gates every
call with a `PRE_RUN` review and names who may decide it; `NoApproval` (the default) runs the call
directly. A review definition must name an audience — `userRoles`, `users`, or both — or the build
fails with `WORKFLOW_164`.
```

packages/ballerina-extension/src/features/ai/agent/skills/workflow-builder/SKILL.md
```diff
@@ -371,11 +371,170 @@ for connection details rather than inventing keys for those modes; only `IN_MEMO
...
+
+Three declaration rules:
+
+- **Assign it to a module-level `final` variable — this is compiler-enforced.** Never a local
+  variable, never a non-`final` one.
+- **The module-level variable name is the agent's stable identity**, so renaming the variable
+  renames the agent.
+- **`check new ({...})`** — `init` takes `*DurableAgentConfig` as an included record, so the whole
+  configuration is a single mapping argument, and the constructor returns an error on an invalid
+  config.
```
hasithaa: Two more compiler-enforced rules are worth stating here, since both are things a model gets wrong and both are hard errors:

- The initializer must be **inline** `new ({...})` (or `new workflow:DurableAgent({...})`) in the variable declaration — a config built into a variable first, or a factory function, fails with `WORKFLOW_151`, because every worker replica has to derive the same registration deterministically.
- Capability names share one flat namespace, and the plugin reports that at compile time as `WORKFLOW_150` (the runtime registration rejects it again at startup). The section below says this happens "when the agent registers"; in 0.10.0 it usually fails the build instead.

Also worth naming `WORKFLOW_149` for the `final`/module-level rule — diagnostic codes give the model something to match when it sees the error.

hasithaa (commented):
Reviewed the new durable-agent section against `ballerina/workflow` **0.10.0** (module `main`, `ballerina/Ballerina.toml` at `0.10.0`). The structure, the driver-method section and the `EventConfig` table all hold up well — `run` returning the instance id, the plain-method (`.`, never `->`) rule, `AgentBusyError`, and the blocking/non-blocking pairs are all exactly right against the source.

The declaration surface, though, is written against 0.9.0, and 0.10.0 reshaped it in a way that is unusually unforgiving for a skill: the removed fields are **hard build errors** (`WORKFLOW_163`), deliberately, because silently ignoring a dropped `requiresApproval: true` would drop a gate. So a model following the tables as they stand generates code that does not compile. Concretely:

- `ActivityDecl` / `ToolDecl`: `requiresApproval` + `userRoles` → one `approvalPolicy` (`ReviewTaskDefinition|NoApproval`).
- `ActivityDecl.retryPolicy`: default is `NoRetry` (`NoAutomaticRetry` is a deprecated alias), and the union gained `RetryBeforeReview`.
- `PeerDecl`: `name`, `'wait`, `callbackChannel`, `requiresApproval`, `userRoles` are all gone; the record is `{agent, description?, allowedEvents?}` and `agent` also takes a `@workflow:Workflow` function. The `'wait` landmine paragraph should go with them.
- `DurableAgentConfig`: missing `maxEventWaits` (default 50).
- `HumanTaskDefinition`: `*ReviewTaskDefinition` is nine fields now, and the audience fields are `string|[string, string...]`, not `string|string[]` — the stated type will make the model write a `string[]` that does not typecheck.

Two suggestions beyond the drift:

1. **Say which version the section is written against.** The rest of `SKILL.md` handles version drift by telling the model to check the resolved signature (`sendData`'s argument order, `awaitHumanTask`'s `taskInput`/`payload`). The durable-agent surface moved more than either of those between 0.9.0 and 0.10.0, so it deserves the same treatment — a line naming 0.10.0 as the baseline and telling the model to verify against the resolved version, rather than a version-specific bug note that will rot.
2. **Prefer diagnostic codes to symptom descriptions.** `WORKFLOW_149/150/151/154/155/163/164` are all reachable from this section; naming them gives the model something to match when a build fails, and they stay true across releases in a way that "the build fails at a line past the end of your file" does not.

Inline suggestions below are directly applicable. Happy to pair on re-verifying once this is rebased if the target version is meant to be 0.9.x rather than 0.10.0 — in that case most of the above still applies in reverse, and the section should say so explicitly.

hasithaa:
Pushed `bd570ec` to this branch applying the review above, since the suggestions overlapped and would have conflicted if applied one at a time. It rewrites the durable-agent section against 0.10.0 and leaves the "Driving the agent" part untouched — that was already accurate.

What changed:

- `ActivityDecl`/`ToolDecl`: `requiresApproval` + `userRoles` → one `approvalPolicy`; `retryPolicy` default is `NoRetry`, union gained `RetryBeforeReview`. Pulled the gating explanation into its own `#### Gating a capability: approvalPolicy` subsection so the removed names are stated once, with the reason they are a hard error.
- `PeerDecl`: `{agent, description?, allowedEvents?}`, `agent` also takes a `@workflow:Workflow` function. The `'wait` landmine paragraph is gone with the field.
- `DurableAgentConfig`: added `maxEventWaits` plus a sentence on how it and `eventTimeout` interact.
- `HumanTaskDefinition`: all nine `*ReviewTaskDefinition` fields; audience type corrected to `string|[string, string...]`.
- Added: the inline-`new` rule (`WORKFLOW_151`), the `@ai:AgentTool auth` rejection (`WORKFLOW_155`), the mapping-vs-array form note (`WORKFLOW_159`), and a version-baseline line at the top of the section.

Throughout I swapped symptom descriptions for diagnostic codes — they give the model something to match against a real build failure and they survive releases.

One open question for you: if this skill is meant to ship against 0.9.x rather than 0.10.0, say so and I will invert the changes — but the section should name whichever version it targets either way.

damsith0:
> Pushed `bd570ec` to this branch applying the review above, since the suggestions overlapped and would have conflicted if applied one at a time. It rewrites the durable-agent section against 0.10.0 and leaves the "Driving the agent" part untouched — that was already accurate.
> 
> What changed:
> 
> * `ActivityDecl`/`ToolDecl`: `requiresApproval` + `userRoles` → one `approvalPolicy`; `retryPolicy` default is `NoRetry`, union gained `RetryBeforeReview`. Pulled the gating explanation into its own `#### Gating a capability: approvalPolicy` subsection so the removed names are stated once, with the reason they are a hard error.
> * `PeerDecl`: `{agent, description?, allowedEvents?}`, `agent` also takes a `@workflow:Workflow` function. The `'wait` landmine paragraph is gone with the field.
> * `DurableAgentConfig`: added `maxEventWaits` plus a sentence on how it and `eventTimeout` interact.
> * `HumanTaskDefinition`: all nine `*ReviewTaskDefinition` fields; audience type corrected to `string|[string, string...]`.
> * Added: the inline-`new` rule (`WORKFLOW_151`), the `@ai:AgentTool auth` rejection (`WORKFLOW_155`), the mapping-vs-array form note (`WORKFLOW_159`), and a version-baseline line at the top of the section.
> 
> Throughout I swapped symptom descriptions for diagnostic codes — they give the model something to match against a real build failure and they survive releases.
> 
> One open question for you: if this skill is meant to ship against 0.9.x rather than 0.10.0, say so and I will invert the changes — but the section should name whichever version it targets either way.

0.10.0 is right — no need to invert. We want the copilot to have the latest workflow knowledge

hasithaa (approved):

packages/ballerina-extension/src/features/ai/agent/skills/workflow-builder/SKILL.md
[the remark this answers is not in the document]
damsith0: confirmed this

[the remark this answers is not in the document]
hasithaa: It is ok, to use `json` in this context for Agent Skill. Other anydata type is allowed to handle complex usecases like xml.

SasinduDilshara (approved):

PR #1095, opened 2026-09-15
Read an agent's events and human tasks in either declared form

hasithaa:
Fixes wso2/product-integrator#2454

## The bug

A durable agent declared the way the workflow module now documents draws **no event or human-task circles** in the overview:

```ballerina
final workflow:DurableAgent deskAgent = check new ({
    ...
    events: {customerMessage: {request: string, response: string}},
    humanTasks: {escalation: {userRoles: "support-lead"}}
});
```

Its activities and model-provider edge still render, which is what makes this look like a layout problem rather than a parse one.

## Cause

Since **0.9.0** these fields are `map<EventConfig>` / `map<HumanTaskDefinition>`, keyed by capability name. In 0.8.x they were lists of records each carrying its own `name`. `DesignModelGenerator` still read only the list:

```java
if (!(valueExpr instanceof ListConstructorExpressionNode list)) {
    continue;   // ← a keyed declaration is skipped entirely
}
```

`activities` is still a list — `(ActivityDecl|function)[]` in both 0.9.0 and 0.10.0 — so it kept working, hence the partial diagram. The overview therefore reads `activities` in list form only, deliberately.

`WorkflowUtil.collectDeclaredEventNames` had the same assumption. `CodeAnalyzer` already handled both forms for the agent panel, so the two halves of the tooling disagreed about what an agent declares.

## The fix

`WorkflowUtil.capabilityEntries` reads either form and decides what an entry is and what it is called. The overview generator, the event-name lookup, and `CodeAnalyzer`'s keyed branch all go through it, so the three cannot drift apart again:

- a keyed entry whose value is a reference rather than an inline mapping (`humanTasks: {audit: auditTask}`) is an entry — its key names a capability that exists;
- one naming rule, `WorkflowUtil.capabilityName`, for both a string-literal key and a quoted identifier (`'wait`), used by every reader.

A capability the tooling cannot see is one it silently drops the next time it writes the declaration back, which is why this belongs in one place rather than three.

**This is read-side only.** The write side still speaks list form: `insertAgentCapabilityEntry` rejects a keyed field, and `DurableAgentHumanTaskBuilder` emits `{name: ..., roles: ...}` where the keyed form wants a `userRoles` config under a key. Editing a keyed declaration through those paths would produce source that does not compile — pre-existing, and not reachable from the overview: the human-task box is `readonly` with no click handler, so nothing here opens a new path to broken source. Making the write side speak both forms is a separate change.

## Verification

A design-model fixture declaring two events and three human tasks in the keyed form, one of them configured by reference. Before the change it produces no workflow capabilities; after, the generated model carries:

```
events:     customerMessage (string), billSubmitted (string)
humanTasks: escalation, signoff, audit
```

`DesignModelGeneratorTest` compared human tasks by count alone, so their names and ranges were never checked; it now compares both, and fails if a name is wrong (verified by mutating the golden). The existing list-form fixture (`project_16`) is untouched and still passes. 16 tests, 0 failures.

Also included, unrelated to capability parsing: `flow-model-generator-core` test heap raised to 2g, because `PackageUtilConcurrencyTest` resolves packages on 16 threads and exhausts the default heap on CI runners.

Based on `staging/5.14.0`. No Try It changes are included; those are a separate PR.

Updates durable-agent capability parsing to support keyed maps and legacy lists for `events` and `humanTasks`.

- Shares `WorkflowUtil.capabilityEntries` across model generation, event lookup, and code analysis.
- Preserves list-form support and existing activity handling.
- Adds validation for human-task names and locations.
- Increases test JVM heap limits to improve CI stability.
- `DesignModelGeneratorTest`: 16 tests passed, 0 failures.


packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/test/resources/get_design_model/config/durable_agent_capability_mapping.json
```diff
@@ -0,0 +1,107 @@
+{
+  "description": "Durable agent declaring events and human tasks keyed by name",
+  "projectPath": "project_18",
+  "output": {
+    "designModel": {
+      "connections": [
+        {
+          "symbol": "deskModel",
+          "location": {
+            "filePath": "/Users/hasitha/Source/workflow/vscode-tryit/packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/test/resources/get_design_model/source/project_18/agent.bal",
```
samithkavishke: Is there a possibility to replace this with a generic file path

hasithaa: Done. `Location.equals` compares only the file *name*, so the absolute path was never load-bearing — replaced with `$PROJECT_ROOT/src/test/resources/...`, and the suite still passes (16 tests, 0 failures).

Worth noting the rest of this directory is in the same shape: the existing goldens carry five different developers' home directories. Happy to sweep those in a follow-up if you want it.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -509,6 +509,70 @@ public static Optional<MappingConstructorExpressionNode> agentConfigLiteral(Expr
...
+     *
+     * @param value the value of an {@code events}, {@code humanTasks} or similar config field
+     * @return the entries it declares, empty when the value is neither form
+     */
+    public static List<CapabilityEntry> capabilityEntries(ExpressionNode value) {
+        List<CapabilityEntry> entries = new ArrayList<>();
+        if (value instanceof MappingConstructorExpressionNode keyed) {
+            for (MappingFieldNode field : keyed.fields()) {
+                if (!(field instanceof SpecificFieldNode entry) || entry.valueExpr().isEmpty()
+                        || !(entry.valueExpr().get() instanceof MappingConstructorExpressionNode config)) {
```
NipunaRanasinghe: This skips a keyed entry whose value isn't a mapping, but `CodeAnalyzer` keeps it — `out.add(...)` at `CodeAnalyzer.java:1670` sits outside the same `instanceof` check. So `humanTasks: {escalation: SOME_CONFIG_CONST}` shows up in the agent panel and is missing from the overview, which is the same kind of mismatch this PR is fixing.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -509,6 +509,70 @@ public static Optional<MappingConstructorExpressionNode> agentConfigLiteral(Expr
...
+                        }
+                    }
+                }
+            }
+        }
+        return entries;
+    }
+
+    // A capability name as written: a string literal carries escapes, a quoted identifier a prefix.
+    private static String literalOrIdentifier(String source) {
```
NipunaRanasinghe: This unescapes the literal body and strips a leading `'`; `CodeAnalyzer.stripQuotes` (`:1713`) does neither. A key written as `'wait: {...}` reads as `wait` here and `'wait` in the panel.

More generally, the description says all three sites now agree, but `CodeAnalyzer` still has its own keyed/list dispatch at `:1554-1672` and doesn't call `capabilityEntries`. Any reason not to move it over as well?

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -509,6 +509,70 @@ public static Optional<MappingConstructorExpressionNode> agentConfigLiteral(Expr
...
+     * by capability name — {@code events: {chat: {request: string}}}, what the module documents —
+     * or the list of records that carry their own {@code name} field, which it still accepts.
+     *
+     * <p>Reading only one form leaves the other invisible, and a capability the tooling cannot see
+     * is one it silently drops the next time it writes the declaration back.
+     *
+     * @param value the value of an {@code events}, {@code humanTasks} or similar config field
+     * @return the entries it declares, empty when the value is neither form
+     */
+    public static List<CapabilityEntry> capabilityEntries(ExpressionNode value) {
```
NipunaRanasinghe, line 535: Reading both forms doesn't stop the drop the doc comment describes. `insertAgentCapabilityEntry` (`:377`) still throws "is not a list literal" for a keyed field, and `replaceAgentCapabilityEntry` splices list-form text over the entry's line range — `DurableAgentHumanTaskBuilder:234` always emits `{name: ..., roles: ...}`, so replacing a keyed entry produces `humanTasks: { {name: "escalation", ...}, signoff: {...} }`, which doesn't compile. (The new fixture also uses `userRoles` where the builder writes `roles`.)

Could the description say this is read-side only? And now that the overview draws these circles, can a user click one through to an edit form? `CodeAnalyzer` already exposed them in the designer panel, so it may well be pre-existing — just want to be sure this doesn't open a new path to broken source.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -509,6 +509,70 @@ public static Optional<MappingConstructorExpressionNode> agentConfigLiteral(Expr
...
+        }
+        if (value instanceof ListConstructorExpressionNode list) {
+            for (Node item : list.expressions()) {
+                if (!(item instanceof MappingConstructorExpressionNode config)) {
+                    continue;
+                }
+                for (MappingFieldNode field : config.fields()) {
+                    if (field instanceof SpecificFieldNode entry && entry.valueExpr().isPresent()
+                            && "name".equals(entry.fieldName().toSourceCode().trim())) {
+                        String name = literalOrIdentifier(entry.valueExpr().get().toSourceCode().trim());
```
NipunaRanasinghe: Nit: no `break` after the first `name` match, so a mapping with two `name` fields now yields two entries where `getMappingStringField` yielded one. That's invalid Ballerina anyway, so low priority. Also `!name.isEmpty()` here vs `isBlank()` in `CodeAnalyzer`.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/DesignModelGenerator.java
```diff
@@ -404,13 +404,17 @@ private void populateAgentDeclaredCapabilities(IntermediateModel intermediateMod
...
                 switch (fieldName) {
-                    case "events" -> populateAgentEvents(agent, list);
-                    case "humanTasks" -> populateAgentHumanTasks(agent, list);
-                    case "activities" -> linkAgentActivities(intermediateModel, agent, list);
+                    // Declared either keyed by name — `events: {chat: {request: string}}`, what the
+                    // module documents — or as the list of records it still accepts. Both are read:
+                    // a capability the overview cannot see is one the diagram simply omits.
+                    case "events" -> populateAgentEvents(agent, valueExpr);
+                    case "humanTasks" -> populateAgentHumanTasks(agent, valueExpr);
+                    case "activities" -> {
```
NipunaRanasinghe, line 413: `CodeAnalyzer`'s keyed branch is generic and handles `capabilityType == "activity"` too. If `activities` also became a map in 0.9.0 the same bug is still here; if it didn't, `CodeAnalyzer` is parsing a shape that can't exist. Worth stating which in the description.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/test/resources/get_design_model/config/durable_agent_capability_mapping.json
```diff
@@ -0,0 +1,107 @@
...
+              "attachedFunctions": []
+            },
+            {
+              "name": "billSubmitted",
+              "type": "string",
+              "attachedServices": [],
+              "attachedFunctions": []
+            }
+          ],
+          "humanTasks": [
```
NipunaRanasinghe, line 63: `assertWorkflows` compares human tasks by size only (`DesignModelGeneratorTest.java:116`), so these names and locations are never checked — two tasks with the wrong names and wrong ranges would still pass. The events half of the fix is covered, since those are compared by name and type.

The ranges themselves are right, for what it's worth: `escalation` at 14:8–14:72 is exactly the `SpecificFieldNode`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/build.gradle
```diff
@@ -123,6 +123,9 @@ test {
     useTestNG() {
         suites "src/test/resources/testng.xml"
     }
+
+    // The concurrency guard resolves packages on 16 threads; the default heap runs out on CI runners.
+    maxHeapSize = "2g"
```
NipunaRanasinghe, line 128: This fixes a `PackageUtilConcurrencyTest` OOM and has nothing to do with capability parsing. Fine change in itself, but the description doesn't mention it — either split it out or add a line about it.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
hasithaa: Fixed. `capabilityEntries` now keeps a keyed entry whose value is not an inline mapping, with `config == null`, so `humanTasks: {escalation: SOME_CONFIG_CONST}` reads as a capability in both places. The event reader falls back to `anydata` for the request type when there is no inline config to read it from.

The new fixture declares one such task (`audit: auditTask`) so this stays covered.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
hasithaa: Both fixed. `CodeAnalyzer.collectKeyedCapabilities` now delegates to `capabilityEntries`, which is also what decides the name — `WorkflowUtil.capabilityName`, one rule for a string literal and for a quoted identifier. `'wait` reads as `wait` everywhere now. The list branch furthermore normalizes its `name` value through the same function, and `collectCapabilityFields` unescapes keyword field names the way the list branch already did.

I kept the rest of `CodeAnalyzer`'s list dispatch where it is: it hydrates form values (the `retryPolicy` decomposition, the `bindings` explosion, the bare function/variable reference form) that the overview has no use for, so moving it would mean pushing form concerns into `WorkflowUtil`. What the two must agree on is *what an entry is and what it is called*, and that part is now shared.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
hasithaa: You are right on both counts, and the description now says so.

It is read-side only. `insertAgentCapabilityEntry` still throws for a keyed field and `replaceAgentCapabilityEntry` splices list-form text, so editing a keyed declaration through those paths produces source that does not compile. Pre-existing — `CodeAnalyzer` has exposed these in the panel since before this PR.

On the new path: no. The overview's human-task box (`GeneralWidget.tsx:169`) renders `readonly` with no click handler, so a circle drawn here does not reach an edit form. Nothing in this PR opens a way to broken source; making the write side speak both forms is a separate change.

On `userRoles` vs `roles`: the fixture is right for the form it uses. `HumanTaskDefinition` includes `ReviewTaskDefinition`, whose field is `userRoles`; the list form's `HumanTaskDecl` used `roles`. That the builder emits `roles` is exactly the write-side mismatch above.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
hasithaa: Both fixed — `break` after the first `name` match, and `isBlank()` to match `CodeAnalyzer`. (`CodeAnalyzer` now takes its name from the same function, so the two cannot diverge here again.)

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/DesignModelGenerator.java
hasithaa: Checked the balas rather than guessing: `activities` is `(ActivityDecl|function)[]` in 0.9.0 and still in 0.10.0 — only `events` and `humanTasks` became maps. So the overview is right to read activities in list form only, and `CodeAnalyzer`'s keyed branch is generic over a shape activities cannot have today. Stated in the description.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/test/resources/get_design_model/config/durable_agent_capability_mapping.json
hasithaa: Good catch — fixed. `assertWorkflows` now calls `assertHumanTasks`, which compares each task's name and location instead of just the count. Verified it bites: renaming `escalation` to `wrongName` in the golden fails the test, where before it passed.

The fixture also gained a third task, `audit: auditTask`, whose config is a reference rather than an inline mapping — the case from your other comment.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/build.gradle
hasithaa: Fair — the description now calls it out: the heap bump is for `PackageUtilConcurrencyTest`, which resolves packages on 16 threads and exhausts the default heap on CI runners. I kept it in this PR rather than splitting it, since without it the branch cannot go green; happy to lift it into its own PR if you would rather keep this one to the parse fix.

NipunaRanasinghe (approved):
LGTM

PR #1097, opened 2026-09-16
[Agent Builder] Add LS design-model support for agent overview diagram

dan-niles:
Adds agent-aware design model generation (`CodeAnalyzer`, `ConnectionFinder`, `DesignModelGenerator`, new `model.AgentCall`/`Connection` fields) that the frontend agents overview canvas ([next PR in the stack](https://github.com/ballerina-platform/ballerina-vscode/pull/1098)) renders.


dan-niles:
@coderabbit review

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/test/resources/get_design_model/source/project_20/tools.bal
```diff
@@ -0,0 +1,27 @@
+import ballerina/ai;
+import ballerina/http;
+
+final http:Client billingClient = check new ("http://localhost:9090");
+
+// Agent-as-tool: delegates to specialistAgent -- exercises Connection.delegatesTo.
+@ai:AgentTool
+isolated function delegateToSpecialist(string query) returns string {
+    string|error result = specialistAgent.run(query);
```
gigara, line 9: **High — fixture split across the wrong packages.** `specialistAgent` (and `supervisorAgent` in `helpers.bal`, `supportModel` in `typedagent.bal`) are defined only in **project_19**, so project_18 has unresolved symbols. Since the expected JSONs were regenerated from this broken state, the features these files claim to exercise are silently asserted as *absent*:

- `delegatesTo` / `agentTools` (agent-as-tool) are empty in **every** config in the suite — the key `delegatesTo` appears in no expectation file — despite the comment above saying "exercises Connection.delegatesTo".
- The typed agent's positional model provider is untested: `typedAgent` expects `modelProvider: null` because `supportModel` is undefined here.
- `ticketTools` is defined here but never used by any project_18 agent, so the named-toolkit (`SimpleNameReferenceNode`) label path is untested too.

Suggested fix: move `tools.bal`/`helpers.bal` into project_19 (their agents live there), give project_18 its own `supportModel` (and keep/duplicate `billingClient` for the typed agent), regenerate expectations, and sanity-check that `delegatesTo`, `agentTools`, and `mcpToolKits: ["ticketTools", ...]` come out non-empty. A cheap regression guard: assert in the test that at least one connection in the topology fixture has a non-empty `delegatesTo`, so a future regeneration can't zero it out unnoticed.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/test/resources/get_design_model/source/project_20/agents.bal
```diff
@@ -0,0 +1,34 @@
+import ballerina/ai;
+
+final ai:Wso2ModelProvider supportModel = check ai:getDefaultModelProvider();
+
+// Delegates to specialistAgent (via delegateToSpecialist) and uses an http:Client (via callHttpTool).
+final ai:Agent supervisorAgent = check new (
+    systemPrompt = {role: "Supervisor", instructions: string `Route to a specialist.`},
+    model = supportModel,
+    tools = [delegateToSpecialist, callHttpTool]
```
gigara, line 9: Same package mix-up from this side: `delegateToSpecialist`, `callHttpTool` (and `ticketTools` below) are defined in **project_18**, so they don't resolve here. The regenerated expectation in `agent_topology.json` shows the fallout: `supervisorAgent` has `dependentFunctions: []`, no `toolConnections`, no `agentTools`, no `delegatesTo`, and `specialistAgent`'s `mcpToolKits` contains only the inline URL — contradicting both the comment above and the config's description ("agent-as-tool delegation, a tool using an http:Client...").

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/test/resources/get_design_model/source/project_20/services.bal
```diff
@@ -0,0 +1,119 @@
...
+                _ = check specialistAgent.run(query);
+            }
+        }
+    }
+}
+
+// Calls through a local helper and through a service method keep their order -- exercises agentCalls expansion.
+service /helpers on new http:Listener(8098) {
+    resource function post pipeline(@http:Payload string query) returns string|error {
+        return runPipeline(query);
```
gigara, line 105: `runPipeline` is defined in project_18, not here, so it doesn't resolve — and the expectation for this resource is `agentCalls: []`. The module-level-helper expansion path this comment says it exercises is effectively untested (the `self.finish` method-helper path in `mixed` below does work). Moving `helpers.bal` into this project fixes it.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/model/Function.java
```diff
@@ -47,4 +50,26 @@ public int hashCode() {
         return Objects.hash(name().hashCode(), location().hashCode(), connections, workflows, workflowSendData,
                 invalidWorkflowSendData);
     }
+
+    // Mirrors hashCode's size-based fields (agentCalls excluded) instead of the record-derived, field-by-field
+    // equals, which would otherwise drift out of sync with hashCode the moment a component is added here.
+    @Override
+    public boolean equals(Object obj) {
```
gigara, line 57: **equals/hashCode contract violation, and the comment is backwards.** `hashCode` hashes the collections by content, while this `equals` compares only their sizes — so two "equal" functions with same-sized but different sets get different hash codes, which is the actual contract break. A *stricter* equals (the record-derived one, whose fields are a superset of hashCode's) can never drift out of sync with hashCode; only a looser one can.

Also, nothing appears to call `Function.equals` — `Service.equals` compares list sizes only, and the test compares agentCalls through its own helpers — so these overrides could simply be dropped. If you do want agentCalls excluded from equality, compare the same components `hashCode` uses, by content. (Same applies to `ResourceFunction.equals`.)

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/IntermediateModel.java
```diff
@@ -106,11 +119,30 @@ public static class FunctionModel {
...
+        // so a helper's agent calls can be spliced in where the helper was called.
+        protected final List<HelperCall> helperCalls = new ArrayList<>();
 
         protected void addSentEvent(String workflowUuid, String eventName) {
             this.workflowSendData.computeIfAbsent(workflowUuid, k -> new HashSet<>()).add(eventName);
         }
 
+        // Belt-and-suspenders against a node being visited twice (see the FunctionBodyBlockNode
+        // traversal note in CodeAnalyzer): a duplicate call at the same source line is dropped.
+        protected void addAgentCall(AgentCall agentCall) {
```
gigara, line 133: **Dedup keys on line alone**, so two calls to *different* agents on one line — e.g. `return check a.run(x) + check b.run(y);` — drop the second call. Dedup on `(connection, line)` instead. And since the double-visitation this guards against was fixed in this same PR (the `super.visit(functionBodyBlockNode)` removal), it may be worth checking whether the guard is still needed at all.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/DesignModelGenerator.java
```diff
@@ -221,6 +233,100 @@ public DesignModel generate() {
...
+        Deque<String> pending = new ArrayDeque<>(tool.connections);
+        while (!pending.isEmpty()) {
+            String uuid = pending.pop();
+            Connection dependentConnection = intermediateModel.uuidToConnectionMap.get(uuid);
+            if (dependentConnection == null || !seen.add(uuid)) {
+                continue;
+            }
+            if (ConnectionKind.AGENT.toString().equals(dependentConnection.getKind())) {
+                agent.addDelegatesTo(uuid);
+                agent.addAgentTool(toolName, uuid);
```
gigara, line 321: `agentTools` is `Map<String, String>` (tool → one agent uuid), so when a single tool function delegates to two agents this `put` overwrites the first edge — `delegatesTo` keeps both agents, but the per-tool attribution silently loses one. `Map<String, Set<String>>` would keep them, unless the frontend contract is deliberately one agent per tool.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/ConnectionFinder.java
```diff
@@ -305,4 +322,172 @@ public void handleInitMethodArgs(Connection connection, ExpressionNode expressio
...
+        String symbol = expr instanceof SimpleNameReferenceNode varRef ? varRef.name().text() : null;
+        connection.setMemory(new Connection.MemoryStore(symbol, CommonUtils.getTypeName(rawType)));
+    }
+
+    // A named variable resolves through its type; an inline `ai:getDefaultModelProvider()` has no variable
+    // and a union return type, so it is recognised by name.
+    private void setModelProvider(Connection connection, ExpressionNode expression) {
+        ExpressionNode expr = unwrapCheck(expression);
+        if (expr instanceof FunctionCallExpressionNode call
+                && call.functionName().toSourceCode().trim().endsWith(DEFAULT_MODEL_PROVIDER_FUNCTION)) {
```
gigara: `endsWith(DEFAULT_MODEL_PROVIDER_FUNCTION)` on source text matches any function whose name ends with `getDefaultModelProvider` — a user's own `getDefaultModelProvider()` or `myGetDefaultModelProvider()` would be misread as the WSO2 default provider. Safer to resolve the call's symbol and check its module, the way `isAgentToolAnnotation` already does with `CommonUtils.isAiModule`.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/DesignModelGenerator.java
```diff
@@ -221,6 +233,100 @@ public DesignModel generate() {
                 .build();
     }
 
+    /**
+     * For every agent connection, resolves its tool functions' own connections and splits them into
```
gigara: This javadoc describes `linkAgentToolTargets` (resolving tool connections into `delegatesTo`/`toolConnections`, running over unreachable agents), but it's attached to `expandedAgentCalls`, which is documented only by the line comment below; `linkAgentToolTargets` itself has no doc. Suggest moving it down. Also, the line comment's "expanded once and reused by every caller" only holds within one entry point's expansion — the memo map is created fresh per `expandedAgentCalls` call.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -124,6 +132,11 @@ public class CodeAnalyzer extends NodeVisitor {
     private IntermediateModel.ServiceClassModel currentServiceClass;
     private String serviceClassName;
     private Workflow currentWorkflow;
+    // Nearest enclosing if/match/fork construct a call site is in, so agent calls in different
+    // branches or workers can be told apart on the overview; null outside any such construct.
```
gigara: Stale leftover: this first sentence describes an older single-field design ("Nearest enclosing ... null outside any such construct" — the Deque is never null, just empty). The sentence below it is the accurate one; suggest deleting this one.

gigara (commented):
The production code looks solid overall and CI is green, but the new test fixtures have a package mix-up that silently removes coverage for the headline features of this PR — flagged inline with a suggested fix. A few smaller logic issues and cleanups are also noted inline, roughly in priority order.

**High**: test fixtures split across the wrong packages (see comments on `project_18/tools.bal`, `project_19/agents.bal`, `project_19/services.bal`). Because the expected JSONs were regenerated from the broken fixtures, `delegatesTo` and `agentTools` are empty in **every** config in the suite (the key `delegatesTo` appears in no expectation file), and the named-MCP-toolkit label, module-level helper expansion, and the typed agent's positional model provider are all asserted as absent — while the comments and the `agent_topology.json` description claim they are exercised.

**Medium**: `equals`/`hashCode` contract violation in `Function`/`ResourceFunction`; agent-call dedup keyed on line alone; `agentTools` losing edges when one tool delegates to two agents; default-model-provider detection by source-text suffix.

**Low**: misplaced javadoc, a stale field comment, an unused `SELF` constant opportunity, test-helper duplication, and a cosmetic escape issue.

Things I checked that are fine: no mutation-during-iteration in `linkAgentToolTargets` (the graph build only reads the connection maps); the `?wso2_icon` suffix in expectations is genuine generator output; the author-machine absolute paths in the fixture JSONs are the pre-existing convention (locations are not compared); the cross-connection config-field leak fix is properly gated in `recordAgentConfigField`; and class-method tools (`escalate`) do resolve into `toolConnections`, as project_18's `typedAgent` expectation confirms.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -124,6 +132,11 @@ public class CodeAnalyzer extends NodeVisitor {
...
+    // Enclosing if/match/fork/loop constructs of the statement being visited, outermost first.
+    private final Deque<AgentCall.Group> agentCallGroups = new ArrayDeque<>();
+    private static final String SELF = "self";
```
gigara, line 137: Nit: `isAiMethodCall` still compares against the literal `"self"` — now that this constant exists, it could be used there too.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/test/java/io/ballerina/designmodelgenerator/extension/DesignModelGeneratorTest.java
```diff
@@ -158,12 +164,106 @@ private boolean assertServices(List<Service> actual, List<Service> expected) {
...
+                            expectedService.getRemoteFunctions())
+                    || !assertResourceFunctionAgentCalls(actualService.getResourceFunctions(),
+                            expectedService.getResourceFunctions())) {
+                return false;
+            }
         }
         return true;
     }
 
+    private boolean assertFunctionListAgentCalls(List<Function> actual, List<Function> expected) {
```
gigara: Nit: this and `assertResourceFunctionAgentCalls` are structurally identical — a single generic helper taking a `Function<T, List<AgentCall>>` extractor would remove the duplication.

More substantively: `assertAgentCalls` never verifies *which* agent a call targets. The uuids are unstable across runs, but they could be resolved to connection symbols (the way `assertConnections` matches by symbol) — as written, a call attributed to the wrong agent would still pass.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/ConnectionFinder.java
```diff
@@ -305,4 +322,172 @@ public void handleInitMethodArgs(Connection connection, ExpressionNode expressio
...
+        return MCP_SERVER;
+    }
+
+    private Optional<String> firstStringArgument(NewExpressionNode newExpression) {
+        return getArgList(newExpression).stream()
+                .filter(arg -> arg instanceof PositionalArgumentNode positional
+                        && positional.expression() instanceof BasicLiteralNode literal
+                        && literal.kind() == SyntaxKind.STRING_LITERAL)
+                .map(arg -> ((BasicLiteralNode) ((PositionalArgumentNode) arg).expression()).literalToken().text())
+                .map(text -> text.substring(1, text.length() - 1))
```
gigara, line 365: Cosmetic: this strips the surrounding quotes but keeps raw escape sequences, so a URL/role containing `\"` surfaces escaped in the label. (Same in `setRoleFromSystemPrompt`.) Fine to leave if fixture strings never contain escapes.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/test/resources/get_design_model/source/project_20/tools.bal
dan-niles: Fixed — moved tools.bal/helpers.bal into project_19 and gave project_18 its own supportModel/billingClient. Regenerated agent_topology.json and ai_agent_async_usage.json; delegatesTo, agentTools and the typed agent's modelProvider now come out non-empty.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/test/resources/get_design_model/source/project_20/agents.bal
dan-niles: Same fix as the tools.bal comment — moved tools.bal/helpers.bal into project_19, both projects compile now.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/test/resources/get_design_model/source/project_20/services.bal
dan-niles: Fixed by the same move — runPipeline resolves now.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/model/Function.java
dan-niles: Not changing this. hashCode() is already size-based like equals() (it hashes .size() of each collection, not the collection itself), so there's no actual mismatch between them. Both exclude agentCalls because AgentCall.connection is a uuid regenerated every test run — a record-derived equals would make the golden comparisons flaky.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/IntermediateModel.java
dan-niles: Fixed — dedup now keys on (connection, line).

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/DesignModelGenerator.java
dan-niles: Leaving this as Map<String, String> for now. Changing the value type is a JSON schema change that needs a matching update in the frontend PR (#1098), not something to do unilaterally in the LS.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/ConnectionFinder.java
dan-niles: Fixed — resolves the call's symbol and checks its module now, same pattern isAgentToolAnnotation already uses.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/DesignModelGenerator.java
dan-niles: Fixed — moved the javadoc onto linkAgentToolTargets and corrected the memo-map comment (it's per entry-point, not global).

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/CodeAnalyzer.java
dan-niles: Fixed — removed the stale first sentence.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/CodeAnalyzer.java
dan-niles: Fixed — uses SELF now.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-ls-extension/src/test/java/io/ballerina/designmodelgenerator/extension/DesignModelGeneratorTest.java
dan-niles: Deduped assertFunctionListAgentCalls/assertResourceFunctionAgentCalls into one generic assertAgentCallsList. Leaving the target-agent verification for later — it needs resolving uuids back to stable symbols, more than a straight fix.

packages/ballerina-language-server/architecture-model-generator/modules/architecture-model-generator-core/src/main/java/io/ballerina/designmodelgenerator/core/ConnectionFinder.java
dan-niles: Leaving as-is, per your note that it's fine given fixture strings don't have escapes.

gigara (approved):

PR #1098, opened 2026-09-16
[Agent Builder] Add UI changes for agent overview diagram

dan-niles:
LS changes: https://github.com/ballerina-platform/ballerina-vscode/pull/1097

Removes the old tab based home screen and replaces it with an agent overview diagram.

https://github.com/user-attachments/assets/38c6fa0e-5383-40c9-8957-9c2b89ea0959


dan-niles:
@coderabbit review

kaumini (commented):
Three issues around the entry-card row folding and the topology hook. All three were verified against this branch with rendering tests.

packages/component-diagram/src/components/AgentTopologyDiagram/index.tsx
```diff
@@ -0,0 +1,482 @@
...
+            readonly,
+            orientation,
+            onAgentSelect,
+            onTriggerSelect,
+            onAddTrigger,
+            onConfigureEntry,
+            onDeleteEntry,
+            focus: (hoveredId ?? pinnedId) && graphRef.current ? focusAround(graphRef.current, hoveredId ?? pinnedId) : undefined,
+            setHovered,
+            visibleRows: layoutRef.current?.visibleRows,
```
kaumini: **"Show N more" never expands the entry card.**

`visibleRows` is read off the `layoutRef` mutable ref inside this `useMemo`, so `ServiceNodeWidget` renders whatever the ref happened to hold at render time.

Clicking the footer calls `onExpandEntry` → `setExpanded` → the memo recomputes (deps include `expanded`) but still reads the **old** `layoutRef.current.visibleRows`, so the widget re-renders with the same row count. `applyLayout` does eventually recompute the ref with the expanded budget (via the re-armed `ResizeObserver`), but that mutates a ref and triggers no re-render, so the card stays folded and the "Show N more" footer stays on screen.

The same staleness affects a canvas resize: `rowBudget` derives a new row count from the new height and `layoutTopology` spaces the cards for it, while the widgets keep drawing the old count — heights and positions then disagree.

Moving the computed `visibleRows` into state (or recomputing the layout in a render-triggering path) would fix both.

packages/component-diagram/src/components/nodes/ServiceNode/ServiceNodeWidget.tsx
```diff
@@ -0,0 +1,304 @@
...
+                    <Subtitle>{entry.subtitle}</Subtitle>
+                </Titles>
+                {isService && <CardMenuButton menu={menu} />}
+            </Header>
+            {isService && <CardMenuPopover menu={menu} />}
+            {isService &&
+                rows.map((handler) => <HandlerRow key={handler.id} handler={handler} dimmed={rowDimmed(handler)} />)}
+            {hidden > 0 && <Footer onClick={() => onExpandEntry?.(entry.id)}>{`Show ${hidden} more`}</Footer>}
+            {isService &&
+                rows.map((handler, index) => (
```
kaumini, line 294: **Folded handler rows lose their agent links entirely.**

`ServiceNodeModel`'s constructor creates a row port for *every* handler, but this only mounts `RowPort` for the visible `rows`. A port with no mounted `PortWidget` never gets `updateCoords`, so `reportedPosition` stays `false` — and `LinkWidget.render` in `@projectstorm/react-diagrams-core` bails outright:

```js
if (link.getSourcePort() && !link.getSourcePort().reportedPosition) return null;
```

So every edge whose `handlerId` belongs to a folded row is silently dropped. With 5 handlers each triggering its own agent on a canvas small enough to fold the card to 3 rows, the diagram draws **5 agent cards but only 3 links** — the two agents behind the folded rows sit disconnected.

Those agents aren't marked `orphan` either (reachability is computed on the graph, not on what got rendered), so they don't get the dashed border or the "No trigger yet · Add Trigger" footer. They just float with no explanation — and combined with the expand issue above, there's no way to reveal them.

The `try {} catch {}` around `getPortCoords` in `reportPorts` is the same unmounted-port condition.

packages/ballerina-visualizer/src/views/BI/AgentTopology/useAgentTopology.ts
```diff
@@ -0,0 +1,89 @@
...
+        startLine: agent.position?.startLine ?? 0,
+        moduleName: agent.moduleName,
+        isDefinition,
+    };
+}
+
+export function useAgentTopology(
+    projectPath: string,
+    agents: ProjectStructureArtifactResponse[],
+    agentDefinitions: ProjectStructureArtifactResponse[] = []
```
kaumini: **This default parameter causes an endless refetch loop.**

`= []` creates a fresh array on every call, so when the optional `agentDefinitions` prop is omitted the identity changes each render: the `artifacts` memo invalidates → `scheduleFetch` is recreated → `useEffect(scheduleFetch, [scheduleFetch])` re-runs and re-arms the timer → the fetch calls `setInput` with a new object → re-render → repeat. The result is a `getDesignModel` RPC every `DEFER_MS` (600ms), indefinitely.

Latent today, since the only caller (`AgentBuilderOverview`) passes a `useMemo`-stable array — but `AgentTopologyProps.agentDefinitions` is declared optional and forwarded straight through, so the next caller that omits it hits this.

Hoisting the default to a module-level constant avoids it. Worth noting that `agentDefinitions` is also dead as it stands: `buildAgentNodes` filters every `isDefinition: true` artifact out, and nothing else reads that field — so dropping the prop entirely would remove this hazard too.

packages/component-diagram/src/components/AgentTopologyDiagram/index.tsx
dan-niles: Fixed in 490822c — moved visibleRows into state, updated whenever applyLayout runs.

packages/component-diagram/src/components/nodes/ServiceNode/ServiceNodeWidget.tsx
dan-niles: Fixed in 490822c — folded rows now mount a port too, parked at the card's centre, so their links stop dropping.

packages/ballerina-visualizer/src/views/BI/AgentTopology/useAgentTopology.ts
dan-niles: Fixed in 490822c — default hoisted to a module-level constant.

kaumini (approved):

PR #1101, opened 2026-09-16
[Agent Builder] Add durable agents support

dan-niles:
## Purpose

Adds support for creating and managing durable agents in agent builder.

Screenshot 2026-09-16 at 2 54 00 PM

Screenshot 2026-09-16 at 2 54 06 PM

PR #1104, opened 2026-09-16
[Agent Builder] Add durable agent support to the design model

dan-niles:
## Summary
- LS-side split of #1101 (closed): extends `getDesignModel` with durable-agent support (`Workflow` role/gating on activities and tasks, tools, peers, delegation; mapping-form `events`/`humanTasks`).
- First of a three-PR stack for durable agents: LS → frontend → frontend tests. Stacked on top of #1099.

## Test plan
- [ ] `architecture-model-generator-ls-extension:test`
- [ ] `checkstyleMain` / `checkstyleTest` on `architecture-model-generator-core` and `-ls-extension`


gigara (commented):
Reviewed at head `814421e` (diff vs `ab-diagram/frontend-tests`). The design-model half is strong — `walkToolConnections` cleanly unifies the two walkers, `declaredEntries` handles both the keyed and list config forms, and the `project_20` topology test is genuinely thorough. The issues below are mostly on the service-model (codegen) side.

### Must fix

1. **Durable + shaped HTTP resource can generate non-compiling code, and mislabels the instance id as the agent's answer** — `HttpAgentTriggerChannel.java:399`. For a durable agent the generated body still calls `claimAgent.run(prompt)` (whose result is the **instance id**), but the forced `Answer.text(answer(shaped).wrapped())` returns it against the user's declared response type: a record-typed response body won't compile (`return result;` returns a `string`), and even a string-typed one hands the caller an instance id presented as the answer. `defaultResource` (line 462) handles this correctly with `Answer.instanceId()` / `http:Accepted` / `{instanceId: result}` — the shaped path should acknowledge the same way, or fall back to `RETURN_ANSWER_UNMAPPED` when the declared type isn't string.

2. **Indentation regression in every generated agent event trigger (plain agents too, not just durable)** — `EventAgentTriggerChannel.java:268`. `REPLY_METHOD` already places `{{resultLog}}` at the body's 4-space indent, and the code then substitutes `AgentTriggerChannel.indent(resultLog)`, adding another 4: the generated `log:printInfo(...)` lands at 8 spaces while its siblings sit at 4. Fix: substitute `resultLog` without the inner `indent(...)`. The tests miss this because they assert with `contains()` on statement text — one exact-block assertion per channel would harden them.

3. **CI is red on this PR's own checks** (base branch is green):
   - Checkstyle: `Workflow.java:299` — missing `@param location` javadoc tag fails `Build repo`.
   - `DesignModelGeneratorTest`: 5 cases fail, including the new `durable_agent_model_connection` / `durable_agent_topology` **and** three pre-existing agent tests (`ai_agent_module_level_usage`, `ai_agent_async_usage`, `agent_topology`). Pre-existing tests failing suggests either the output changed for plain agents or the CI distribution lacks the `workflow` module version the test projects need — worth confirming which before merge either way.

### Should fix

4. **Mutable map keys** — `DesignModelGenerator.java:435,460,465`. `Map<Workflow, LineRange>` uses keys whose `hashCode()`/`equals()` depend on the sizes of `events`/`humanTasks`/`activities` — exactly what `populateAgentDeclaredCapabilities` mutates during the loop. It works today only because nothing re-looks-up a key after mutation; an ordered list of pairs would be safer. Also worth a one-line comment on *why* capability population is deferred (forward `peers` references) — and note the test project only exercises a backward peer reference, so the deferral isn't actually proven by a test.

5. **Mapping-form `tools` entries drop gating, asymmetrically with `activities`/`peers`** — `DesignModelGenerator.java:590`. `populateAgentTools` reads only the `tool` field of a `{tool: x, ...}` entry, while the sibling handlers capture `requiresApproval`/`userRoles`. If the `workflow:DurableAgent` config supports gated tool entries, the design model silently under-reports them — please confirm against the workflow library.

### Minor

6. Generated `instanceFor` holds the service lock across the potentially slow `{{agent}}.run(...)` call (`AgentTriggerContext.java:66-76`) — one session's first turn blocks every other session's `durableTurn`. Either narrow the lock or add a comment acknowledging the serialization, since this is emitted into user projects.
7. The event resource silently takes the **first** path parameter as the instance id (`HttpAgentTriggerChannel.java:433-440`) — for `[string tenant]/instances/[string instanceId]`, `sendData` targets `tenant`. Prefer a param named `instanceId` when present, or document the convention in the `GenerationRefusedException` message.
8. Two `isEventTrigger()` methods with different semantics: `GetServiceInitModelContext.isEventTrigger()` is "eventChannel non-blank" while `AgentTriggerContext.isEventTrigger()` also requires durable. `AgentTriggerServiceBuilder:114` compensates with `durable && ...`, but `HttpAgentTriggerChannel.additionalProperties:324` uses the unchecked one. Folding the durable check into the context record would align them.
9. `DesignModelGenerator.isMcpToolKit` (line 608) duplicates `ConnectionFinder`'s private helper byte-for-byte — the PR already opened up `mcpToolKitLabel`/`getArgList` for sharing, so this one could be shared the same way.
10. Test gap: `assertDurableAgentFacts` (`DesignModelGeneratorTest.java:135-146`) asserts human-task roles but never titles, even though both configs carry stable `"title"` values — mapping to `userRoles() + "/" + title()` would close it for free.

### Javadoc / docs

- `@param` lists not updated for new record components: `AgentTriggerContext` (missing `agentKind`), `GetServiceInitModelContext` and `ServiceModelRequest` (missing `agentKind`, `eventChannel`, `eventResponse`).
- `Workflow.HumanTask` lost its `@param` docs when gaining `userRoles`/`title`, and its description ("awaited inside a workflow function via `ctx->awaitHumanTask(...)`") is stale for durable agents, where the task comes from the config's `humanTasks` mapping. `ActivityDecl` / `PeerDecl` have no javadoc at all, unlike every sibling record.

dan-niles:
Addressed in 33e9964e7d:

- Durable + shaped HTTP resource: now falls back to the unmapped-response TODO instead of returning the instance id as the answer.
- Event-trigger indent: removed the double `indent()` call.
- Checkstyle: added the missing `@param` tags on `HumanTask`/`ActivityDecl`/`PeerDecl`.
- The 5 `DesignModelGeneratorTest` failures were stale goldens (local `ballerina/ai`/`ballerina/http` versions moved since they were captured), not a regression — regenerated and verified.
- Mutable map key: `durableAgents` is now a list of pairs instead of a `Map<Workflow, LineRange>`.
- Path-parameter instance id: prefers a param literally named `instanceId` over the first one.
- Unified the two `isEventTrigger()` methods; removed the now-redundant `durable &&` guard.
- `isMcpToolKit` is now a single shared helper instead of duplicated.
- Filled in the missing `@param` tags on `AgentTriggerContext`, `GetServiceInitModelContext`, `ServiceModelRequest`.

Not addressed: mapping-form `tools` entries dropping gating is real (confirmed against the `workflow` library's `ToolDecl`), but fixing it needs a wire-type change into the frontend PR, so left open for now. The `instanceFor` lock serialization got an acknowledging comment rather than a restructure, since narrowing it risks a race between concurrent first turns.

gigara (approved):

PR #1105, opened 2026-09-16
[Agent Builder] Draw durable agents on the overview canvas and Add Agent flow

dan-niles:
## Summary
- Frontend split of #1101 (closed): the durable-agent box widget (bi-diagram), overview canvas durable-agent card support (component-diagram), the Find panel replacing the Entry points chip, and the Add Agent / Add Trigger UI for durable agents.
- Second of a three-PR stack for durable agents: LS → frontend → frontend tests. Based on #1104.

## Test plan
- [ ] `component-diagram` tsc (both tsconfigs) and eslint
- [ ] `bi-diagram` tsc (both tsconfigs)
- [ ] `ballerina-visualizer` tsc
- [ ] Manual: durable page, overview canvas, Add Agent popup, Find panel — see `specs/agent-builder/durable-agents/README.md` "Resume here" checklist


gigara (commented):
Reviewed at head `ec475a3` (diff vs `ab-durable-agents-split/ls-changes`). Large but well-organized: `DurableAgentRunNodeModel` extends `BaseAgentNodeModel`, the durable widget reuses `EdgeAddButton`/`UsageIcon`/`usageFadeIn` instead of copying, the sizing math is centralized in `agentNodeLayout.ts` and matches on both the widget and `SizingVisitor` sides (verified, including reference mode), and `useDurableAgentUsages` is careful about caching/staleness/animation. The blockers are all seam issues rather than design issues.

### Blocking — the PR doesn't compile on its own

1. **The stack is cut at the wrong seam.** `component-diagram` builds with plain `tsc` and its `tsconfig.json` includes `src/test`, but this PR makes `TopologyAgentNode` fields (`kind`, `channels`, `people`, `activities`, `gatedActivities`, `humanTasks`, `peers`), `TopologyHandler.wired`, `TopologyLayout.edgeLanes`, `TopologyFocus.inlets`, and the new `estimateAgentCardHeight(node)` signature **required**, while the matching test updates live in #1106. CI confirms: `topologyLayout.test.ts` / `topologyFocus.test.ts` / `topologyBounds.test.ts` fail tsc on this PR. Either fold the existing-test updates into this PR or make the new fields optional with defaults.
2. **`AgentCardNodeWidget.tsx:665`** still passes `orphan={node.orphan}` to `Card`, whose styled-component props were narrowed to `{hovered, receded, readonly?, railless?}` when the orphan border styling was removed — this is the `TS2322` failing CI. Remove the attribute (or re-add the prop if dropping the orphan border was unintentional).
3. **`MainPanel.tsx:351-357` and `:912`** pass props the receiving components don't declare: `isInDevant`/`isICPSupported` on `AgentBuilderOverview` (its props are only `projectPath`) and `initialModuleIdentifier` on `ViewConfigurableVariables`. Compile errors under ts-loader, and the values are silently dropped — these look like stray changes from another branch; either complete the wiring or drop them.

### Should fix

4. **Skeleton can hang forever** — `durableAgentUsages.ts`: when `getDesignModel` *resolves* but `designModel` is absent (LS not ready) and nothing was shown before, `durableUsagesReady` stays false and `FlowDiagram` keeps rendering `DiagramSkeleton` with no retry until an unrelated project-content update. The `catch` branch falls back to `show([])`; the missing-model branch should do the same (or retry).
5. **Wrong workflow binding** — `topologyModel.ts`: `buildWorkflowNodes` passes *all* workflows into `findDurableWorkflow`, whose name-based fallback can bind a durable-agent artifact to a **plain** workflow sharing the name (wrong capabilities on the card + a spurious duplicate node for the real durable workflow). `FocusFlowDiagram/agentUsages.ts`'s `findDurableAgentWorkflow` filters `kind === "DURABLE_AGENT"` first — do the same here, ideally via one shared helper.
6. **Dead code from the settings-button removal**: `onConfigureAgent?` in `AgentNodeActions.ts:44`, plus `handleOnConfigureAgentIdentifier` and its wiring at `FlowDiagram/index.tsx:3983,4224`, have no remaining caller after `onConfigureAgentClick` was removed from `DurableAgentRunNodeWidget`. Remove them — or confirm that losing "Configure Agent Identifier" with no replacement is intended.

### Minor

7. `methodColor` / `DEFAULT_METHOD_COLOR` (and the method-pill styling) are duplicated between `FindPanel.tsx:28-29` and `ServiceNodeWidget.tsx:41-42`; `PinBanner` already imports `MethodPill` from `FindPanel`, so `FindPanel` is becoming the accidental home of shared bits — worth extracting one helper next to the shared node styles.
8. Theme captured at module load — `AddAgentPopup/CreateNewSection.tsx:71-83`: `agentIcon`/`durableAgentIcon` call `getNodeChartColor(...)` once at import, so the color goes stale after a VS Code theme switch. Evaluate per render like the other call sites.
9. Truncated comment — `AgentBuilderOverview/index.tsx:230` ends mid-sentence: "…so opening a".
10. Dead anchors — `DurableAgentRunNodeWidget.tsx:351-375`: after replacing the corner "+" affordances with `CapabilityAddTile`, only `"topRight"` of the 9 `AffordanceAnchorName` members is used; the rest (and the stacked-pairs comment) are dead.
11. `withDurableUsages` recomputes `findDurableAgentBox(flow)` inside the `.map` for every node (`durableAgentUsages.ts`) — hoist it out of the loop.

### Copy

- `CreateNewSection.tsx:636`: "Can wait days for a person or an event and **resumes** after a restart" → "**resume**" (parallel with "wait").
- `AddAgentTriggerPopup.tsx:44-53`: only one of the four subtitle variants ends with a period — make them consistent.

dan-niles:
Replying to https://github.com/ballerina-platform/ballerina-vscode/pull/1105#pullrequestreview-5238784088

Pushed fixes for the compile breaks (orphan prop, stray MainPanel props), the skeleton-hang and workflow-binding issues, the dead onConfigureAgent code, the stale theme color, the truncated comment, the dead anchors, the map-loop hoist, the copy nits, and the methodColor duplication (`61b54d6e11`).

On the stack seam (component-diagram not building standalone since the field/type updates live in the tests PR): confirmed real. Since the stack is meant to land together, leaving this as-is for now rather than folding those test changes in here or loosening the types. Open to either if you'd rather not carry it this way.

gigara (approved):

PR #1106, opened 2026-09-16
[Agent Builder] Add tests for durable agent canvas support and the Find panel

dan-niles:
## Summary
- Test split of #1101 (closed): unit tests and snapshots for the durable-agent overview card, the Find panel, and the durable usage rail added in #1105.
- Third of a three-PR stack for durable agents: LS → frontend → frontend tests. Based on #1105.

## Test plan
- [ ] `component-diagram` jest (incl. refreshed snapshots)
- [ ] `bi-diagram` jest
- [ ] `ballerina-visualizer` jest (`FlowDiagram`, `FocusFlowDiagram`, `AIChatAgent`)
- [ ] `ballerina-extension` jest (`test-support/durableAgentArtifacts.test.ts`)


gigara (commented):
Reviewed at head `c71f0d0` (diff vs `ab-durable-agents-split/frontend`). Strong test PR overall: pure-function tests for findRows/focus/layout/model geometry, a correctly choreographed fake-timer test for the usage-rail refetch race (hand-traced against the hook), RTL tests for the Find panel's pin/isolate/keyboard flows, and one conventional snapshot for the durable card. Every assertion I spot-checked matches production exactly, which makes the test plan credible. Two substantive issues and some hygiene:

### Should fix

1. **Vacuous test — the "bare canvas" click never reaches the canvas.** `AgentTopology.test.tsx:366`: `data-testid='diagram-canvas'` exists nowhere in production, so the `??` fallback always clicks `dom.container.firstElementChild` — the outer Root wrapper, an *ancestor* of the `Glide` element carrying `onCanvasClick` (`AgentTopologyDiagram/index.tsx:491,592`). "keeps the pin when the bare canvas is clicked" passes no matter what the canvas click does. Either add the testid to the real canvas element in production or click the element the engine renders.
2. **The tested function is a pass-through; the real branching is untested.** `addAgentTrigger.test.ts:42` calls `openAddAgentTrigger`, which just forwards its four arguments into `openView` (`utils.ts:592-607`). The load-bearing durable logic lives in `startAddAgentTrigger` (`utils.ts:555-563`): forcing `agentOrgName` to `"ballerina"` for durable agents vs. `node.codedata?.org` otherwise, plus the missing-name guard — wired into both FlowDiagram and FocusFlowDiagram, and a regression there wouldn't be caught. Test `startAddAgentTrigger(box, rpcClient)` instead of (or alongside) the pass-through.

### Minor

3. **Fixture duplication is compounding**: the `agent()` default builder is copy-pasted in `findRows.test.ts:23`, `topologyFocus.test.ts:25`, and `topologyLayout.test.ts:47` — this PR had to add the same six new default fields to each copy, which is exactly the drift a shared builder module in `src/test/` would prevent. Likewise the durable-claims fixture exists in three near-identical variants (`AgentTopology.test.tsx:109-140`, `topologyModel.test.ts:101-129`, `topologyDescribe.test.ts:74+`).
4. **`HumanTask` titles asserted nowhere** despite the fixtures carrying stable values ("Manager sign-off", "Release payment") — cheap to add alongside the role assertions.
5. Only one path of `useDurableAgentUsages` is tested (refetch-after-update); the rpc-failure path (`durableAgentUsages.ts:127-131`) and the cached-hit/first-show behavior aren't. Also the test dodges the never-cleared module-level `usageCache` with `agent-${Date.now()}` (line 131) — works, but will silently bite the next hook test; a test-only cache reset or `jest.isolateModules` would be sturdier.
6. `AgentTopology.test.tsx:238-247` awaits the snapshot helper then renders a second diagram instance for the direct assertions — split into two tests or reuse the first render. The popover test at line 249 is fully synchronous but keeps a copy-pasted 15000ms timeout and sits under the "Snapshot Tests" describe while snapshotting nothing.
7. Positional fragility — `AgentTopology.test.tsx:251`: `const [people, channels] = view.getAllByText("1")` relies on exactly two circles showing "1" and their DOM order (currently correct); if the fixture gains a peer or drops an activity, the destructure hovers the wrong circle with a baffling failure. An accessible name per capability circle would make this robust.
8. Magic `40` — `agentFocusFit.test.ts:65` hardcodes the un-exported `verticalBias` from `fitAgentFocus`; the test already imports the other fit constants, so export the bias or name it in a comment.

### Test-name wording

- `topologyLayout.test.ts:333` "gives a row's edge no lane left to right" is hard to parse — it means no lanes are assigned in the horizontal orientation; say that.
- `topologyModel.test.ts:747` "…a durable workflow the artifact list does not know" → "does not list".

dan-niles:
Replying to https://github.com/ballerina-platform/ballerina-vscode/pull/1106#pullrequestreview-5238784352

Pushed fixes (2d7fe211f8):
- Bare-canvas-click test: added a real testid to the canvas element so it exercises onCanvasClick instead of an ancestor that never bubbles into it.
- Tested startAddAgentTrigger's org-forcing and missing-name-guard branches directly, not just the pass-through.
- Added coverage for useDurableAgentUsages' rpc-failure and missing-design-model paths.
- Extracted the duplicated agent() fixture builder into a shared file.
- Reused the snapshot render instead of a second one, moved the popover test out of the snapshot describe block, dropped its unneeded 15s timeout.
- Named the agentFocusFit test's magic 40.
- Two test-name wording fixes.

Left the HumanTask title suggestion out: title isn't read anywhere in production, so an assertion on it wouldn't test anything real. The positional getAllByText fragility also stays as-is for now, since fixing it needs an accessible name added to production markup.

gigara (approved):

gigara (approved):

PR #1108, opened 2026-09-16
Follow the workflow 0.10 task model in the designer's source-side forms

hasithaa:
## Purpose

First cut of the designer's side of the workflow 0.10 task model (Phase 5 of the module plan): everything that reads or writes *source* rather than the pinned bala's signatures, so it can land ahead of the 0.10 pin.

## What changes

- **Approval gates emit `approvalPolicy`.** The activity and tool capability forms keep their *Requires Approval* flag and reviewer roles, and gain *Users*, *Excluded Users* and *Excluded Roles*; on save they write `approvalPolicy: {userRoles: …, users: …}` instead of `requiresApproval`/`userRoles`. Opening a declaration hydrates `approvalPolicy` back into those fields; the removed 0.9 fields no longer hydrate.
- **Peers follow `PeerDecl`.** The peer form is `agent`, `description`, `allowedEvents`; `name`, `'wait`, `callbackChannel` and the gate are gone (the peer is named by its variable; each allowed event becomes a tool of its own).
- **Human-task capabilities** write `userRoles` (not the legacy `roles`) and accept `users`/`excludedUsers`/`excludedRoles`; roles may be empty when users are named.
- **Retry policy** gains *Retry, then Review* (`RetryBeforeReview`: attempts, then a review) beside *Auto Retry* and *Human Review*, and the review sub-form carries the audience fields. Parsing tells the three shapes apart the way the runtime does; a review naming nobody writes `userRoles: ()` so the compiler names the omission. The first option is now labelled *No Retry* (`NoRetry` is canonical in 0.10; `NoAutomaticRetry` still hydrates).
- **Migration quick-fix** (`MigrateWorkflowDeclarationCodeAction`) on `WORKFLOW_163`: `requiresApproval: true` + `userRoles: X` → `approvalPolicy: {userRoles: X}`; a peer's `name`/`'wait`/`callbackChannel` → removed. It can only be exercised once the pinned bala's plugin emits `WORKFLOW_163`, so it has no LS test yet.

## Gated on the 0.10 pin (not here)

`awaitHumanTask`'s audience fields and `approvalPolicy` on `callActivity` are signature-derived from the pinned bala (0.9.0); they follow the pin bump, together with the code-action test, the LS goldens for 0.10-syntax sources, and the peer diagram entries.

## Goldens

Regenerated only the goldens the form work changes (approval/peer/retry); the machine-dependent drift set (`currency_converter*`, `new_connection*`, …) was left untouched. Two peer to_source fixtures were rewritten around `allowedEvents`; new fixture for retry-then-review emission.

- Updated designer forms for the workflow 0.10 task model.
- Added `approvalPolicy` support with reviewer audiences and review details.
- Updated peer declarations to use `agent`, `description`, and `allowedEvents`.
- Added audience fields to human tasks and retry policies.
- Added the “Retry, then Review” option and support for 0.10 retry shapes.
- Added the `WORKFLOW_163` migration quick fix for legacy approval and peer fields.
- Prevented unsupported open-record fields and empty values from producing invalid activity output.
- Updated timeout guidance, tests, fixtures, goldens, and pinned workflow dependencies to 0.10.0.


packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1685,11 +1693,15 @@ private void collectCapabilityFields(MappingConstructorExpressionNode config, St
             if ("name".equals(fieldName) || fieldName.equals(refField)) {
                 continue;
             }
+            String rawValue = specificField.valueExpr().get().toSourceCode().trim();
+            if (WorkflowUtil.APPROVAL_POLICY_FIELD.equals(fieldName)) {
```
NipunaRanasinghe, line 1696: **Blocker: `approvalPolicy` never hydrates for activities or tools — the only two capabilities that write it.**

`hydrateApprovalPolicy` is reachable only from `collectCapabilityFields`, which is called only by `collectKeyedCapabilities` — the `key: {...}` mapping form, used for `events` and `humanTasks`. But `activities` and `tools` are written as **list** constructors (`upsertAgentCapabilityEntry` appends `{activity: a, approvalPolicy: {...}}` to `activities: [...]`), so they are read by the list branch of `collectDeclaredCapabilities`, which does `fieldToPropertyKey.get(fieldName)` and skips on null. `approvalPolicy` is not in either map — they are now just `Map.of("activity","activity","name","name","description","description")`.

Failure: gate an activity (*Requires Approval* + reviewer roles) → save writes `{activity: a, approvalPolicy: {userRoles: "mgr"}}` → reopen the activity's form → the gate is off and the roles are empty → save again → `approvalPolicy` is gone from source.

This contradicts "Opening a declaration hydrates `approvalPolicy` back into those fields" in the PR description. Nothing catches it because no `.bal` fixture in the repo contains `approvalPolicy` — the list path needs the same branch, and a source fixture with a gated activity and a gated tool would pin both.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -2400,18 +2430,30 @@ static RetryPolicyForm normalizeRetryPolicy(String rawValue) {
...
-                // rule the compiler plugin and the runtime apply.
+                // Three shapes share one union, told apart the way the compiler plugin and the
+                // runtime tell them: an audience makes a review, attempts make retries, both make
+                // retries followed by a review.
                 Map<String, String> fields = WorkflowUtil.parseRecordLiteral(rawValue);
-                if (fields.containsKey(USER_ROLES_FIELD)) {
-                    dropdownValue = ActivityCallBuilder.MANUAL_RETRY_VALUE;
+                boolean audience = fields.containsKey(USER_ROLES_FIELD)
+                        || fields.containsKey(WorkflowUtil.USERS_KEY);
+                boolean attempts = fields.containsKey(ActivityCallBuilder.MAX_RETRIES_KEY);
```
NipunaRanasinghe: **Blocker: the attempt fields are dropped unless `maxRetries` is present.**

`attempts` keys off `maxRetries` alone, and the four attempt reads moved inside `if (attempts)`. But every AutoRetry field has a default (`MAX_RETRIES_DOC`: "default: 3", `RETRY_DELAY_DOC`: "default: 1.0"), so a record can legitimately omit `maxRetries`:

- `retryPolicy = {retryDelay: 5.0, retryBackoff: 3.0}` → `audience` false, `attempts` false → AutoRetry selected with all four sub-fields blank → `autoRetryRecordLiteral` re-emits `{}` on save. The retry configuration is silently wiped. Before this change the `else` branch read them unconditionally, so this is a regression.

The same defect hits the shape this PR adds: select *Retry, then Review*, leave *Max Retries* empty and fill *Retry Delay* → `retryBeforeReviewRecordLiteral` writes `{retryDelay: 1.5, userRoles: "ops"}` → reading that back gives `attempts` false → plain *Human Review*, delay gone. The new option does not round-trip through its own writer.

Discriminating on "any of `maxRetries`/`retryDelay`/`retryBackoff`/`maxRetryDelay` present" fixes both, and `workflow_retry_policy_parsing.json` is the natural place to pin a `{retryDelay: …}`-only record.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1702,7 +1714,25 @@ private void collectCapabilityFields(MappingConstructorExpressionNode config, St
...
+    // or a policy the form cannot read, leaves the gate off.
+    private static void hydrateApprovalPolicy(ExpressionNode policy, Map<String, String> values) {
+        if (policy.kind() != SyntaxKind.MAPPING_CONSTRUCTOR) {
+            return;
+        }
+        values.put("requiresApproval", "true");
+        for (MappingFieldNode field : ((MappingConstructorExpressionNode) policy).fields()) {
+            if (field instanceof SpecificFieldNode specific && specific.valueExpr().isPresent()) {
+                String key = specific.fieldName().toSourceCode().trim();
+                if (TEXT_MODE_CAPABILITY_FIELDS.contains(key)) {
```
NipunaRanasinghe: **The policy's `description` (and `title`) overwrites the capability's own.**

The policy's fields are written into the shared `values` map under their bare source names, and `description` is also the activity/tool capability's own form key (`Map.of(…, "description", "description")`). `TEXT_MODE_CAPABILITY_FIELDS` contains both `description` and `title`, so for

```ballerina
{activity: chargeCard, description: "Charges the card",
 approvalPolicy: {userRoles: ["manager"], description: "Approve the charge"}}
```

the activity's *Description* form field ends up holding `"Approve the charge"`, and saving writes that back to the activity. Which one wins is decided by field order in source, since `putIfNotBlank` does not check for an existing entry.

Restricting the copy to the audience keys (`userRoles` + `AUDIENCE_KEYS`) avoids the collision — and those are the only policy fields the gate form can show today anyway (see the separate note on `approvalPolicyLiteral`).

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -1241,6 +1264,88 @@ public static void addApprovalGateProperties(NodeBuilder nodeBuilder, String app
...
+     * @param approvalKey   property key of the gate flag
+     * @param userRolesKey  property key of the reviewer roles
+     * @return the record literal, or {@code ""}
+     */
+    public static String approvalPolicyLiteral(SourceBuilder sourceBuilder, String approvalKey,
+                                               String userRolesKey) {
+        boolean gated = sourceBuilder.getProperty(approvalKey)
+                .map(p -> p.value() != null && "true".equals(p.value().toString()))
+                .orElse(false);
+        return gated ? reviewAudienceLiteral(sourceBuilder, userRolesKey, List.of()) : "";
```
NipunaRanasinghe: **An existing policy's `title`, `description` and `timeout` are dropped on edit-save.**

The gate form carries the flag, `userRoles` and the five audience fields, and this literal writes exactly those. So a hand-written (or runtime-documented) policy

```ballerina
approvalPolicy: {userRoles: ["manager"], title: "Approve refund", timeout: {hours: 4}}
```

comes back as `approvalPolicy: {userRoles: ["manager"]}` the first time someone opens that activity's form and saves it. The `retryPolicy` review sub-form carries title/description/timeout for exactly this reason, and `collectDeclaredCapabilities` already states the principle — "a capability the designer cannot see is one it silently drops on the next save".

Either give the gate form the same three fields, or say in the PR description that they are deliberately out of scope for this cut.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -2442,6 +2484,11 @@ static RetryPolicyForm normalizeRetryPolicy(String rawValue) {
                 maxRetryDelay, review);
     }
 
+    // `userRoles: ()` says the users alone decide; the form shows that as an empty roles field.
+    private static String nilAsBlank(String value) {
```
NipunaRanasinghe, line 2508: Nit: `nilAsBlank` is applied only on the retry path. A human-task capability or an `approvalPolicy` written as `userRoles: ()` — which `DurableAgentHumanTaskBuilder` and `reviewAudienceLiteral` both emit whenever roles are empty — hydrates the literal `()` into the *User Roles* text box via `collectCapabilityFields`/`hydrateApprovalPolicy`. It does round-trip (`EXPRESSION_LIKE` matches the paren, so `roleSource` leaves it alone), but the user sees `()` sitting in a roles field. Worth applying the same helper in both hydration paths.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentHumanTaskBuilder.java
```diff
@@ -232,7 +234,13 @@ public Map<Path, List<TextEdit>> toSource(SourceBuilder sourceBuilder) {
         String timeout = sourceBuilder.getProperty(TIMEOUT_KEY)
                 .map(p -> p.value() == null ? "" : p.value().toString().trim()).orElse("");
         StringBuilder entry = new StringBuilder("{name: ").append(WorkflowUtil.constantNameLiteral(name))
-                .append(", roles: ").append(WorkflowUtil.quoteIfBareRole(roles));
+                .append(", userRoles: ").append(roles.isBlank() ? "()" : roles);
+        for (String key : WorkflowUtil.AUDIENCE_KEYS) {
```
NipunaRanasinghe: Nit: this is `WorkflowUtil.reviewAudienceLiteral` re-implemented — the same `userRoles: ()` fallback followed by the same `AUDIENCE_KEYS` loop. And `reviewAudienceLiteral`'s `moreFields` parameter is currently dead (`approvalPolicyLiteral` is its only caller and always passes `List.of()`), while the tail this method builds — `resultType`, `title`, `timeout` — is exactly what that parameter is shaped for. Either call the helper here, or drop `moreFields`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1685,11 +1693,15 @@ private void collectCapabilityFields(MappingConstructorExpressionNode config, St
             if ("name".equals(fieldName) || fieldName.equals(refField)) {
                 continue;
             }
+            String rawValue = specificField.valueExpr().get().toSourceCode().trim();
```
NipunaRanasinghe: Nit: `rawValue` is computed above the `approvalPolicy` branch but that branch `continue`s without using it. It only needs to move up if something below the `propertyKey == null` guard reads it — which nothing does. Leaving it where it was keeps the read next to its uses.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentPeerBuilder.java
```diff
@@ -219,46 +201,15 @@ public Map<Path, List<TextEdit>> toSource(SourceBuilder sourceBuilder) {
...
-        // dropped a channel a waiting peer had declared, which now hydrates into the form.
-        if (!callbackChannel.isBlank()) {
-            entry.append(", callbackChannel: ").append(WorkflowUtil.quoteIfPlain(callbackChannel));
-        }
-        if (requiresApproval) {
-            entry.append(", requiresApproval: true");
-        }
-        if (!userRoles.isBlank()) {
-            entry.append(", userRoles: ").append(userRoles);
+        if (!allowedEvents.isBlank() && !"[]".equals(allowedEvents)) {
```
NipunaRanasinghe: Nit: `"[]".equals(...)` is an exact-text check, so `[ ]` or `[\n]` from the expression editor gets written through as an empty `allowedEvents`. Stripping the brackets and testing the remainder for blankness covers the spacing the editor can produce.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/MigrateWorkflowDeclarationCodeAction.java
```diff
@@ -0,0 +1,151 @@
...
+        String name = field.fieldName().toSourceCode().trim();
+        return name.startsWith("'") ? name.substring(1) : name;
+    }
+
+    private String valueSource(SpecificFieldNode field) {
+        return field.valueExpr().map(expr -> expr.toSourceCode().trim()).orElse("");
+    }
+
+    // Removes the field with the comma that separated it, so the mapping stays well-formed.
+    private static TextEdit removal(MappingConstructorExpressionNode mapping, SpecificFieldNode field) {
```
NipunaRanasinghe: The separator arithmetic here (three cases: preceded by a separator, first of several, sole field) is the part of this class most likely to produce a malformed mapping, and it does not depend on `WORKFLOW_163` being emitted — it is a pure function of a parsed `MappingConstructorExpressionNode`. A unit test over a synthetic declaration covering all three positions, plus the gated case where `userRoles` precedes `requiresApproval` (the removal edit then abuts the replacement edit), would pin it ahead of the 0.10 pin rather than after it.

Also `valueSource` at line 128 is the only non-static private helper in the class.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/MigrateWorkflowDeclarationCodeAction.java
```diff
@@ -0,0 +1,151 @@
...
+            } else {
+                edits.add(removal(mapping, field));
+                title = String.format(REMOVE_TITLE, name);
+            }
+        } else {
+            edits.add(removal(mapping, field));
+            title = String.format(REMOVE_TITLE, name);
+        }
+        return List.of(CodeActionUtil.createCodeAction(title, edits, context.fileUri(), CodeActionKind.QuickFix));
+    }
```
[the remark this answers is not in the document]
hasithaa: Partly in 70fd6ca2ae: the removal edit is now unit-tested over a parsed mapping (first, middle, sole and keyword-escaped positions). A diagnostic-driven L3 test needs a bala that emits WORKFLOW_163, which is the 0.10 module; it is queued with the pin bump that this PR lists as gated.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Fixed in 70fd6ca2ae: the list branch of `collectDeclaredCapabilities` decomposes `approvalPolicy` the same way the keyed branch does, so a gated activity or tool reopens gated.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Fixed in 70fd6ca2ae: `attempts` is true for any of the four tuning fields, and they are always read, so `{retryDelay: 5.0}` stays AutoRetry with its delay and a retry-then-review without `maxRetries` round-trips.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Fixed in 70fd6ca2ae: the hydration copies `userRoles` and the audience keys under their own names, and the policy's title/description/timeout under the gate's own keys (`approvalTitle`, `approvalDescription`, `approvalTimeout`), so the capability's description is untouched.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
hasithaa: Fixed in 70fd6ca2ae: the gate form carries Review Title, Review Description and Review Timeout, hydrated from and written back to the policy, so a hand-written policy survives an edit-save.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Fixed in 70fd6ca2ae: a `()` roles value hydrates blank on the keyed-capability and approval-policy paths too.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentHumanTaskBuilder.java
hasithaa: Fixed in 70fd6ca2ae: `reviewAudienceFields` is the shared piece; the human-task literal leads with `name` and appends it, and `approvalPolicyLiteral` now passes the review details through `moreFields`, so the parameter is no longer dead.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Fixed in 70fd6ca2ae: `rawValue` is read after the guard, next to its uses.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentPeerBuilder.java
hasithaa: Fixed in 70fd6ca2ae: an `allowedEvents` value that is empty once brackets and whitespace are stripped is treated as no events.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/MigrateWorkflowDeclarationCodeAction.java
hasithaa: Added in 70fd6ca2ae: `MigrateWorkflowDeclarationRemovalTest` applies the removal edit to parsed mappings in all three positions plus a keyword-escaped field, and checks the resulting source. The gated case where `userRoles` precedes `requiresApproval` produces two non-overlapping edits (replace the flag, delete the roles with its own separator); the test covers the deletion side, and the L3 test that exercises both together is gated on the 0.10 pin.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1617,6 +1625,11 @@ private void collectDeclaredCapabilities(ExpressionNode listExpr, String capabil
                         }
                         continue;
                     }
+                    // The gate is a composite too: it decomposes into the flag and the review fields.
+                    if (WorkflowUtil.APPROVAL_POLICY_FIELD.equals(fieldName)) {
```
NipunaRanasinghe, line 1629: The fix is right, but nothing pins it: no `.bal` fixture in the repo contains `approvalPolicy`, so if this branch is dropped again the goldens stay green. The four `node_template` goldens that changed in `70fd6ca2ae` pin the gate form's new properties — Review Title/Description/Timeout — not hydration from source, which is where the bug was.

A source fixture with a gated activity and a gated tool, e.g.

```ballerina
activities: [{activity: chargeCard, approvalPolicy: {userRoles: ["manager"], users: "alice", title: "Approve the charge", timeout: {hours: 4}}}],
tools: [{tool: refund, approvalPolicy: {userRoles: ()}}]
```

would pin the flag coming back on, the audience landing under its own keys, and the details landing under `approvalTitle`/`approvalDescription`/`approvalTimeout` rather than on the capability's own `description` — all three of the things this commit fixed.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -2400,18 +2451,34 @@ static RetryPolicyForm normalizeRetryPolicy(String rawValue) {
...
+                // Three shapes share one union, told apart the way the compiler plugin and the
+                // runtime tell them: an audience makes a review, attempts make retries, both make
+                // retries followed by a review.
                 Map<String, String> fields = WorkflowUtil.parseRecordLiteral(rawValue);
-                if (fields.containsKey(USER_ROLES_FIELD)) {
-                    dropdownValue = ActivityCallBuilder.MANUAL_RETRY_VALUE;
+                boolean audience = fields.containsKey(USER_ROLES_FIELD)
+                        || fields.containsKey(WorkflowUtil.USERS_KEY);
+                // Every tuning field has a default, so any one of them alone still declares attempts.
+                boolean attempts = fields.containsKey(ActivityCallBuilder.MAX_RETRIES_KEY)
```
NipunaRanasinghe, line 2456: Same gap here: every retry fixture in the suite still carries `maxRetries` (`durable_agent_box.bal` has `{maxRetries: 2, retryDelay: 1.5}` on both the capability entry and the `callActivity`), so no test exercises the case that motivated widening this predicate. Narrowing it back to `containsKey(MAX_RETRIES_KEY)` would not turn anything red.

Two cases in `workflow_retry_policy_parsing` would cover it:

- `retryPolicy = {retryDelay: 5.0, retryBackoff: 3.0}` → AutoRetry with both sub-fields populated (before the fix: AutoRetry with everything blank, re-emitted as `{}`).
- `retryPolicy = {retryDelay: 1.5, userRoles: "ops"}` → RetryBeforeReview with the delay kept (before the fix: plain Human Review, delay dropped).

NipunaRanasinghe (dismissed):
Approving. Two follow-ups inline, both about test coverage rather than behaviour.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Added in f2589ee32b: `diagram_generator/source/durable_agent_gate_policy.bal` declares a gated activity (`userRoles`, `users`, `title`, `timeout`, plus the capability's own `description`) and a gated tool (`userRoles`, `excludedUsers`, `description`); its golden pins the flag, each audience field, `approvalTitle`/`approvalDescription`/`approvalTimeout` under their own keys, and the activity's description untouched. Dropping the list-branch hydration turns it red.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Added in f2589ee32b: `reconcileFlow` in `durable_agent_box.bal` gains `retryPolicy = {retryDelay: 5.0, retryBackoff: 3.0}` (golden: AutoRetry with both sub-fields, no maxRetries) and `retryPolicy = {retryDelay: 1.5, userRoles: "ops"}` (golden: Retry, then Review with the delay kept); `workflow_retry_policy_parsing` covers both lines. Narrowing the predicate back to `maxRetries` alone fails both.

NipunaRanasinghe (dismissed):
LGTM

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1509,26 +1509,25 @@ private void populateAgentDeclarationMetadata(NonTerminalNode varNode) {
...
                 case "events" -> collectDeclaredCapabilities(valueExpr, "event", null,
                         Map.of("name", "name", "request", "requestType", "response", "responseType",
                                 "cardinality", "cardinality"), updateEvents);
                 case "humanTasks" -> collectDeclaredCapabilities(valueExpr, "humanTask", null,
-                        Map.of("name", "taskName", "roles", "userRoles", "title", "title",
-                                "description", "description", "resultType", "resultType", "timeout", "timeout"),
+                        Map.of("name", "taskName", "roles", "userRoles", "userRoles", "userRoles",
+                                "users", "users", "excludedUsers", "excludedUsers", "excludedRoles", "excludedRoles",
+                                "title", "title", "description", "description", "resultType", "resultType",
+                                "timeout", "timeout"),
```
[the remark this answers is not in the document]
hasithaa: Addressed in #1117 (stacked on this PR): the list-form `humanTasks` mapping is `Map.ofEntries` and carries `administratorRoles` and `administratorUsers`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -1300,6 +1335,151 @@ public static void addApprovalGateProperties(NodeBuilder nodeBuilder, String app
...
+     */
+    public static List<String> approvalDetailFields(SourceBuilder sourceBuilder) {
+        List<String> fields = new ArrayList<>();
+        String title = trimmedProperty(sourceBuilder, APPROVAL_TITLE_KEY);
+        if (!title.isBlank()) {
+            fields.add("title: " + quoteIfPlain(title));
+        }
+        String description = trimmedProperty(sourceBuilder, APPROVAL_DESCRIPTION_KEY);
+        if (!description.isBlank()) {
+            fields.add("description: " + quoteIfPlain(description));
```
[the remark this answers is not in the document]
hasithaa: Addressed in #1117: `approvalDetailFields` writes an expression-mode title or description through unchanged and quotes only text mode.

hasithaa:
On the latest CodeRabbit round:

- **Peer approval policy** – not applicable: the module's `PeerDecl` record (workflow 0.10.0) has only `agent`, `description` and `allowedEvents`, so a peer cannot carry an approval policy and nothing is dropped on save. The gate helper's doc that still named peer delegation is trimmed in #1117.
- **Retry, then Review with blank attempts** – valid, fixed in #1117 (a04855967c): `RetryBeforeReview` requires `maxRetries`, so the form now writes the AutoRetry default when the field is blank, with a to_source test.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
```diff
@@ -1348,7 +1435,14 @@ public static void populateAdvancedArgs(SourceBuilder sourceBuilder, Map<String,
...
                     Property paramData = Property.convertToProperty(paramProp);
-                    if (paramData.value() != null && !paramData.value().toString().isEmpty()) {
+                    // The open rest never becomes a named argument, and a value that renders to nothing
+                    // (an empty map) must not leave `name = ` behind.
+                    if (ADDITIONAL_VALUES_KEY.equals(paramName) || isIncludedRecordRest(paramData)
+                            || paramData.value() == null || paramData.value().toString().isEmpty()
+                            || paramData.toSourceCode() == null || paramData.toSourceCode().isBlank()) {
+                        continue;
+                    }
+                    {
```
NipunaRanasinghe, line 1445: Leftover braces from flipping the `if` into a `continue` — the block does nothing now and can go.

Separately, the open rest is guarded three times here: it's in `EXCLUDED_CALL_ACTIVITY_PARAMS`, then checked by name, then by kind. If that's deliberate because a property can reach this point without its codedata kind set, a short comment saying so would help; otherwise one check is enough.

NipunaRanasinghe (approved):
Re-approving. 

Three things to sort out, but none of them are blocking. @hasithaa +1 to have those addressed in a future PR if applicable.

**`WORKFLOW_163` still has no test.** The reason given last time was that it needs a bala that emits the diagnostic, and that bala is now pinned. Worth checking whether 0.10.0 actually emits it, and adding the test if it does.

**`callActivity`'s approval gate looks different from the declaration's.** On the activity and tool forms it's a flag plus reviewer roles, the audience fields and review title/description/timeout. On `callActivity` it arrives as a raw record editor over `workflow:ReviewTaskDefinition` under advanced options. Same idea, two quite different forms. Fine for this cut if it's a conscious call, but it's inherited from the signature rather than chosen.

One inline nit on the `additionalValues` fix. That fix looks right, by the way, though it is a second unrelated bug riding in an already large PR.

hasithaa:
On the re-approval's three points, all addressed in #1117:

- **`WORKFLOW_163` test** – 0.10.0 does emit it (checked with `bal build` on the test distribution). Plugin diagnostics do not reach the language-server test harness, so the migration's edit computation is extracted and unit-tested instead: flag plus roles become an `approvalPolicy`, a false flag or a removed peer field is dropped.
- **`callActivity`'s approval gate** – it was inherited from the signature, not chosen. Approval Policy is now one dropdown on the activity call, the agent activity and the tool, built like Retry Policy, with the review fields nested under it, and the advanced section is gone from the activity forms.
- **`additionalValues` nit** – the advanced-parameter path is removed altogether, and with it the leftover braces and the triple guard; the open rest is excluded once, by name.

PR #1110, opened 2026-09-17
[Agent Builder] Add option to enable Agent Manager instrumentation in the deployment sidebar

dan-niles:
## Summary

Resolves https://github.com/wso2/product-integrator/issues/2460

https://github.com/user-attachments/assets/fbbaa9c5-525e-47cd-a488-966fb5864277

- Adds an "Enable Agent Manager instrumentation" checkbox to the deployment sidebar, shown for projects with agents, to publish traces to WSO2 Agent Manager.
- Mutually exclusive with dev-time tracing: switching between them warns first, since only one provider can be active in `trace_enabled.bal`/`Config.toml` at a time.
- Deep-links to Configurable Variables (preselecting `ballerinax/amp`) when `otelEndpoint`/`apiKey` are unset; warns once per project until fixed, not on every Run.
- Fixes: dev-time tracing indicators no longer show "On" when only Agent Manager is active; the local trace server no longer starts on Run/Debug/Try It when Agent Manager is the active provider; the checkbox no longer hides in a Devant-hosted session.

## Summary

- Added Agent Manager instrumentation controls for projects with agents.
- Added provider tracking for `amp` and `idetraceprovider`.
- Added provider-switching warnings and guidance for missing `otelEndpoint` or `apiKey`.
- Added deep-linking to Configurable Variables with `ballerinax/amp` preselected.
- Prevented local tracing from starting or appearing active when Agent Manager tracing is enabled.
- Added provider-aware tracing status updates and initialization handling.
- Kept the Agent Manager option visible in Devant-hosted sessions.


packages/ballerina-core/src/rpc-types/agent-chat/interfaces.ts
```diff
@@ -75,8 +75,11 @@ export interface ExecutionStep {
     hasError?: boolean;
 }
 
+export type TracingProvider = 'idetraceprovider' | 'amp';
+
 export interface TraceStatus {
     enabled: boolean;
+    provider?: TracingProvider;
```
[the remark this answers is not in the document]
dan-niles: Not changing this. The extension and webview ship together in one VSIX, so there's no version-skew scenario between them for this field to break.

packages/ballerina-extension/src/features/tracing/tracer-machine.ts
```diff
@@ -233,7 +261,12 @@ function createTracerMachine(projectPath?: string, childProjectPaths?: string[])
                     on: {
                         // ENABLE while already enabled: write trace_enabled.bal for the new project, stay in enabled.
                         ENABLE: {
-                            actions: [enableTracingInProject],
+                            actions: [
+                                enableTracingInProject,
+                                assign({
+                                    provider: (context, event) => resolveProvider(event),
+                                }),
+                            ],
```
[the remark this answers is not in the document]
dan-niles: Fixed. Switching to amp now stops the local server, unless another known project in the workspace is still on idetraceprovider and needs it.

packages/ballerina-extension/src/features/tracing/utils.ts
```diff
@@ -305,6 +310,75 @@ export async function removeTracingConfig(workspaceDir: string): Promise<void> {
...
+export function isAmpConfigIncomplete(workspaceDir: string): boolean {
+    const configFilePath = path.join(workspaceDir, 'Config.toml');
+    if (!fs.existsSync(configFilePath)) {
+        return true;
+    }
+    try {
+        const content = fs.readFileSync(configFilePath, 'utf-8');
+        const parsedConfig: any = parse(content);
+        const amp = parsedConfig?.ballerinax?.amp;
+        return !amp?.otelEndpoint || !amp?.apiKey;
```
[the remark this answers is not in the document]
dan-niles: Fixed. Now trims otelEndpoint and apiKey before checking, so whitespace-only values count as incomplete.

packages/ballerina-visualizer/src/views/AgentChatPanel/Components/ChatInterface.tsx
```diff
@@ -618,7 +619,7 @@ const ChatInterface: React.FC = () => {
     // Listen for tracing status changes pushed from the extension (toggled from the diagram)
     useEffect(() => {
         rpcClient.getAgentChatRpcClient().onTracingStatusChanged((status) => {
-            setIsTracingEnabled(status.enabled);
+            setIsTracingEnabled(status.enabled && status.provider !== 'amp');
```
kaumini, line 622: **The pushed `tracingStatusChanged` payload never carries `provider`, so this always evaluates to `true` for Agent Manager.**

This handler reads `status.provider` off the pushed notification, but the only producer omits it:

```ts
// packages/ballerina-extension/src/features/tracing/activate.ts:100
sendTracingStatusChangedNotification({ enabled: isEnabled });
```

`TraceStatus.provider` is optional, so this type-checks — but at runtime the webview sees `provider === undefined` and `status.enabled && status.provider !== 'amp'` collapses to `true && true`.

**Repro:** enable Agent Manager tracing on a project with agents. `TracerMachine` transitions `disabled → enabled`, `onUpdate` fires, and this line sets the indicator to On and offers local trace details that will never arrive (traces go to the remote AMP endpoint). The mount-time `checkTracingStatus()` at L594 *is* correct, because `getTracingStatus()` does populate `provider` — so the panel is right until the first push, then wrong until it remounts.

This is the "dev-time tracing indicators no longer show On when only Agent Manager is active" item in the PR description: the pull path was fixed, the push path wasn't.

**Fix at the producer**, so every subscriber benefits — `useTracingStatus` happens to be immune only because it ignores the payload and re-fetches:

```ts
// activate.ts:100
sendTracingStatusChangedNotification({ enabled: isEnabled, provider: TracerMachine.getProvider() });
```

`TracerMachine` is already imported there, and that is the same source `getTracingStatus()` uses for its no-`projectPath` branch, so push and pull agree.

packages/ballerina-extension/src/features/config-generator/configGenerator.ts
```diff
@@ -167,7 +167,20 @@ export async function checkConfigUpdateRequired(
...
+            // Only block Run the first time; cleared below once fixed, so a later regression re-warns.
+            if (!ballerinaExtInstance.context?.globalState.get<boolean>(ampWarningKey, false)) {
+                hasWarnings = true;
+                await ballerinaExtInstance.context?.globalState.update(ampWarningKey, true);
+            }
+        } else {
+            await ballerinaExtInstance.context?.globalState.update(ampWarningKey, undefined);
+        }
+
+        return { hasWarnings, ampConfigMissing };
```
kaumini: **`ampConfigMissing` is returned even when the amp check wasn't what raised the warning.**

It is computed independently of whether the one-shot actually set `hasWarnings`, but `handleOnUnSetValues` uses it to swap *both* the dialog text and the deep-link target (L258-260, L304-306).

**Repro:** amp provider active with blank `otelEndpoint`/`apiKey`, and `ballerina.ampConfigWarningShown:<projectPath>` already `true` because the user hit **Run Anyway** once. They now add `configurable string dbUrl = ?;`. On Run, the loop at L148-168 sets `hasWarnings = true` for `dbUrl`; the amp block skips its inner `if` (flag already set) so it contributes nothing — yet still returns `ampConfigMissing: true`. The modal says the run is blocked on the OTel endpoint and API key, and the config panel opens preselected on `ballerinax/amp`. `dbUrl`, the variable that actually blocked the run, is never surfaced.

Same shape on a *first* run where both are missing: the amp text wins and hides the real configurables.

Suggested: report amp as the cause only when it is the sole reason for prompting.

```suggestion
        // Publishing traces to Agent Manager requires otelEndpoint/apiKey in [ballerinax.amp] of Config.toml
        const ampConfigMissing = getActiveTracingProvider(projectPath) === 'amp' && isAmpConfigIncomplete(projectPath);
        const ampWarningKey = `${AMP_CONFIG_WARNING_SHOWN_PREFIX}${projectPath}`;
        // Only true when the amp check is the *sole* reason we're prompting; otherwise the
        // generic dialog must win, or real unset configurables get hidden behind the amp text.
        let ampIsSoleWarning = false;
        if (ampConfigMissing) {
            // Only block Run the first time; cleared below once fixed, so a later regression re-warns.
            if (!ballerinaExtInstance.context?.globalState.get<boolean>(ampWarningKey, false)) {
                ampIsSoleWarning = !hasWarnings;
                hasWarnings = true;
                await ballerinaExtInstance.context?.globalState.update(ampWarningKey, true);
            }
        } else {
            await ballerinaExtInstance.context?.globalState.update(ampWarningKey, undefined);
        }

        return { hasWarnings, ampConfigMissing: ampIsSoleWarning };
```

packages/ballerina-extension/src/features/tracing/tracer-machine.ts
```diff
@@ -375,7 +425,21 @@ function createTracerMachine(projectPath?: string, childProjectPaths?: string[])
...
                                         }),
                                     ],
-                                }
+                                },
+                                // amp needs no local receiver, unless another project still depends on it.
+                                ENABLE: {
+                                    target: "serverStopping",
+                                    cond: (context, event) =>
+                                        resolveProvider(event) === 'amp'
+                                        && context.provider !== 'amp'
```
kaumini: **This guard reads the machine-wide `context.provider`, so the switch can fail to stop the very server it exists to stop.**

`context.provider` is machine-wide and can already be `'amp'` while the project being switched is still on `idetraceprovider` — `resolveInitialProvider` returns the *first* project with a trace file, scanning `[currentProjectPath, ...childProjectPaths]`.

**Repro** — workspace with children Q (amp) and P (idetraceprovider), opened in workspace view where `StateMachine.context().projectPath` is initially undefined (the case `activate.ts:58` exists to handle):

1. Init: `resolveInitialProvider` skips the undefined `currentProjectPath`, hits Q first → `context.provider = 'amp'`.
2. User runs P → `startServer(P)` reads `'idetraceprovider'` off disk → `START_SERVER` → `enabled.serverStarted`, local OTLP receiver listening.
3. User enables Agent Manager for P. `resolveProvider(event) === 'amp'` ✅, `!hasOtherIdeTracingProjects` ✅ (Q is amp), but `context.provider !== 'amp'` ❌ → guard fails.

I confirmed the fallout against the installed xstate (4.38.3) with a reduction of this exact structure — when the guard fails the event bubbles to the parent `enabled.on.ENABLE` at L241, which only writes the file:

```
context.provider = idetraceprovider  -> state: {"enabled":"serverStopping"} | actions: [ 'CHILD stop-server amp' ]
context.provider = amp (stale)       -> state: {"enabled":"serverStarted"}  | actions: [ 'PARENT enable amp' ]
```

So no project uses `idetraceprovider` any more and the local trace server keeps running — the leak this transition was added to prevent.

Read the target project's provider off disk instead, which is what the rest of the switch logic already does (`enableTracingInProject` at L77 reads `getActiveTracingProvider(event.projectPath)` for exactly this reason):

```suggestion
                                        && getActiveTracingProvider((event as any).projectPath) !== 'amp'
```

(`(event as any).projectPath` matches the file's existing idiom, e.g. L508. If you'd rather not repeat the cast, a `previousProviderOf(event)` helper next to `resolveProvider` would read better and could be reused at L77.)

packages/ballerina-extension/src/features/tracing/tracer-machine.ts
```diff
@@ -531,22 +596,34 @@ export const TracerMachine = {
...
+    // projectPath scopes the amp check per-project; context.provider is machine-wide.
+    startServer: (projectPath?: string) => {
+        // Agent Manager exports traces remotely; the local OTLP receiver has nothing to catch.
+        const context = ensureInitialized().getSnapshot().context as TracerMachineContext;
+        const provider = projectPath ? getActiveTracingProvider(projectPath) : context.provider;
+        if (provider === 'amp') {
+            return;
+        }
         ensureInitialized().send({ type: 'START_SERVER' });
     },
```
kaumini: **The `context.provider` fallback can be stale, which both starts the receiver when it shouldn't and — worse — suppresses it when it should start.**

`context.provider` is never reset on `DISABLE`: the `hasOtherEnabledProjects` branch at L251-254 runs `disableTracingInProject` only, with no `assign`. So it can outlive the project it described.

**Regression vs. the previous `startServer()`** — workspace with Q (amp) and P (idetraceprovider):

1. `context.provider === 'amp'` (from Q).
2. User disables tracing on Q. P still has `trace_enabled.bal`, so the machine stays `enabled` and `context.provider` stays `'amp'`.
3. User debugs P. `handleDebugCommandWithContext` (`features/bi/activator.ts:391`) calls `TracerMachine.startServer(projectPath)` with `projectPath` undefined in workspace view → the ternary falls back to `context.provider` → `'amp'` → early return.

`START_SERVER` is never sent, the receiver never starts, and no traces are captured for a project that is explicitly on `idetraceprovider`. Before this PR that call always started the server.

The check only has authority when it is scoped to a known project, so fail open otherwise (that restores the pre-PR behaviour for the unknown-project case, and keeps the intended skip for the known-amp case):

```suggestion
    // The amp check is per-project. context.provider is machine-wide and is never reset on
    // DISABLE, so it can be stale — fail open and start the receiver when the project is unknown.
    startServer: (projectPath?: string) => {
        // Agent Manager exports traces remotely; the local OTLP receiver has nothing to catch.
        if (projectPath && getActiveTracingProvider(projectPath) === 'amp') {
            return;
        }
        ensureInitialized().send({ type: 'START_SERVER' });
    },
```

Worth noting this still starts the receiver when `projectPath` names a project with *no* `trace_enabled.bal` while some other project keeps the machine `enabled` (also pre-existing). If you want to close that too, gate on "no known project uses `idetraceprovider`" — `hasOtherIdeTracingProjects` is nearly that predicate already.

packages/ballerina-visualizer/src/views/AgentChatPanel/Components/ChatInterface.tsx
dan-niles: Fixed. The push notification now includes provider, matching the initial fetch.

packages/ballerina-extension/src/features/config-generator/configGenerator.ts
dan-niles: Fixed. ampConfigMissing is only true when amp is the sole reason we're prompting, so a different unset configurable no longer gets hidden behind the amp text.

packages/ballerina-extension/src/features/tracing/tracer-machine.ts
dan-niles: Fixed. The guard now reads the switching project's own on-disk provider instead of context.provider, so it no longer misfires when a different project already set that to amp.

packages/ballerina-extension/src/features/tracing/tracer-machine.ts
dan-niles: Fixed. startServer() now fails open (starts the server) when projectPath is unknown, instead of trusting the same stale context.provider — matches pre-PR behavior for that case, while still skipping when a specific project's own provider is confirmed amp.

kaumini (approved):

PR #1111, opened 2026-09-17
Keep long Copilot threads inside the context window, and say so when they aren't

ravindu25:
Fixes wso2/product-integrator#2317

### Problem

After several integrations in one chat thread, Copilot stopped responding — the
turn ended with no text and no error, and every later message in that thread did
the same.

The thread had outgrown the model's context window. The provider reports that as
`model_context_window_exceeded`, which unifies to the same `finishReason: 'length'`
as the per-response output limit. The output limit is recoverable and is resumed
automatically; a full context window is not, so that branch fell through to a
console warning and ended the turn producing nothing. Nothing reached the user.

Two things drove threads into the window in the first place. Every turn's user
message carries a full dump of every file in the workspace, and that message is
replayed verbatim by every later turn, so an N-turn thread re-sends N copies of
the project. And compaction — the mechanism meant to contain this — was switched
off entirely for exactly the large workspaces that needed it, because the fixed
per-turn cost exceeded its trigger.

### Fix

- Report context exhaustion in the chat instead of ending the turn silently, and
  suppress follow-up chips that would only re-send the same oversized thread.
- Replace superseded codebase dumps in replayed history, so a thread carries one
  copy of the project rather than one per turn.
- Raise the compaction trigger above the per-turn floor instead of disabling
  compaction, so large workspaces keep it.
- Warn once per thread at 80% of the window, on every provider — AWS Bedrock has
  no compaction at all and previously got no notice.
- Bound `web_fetch` and whole-file `file_read` results, which were unbounded and
  persisted in context for the rest of the thread.

### Notes

Two of these paths were already partly mitigated after this issue was filed
(server-side compaction, and moving library lookups behind a capped subagent
report). This covers what remained.

Not addressed here: the codebase dump is still re-sent in full on the live turn.
Removing that means moving to a file tree plus on-demand reads, which is a
separate change.

## Summary

- Improve chat stability when threads approach or exceed the model context window.
- Report context pressure and context overflow in the chat and mini chat interfaces.
- Suppress follow-up suggestions after context overflow.
- Prune superseded codebase snapshots and analysis blocks from replayed history.
- Keep compaction enabled for large workspaces when sufficient capacity remains.
- Limit `web_fetch` and whole-file `file_read` content retained in thread context.
- Add tests for compaction policies, history pruning, and context-overflow serialization.


IsuruMaduranga:
@ravindu25 How did we hit this issue? Sonnet has 1M context and we compact way before hitting that ( 500K to be exact )

ravindu25:
closing this PR as it needs to be refactored

PR #1112, opened 2026-09-17
Improve Node Positioning in the Integration Overview Diagram

madushajg:
## Purpose
$title
Addresses: https://github.com/wso2/product-integrator/issues/2465

Screenshot 2026-09-17 at 11 28 44

## Summary

- Improve node positioning in the Integration Overview Diagram.
- Center listeners, workflows, and connection nodes on link anchors.
- Add collision handling and refinement passes for consistent node distribution.
- Align node sizing and port-anchor calculations with rendered row layouts.
- Standardize port rendering and remove unnecessary margin offsets.
- Centralize entry-node and workflow play-button sizing constants.
- Flatten near-level links to prevent unwanted connector bends.
- Add regression and layout tests for positioning, sizing, and link rendering.


packages/component-diagram/src/components/NodeLink/NodeLinkModel.ts
```diff
@@ -105,8 +112,24 @@ function buildBezierSegment(p0: Point2D, p1: Point2D): BezierSegment {
...
+ * rather than drawing it through their own slightly different Y's - see STRAIGHT_TOLERANCE. A
+ * multi-point detour's bend points are deliberately curved regardless of how close together they
+ * land, so this only ever applies to the 2-point case (identity otherwise).
+ */
+function flattenIfNearLevel(points: Point2D[]): Point2D[] {
+    if (points.length !== 2 || Math.abs(points[0].y - points[1].y) >= STRAIGHT_TOLERANCE) {
+        return points;
+    }
+    const flatY = (points[0].y + points[1].y) / 2;
+    return [{ x: points[0].x, y: flatY }, { x: points[1].x, y: flatY }];
```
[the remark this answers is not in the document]
madushajg: This is intentional, not a bug. Flattening both endpoints to the shared average `Y` is what removes the faint bow (added in a previous commit, with a regression test covering exactly this behavior). The max shift is `4px`, well within the `10px` port dot's radius, so there's no visible gap. Leaving as-is.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -190,6 +174,182 @@ export function autoDistribute(engine: DiagramEngine) {
...
+    });
+
+    const result = new Map<T, number>();
+    blocks.forEach((block) => {
+        const blockCenter = block.sumDesired / block.count;
+        let cursor = blockCenter - block.totalHeight / 2;
+        block.members.forEach(({ item, height }) => {
+            result.set(item, cursor + height / 2);
+            cursor += height + gap;
+        });
```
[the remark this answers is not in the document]
madushajg: The block is already centered correctly. The packed span always runs from `blockCenter - totalHeight/2` to `blockCenter + totalHeight/2` by construction, so it's symmetric around `blockCenter` regardless of unequal member heights. Shifting by the difference from the arithmetic mean of centers would actually miscenter it, since that mean isn't the right centering criterion when heights differ. No change needed.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -203,11 +363,30 @@ export const LINK_DETOUR_MARGIN = 16;
...
+ * These are real rendered pixel measurements, not a "content + padding" guess. `Box` and
+ * `StyledServiceBox` (see styles.ts) are declared `box-sizing: border-box`, so a row's declared
+ * 40px height already includes its own border - it does not add on top of it. Confirmed against
+ * the real production webview DOM (not jsdom, which never lays out real pixel sizes).
  */
-const ROW_PADDING = 8;
-const ENTRY_HEADER_HEIGHT = 64 + ROW_PADDING;
-const ENTRY_ROW_HEIGHT = 40 + ROW_PADDING;
-const ENTRY_VIEW_ALL_BUTTON_HEIGHT = 40;
+const ROW_PADDING = 8; // gap between stacked rows - `Box`'s own `gap: 8px`, already exact
```
samithkavishke: **LOW** — `ROW_PADDING`/`BOX_INSET`/`ENTRY_HEADER_CONTENT_HEIGHT`/`ENTRY_ROW_CONTENT_HEIGHT` (lines 372–383) are hand-derived magic numbers manually kept in sync with the CSS declared in `styles.ts` (`Box`'s `gap: 8px`, `StyledServiceBox`'s `height: 40px`, etc.), with no shared source of truth.

The comment above even says these were "confirmed against the real production webview DOM" by manual inspection — exactly the kind of drift that caused the bug this PR fixes. A future edit to `StyledServiceBox`'s height or `Box`'s padding/border in `styles.ts` will silently reintroduce the same port-anchor misalignment with no compiler or test signal until someone eyeballs the rendered diagram again.

Suggestion: consider importing these pixel values from a single shared constants module referenced by both `styles.ts` and `diagram.ts` (or adding a lightweight snapshot/lint check that fails if they diverge), so the two can't drift apart unnoticed.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -190,6 +174,182 @@ export function autoDistribute(engine: DiagramEngine) {
...
+ * has two incoming links that don't agree on what "this node's center" should even mean unless
+ * each is first translated through its own arrival row's offset - averaging their raw source
+ * anchors without that would silently target neither row correctly.
+ *
+ * Nodes with nothing pointing at them keep their existing center exactly. Siblings that end up
+ * wanting the same (or too-close) center - e.g. two connections both fed solely by the same
+ * automation node - settle symmetrically around that shared center rather than one keeping it
+ * outright and the other being shoved aside; see `resolveMinGapPositions`.
+ */
+function positionColumnByIncomingLinks(nodes: NodeModel[], links: NodeLinkModel[], x: number, defaultHeight: number) {
```
samithkavishke: **LOW** — `positionColumnByIncomingLinks` (here) and `refineNodesByAllLinks` (line 315) duplicate ~25 lines of `desiredCenter`/`rowOffset` computation logic, differing only in which links are selected (incoming-only vs. both directions) and the sort key (by desired center vs. existing Y).

A future fix to the row-offset math (e.g. another GraphQL/AI-service port edge case) has to be applied in both functions or they silently diverge.

Suggestion: factor the shared `desiredCenter` computation into a helper, e.g. `computeDesiredCenter(node, relevantLinks): number`, parametrized by the link-selection/direction logic, and have both functions call it before their own sort.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -134,36 +126,28 @@ export function autoDistribute(engine: DiagramEngine) {
...
+    const links = model.getLinks().filter((linkModel): linkModel is NodeLinkModel => linkModel instanceof NodeLinkModel);
 
-    // Position connection nodes
-    connectionNodes.forEach((node, index) => {
-        const connectionNode = node as ConnectionNodeModel;
-        connectionNode.setPosition(connectionX, node.getY());
-    });
+    // Round 1: position everything downstream of entry nodes from their current (creation-order)
+    // Y, exactly as before this round existed.
+    connectedListeners.forEach((listenerNode) => positionConnectedListener(listenerNode, entryNodes as NodeModel[], listenerX));
```
samithkavishke, line 159: **LOW** — `autoDistribute`'s two rounds re-filter `links` from scratch per node inside `positionConnectedListener`, `positionColumnByIncomingLinks` (called here and at line 137/150), and `refineNodesByAllLinks` (line 147/148) — 5 separate `O(nodes*links)` scans total across the full pass.

Negligible at typical diagram sizes, but since `autoDistribute` re-runs on every model edit, for a large integration diagram this is real repeated work.

Suggestion: bucket `links` once into a `Map<NodeModel, NodeLinkModel[]>` (or two maps, by source/target) at the top of `autoDistribute` and pass the pre-filtered lists down, instead of re-scanning the full link array in each helper call.

samithkavishke (dismissed):
LGTM, Only LOW priority ones found

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -190,6 +174,182 @@ export function autoDistribute(engine: DiagramEngine) {
...
+        if (incomingLinks.length === 0) {
+            return { item: node, desiredCenter: node.getY() + (node.height || defaultHeight) / 2, height: node.height || defaultHeight };
+        }
+        const nodeBox = getNodeBoundingBox(node);
+        const currentCenter = (nodeBox.top + nodeBox.bottom) / 2;
+        const desiredCenter =
+            incomingLinks.reduce((sum, link) => {
+                // How far this link's own arrival row sits from the node's current center - 0 for
+                // the generic in port, some real row offset for a specific event/function port.
+                const rowOffset = getPortAnchorY(node, link.getTargetPort(), nodeBox) - currentCenter;
```
rosensilva: Will this calculate workflow links align to the card’s center, but the input port is in the header? Can we calculate the input anchor from the header’s play-button position?

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -190,6 +174,182 @@ export function autoDistribute(engine: DiagramEngine) {
...
+            prev.count += curr.count;
+            prev.totalHeight += curr.totalHeight + gap;
+            prev.members.push(...curr.members);
+            blocks.pop();
+        }
+    });
+
+    const result = new Map<T, number>();
+    blocks.forEach((block) => {
+        const blockCenter = block.sumDesired / block.count;
```
rosensilva, line 350: Will different-height nodes shift away from their intended average position during collision handling? Can we include each node’s height and spacing offsets when calculating the group’s center?

packages/component-diagram/src/utils/diagram.ts
madushajg: Good catch. The input port is actually inside the workflow's play button, not centered on the card. Fixed `getPortAnchorY` to anchor a workflow node's "in" port to the play button's real position (new `WORKFLOW_PLAY_BUTTON_TOP`/`SIZE` constants shared with the button's CSS) instead of the box center, and updated the positioning tests that had baked in the old center assumption.

packages/component-diagram/src/utils/diagram.ts
madushajg: Fixed. Blocks were centered at `blockCenter - totalHeight/2`, which only equals the group's real average position when all members share the same height; with mixed heights it drifts toward the taller node. Fixed by anchoring on the block's actual packed-offset mean instead. Snapshot tests (which include mixed-height fixtures) updated accordingly.

packages/component-diagram/src/utils/diagram.ts
madushajg: Fair point. Extracted the shared `desiredCenter`/`row-offset` logic into `computeDesiredCenterItem(node, relevantLinks, defaultHeight)`, and both `positionColumnByIncomingLinks` and `refineNodesByAllLinks` now just build their own link-selection list and call it before their respective sort/order step.

packages/component-diagram/src/utils/diagram.ts
madushajg: Bucketed links once into `incomingLinksByNode`/`outgoingLinksByNode` maps in `autoDistribute`; both helpers now do `O(1)` lookups instead of re-scanning per node. 

Note: `positionConnectedListener` doesn't actually scan links (it matches via `entryNodes`), so left that alone as out of scope.

packages/component-diagram/src/utils/diagram.ts
madushajg: Added `ENTRY_ROW_GAP`/`ENTRY_ROW_CONTENT_HEIGHT` to `resources/constants.ts`and pointed both `styles.ts`'s CSS and `diagram.ts`'s row math at them instead of separate hardcoded 8/40s.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -190,6 +197,207 @@ export function autoDistribute(engine: DiagramEngine) {
...
+ */
+function resolveMinGapPositions<T>(items: Array<{ item: T; desiredCenter: number; height: number }>, gap: number): Map<T, number> {
+    interface Block { sumDesired: number; count: number; members: Array<{ item: T; height: number }>; totalHeight: number; }
+    const blocks: Block[] = [];
+    items.forEach(({ item, desiredCenter, height }) => {
+        blocks.push({ sumDesired: desiredCenter, count: 1, members: [{ item, height }], totalHeight: height });
+        while (blocks.length >= 2) {
+            const curr = blocks[blocks.length - 1];
+            const prev = blocks[blocks.length - 2];
+            const minDistance = prev.totalHeight / 2 + gap + curr.totalHeight / 2;
```
NipunaRanasinghe: This can place nodes closer together than `gap`, and I was able to reproduce a real overlap.

The packing code below already explains why `totalHeight / 2` is the wrong half-extent for a block whose members have different heights, and correctly anchors on the mean member center instead. But this merge test still uses `totalHeight / 2`, so the extent assumed here and the extent actually produced below disagree.

The real bottom edge of a packed block is `center + totalHeight - mean(memberCenterOffsets)`. When the taller member ends up last in a block, that sits below what this line assumes, and the next block gets admitted too close.

I ran the function as written with realistic entry-node heights (a header-only service ~67px, an 11-row service ~595px) and `gap = NODE_GAP_Y / 2 = 50`, scanning desired centers: worst case was **82px of overlap**. A smaller repro (heights 40/200/40, gap 25) gives 12.5px clearance where 25 was required.

This matters more now than before, because round 2 runs this pass over `entryNodes`, which is where heights vary the most and whose Y was previously never recomputed.

Suggested fix: track the block's real top/bottom extents instead of deriving them from `totalHeight`:

```ts
// topExtent = mean(memberCenterOffsets), bottomExtent = totalHeight - topExtent
const minDistance = prev.bottomExtent + gap + curr.topExtent;
```

recomputed on each merge. A mixed-height regression test would be good too — the current tests only cover equal-height siblings, which is exactly the case that works.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -273,6 +500,13 @@ export function getPortAnchorY(node: NodeModel, port: PortModel | null | undefin
         return center;
     }
 
+    if (node.type === "workflow" && port === node.getInPort()) {
+        // Unlike other entry nodes, a workflow's "in" port renders inside its play button
+        // (PlayButtonCircle), pinned near the header rather than centered on the box - so it
+        // needs its own anchor instead of falling into the generic center case below.
+        return box.top + WORKFLOW_PLAY_BUTTON_TOP + WORKFLOW_PLAY_BUTTON_SIZE / 2;
```
NipunaRanasinghe: This lands at `top + 36`, but the real DOM position is `top + 39`:

- `PlayButtonCircle` is `position: absolute` inside `Box` (`position: relative`), and `top` resolves against Box's **padding** edge, so add Box's 1.5px border.
- `PlayButtonPortWrapper`'s `top: 50%` resolves against `PlayButtonCircle`'s padding box, which is another 1.5px inside its own border — `PlayButtonCircle` has no `box-sizing: border-box`, so its `28px` is content height and it renders 31px tall.

`1.5 + 22 + 1.5 + 14 = 39`.

It's only 3px, but it's the same slope this PR is fixing everywhere else, and right now it's invisible because `STRAIGHT_TOLERANCE = 8` absorbs it. Either add the two border terms here, or give `PlayButtonCircle` `box-sizing: border-box` and account for Box's border.

packages/component-diagram/src/components/NodeLink/NodeLinkModel.ts
```diff
@@ -45,6 +45,13 @@ const LINK_CURVATURE = 0.5;
  */
 const LINK_MIN_CURVE_OFFSET = 20;
 
+/**
+ * Below this vertical gap between a plain link's two ports, the curve's natural bow (however
+ * gentle) reads as a faint, pointless wobble rather than a real S-curve - especially over a long
+ * horizontal span, where even a few px of Y difference stays visible the whole way across.
+ */
+const STRAIGHT_TOLERANCE = 8;
```
NipunaRanasinghe: 8px is large enough to hide real anchor bugs rather than just float noise. Flattening moves each endpoint up to 4px off its actual port, in both the rendered path and the `avoidLinkObstructions` sampling. Two effects worth deciding on deliberately:

- the line visibly detaches from the port dot near the tolerance boundary;
- any anchor math that drifts by less than 8px becomes undetectable — including the workflow play-button offset I flagged in `diagram.ts`, which this is currently masking.

I'd suggest 1–2px, plus a test that anchors land exactly, so the tolerance only absorbs rounding. If 8 is deliberate, it'd help to say here why it needs to be that wide.

packages/component-diagram/src/components/NodeLink/NodeLinkModel.ts
```diff
@@ -105,8 +112,24 @@ function buildBezierSegment(p0: Point2D, p1: Point2D): BezierSegment {
...
  * numeric checks of all three claims.
  */
+/**
+ * A plain port-to-port link (no avoidLinkObstructions detour) whose two ends are already
+ * near-level: flattens both ends to their shared average Y so the curve skips the bow entirely,
+ * rather than drawing it through their own slightly different Y's - see STRAIGHT_TOLERANCE. A
+ * multi-point detour's bend points are deliberately curved regardless of how close together they
+ * land, so this only ever applies to the 2-point case (identity otherwise).
+ */
+function flattenIfNearLevel(points: Point2D[]): Point2D[] {
```
NipunaRanasinghe: The long `getBezierSegments` doc block is now stranded above this function — two JSDoc blocks stack here, and the first one ends up documenting `flattenIfNearLevel`. Please move it back down onto `getBezierSegments`.

packages/component-diagram/src/components/NodeLink/NodeLinkModel.ts
```diff
@@ -115,10 +138,15 @@ function getBezierSegments(points: Point2D[]): BezierSegment[] {
  * detour `avoidLinkObstructions` (utils/diagram.ts) routed around an obstruction.
  */
 export function buildBezierPath(points: Point2D[]): string {
-    const commands = getBezierSegments(points).map(
+    const segments = getBezierSegments(points);
+    // Read the start back off the first segment - not the raw `points[0]` - so the "M" always
+    // matches whatever getBezierSegments actually drew from (its own near-level flattening
+    // included), rather than the two silently disagreeing on where the path begins.
+    const start = segments[0].start;
```
NipunaRanasinghe: Minor: `segments[0]` is unguarded here and on line 178. A 1-point array used to produce `M x y`; now it throws. Not reachable from `getCurvePath` today since react-diagrams links always have two points, but these two functions are exported and used to be total.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -134,36 +130,47 @@ export function autoDistribute(engine: DiagramEngine) {
...
-        const connectionNode = node as ConnectionNodeModel;
-        connectionNode.setPosition(connectionX, node.getY());
-    });
+    // Round 1: position everything downstream of entry nodes from their current (creation-order)
+    // Y, exactly as before this round existed.
+    connectedListeners.forEach((listenerNode) => positionConnectedListener(listenerNode, entryNodes as NodeModel[], listenerX));
+    positionColumnByIncomingLinks(workflowNodes as NodeModel[], incomingLinksByNode, workflowX, ENTRY_NODE_HEIGHT);
+    positionColumnByIncomingLinks(connectionNodes as NodeModel[], incomingLinksByNode, connectionX, CON_NODE_HEIGHT);
+
+    // Round 2: entry and workflow nodes sit *between* two neighbors (a listener/entry node on one
```
NipunaRanasinghe, line 163: "their sequence still reflects the source file, not link geometry" isn't true for workflow nodes — `positionColumnByIncomingLinks` already re-sorted them by `desiredCenter` in round 1, so by the time round 2 runs, only `entryNodes` still hold source-file order.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -190,6 +197,207 @@ export function autoDistribute(engine: DiagramEngine) {
...
+): void {
+    const items = nodes.map((node) => {
+        const neighborLinks = [...(incomingLinksByNode.get(node) ?? []), ...(outgoingLinksByNode.get(node) ?? [])];
+        return computeDesiredCenterItem(node, neighborLinks, ENTRY_NODE_HEIGHT);
+    });
+
+    // Kept in the nodes' EXISTING vertical order (not re-sorted by desiredCenter), so a node
+    // never leapfrogs another - only its own Y shifts, same as autoDistribute already keeps
+    // entry/workflow nodes in source-file order today.
+    items.sort((a, b) => a.item.getY() - b.item.getY());
```
NipunaRanasinghe: This sorts by `getY()` (the node's top edge) while everything else in this path works in center space. With mixed heights, top order and center order aren't the same, so the order handed to `resolveMinGapPositions` can differ from the visual order it's meant to preserve. Sorting by `getY() + height / 2` would be consistent.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -190,6 +197,207 @@ export function autoDistribute(engine: DiagramEngine) {
...
+ * touches keeps its own current center.
+ */
+function computeDesiredCenterItem(
+    node: NodeModel,
+    relevantLinks: NodeLinkModel[],
+    defaultHeight: number
+): { item: NodeModel; desiredCenter: number; height: number } {
+    const height = node.height || defaultHeight;
+    if (relevantLinks.length === 0) {
+        return { item: node, desiredCenter: node.getY() + height / 2, height };
```
NipunaRanasinghe: This branch uses `node.getY() + height / 2` as the current center, while the branch below uses `getNodeBoundingBox`'s center. They agree today only because both fall back to the same default height. Worth using the box in both so they can't diverge.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -203,11 +411,30 @@ export const LINK_DETOUR_MARGIN = 16;
...
-const ENTRY_VIEW_ALL_BUTTON_HEIGHT = 40;
+const ROW_PADDING = ENTRY_ROW_GAP; // gap between stacked rows - `Box`'s own `gap`
+// `Box`'s own border + padding: the inset between a node's outer edge and its first/last row,
+// counted once at the top and once at the bottom. Border-box sizing doesn't apply here since Box
+// has no explicit width/height of its own (it's sized by its content), so this is added on top
+// regardless.
+const BOX_INSET = NODE_BORDER_WIDTH + NODE_PADDING; // 1.5 + 8 = 9.5
+// ServiceBox's real rendered height - matches its own CSS declaration exactly (no border of its
+// own either way).
+const ENTRY_HEADER_CONTENT_HEIGHT = ENTRY_NODE_HEIGHT - NODE_PADDING; // 56
```
NipunaRanasinghe: This repeats `ServiceBox`'s own `height: ${ENTRY_NODE_HEIGHT - NODE_PADDING}px` in a second file — the exact kind of duplication the rest of the PR removes. Better to export it from `resources/constants` next to `ENTRY_ROW_CONTENT_HEIGHT` and have `styles.ts` use it too.

packages/component-diagram/src/components/nodes/EntryNode/components/styles.ts
```diff
@@ -44,13 +54,13 @@ export const Header = styled.div<NodeStyleProp>`
...
 `;
 
-export const TopPortWidget = styled(PortWidget)`
-    margin-top: -3px;
-`;
+// PortWidget itself renders a bare, zero-height div, so its reported link-anchor position is
+// exactly wherever the flex row centers it - no margin nudge here, or "in"/"out" would sit off
+// that center by different amounts (see getPortAnchorY, which assumes dead center for both) and
+// every link's straight leg would render with a small, otherwise-unexplained slope.
+export const TopPortWidget = styled(PortWidget)``;
```
NipunaRanasinghe: An empty `styled(PortWidget)` with no rules is just `PortWidget` with extra indirection — same in `ListenerNodeWidget.tsx` and `ConnectionNodeWidget.tsx`. I'd use `PortWidget` directly and move the comment to where it's rendered. That would also cut a good chunk of the ~1.5k-line snapshot diff, most of which is emotion re-numbering its generated classes rather than any real change.

packages/component-diagram/src/components/nodes/EntryNode/components/styles.ts
```diff
@@ -106,28 +116,30 @@ export const Description = styled(StyledText)`
 `;
 
 export const Box = styled.div<NodeStyleProp>`
+    box-sizing: border-box;
```
NipunaRanasinghe, line 110: Worth calling out in the PR description: this changes width too, not just height. `StyledServiceBox` previously rendered 267×43 (240 width + 24 padding + 3 border) and overflowed `Box`; it now renders 240×40 and fits. That's an improvement, but every entry node's rows change size visually, which the summary doesn't mention.

NipunaRanasinghe (changes requested):
One thing should block: `resolveMinGapPositions` can place nodes closer than `gap` when a block's members have different heights. I reproduced 82px of overlap with realistic entry-node heights — details inline. It's more exposed now that round 2 runs this pass over the entry column.

Two smaller things go against the PR's own goal: the workflow play-button anchor is 3px off, and `STRAIGHT_TOLERANCE = 8` is wide enough to hide that (and any future drift of the same size). The rest of the comments are cleanup.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -190,6 +197,207 @@ export function autoDistribute(engine: DiagramEngine) {
...
+        let offset = 0;
+        let sumOffset = 0;
+        block.members.forEach(({ height }) => {
+            sumOffset += offset + height / 2;
+            offset += height + gap;
+        });
+        let cursor = blockCenter - sumOffset / block.members.length;
+        block.members.forEach(({ item, height }) => {
+            result.set(item, cursor + height / 2);
+            cursor += height + gap;
```
rosensilva: New regression: nodes can overlap. Spacing is checked before the new centering adjustment moves each group. That movement can push it into the next group. I reproduced roughly 58px of overlap in the positioning pass. Use the adjusted group boundaries when checking collisions, or repeat collision resolution after recentering. Test three or more unequal-height nodes. Can you please check this?

packages/component-diagram/src/components/NodeLink/NodeLinkModel.ts
madushajg: Agreed, lowered to `2px` with a comment on why. Updated the existing test's fixture and added a boundary test for the `≥ tolerance` case.

packages/component-diagram/src/utils/diagram.ts
madushajg: Confirmed via box-model trace. Real offset is `+39` not `+36`. Fixed the arithmetic (`+2×NODE_BORDER_WIDTH`), updated affected test numbers.

packages/component-diagram/src/utils/diagram.ts
madushajg: Good Catch. Merge test still assumed symmetric extents. Fixed by tracking real packed offsets incrementally; added a regression test with your exact repro (67/595/67, gap=50).

packages/component-diagram/src/components/NodeLink/NodeLinkModel.ts
madushajg: Fixed. Moved `getBezierSegments`'s doc back above it.

packages/component-diagram/src/components/NodeLink/NodeLinkModel.ts
madushajg: Fixed. Both functions now fall back to `points[0]` for <2-point input.

packages/component-diagram/src/utils/diagram.ts
madushajg: Fixed. corrected; only entry nodes keep source order into round 2, workflows don't.

packages/component-diagram/src/utils/diagram.ts
madushajg: Fixed. Now sorts by `getY() + height/2`.

packages/component-diagram/src/utils/diagram.ts
madushajg: Fixed. Both branches now derive it from `getNodeBoundingBox`.

packages/component-diagram/src/utils/diagram.ts
madushajg: Fixed. Extracted as a shared constant, same as `ENTRY_ROW_CONTENT_HEIGHT`.

packages/component-diagram/src/components/nodes/EntryNode/components/styles.ts
madushajg: Fixed. Removed all 6 wrappers, use `PortWidget` directly, moved the comment to each render site.

packages/component-diagram/src/components/NodeLink/NodeLinkModel.ts
```diff
@@ -115,10 +142,18 @@ function getBezierSegments(points: Point2D[]): BezierSegment[] {
...
  */
 export function buildBezierPath(points: Point2D[]): string {
-    const commands = getBezierSegments(points).map(
+    const segments = getBezierSegments(points);
+    // Read the start back off the first segment - not the raw `points[0]` - so the "M" always
+    // matches whatever getBezierSegments actually drew from (its own near-level flattening
+    // included), rather than the two silently disagreeing on where the path begins. `segments` is
+    // only ever empty for a <2-point `points` (getBezierSegments/flattenIfNearLevel never drop
+    // points otherwise), in which case there's nothing to have flattened - falling back to
+    // `points[0]` keeps this total rather than throwing on that input.
```
dulajdilshan: **[Min] Empty path is still not total.** The new `<2-point` fallback only handles one point: `[]` still makes `start` undefined before `start.x` is read (and `sampleBezierPath` returns `[undefined]`). Could we handle the empty case explicitly, or narrow the documented contract to a one-point path?

dulajdilshan (dismissed):
Nothing blocking. finding (1 Min)

### Done well
- The follow-up narrows near-level flattening and adds a boundary regression test.
- Unequal-height collision packing is covered with a focused regression case.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -190,6 +200,222 @@ export function autoDistribute(engine: DiagramEngine) {
...
+    if (serviceNodes.length === 0) {
+        return;
+    }
+    const avgCenterY =
+        serviceNodes.reduce((sum, n) => {
+            const entryNode = n as EntryNodeModel;
+            return sum + getPortAnchorY(entryNode, entryNode.getInPort());
+        }, 0) / serviceNodes.length;
+    const listenerHeight = listenerNode.height || LISTENER_NODE_HEIGHT;
+    listenerNode.setPosition(listenerX, avgCenterY - listenerHeight / 2);
```
NipunaRanasinghe, line 379: Non-blocking, and pre-existing — not something this PR introduced.

Listeners are the one column that never goes through `resolveMinGapPositions`. Each connected listener is positioned independently, so two listeners attached to the same service land on exactly the same center. I measured a full 64px overlap (both at `y = 568`), and I get the same result on `staging/5.14.0`, so it's not a regression.

Just flagging it because it's the same class of problem this PR fixes for workflows, connections and entry nodes, and the machinery is already here — running the connected listeners through `resolveMinGapPositions` after `positionConnectedListener` would close it. Fine as a follow-up.

NipunaRanasinghe (dismissed):
Approving. Everything from the last round is addressed, and the two substantive ones are properly fixed rather than worked around.

One non-blocking note inline about listener collisions, which is pre-existing rather than anything this PR changed.

packages/component-diagram/src/utils/diagram.ts
madushajg: Already fixed. The merge/admission test now uses each block's real packed extents instead of assuming `totalHeight/2`. Sanity-checked separately (a standalone script outside the test suite, not a Jest test) by hand-copying the algorithm and running 200k random cases (3-6 nodes, mixed heights/gaps): no overlap beyond float noise. The actual Jest regression test using your fixture numbers passes. No further code change made here.

packages/component-diagram/src/utils/diagram.ts
```diff
@@ -190,6 +197,207 @@ export function autoDistribute(engine: DiagramEngine) {
...
+            const minDistance = prev.totalHeight / 2 + gap + curr.totalHeight / 2;
+            if (curr.sumDesired / curr.count - prev.sumDesired / prev.count >= minDistance) {
+                break;
+            }
+            prev.sumDesired += curr.sumDesired;
+            prev.count += curr.count;
+            prev.totalHeight += curr.totalHeight + gap;
+            prev.members.push(...curr.members);
+            blocks.pop();
+        }
```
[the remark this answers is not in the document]
madushajg: Already fixed. Same root cause. Verified with the exact numbers (64@0, 359@0, 64@320, gap=50): old code overlaps by 72px, current code by 0.

packages/component-diagram/src/components/NodeLink/NodeLinkModel.ts
madushajg: Valid. The `?? points[0]` fallback only restores totality for the 1-point case; `points = []` still let start be `undefined`. Added explicit `points.length === 0` guards to both functions `("" / [])`, and a test locking in both the 0-point and 1-point cases.

NipunaRanasinghe (approved):

PR #1113, opened 2026-09-17
Fix issues with parameter chip editing

madushajg:
## Purpose
Addresses: https://github.com/wso2/product-integrator/issues/2208

https://github.com/user-attachments/assets/0b7e6dac-ad0c-4e4e-b1e7-1fd32ad2e16d

## Summary

- Improved parameter-chip editing in the expression editor.
- Added character-by-character Backspace behavior.
- Preserved spaces and sibling placeholder chips during edits.
- Added Enter, blur, boundary-click, and active-chip selection handling.
- Added Tab and Shift-Tab navigation between editable chips.
- Added end-to-end coverage with a dedicated Ballerina test fixture.


madushajg:
@coderabbitai review

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
```diff
@@ -59,6 +60,34 @@ export const ProgrammerticSelectionChange = Annotation.define<boolean>();
...
+export const activeEditableTokenField = StateField.define<number | undefined>({
+    create: () => undefined,
+    update(value, tr) {
+        let mapped = value === undefined ? undefined : tr.changes.mapPos(value, 1);
+        for (const effect of tr.effects) {
+            if (effect.is(setActiveEditableTokenEffect)) {
+                mapped = effect.value;
+            }
+        }
+        return mapped;
```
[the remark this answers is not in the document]
madushajg: Fixed with b7717e62fcdd60de23ac3cc0f9b985301c460382

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
```diff
@@ -229,21 +267,40 @@ export const tokenField = StateField.define<TokenFieldState>({
...
+        // uses assoc=-1 for the mirror-image reason: it keeps typing at the very start of a
+        // chip (e.g. after pressing Home) inside the tracked range instead of excluding it.
+        // Both boundaries must stay in sync with activeEditableTokenField's own mapping above
+        // so the "is this the active chip" comparisons in buildDecorations keep matching.
         let tokens = oldState.tokens.map(token => ({
             ...token,
-            start: tr.changes.mapPos(token.start, 1),
-            end: tr.changes.mapPos(token.end, -1)
+            start: tr.changes.mapPos(token.start, -1),
+            end: tr.changes.mapPos(token.end, 1)
```
samithkavishke: **[HIGH] Regression — boundary typing can silently corrupt non-editable chips**

The `assoc` flip (start: `-1`, end: `1`) is applied to *every* token/compound in `tokenField.update`, not just the one in `activeEditableTokenField`. For a non-editable chip (e.g. a `VARIABLE` reference), this makes typing at its start/end boundary get absorbed into its tracked `{start, end}` range instead of landing as separate plain text — `buildDecorations` then renders the typed character as part of the chip's label via `createChip(docContent.slice(token.start, token.end), ...)`. A following Backspace matches the token and deletes the whole merged span in one keystroke (`removeChipEffect`), destroying the original chip content along with the new text.

Suggested fix: only apply the wider assoc to the token/compound whose `start` equals `activeStart` (mirroring `activeEditableTokenField`'s own mapping); keep `assoc=1`/`assoc=-1` (start/end) for every other token so inactive chips' boundaries don't grow.

```suggestion
            start: tr.changes.mapPos(token.start, token.start === activeStartBeforeChange ? -1 : 1),
            end: tr.changes.mapPos(token.end, token.start === activeStartBeforeChange ? 1 : -1)
```
(`activeStartBeforeChange` = `tr.startState.field(activeEditableTokenField, false)`, computed once above and applied to both the `tokens` and `compounds` maps.)

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/components/ChipExpressionEditor.tsx
```diff
@@ -24,9 +24,12 @@ import {
...
     buildOnChangeListner,
     chipPlugin,
+    chipBoundaryClickHandler,
+    activeChipSelectionGuard,
     chipTheme,
     completionTheme,
     tokenField,
+    activeEditableTokenField,
     tokensChangeEffect,
     expressionEditorKeymap,
```
samithkavishke, line 35: **[MEDIUM] Regression — Enter is swallowed by list continuation before a chip can commit**

This new `expressionEditorKeymap` Enter binding (commits the active chip) is registered *after* `listContinuationKeymap` in the `keymap.of([...])` array (lines 307-308 below):

```ts
...(props.enableListContinuation ? listContinuationKeymap : []),
...expressionEditorKeymap
```

CodeMirror keymaps run in array order, so when `enableListContinuation` is true (the only current caller is `PromptMode`), a list-pattern line's Enter handler (`handleEnterForListContinuation`) runs first and consumes the keystroke before the chip-commit binding ever sees it.

Repro: in a Prompt, author a bullet line with an interpolated call, e.g. `- Greet the user: ${getUserName($1)}`. Click into the `$1` chip, type a value, press Enter to commit — because the line matches `LIST_PATTERNS` with non-empty content, the list handler splits/continues the list instead, leaving the chip stuck in edit mode or corrupting the document with an unwanted line break.

Suggested fix — swap the order so the active-chip commit takes priority:

```ts
...expressionEditorKeymap,
...(props.enableListContinuation ? listContinuationKeymap : [])
```

(or have the Enter handler in `expressionEditorKeymap` check `activeEditableTokenField` and short-circuit list continuation explicitly, if list continuation should still win when no chip is active).

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
madushajg: Fixed. Scoped the extend-friendly assoc (`start=-1, end=1`) to only the actively-edited chip; all other tokens/compounds keep the original safe mapping (`start=1, end=-1`), so boundary typing no longer merges into neighboring chips.

samithkavishke: This doesn't look applied yet — the PR head is still `e3aca503c` (unchanged since this thread started), and the current diff still maps *every* token/compound with the wider assoc:

```ts
start: tr.changes.mapPos(token.start, -1),
end: tr.changes.mapPos(token.end, 1)
```

unconditionally, with no check against `activeEditableTokenField`/`activeStart`. Could you push the scoped version and confirm the commit hash, same as you did for the other thread (`b7717e62f`)?

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/components/ChipExpressionEditor.tsx
madushajg: Fixed. Split the chip-commit Enter binding into its own `chipCommitKeymap`, registered ahead of `listContinuationKeymap`, so committing an active chip takes priority without breaking the existing list-continuation/newline fallback when no chip is active.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
madushajg: Sorry the [commit](https://github.com/ballerina-platform/ballerina-vscode/pull/1113/commits/e6faf302e7794a60e3e51e7b7f87164fccd64283) didn't push.

samithkavishke: Confirmed fixed in `e6faf302e` — `activeStartBeforeChange` is now computed and only the active chip gets the extend-friendly assoc; all other tokens/compounds keep the safe `start=1, end=-1` mapping. Boundary typing next to an inactive chip no longer merges into it. ✅

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/components/ChipExpressionEditor.tsx
samithkavishke: Confirmed fixed in `ebf67592f` — pulling the commit-chip Enter binding into a separate `chipCommitKeymap` and registering it ahead of `listContinuationKeymap` (while it still falls through to `false` when no chip is active) is a cleaner fix than my suggested reorder; list continuation keeps working normally when no chip is being edited. ✅

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
```diff
@@ -229,21 +267,50 @@ export const tokenField = StateField.define<TokenFieldState>({
...
+            };
+        });
+
+        // A chip is actively being edited: its containing expression is typically mid-edit
+        // (often syntactically incomplete), so an LS-backed token refresh landing right now
+        // would recompute the *whole* token stream from that transient/invalid parse - which
+        // can misclassify or drop tokens for chips the user isn't even touching. Ignore it and
+        // keep the locally-mapped tokens/compounds; the real refresh runs once editing commits
+        // (Enter/blur, see buildNeedTokenRefetchListner and buildOnFocusOutListner).
+        const isEditingChip = activeStartBeforeChange !== undefined;
```
samithkavishke: **[LOW] Non-blocking — chip can get stuck mid-edit on an external value sync**

`isEditingChip` (used to skip `tokensChangeEffect`) and `activeChipSelectionGuard` (which clears `activeEditableTokenField` when the selection leaves the active chip) both bail out whenever the transaction has `docChanged=true` — not just ones confined to the active chip's own range. This predates the two most recent fix commits; it isn't something they introduced.

If the editor's `value` prop is externally replaced (e.g. a parent re-render pushes a fresh serialized expression) while a chip is mid-edit, the whole-document replace transaction has `docChanged=true`, so:
- `activeChipSelectionGuard` never runs to clear edit mode, and
- `tokenField.update` ignores the incoming LS-backed token refresh (`if (isEditingChip) continue;`).

The chip can then stay visually stuck in edit mode with a stale/mismapped range until the user manually forces a commit (Enter/blur).

I haven't been able to confirm this fires in the real UI today (it may well be masked by blur-triggered commits in practice), so flagging as a low-priority, non-blocking follow-up rather than a confirmed bug — worth a quick look if there's an existing repro path for external value updates arriving mid-edit.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
```diff
@@ -414,13 +633,16 @@ export const expressionEditorKeymap = [
...
 
             // Check if cursor is within a compound token
             const affectedCompound = compounds.find(
                 compound => compound.start < cursor && compound.end >= cursor
             );
 
-            if (affectedCompound) {
+            // While the compound is in edit mode its text is live/editable - let Backspace
+            // delete a single character normally instead of yanking the whole thing.
+            if (affectedCompound && affectedCompound.start !== activeStart) {
```
samithkavishke: **[LOW] Non-blocking — dead branch / misleading comment**

The guard `affectedCompound && affectedCompound.start !== activeStart` (and the comment above it, "While the compound is in edit mode...") implies a compound can be the actively-edited chip. But `CompoundTokenSequence.tokenType` is type-restricted to `VARIABLE | DOCUMENT`, while `isEditableValueChip` — the only gate that ever sets `activeEditableTokenField` — matches `PARAMETER | VALUE` only. So `activeStart` can never equal a compound's `start`, and `affectedCompound.start !== activeStart` is always `true` today.

Not a functional bug (the branch just always takes the "delete the whole compound" path, which is correct for now since compounds aren't editable). But the comment documents behavior that can't currently occur, and could mislead a future maintainer who extends `isEditableValueChip` to compounds into assuming single-character Backspace-in-place is already exercised for them. Worth a short comment update (or removing the dead condition) whenever compounds become editable, not urgent now.

samithkavishke (dismissed):
LGTM

gigara (commented):
Reviewed the chip-editing changes — the assoc-mapping reasoning and the keymap-ordering are solid and well documented. Left 6 inline comments, ordered by priority (P1 is the one I'd address before merging; P2–P4 are edge-case robustness issues that could be follow-ups).

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
```diff
@@ -404,7 +493,137 @@ export const chipPlugin = ViewPlugin.fromClass(
...
+        }
+    }
+];
+
 export const expressionEditorKeymap = [
+    {
+        // Jumps to (and activates) the next editable chip after the current one, so a
+        // multi-argument function call's placeholders can be filled without ever touching
+        // the mouse. Wraps back to the first chip past the last one.
+        key: "Tab",
```
gigara, line 641: **P1 — Tab/Shift-Tab create a keyboard trap (a11y/UX regression).**

These bindings run whenever the document contains *any* editable chip (not just while one is being edited), and the wrap-around (`?? chips[0]`) means they always return `true`. Since CodeMirror's `defaultKeymap` doesn't bind Tab, Tab previously moved focus out of the editor to the next form field. Now, once a field contains a single value/parameter chip, a keyboard user can never Tab out of it — and Escape here only closes the helper pane. That's a WCAG 2.1.2 "no keyboard trap" violation, and it applies to every editor using `expressionEditorKeymap` (including prompt mode), not just this flow.

Suggestion: only handle Tab when a chip is currently active (`activeStart !== undefined`), and instead of wrapping past the last/first chip, return `false` so focus can leave the editor.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
```diff
@@ -404,7 +493,137 @@ export const chipPlugin = ViewPlugin.fromClass(
...
+        if (event.button !== 0) return false;
+        if ((event.target as HTMLElement)?.closest('[data-chip-widget]')) return false;
+
+        const pos = view.posAtCoords({ x: event.clientX, y: event.clientY });
+        if (pos == null) return false;
+
+        const tokenState = view.state.field(tokenField, false);
+        if (!tokenState) return false;
+
+        const hit = tokenState.tokens.find(token =>
```
gigara: **P2 — Boundary clicks and Tab can activate tokens that are never rendered as chips.**

This handler (and `getEditableChipRanges` below) scans the raw `tokenState.tokens` list, but rendering via `iterateTokenStream` excludes tokens consumed by a compound (`compoundTokenIndices`) and orphan tokens inside an unclosed `${` in interpolation editors. A VALUE/PARAMETER token hidden inside a `${...}` compound chip can therefore be "activated" by Tab or a boundary click: no active box appears (the compound's replace decoration still hides the text, since compounds are never `isEditableValueChip` — see the comment on `buildDecorations`), yet the editor silently enters edit mode and starts suppressing LS token refreshes.

Suggestion: derive the activatable ranges from the same filtering the renderer uses (e.g. reuse `iterateTokenStream`).

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
```diff
@@ -368,6 +439,15 @@ export const chipPlugin = ViewPlugin.fromClass(
                         return;
                     }
 
+                    if (
+                        isEditableValueChip(compound.tokenType) &&
```
gigara: **P3 — The compound "editable chip" branches are unreachable.**

`CompoundTokenSequence.tokenType` is typed `TokenType.VARIABLE | TokenType.DOCUMENT` (types.ts), so `isEditableValueChip(compound.tokenType)` is always `false`. That makes this active-mark branch, the `compoundRanges` half of `getEditableChipRanges`, and the compound `activeStart` comparisons in `tokenField.update` / the Backspace handler dead code today (the `getEditableChipRanges` comment even admits "once reachable").

Suggestion: either drop these branches for now or widen the compound type deliberately — as written, a reader can't tell whether compound editing is supported.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
```diff
@@ -404,7 +493,137 @@ export const chipPlugin = ViewPlugin.fromClass(
...
+
+    const activeStart = update.state.field(activeEditableTokenField, false);
+    if (activeStart === undefined) return;
+
+    const tokenState = update.state.field(tokenField, false);
+    if (!tokenState) return;
+
+    const activeRange = tokenState.tokens.find(token => token.start === activeStart)
+        ?? tokenState.compounds.find(compound => compound.start === activeStart);
+    if (!activeRange) return;
```
gigara: **P4 — Stale active-edit state can wedge token refreshes.**

Two related holes:

1. `if (!activeRange) return;` — when the tracked token no longer exists (removed via `removeChipEffect`, or wiped by an external doc sync), the guard silently gives up instead of clearing, so `isEditingChip` stays `true` and every `tokensChangeEffect` is dropped until an Enter or blur happens.
2. The external value sync in `ChipExpressionEditor.tsx` (`updateEditorState`) dispatches a full-document replacement together with `tokensChangeEffect`. If a chip is active at that moment, the effect is skipped by the `isEditingChip` check while the old token positions get mapped through a full replace — producing garbage ranges — and `isTokenUpdateScheduled` is already `false`, so nothing retries.

Suggestion: treat a transaction annotated `SyncDocValueWithPropValue` as authoritative (apply the refresh and clear the active field), and have this guard clear the active state when the tracked range can't be found.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
```diff
@@ -404,7 +493,137 @@ export const chipPlugin = ViewPlugin.fromClass(
...
+        if ((event.target as HTMLElement)?.closest('[data-chip-widget]')) return false;
+
+        const pos = view.posAtCoords({ x: event.clientX, y: event.clientY });
+        if (pos == null) return false;
+
+        const tokenState = view.state.field(tokenField, false);
+        if (!tokenState) return false;
+
+        const hit = tokenState.tokens.find(token =>
+            isEditableValueChip(token.type) && token.start < token.end && (pos === token.start || pos === token.end)
```
gigara: **P5 — Clicking after a trailing chip can no longer place the caret after it.**

Because of the `pos === token.end` half of this check: when the document *ends* with a value chip (a common single-value field state), any click in the blank space to the right resolves to doc end == chip end and activates the chip with its text selected. There's then no mouse-only way to put the cursor after the chip to append text.

Worth deciding consciously — e.g. only treat `pos === token.start` plus clicks whose x-coordinate falls left of the chip's rendered end as boundary hits.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
```diff
@@ -229,21 +267,50 @@ export const tokenField = StateField.define<TokenFieldState>({
...
+        // (often syntactically incomplete), so an LS-backed token refresh landing right now
+        // would recompute the *whole* token stream from that transient/invalid parse - which
+        // can misclassify or drop tokens for chips the user isn't even touching. Ignore it and
+        // keep the locally-mapped tokens/compounds; the real refresh runs once editing commits
+        // (Enter/blur, see buildNeedTokenRefetchListner and buildOnFocusOutListner).
+        const isEditingChip = activeStartBeforeChange !== undefined;
 
         for (let effect of tr.effects) {
             if (effect.is(tokensChangeEffect)) {
+                if (isEditingChip) continue;
```
gigara, line 322: **P6 — Wasted LS round-trips while editing (efficiency, low).**

While a chip is active, typing a space still schedules a token refetch (`needTokenRefetchListner`'s `endsWith(' ')` rule), the fetch runs, and the resulting `tokensChangeEffect` is then discarded here. Cheaper to not schedule at all when `activeEditableTokenField` is set (the listener can read it via `update.state.field(...)`).

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
madushajg: Fixed the keyboard trap. Tab/Shift-Tab now only act while a chip is already active and never wrap; otherwise they return false so focus leaves the editor normally.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
madushajg: Fixed. Both boundary-click and Tab/Shift-Tab now derive activatable ranges from `iterateTokenStream` (same filtering `buildDecorations` uses), so a token hidden inside a compound/interpolation can no longer be silently activated.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
madushajg: Confirmed dead code (compounds can never satisfy `isEditableValueChip`). Removed the four always-false branches and left comments pointing to why, rather than widening the type.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
madushajg: Fixed both. External sync now clears/overrides stale active-edit state instead of getting dropped, and the selection guard clears when the tracked chip no longer exists instead of leaving refreshes stuck.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
madushajg: Fixed. Clicks past a trailing chip's rendered edge now place the caret normally instead of activating/selecting it.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
madushajg: Fixed. Skips scheduling the LS refetch entirely while a chip is active, since it'd be discarded anyway.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
madushajg: Already resolved. The earlier P4 fix (`SyncDocValueWithPropValue` handling in `activeEditableTokenField.update` and `tokenField.update`) closes this exact repro directly, without needing `activeChipSelectionGuard` to run.

packages/ballerina-side-panel/src/components/editors/MultiModeExpressionEditor/ChipExpressionEditor/CodeUtils.ts
madushajg: Already fixed. Current code has no `!== activeStart` guard; it's just `if (affectedCompound)` with a comment stating compounds aren't individually editable today.

samithkavishke (approved):
LGTM

NipunaRanasinghe (approved):

PR #1117, opened 2026-09-17
Fix the durable workflow designer's forms, agent listing, and icons

hasithaa:
## Purpose

Follow-ups to #1108 for the durable agent designer, each a separate commit. #1108 is merged, so this branch now carries only its own commits.

### Issues

GitHub does not honour closing keywords across repositories, so these need closing by hand once this merges.

- Fixes wso2/product-integrator#2479 — the agent box's gear and menu opened the run node's form; the box now edits the declaration's configuration and the run keeps its arguments.
- Fixes wso2/product-integrator#2113 — `inputType` and `resultType` are on the configuration form, hydrated from the declaration and written back to it.
- Fixes wso2/product-integrator#2111 — Query and Input are gone from the box form and its hydration, which is the second of the two options the issue asks for; the run node keeps them and does write them.
- Fixes wso2/product-integrator#2440 — the configuration form orders Role and Instructions first.

Already fixed by #1108 and verified here: wso2/product-integrator#2443 (the dangling `additionalValues =`), whose advanced-parameter path this PR removes entirely, and wso2/product-integrator#2454 (agent events and human tasks on the overview diagram).

**One form per thing** – Editing a durable agent from its own diagram (gear or menu) opened the run node's form, which mixed a run's `query` with the declaration's configuration; creating one asked for a different set again. The agent box's form is now the declaration's configuration only: Role, Instructions, Model, Input Type, Maximum Iterations, Event Timeout, Maximum Event Waits, hydrated from the config literal and written back to it. In a caller's flow the box keeps the run form, whose fields are the run's arguments: agent, query, input. New config fields are appended one per line after the last field, and a new capability list follows the last config field, instead of a leading-comma line before the closing brace.

**Approval policy is a dropdown, and dependent fields nest** (wso2/product-integrator#2479) – The activity call, the durable agent activity and the tool form had a Requires Approval flag with nine review fields beside it, next to the retry dropdown's own nine, all at one level. Approval Policy is now a dropdown (No Approval / Human Approval) built like Retry Policy, with the review fields under it; a policy the form cannot edit (a const or variable) is carried verbatim. The frontend indents a dropdown's or checkbox's dependent fields under their parent with a left rule, so the two sets read apart.

**Policies refuse to save when they name nothing** – A dropdown branch could be saved with its fields empty, and the form then wrote what the record demanded: `userRoles: ()`, a review nobody can decide and a task nobody can act on, or an invented attempt count. Both are refused now. Max Retries is required under Retry, then Review, so the Save button blocks, and a review with neither roles nor users is refused with a message naming both fields. Nil roles stay legal beside named users, which is how the module scopes a review to people rather than roles. The compiler plugin rejects both shapes too; this is the editor-level check ahead of it.

**Fields in source order, with a slimmer advanced section** – The activity call form shows the two policy dropdowns in the signature's order, and the agent activity form follows `ActivityDecl`: activity, name, description, bindings, approval policy, retry policy. Advanced Configurations is no longer a nested parameter group; it now holds only the low-priority fields, Step Id and Check Error.

**Loading an activity call left its arguments empty** – Arguments written with quoted keys (`{"claimId": claim.claimId, ...}`, as the examples do) did not match their parameters, so the form opened blank. The key is unquoted before the match.

**Durable agents are agents** (wso2/product-integrator#2479) – A `workflow:DurableAgent` declaration publishes under the Agents artifact category, lists in the explorer's Agents section after the AI agents, and its create card sits in the AI Integration panel and wizard category. The entry keeps `kind: DURABLE_AGENT`, so clicking it opens the agent model. Its icon is the agent glyph with a small workflow badge (palette, overview diagram, artifact card); the explorer tree takes a single font glyph and keeps the plain agent icon.

**Creating a type from the input type field wiped the form** – The creation form rebuilt its fields on every identifier update, and saving the new type is one. It now follows only its own function's update.

**CodeRabbit follow-ups from #1108**
- List-form `humanTasks` hydrate `administratorRoles` and `administratorUsers`.
- A review title or description in expression mode is written through unchanged.
- `RetryBeforeReview` requires `maxRetries`: a blank attempt count writes the AutoRetry default of 3 instead of an invalid record that reopened as Human Review.
- The gate helper's doc no longer names peer delegation; `PeerDecl` has no approval policy.
- `WORKFLOW_163` is emitted by the 0.10.0 compiler plugin (verified with `bal build` on the test distribution), but plugin diagnostics do not reach the language-server test harness, so the migration's edit computation is extracted and unit-tested: flag plus roles become an `approvalPolicy`, a false flag or a removed peer field is dropped.
- The leftover braces and the triple guard on the open rest went with the advanced-parameter path itself.

**0.10 fields the designer could not reach** – A gap analysis against the released module found three, now closed. An agent-declared human task carries `taskInputType`, which no form offered and the entry writer never wrote, so editing such a task deleted it. The agent's own `resultType` was equally invisible; it joins the configuration form beside the input type. And the module takes a step id on `sleep`, both child workflow starters, the human task and the activity call, while only the activity call offered one: all five do now, as an advanced field through one helper. That helper also quotes a step id typed as text, which every call site previously wrote raw, producing source that did not compile.

**Tests** – Node templates (empty forms) for the activity call, the agent activity and the tool regenerate with the dropdown shape; a diagram fixture loads a call with quoted keys, step id, approval and retry set; to_source cases cover a human-approved call with a step id, an agent activity with both policies, the blank attempt count and the configuration form; the artifacts golden lists the durable agent under Agents. All generated against `ballerina/workflow` 0.10.0.


packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/workflowAudienceValidation.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    it("refuses a declared task that names nobody, by node kind", () => {
+        expect(validateWorkflowAudience({}, "DURABLE_AGENT_HUMAN_TASK")?.fieldKey).toBe("userRoles");
+        expect(validateWorkflowAudience({ users: "alice" }, "DURABLE_AGENT_HUMAN_TASK")).toBeUndefined();
+        // A node that declares no task is not subject to the rule.
+        expect(validateWorkflowAudience({}, "ACTIVITY_CALL")).toBeUndefined();
+    });
+
+    it("treats the empty shapes a cleared field leaves behind as naming nobody", () => {
+        for (const empty of ["", "   ", "()", "[]", [], null, undefined]) {
```
NipunaRanasinghe: This is what's failing CI (`Build / Fast tests`, the "Type-check test code" step). Reproduced locally:

```
workflowAudienceValidation.test.ts(39,20): error TS7005: Variable 'empty' implicitly has an 'string | any[]' type.
```

The mixed array collapses to a union that `noImplicitAny` won't infer. Annotating it fixes it:

```ts
const empties: unknown[] = ["", "   ", "()", "[]", [], null, undefined];
for (const empty of empties) {
```

packages/ballerina-visualizer/src/views/BI/ComponentListView/AIAgentPanel.tsx
```diff
@@ -80,13 +81,24 @@ export function AIAgentPanel(props: AIAgentPanelProps) {
...
-    if (q?.trim() && !agentMatches && mcpTriggers.length === 0) {
+    if (q?.trim() && !agentMatches && !durableAgentMatches && mcpTriggers.length === 0) {
         return null;
     }
 
+    // A durable agent is a workflow as much as an agent, so it stays creatable in any scope: the
+    // panel is dimmed as a whole only when it has no card left enabled.
     return (
-        <PanelViewMore disabled={isDisabled}>
+        <PanelViewMore disabled={isDisabled && !durableAgentMatches}>
```
NipunaRanasinghe: This un-dims the whole panel, not just the durable agent card.

`cardMatchesSearch` returns `true` when the query is empty (`componentListUtils.ts:34`), so with no search — the normal case — `durableAgentMatches` is `true` and `disabled` is always `false`. `PanelViewMore` sets `opacity: 0.5; pointer-events: none` when disabled, so in a non-`AI_AGENT` scope the AI Chat Agent card and the MCP cards go from blocked to fully clickable.

The intent in the comment above is right, but it needs the gating moved down to the cards: keep `disabled={isDisabled}` on the panel and render the durable agent card outside it, or drop it from the panel and disable the AI-chat/MCP cards individually.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/index.tsx
```diff
@@ -1295,6 +1296,19 @@ export const FlowNodeForm = forwardRef<FormExpressionEditorRef, FlowNodeFormProp
     };
 
     const handleFormValidation = async (data: FormValues, dirtyFields?: any): Promise<boolean> => {
+        // A task naming nobody is refused when the source is generated, which is after the save has
+        // left the panel. Catching it here keeps the panel open with what was typed.
+        const audienceError = validateWorkflowAudience(data, node?.codedata?.node);
```
NipunaRanasinghe, line 1307: Can you check by hand that the panel recovers from this?

The early return skips `setDiagnosticsToFields`, which is the only thing that clears `field.diagnostics`. The diagnostic you set here flows into `diagnosticsInfo` → `setError(key, { type: "expression_diagnostic" })` → `isValid` false → `disableSaveButton` (`Form/index.tsx:1028-1125`). That error is cleared only when `diagnosticsInfo` for the key goes empty, which needs either a `formFields` rebuild or an expression editor pushing `setDiagnosticsInfo({ key, diagnostics: [] })`. The roles field renders in text mode by default (`buildReviewerRolesSubProperty`), so I'm not sure either happens.

Worth trying: trigger the error, type a role, see if Save re-enables. If it doesn't, the form is stuck and the diagnostic needs clearing on the next validation pass.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1716,38 +1723,30 @@ private void collectCapabilityFields(MappingConstructorExpressionNode config, St
...
+
+    private static void putReviewValues(Map<String, String> values, ActivityCallBuilder.ReviewKeys keys,
+                                        ActivityCallBuilder.ReviewFormValues review) {
+        putIfNotBlank(values, keys.userRoles(), review.userRoles());
+        putIfNotBlank(values, keys.users(), review.users());
+        putIfNotBlank(values, keys.excludedUsers(), review.excludedUsers());
+        putIfNotBlank(values, keys.excludedRoles(), review.excludedRoles());
+        putIfNotBlank(values, keys.administratorRoles(), review.administratorRoles());
+        putIfNotBlank(values, keys.administratorUsers(), review.administratorUsers());
+        putIfNotBlank(values, keys.title(), review.title().value());
```
NipunaRanasinghe: `.value()` drops the `isExpression` flag that `ApprovalPolicyForm.normalize` just worked out. The activity call path keeps it — `addHiddenReviewTextProperty` takes the whole `ReviewText` — but the capability path only has a `Map<String, String>` to write into.

So for a tool or agent activity whose `approvalPolicy.title` is a variable reference, the value lands in a text-mode field and `addQuotedRecordField` should re-quote it on save, turning `titleVar` into `"titleVar"`. The PR body lists this as fixed, but the fix looks like it only covers the activity call.

A fixture with an expression title on a `ToolDecl` would settle it either way.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ChildWorkflowCallBuilder.java
```diff
@@ -138,6 +139,7 @@ public void setConcreteTemplateData(TemplateContext context) {
                 .editable(true)
                 .stepOut()
                 .addProperty(Property.VARIABLE_KEY);
+        WorkflowUtil.addStepIdProperty(this);
```
NipunaRanasinghe, line 142: Sleep and Human Task both call `markStepIdAdvanced` so the template and CodeAnalyzer's source re-read render the same form. The two child workflow builders add the property to the template but go through the generic remote-call path on re-read, so an existing `runChildWorkflow(..., stepId = "x")` looks like it would come back as a plain non-advanced field rather than under Advanced Configurations.

`child_workflow_run_step_id_to_source.json` covers the write direction only — is there a diagram-generator fixture for reading one back? Same question for `ChildWorkflowRunBuilder`.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/workflowAudienceValidation.ts
```diff
@@ -0,0 +1,89 @@
...
+const TASK_AUDIENCE: AudiencePair = { roles: "userRoles", users: "users" };
+
+/** Retry policy values that raise a review. */
+const REVIEWING_RETRY_POLICIES = ["ManualRetry", "RetryBeforeReview"];
+const HUMAN_APPROVAL = "HumanApproval";
+
+/** Node kinds whose own form declares a task, rather than a policy that raises one. */
+const TASK_NODES = ["DURABLE_AGENT_HUMAN_TASK", "HUMAN_TASK"];
+
+const MESSAGE = "Name who may decide this: fill in the roles, the users, or both";
```
NipunaRanasinghe, line 47: Nit: the same rule now has three messages. This one says "Name who may decide this: fill in the roles, the users, or both", `ActivityCallBuilder.NO_REVIEWER_MESSAGE` says "A review needs someone who may decide it: fill in Reviewer Roles, Reviewers, or both", and the compiler plugin says a third thing.

Three layers of the check is fine and you explain why. The three wordings aren't — the person sees one or the other depending on how far the save got. Worth aligning at least the two in this repo.

packages/ballerina-side-panel/src/components/editors/DropdownChoiceForm.tsx
```diff
@@ -95,7 +98,7 @@ export function DropdownChoiceForm(props: DropdownChoiceFormProps) {
                     }}
                 />
             </ChoiceSection>
-            <FormSection>
+            {dynamicFields.some(dfield => !dfield.hidden) && <FormSection>
```
NipunaRanasinghe: Nit: this guard doesn't match the filter below, which also drops `advanced` fields and `GROUP_SECTION`s with no `advanceProps`. A branch made only of advanced fields would render an empty indented box with a left rule.

No branch in this PR hits it, but matching the predicate to the filter costs nothing.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -1588,23 +1541,44 @@ public static Map<Path, List<org.eclipse.lsp4j.TextEdit>> setAgentConfigFields(
...
+            LinePosition closeBrace = config.closeBrace().lineRange().startLine();
+            position = new org.eclipse.lsp4j.Position(closeBrace.line(), closeBrace.offset());
+            appendFields(insertion, missing, ", ", false);
+        } else {
+            MappingFieldNode last = config.fields().get(config.fields().size() - 1);
+            LinePosition end = last.lineRange().endLine();
+            position = new org.eclipse.lsp4j.Position(end.line(), end.offset());
+            LinePosition start = last.lineRange().startLine();
+            boolean multiLine = start.line() != config.openBrace().lineRange().startLine().line();
+            appendFields(insertion, missing, multiLine ? ",\n" + " ".repeat(start.offset()) : ", ", true);
```
NipunaRanasinghe: Nit: `" ".repeat(start.offset())` reproduces the column as spaces, so a tab-indented declaration gets space-indented appends. Same on line 389.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/workflowAudienceValidation.test.ts
hasithaa: Fixed in d79bc5bbd — annotated as `unknown[]`. I also found the gate I was missing locally: `node scripts/typecheck-tests.js` reproduces it, and it is clean now.

packages/ballerina-visualizer/src/views/BI/ComponentListView/AIAgentPanel.tsx
hasithaa: Right, and thanks for catching it. `cardMatchesSearch` returning true on an empty query made `disabled` always false, so the chat agent and MCP cards went from blocked to clickable. Taken your second option: the panel no longer passes `disabled` at all, and the two cards keep the `disabled`+tooltip they already had. The durable agent card stays enabled in every scope, as the Durable Workflow panel is.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/index.tsx
hasithaa: You were right, and it would have wedged. `field.diagnostics` seeds `diagnosticsInfo` through the rebuild effect (Form/index.tsx:765), that sets an RHF error, and `disableSaveButton` includes `Object.keys(errors).length > 0` — so Save goes dead and the only thing that clears the diagnostic is a validation pass the disabled button can no longer start. The message is a notification now (`showErrorMessage`, as `useCreateNode` does): the submit is still blocked and the panel keeps every value, but nothing latches.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Confirmed, and it was my regression: routing the capability gate through `reviewSubProperties` gave it dual-mode wording fields, and `ReviewText.fromSource` decodes the literal, so the mode was lost on a path whose values are only strings. The capability forms now declare the title and description as expressions only, and `putReviewValues` hands on `sourceForm()` rather than the decoded text. Two fixtures: `durable_agent_gate_policy` reads `title: refundTitle` back as `refundTitle`, and `durable_agent_register_tool_expression_title_to_source` writes it out unquoted.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ChildWorkflowCallBuilder.java
hasithaa: Good catch — the two directions did disagree. Re-reading any workflow context call now runs `markStepIdAdvanced`, so the generic remote-call path renders what the templates do. `durable_agent_task_types` reads back a `runChildWorkflow(..., stepId = "audit#1")` alongside a sleep that names its step.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/workflowAudienceValidation.ts
hasithaa: Agreed. Both now come from `WorkflowUtil.AUDIENCE_REQUIRED_MESSAGE`, with the designer's copy commented as mirroring it. The plugin's own wording stays as the module states it.

packages/ballerina-side-panel/src/components/editors/DropdownChoiceForm.tsx
hasithaa: Taken — the guard matches the filter now, so a branch of only advanced fields or an empty GROUP_SECTION renders no box.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
hasithaa: Taken. Both sites call `indentOf`, which reads the node's own leading whitespace minutiae and falls back to the column in spaces, so a tab-indented declaration keeps its tabs.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1716,38 +1729,32 @@ private void collectCapabilityFields(MappingConstructorExpressionNode config, St
...
+        putIfNotBlank(values, keys.userRoles(), review.userRoles());
+        putIfNotBlank(values, keys.users(), review.users());
+        putIfNotBlank(values, keys.excludedUsers(), review.excludedUsers());
+        putIfNotBlank(values, keys.excludedRoles(), review.excludedRoles());
+        putIfNotBlank(values, keys.administratorRoles(), review.administratorRoles());
+        putIfNotBlank(values, keys.administratorUsers(), review.administratorUsers());
+        // Source, not the decoded text: these values reach a capability form as plain strings, and a
+        // reference decoded into text would be quoted into a literal of the same spelling on save.
+        putIfNotBlank(values, keys.title(), review.title().sourceForm());
+        putIfNotBlank(values, keys.description(), review.description().sourceForm());
```
NipunaRanasinghe, line 1756: The title and description are fixed, but the six role and user fields above them have the same exposure and are still dual-mode.

They're seeded with source too — from `durable_agent_gate_policy`:

```json
"approvalUserRoles": "\"finance\"",
"approvalTitle": "\"Approve the charge\"",
"description": "Charges the card"
```

Two conventions in one form: the capability's own fields arrive decoded, the approval review's arrive as source. The roles survive today only because `roleSource` falls through to `quoteIfBareRole`, whose `EXPRESSION_LIKE` is `[(){}:.]`. A bare identifier has none of those, so a gated tool with `userRoles: financeRoles` hydrates as `financeRoles`, keeps TEXT selected, and saves back as `"financeRoles"` — the reference rewritten into a literal of the same spelling. That's the bug you just fixed for the title. Your own comment on `roleSource` names it: "it does not recognise a bare identifier as a reference, which is why the mode is checked first" — and on this path nothing sets the mode.

The reason the mode is lost isn't really that the values are strings. It's `FlowDiagram/index.tsx:3916`, where the capability form is seeded from a fresh template:

```ts
for (const [key, value] of Object.entries(values)) {
    if (nodeProps?.[key]) {
        nodeProps[key].value = value;   // only .value, never types[].selected
    }
}
```

If that loop set the mode — string literal → TEXT with the quotes stripped, anything else → EXPRESSION, which is exactly what `ReviewText.fromSource` already does — every dual-mode capability field would round-trip, and the title and description could keep their text box.

Fine to leave for a follow-up, this PR is already large. But worth filing, because the roles are still exposed either way.

NipunaRanasinghe (approved):
Approving with one finding. +1 to have it in a follow-up PR.

PR #1118, opened 2026-09-17
Show pulling progress and load failures when creating a workflow artifact

NipunaRanasinghe:
## Purpose
Resolves wso2/product-integrator#2487

## Goals
Tell the user a package is being downloaded, reusing the status card the service artifacts (Event Integrations, connectors) already show for the same wait — and say so when it fails, instead of leaving the wait running forever.

### Success Scenario
https://github.com/user-attachments/assets/9a943d28-3173-4089-b536-2b77f4817e04

### Error Scenario
https://github.com/user-attachments/assets/b1e44da0-0fda-43fa-95bf-f3c2ce7a1a9e

**Pulling status.** A 3-second timer starts while a *new* workflow-family artifact's template is loading — Durable Workflow, Durable Agentic Workflow or Workflow Activity, the three that pull `ballerina/workflow`. A cached package resolves well inside that window, so the message only appears on a real pull. Same race as `useServiceInitModel` and `ServiceCreationView`. Once it fires, the bare ring is replaced by the download-icon status card: *"Please wait while the workflow package is being pulled..."*

The card matches `ServiceConfigureForm` (the Event Integrations card the issue points at) and `ServiceCreationView`. It deliberately does **not** match `AddConnectionWizard`/`ConnectionConfigurationPopup`, which use a column layout with a larger blue icon because they render in a ~400px side panel.

Gated on creation (`!functionName`): opening an existing workflow resolves a package the project already has, so claiming a pull there would be false.

**Failure reporting.** `getFunctionNode` had no try/catch and set `isLoading(false)` after its awaits, so a rejection — offline, unresolvable version, i.e. exactly what this PR exists for — parked the user on the loader indefinitely. It now has try/catch/finally and renders an error card with a Retry button that re-runs the load. This card is not workflow-gated: it covers every kind this form handles, so its wording names neither the template nor the package (*"Failed to load the durable workflow. Please try again."*).

**Load supersession.** If the effect re-fires while a request is in flight, a slow failure could raise the error card over an already-loaded form, or a stale success could overwrite a newer template. A `loadSeqRef` stamp — the ref equivalent of `useServiceInitModel`'s `cancelled` flag — means only the newest load writes state. Both loaders claim a sequence; the check is folded together with the file's existing `isMountedRef` convention as `isCurrentLoad(seq)`.

The edit path takes the guard only, not the error handling, to keep this off a path all seven artifact kinds use on a stabilizing release branch. **Known gap accepted in doing so:** "create-load superseded by an edit-load that then rejects" now hangs, because the create path's `finally` defers rather than unconditionally clearing `isLoading`. It needs a create→edit prop change mid-request *and* a rejection. Narrower than the case this PR addresses, but newly reachable rather than pre-existing.

**Incidental.** Dropped two dead CSS rules from the card (`DownloadIcon` renders `<svg width="24" height="24">` with `stroke` from its prop, and ui-toolkit's `Icon` renders an `<i>`, so `& > svg { font-size; color }` reached neither) and inlined a single-use constant.

Diff is 150/59, or **94/3 under `?w=1`** — most of the rest is the try block re-indenting.

No screenshot: the pulling card is an existing, already-shipped visual, and the states only reproduce against a cold Central cache or a forced failure. See Automation tests.

## UI Component Development
- [ ] Added reusable UI components to the ui-toolkit — no new component here. It reuses `DownloadIcon` and the local `StatusCard` pattern. That pattern is copy-pasted across five screens now; #1119 extracts it into a shared `PackagePullStatus` against `main`.
- [x] Use ui-toolkit components wherever possible — uses `Icon`, `Button`, `Typography` and `ThemeColors` from the ui-toolkit.
- [x] Matches with the native VSCode look and feel — the same cards the service creation paths already render.

## User stories
As an integration developer creating my first workflow in a project, I want to know the editor is downloading a package rather than hung, so I wait instead of reloading the window — and to be told when it fails, so I can retry instead of waiting on something that will never finish.

## Release note
Show the package pulling status when creating a workflow artifact for the first time, instead of a blank loading screen, and report a retryable error when the template fails to load.

## Documentation
N/A — no change to behaviour beyond progress and failure messaging on an existing screen.

## Training
N/A

## Certification
N/A — progress and error messaging, no functional or conceptual change to certify.

## Marketing
N/A

## Automation tests
- Unit tests
  > None. See below — I attempted them and reverted.
- Integration tests
  > None. Reproducing the states needs a cold Ballerina Central cache or a forced failure.

**Verified:** `tsc --noEmit -p tsconfig.spec.json` in `packages/ballerina-visualizer`, run before and after — the error set is byte-identical, the same 11 pre-existing `Type 'Timeout' is not assignable to type 'number'` errors in other files, none introduced. `jest`: 44 suites, 609 tests, all passing — though **no existing test covers this file**, so that is a guard on the rest of the package, not evidence about this change.

**Attempted and reverted:** a `createRoot` + `act` test following `AIChat/Footer/index.test.tsx`, stubbing the RPC client and driving the template promise by hand, to cover the plain loader, the pulling card at 3s, the create-only gate, the error card and Retry, and the supersession case. It doesn't run: the shared jest preset pins `jsx: 'react'`, so a component importing only hooks needs React on `globalThis`; past that, React 18's async `act` never settles under the fake timers the 3-second delay needs, and the first hang corrupts the act environment for the rest. Solvable — probably by driving the delay without faking timers — but it is real work and I did not want to hold the fix on it.

**Not verified:** I have not run the extension. Three render states have never been executed by me or by any test — the pulling card, the error card and the Retry button — nor has the supersession guard. Forcing the error state for a manual pass needs a temporary throw in `getFunctionNode`.

## Security checks
- Ran FindSecurityBugs plugin and verified report? N/A — TypeScript/React change, no Java code.

## Samples
N/A

## Related PRs
- #1119 (draft, against `main`) — extracts the duplicated pulling card into a shared `PackagePullStatus` and converts the four pre-existing copies. This PR adds a fifth copy; it should be converted when `staging/5.14.0` and `main` next cross-merge. Not folded in here because `FunctionForm` has no pulling card on `main` yet and the branches have diverged.

## Migrations (if applicable)
N/A

## Test environment
macOS 15 (Darwin 25.5.0), Node via the pnpm workspace. Typecheck and jest only — see Automation tests.

## Learning
Followed the existing pulling-status implementations in `ServiceCreationView.tsx` and `ServiceConfigureForm.tsx`, which is what the linked issue asked to reuse, and `useServiceInitModel`'s `cancelled` flag for the supersession guard.

## Summary

`FunctionForm` now shows package-pulling progress during creation of Durable Workflow, Durable Agentic Workflow, and Workflow Activity artifacts.

After three seconds of loading, the UI displays: “Please wait while the workflow package is being pulled...” Template load failures now show an error card with a Retry button.

Other artifact types retain the existing loading behavior. No new UI components or automated tests were added.


madushajg (dismissed):

NipunaRanasinghe:
Thanks — worked through all eight. Fixed 1, 2, 5 and 7 in df51c8f. Two I'd push back on.

### Fixed

**1. Failed pull left a permanent false claim.** Confirmed: `getFunctionNode` (`:292`) had no try/catch around either await, and `setIsLoading(false)` sat after them at `:357` — the only try/catch inside (`:344`) wraps the durable-agent name lookup, not `getNodeTemplate`. You're right that the card made the stuck state worse than the spinner it replaced, since it asserts something untrue. Wrapped the body in try/catch/finally and added the failure branch, mirroring `ServiceCreationView:337`: an error card reading "Failed to load the durable workflow template. Please try again." with a Retry that bumps a `loadAttempt` the load effect depends on (same pattern as `AIChatAgentWizard`).

**2. Edit path triggered it.** Confirmed — `getExistingFunctionNode` (`:361`) sets `isLoading` too. Gate is now `!functionName && (isWorkflow || isDurableAgent || isActivity)`.

**5. Both CSS rules dead.** Confirmed in the source, and it's worse than "no em units": `DownloadIcon` renders `<svg width="24" height="24">` with `stroke={color}` on the `<g>`, so neither `font-size` nor `color` reaches it — and ui-toolkit's `Icon` renders an `<i>`, not an `<svg>`, so `& > svg` never matched a font icon either. Both rules deleted. (No visual change: the icon renders at its natural 24px, which is what the dead rule was asking for.)

**7.** Inlined the constant.

### Pushing back

**6. "Icon color is `ON_SURFACE` vs `var(--vscode-progressBar-background)` at every other site; layout is row vs column."** It's two of four sites, not every other one. Scoped to each `StatusCard` block:

| Site | Layout | Icon colour |
|---|---|---|
| `AddConnectionWizard` | column, 32px | `progressBar-background` |
| `ConnectionConfigurationPopup` | column, 32px | `progressBar-background` |
| `ServiceCreationView` | row, 24px | `ON_SURFACE` |
| `ServiceConfigureForm` | row, 24px | `ON_SURFACE` |
| `FunctionForm` (this PR) | row, 24px | `ON_SURFACE` |

The two column copies are that shape because connectors render in a ~400px side panel. This copy is byte-identical to `ServiceConfigureForm` — the Event Integrations card the issue names as the reference.

That said, you caught a real error: **the PR description** claimed "the same wording and layout as `ServiceConfigureForm` and `AddConnectionWizard`", and the `AddConnectionWizard` half was wrong. Corrected.

**8. `isPullingPackage && !saving` can't occur.** The outer condition is `(isLoading || (saving && !functionName))` — an OR, so the saving branch isn't gated on `isLoading` being false. And the form renders as a sibling of the loader, independent of `isLoading`, so a submit during a re-load puts both flags up; `!saving` is what keeps "Creating…" from being overridden by the pull message. Narrow, but reachable. Happy to drop it if you'd rather — you did call it harmless.

### 3 and 4 — needs a call on sequencing

Agreed these resolve together, and rebasing onto `PackagePullStatus` beats the `LoadingRing` one-liner: it keeps the card the issue asks for *and* collapses this to ~5 lines. The blocker is base branches — this PR targets `staging/5.14.0`, #1119 targets `main`, and they've diverged (42 commits on main, 70 on staging). So one of:

1. Land this as-is for 5.14.0, convert at the next cross-merge (fifth copy exists temporarily).
2. Land the extraction on `staging/5.14.0` too, then rebase this onto it.
3. Retarget this to `main`, if the fix doesn't need to ship in 5.14.0.

Which do you want? I'd default to (1) if 5.14.0 is close, (2) otherwise.

### Still not verified

I have not run the extension — typecheck (no new errors against baseline) and jest (44 suites / 609 tests, none covering this file) only. The new error card and Retry in particular have not been seen rendered.

Also flagging one I did **not** touch, to keep this diff to the review's scope: `getExistingFunctionNode` has the identical missing-try/catch hang. It's pre-existing, and after fix 2 it no longer shows a misleading message — just the old neutral spinner. Say the word if you want it in here rather than a separate issue.

hasithaa (commented):
functionality looks good to me.

hasithaa (approved):

dulajdilshan (approved):

packages/ballerina-visualizer/src/views/BI/FunctionForm/index.tsx
madushajg, line 436: `getExistingFunctionNode` has no `try/catch/finally`, so any rejection of the `getFunctionNode` RPC call leaves `isLoading` stuck true forever, no error card (`loadError` is only ever set in the create path) and no Retry button (it only renders when `loadError && !isLoading`). This is broader than "create-load superseded by a rejecting edit-load": a single, ordinary open of the edit form for any of the seven artifact kinds whose RPC call fails for any reason (LS not ready, transient error, timeout) hangs the UI on a bare spinner with no recovery short of closing and reopening the panel - reachable via completely normal use, not just rapid double-navigation.

NipunaRanasinghe: Good catch — the behaviour you describe is real, and it's fixed in #1135 against the same `staging/5.14.0`.

One correction on scope: the hang is pre-existing, not introduced here. On the base commit `getExistingFunctionNode` already had no `try`/`catch` and already called `setIsLoading(false)` after an unguarded await, so an ordinary edit-form open whose RPC rejects hung exactly the same way before this PR. Diffing that function base→head with `-w`, everything this PR added to it is:

```diff
+ const seq = ++loadSeqRef.current;
  setIsLoading(true);
  ...
+ if (!isCurrentLoad(seq)) { return; }
  setFunctionNode(flowNode);
  setIsLoading(false);
```

One narrower case *was* newly reachable: a create-path load superseded by a rejecting edit-path load. The create path's `finally` now defers to the newer load instead of clearing `isLoading` unconditionally, and the newer load had no handler to clear it — so that interleave hangs where it previously didn't. It needs a create→edit transition mid-flight, so it's much narrower than what you describe, but it's ours.

For context on why the asymmetry was there at all: an earlier revision of this PR did give the edit path full `try`/`catch`/`finally`. I pulled it back to keep the edit path — which every artifact kind goes through — out of scope on a stabilizing branch. Your comment is a fair argument that that was the wrong trade, and #1135 puts it back.

#1135 gives `getExistingFunctionNode` the same `try`/`catch`/`finally` with `isCurrentLoad(seq)` guards on all three state writes, reusing the existing error card and Retry. No new state, no rendering change, and the package-pulling status stays create-only. Both cases above are covered.

Raising the pre-existing/new distinction only so the history reads right — your finding stands either way, and it's why the edit path has any handling now.

PR #1123, opened 2026-09-18
Carry the field mode when seeding a capability form, and three deferred designer fixes

hasithaa:
## Purpose

#1117 is merged, so this branch carries only its own commits.

Opening a durable agent capability — an activity, tool, peer, event or human task — builds a fresh node template and seeds it with the entry's values. The seeding set the value and nothing else, so a field offering both a text box and an expression editor kept the mode its template declared, which is text. A value that was a reference was then quoted into a literal of the same spelling when the form was saved:

```ballerina
// before: a gated tool whose reviewer roles name a variable
tools: [{tool: refund, approvalPolicy: {userRoles: financeRoles}}]
// after opening the tool and saving it unchanged
tools: [{tool: refund, approvalPolicy: {userRoles: "financeRoles"}}]
```

It survived only where `roleSource` happened to read the value as an expression, and its test looks for `[(){}:.]`, which a bare identifier has none of.

**Why the value alone could not decide it.** The two sides used different conventions: a capability's own fields arrived decoded, with their quotes stripped, while the approval review's arrived as source. One rule could not serve both, so the mode was unrecoverable at the point it was needed.

**The fix.** Hydration is one convention now — source for every capability field — and the seeding chooses the mode from it: a string literal fills the text box with its quotes stripped and its escapes resolved, anything else selects the expression editor. With the mode travelling again, the review title and description take back the text box that #1117 had to remove from them for this reason.

**Tests.** Eleven cover the rule, including the bare identifier that is the bug, an interpolated string template, a concatenation that merely starts and ends with a quote, nil, and a single-mode field. The capability goldens are regenerated to the one convention.

## The deferred fixes

Three of the issues moved to 5.1.0-RC1 ride along, each its own commit.

**The line below an agent run** — Fixes wso2/product-integrator#2472. The agent box is the one node whose link update is conditional, because the synthetic copy that floats above the chain must not join it. An `agent.run(...)` statement carries the same marker, and when it opened a block there was no preceding node to link from, so no branch of the condition held and the box never became the chain's tail. Nothing linked down from it, and since a block's terminator is only drawn when it receives a link, that disappeared too. The start a branch hangs from now counts as a link source, which the link update already knew how to use. Three tests, one of which fails without the change.

**The request body on a DELETE** — Fixes wso2/product-integrator#2112. The defect is wider than reported: `activity:callRestAPI` forwards the body on POST, PUT, PATCH **and** DELETE, and calls `connection->get(path, headers)` without it. The form offered the body on three of the four and the emitter wrote it on the same three, so a DELETE that carried one lost it on save. Both sides follow the module now, with a fixture for the generated call. A body is still not offered on GET, because the module's dispatch ignores it there.

**The guided type editor** — Fixes wso2/product-integrator#2079. It was gated on the type helper's entry alone, so a host that disallows creating types still opened the guided editor from a type field.

Not taken here: wso2/product-integrator#2114, the child workflow input not retyping when the selection changes. Doing it properly means the workflow dropdown carrying a typed `input` field per workflow with hidden carriers, which is a change of the form's shape rather than a fix, so it deserves its own PR.

## Issues

GitHub does not honour closing keywords across repositories, so these need closing by hand once this merges: wso2/product-integrator#2489, #2472, #2112 and #2079.

## Summary

- Preserve capability field values and select text or expression modes during form seeding.
- Restore text-box support for capability titles and descriptions.
- Improve approval-policy handling for Durable Agent capabilities.
- Fix links after `agent.run(...)` blocks.
- Preserve DELETE request bodies in `activity:callRestAPI`.
- Prevent guided type editors from opening when `typeCreation` is disabled.
- Add tests and regenerated fixtures.


packages/ballerina-visualizer/src/views/BI/FlowDiagram/capabilityFieldValues.test.ts
```diff
@@ -0,0 +1,104 @@
+// A capability form is a fresh template seeded with source. These pin the half that was missing:
```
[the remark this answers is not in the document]
hasithaa: Added, thanks — the file now carries the 2026 header like the other two test files in this PR.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/capabilityFieldValues.ts
```diff
@@ -0,0 +1,116 @@
...
+
+/** The text a string literal denotes, with the escapes it carries resolved. */
+function literalText(source: string): string {
+    return source
+        .slice(1, -1)
+        .replace(/\\n/g, "\n")
+        .replace(/\\t/g, "\t")
+        .replace(/\\r/g, "\r")
+        .replace(/\\"/g, '"')
+        .replace(/\\\\/g, "\\");
```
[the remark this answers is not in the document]
hasithaa: Real bug, thanks. Traced it: for `"C:\\\\new"` the first replacement matched the backslash that the escaped backslash had already consumed, so the value became a path with a line break in it, which the form showed and a save would have written back.

It reads the escapes in order now, consuming the character after an escaping backslash, which is what the language server's own `stringLiteralText` does — so the two decoders agree. Two regression cases: that input, and a literal ending in an escaped backslash. I checked the first one fails against the old decoder.

packages/bi-diagram/src/visitors/NodeFactoryVisitor.ts
```diff
@@ -971,9 +971,12 @@ export class NodeFactoryVisitor implements BaseVisitor {
             this.updateNodeLinks(node, nodeModel);
         } else if (isDeclarationCanvasBox && this.lastNodeModel instanceof StartNodeModel) {
             this.updateNodeLinks(node, nodeModel, { showAddButton: false });
-        } else if (this.lastNodeModel) {
+        } else if (this.lastNodeModel || node.viewState?.startNodeId) {
```
NipunaRanasinghe: `viewState.startNodeId` is never set anywhere in the repo — searching for it finds only reads (line 91 here, `utils/diff.ts:444`) and the new test. On top of that, `InitVisitor.getDefaultViewState()` (`InitVisitor.ts:34-45`) replaces every node's `viewState` with `{x,y,lw,rw,h,clw,crw,ch}`, and it runs before this visitor (`Diagram.tsx:177-189`), so even if the LS sent the field it would be wiped.

So for a real `agent.run(...)` opening a block we still get `isAgentBox=true`, `isDeclarationCanvasBox=false`, `lastNodeModel=undefined`, `startNodeId=undefined` — all three branches miss, `updateNodeLinks` is never called, the box never becomes `lastNodeModel`, and #2472 is unchanged.

The marker that actually separates the two cases is already here: the LS sets `agentDeclarationCanvas` only on the synthetic box (`CodeAnalyzer.java:4295`), never on the in-chain statement (`CodeAnalyzer.java:1051`). So this should be:

```ts
} else if (!isDeclarationCanvasBox || this.lastNodeModel) {
```

packages/bi-diagram/src/test/durable-agent-run-links.test.ts
```diff
@@ -0,0 +1,71 @@
...
+        expect(sources).toContain("run-in-branch");
+    });
+
+    it("keeps the synthetic box above the chain out of it", () => {
+        const visitor = new NodeFactoryVisitor();
+        // The floating copy is visited before anything else: no preceding node and no start.
+        visitor.beginVisitDurableAgentRun(agentBox("floating-box"));
+        expect((visitor as any).links).toHaveLength(0);
+        expect((visitor as any).lastNodeModel).toBeUndefined();
+    });
```
NipunaRanasinghe: This fixture has `agentBox: true` and no `agentDeclarationCanvas`, which is an in-chain `agent.run(...)` statement, not the synthetic copy — the LS only marks the synthetic one (`CodeAnalyzer.java:4295`). So asserting `lastNodeModel` stays undefined pins the bug rather than the intended behaviour: that is exactly the state where the box fails to become the chain's tail.

The test above it also only passes because it hand-injects `startNodeId`, which never exists at runtime (see my comment on `NodeFactoryVisitor.ts:974`). Between that and reaching into `(visitor as any)` for private fields, the suite currently verifies a no-op. Worth driving these through `Diagram`'s visitor pipeline instead, or at least building the fixtures from what the LS actually emits.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1719,20 +1712,17 @@ private void collectCapabilityFields(MappingConstructorExpressionNode config, St
...
                 // `userRoles: ()` says "only the named users decide"; the roles box stays empty for it.
-                values.put(propertyKey, nilAsBlank(stripQuotes(rawValue)));
-            } else if (TEXT_MODE_CAPABILITY_FIELDS.contains(fieldName)) {
-                values.put(propertyKey, stripQuotes(rawValue));
+                values.put(propertyKey, nilAsBlank(rawValue));
             } else {
+                // Source, one convention for every field: the form decides the mode from it, and a
+                // value decoded here would reach a dual-mode field with no way to tell a reference
+                // from the text that spells it the same.
                 values.put(propertyKey, rawValue);
```
NipunaRanasinghe, line 1723: It isn't one convention yet — the keyed form's name is still decoded. `collectKeyedCapabilities` puts `entry.name()` straight into the name property, and that's the mapping key already unquoted by `WorkflowUtil.capabilityName`, while every other field here is now source.

The name field is dual-mode TEXT+EXPRESSION (`DurableAgentRegisterEventBuilder.java:104-110`, `DurableAgentHumanTaskBuilder.java:212-221`), so `seedCapabilityValue` sees a non-literal and switches it to the expression editor. The goldens in this PR show the split:

- `durable_agent_box.json` (list form): `"taskName": "\"approveItinerary\""`
- `durable_agent_task_types.json` (keyed form): `"taskName": "approveClaim"` — still bare

So opening a keyed-form human task or event — the form the module documents — puts its name in the expression editor holding a bare identifier that resolves to nothing, with a diagnostic on it. Saving is still safe only because `quoteIfPlain` quotes it whatever the mode says.

Fix it at the source so the convention really is one:

```java
values.put(fieldToPropertyKey.getOrDefault("name", "name"), WorkflowUtil.stringLiteral(entry.name()));
```

packages/ballerina-visualizer/src/views/BI/FlowDiagram/index.tsx
```diff
@@ -3916,10 +3917,12 @@ export function BIFlowDiagram(props: BIFlowDiagramProps) {
...
             // Seed the form with the existing statement's values and point it at that statement.
             const values = (capability?.values || {}) as Record<string, string>;
-            const nodeProps = node.properties as Record<string, { value: unknown }>;
+            const nodeProps = node.properties as Record<string, SeedableProperty>;
             for (const [key, value] of Object.entries(values)) {
                 if (nodeProps?.[key]) {
-                    nodeProps[key].value = value;
+                    // Both halves: the value and the mode it belongs in. The values arrive as source,
+                    // so a reference stays an expression instead of being quoted into a literal on save.
+                    seedCapabilityValue(nodeProps[key], value);
```
NipunaRanasinghe, line 3925: This seeds `node.properties[key]`, which for an approval policy is the *hidden root* property. The title, description and roles the user actually sees are the sub-properties rendered under the dropdown (`reviewSubProperties`), and `withHeldValue` (`ballerina-side-panel/src/components/editors/utils.ts:157-166`) copies only `value` into them — never `types`/`selected`. The sub-field keeps its template mode, which `buildReviewTextSubProperty` (`ActivityCallBuilder.java:1073`) sets to TEXT.

So `approvalPolicy: {title: titleVar}` on a capability still opens in the text box and still saves as `"titleVar"`. That's the same bug, on the two fields this PR hands the text box back to. I'd hold the title/description restore until the mode reaches the dynamic sub-fields.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/capabilityFieldValues.ts
```diff
@@ -0,0 +1,128 @@
...
+            text += body[i];
+            continue;
+        }
+        const escaped = body[++i];
+        switch (escaped) {
+            case "n": text += "\n"; break;
+            case "t": text += "\t"; break;
+            case "r": text += "\r"; break;
+            default: text += escaped; break;
+        }
```
NipunaRanasinghe, line 99: This decoder already exists in Java — `WorkflowUtil.stringLiteralText` plus `appendEscaped` — and it handles `\u{...}` (tested at `WorkflowUtilLiteralTest.java:138`). This copy doesn't: the `default` branch drops the backslash, so `"grin \u{1F600}"` becomes `grin u{1F600}` and saves back as `"grin u{1F600}"`. Same literal, two answers depending on which side decodes it, from day one.

Either add the `u` case, or have the LS send the decoded text alongside an explicit mode flag and drop the client-side decoder — it already knows the property's types when it builds the template.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1719,20 +1712,17 @@ private void collectCapabilityFields(MappingConstructorExpressionNode config, St
                 values.put(propertyKey, WorkflowUtil.stripModulePrefix(rawValue));
             } else if (ROLE_FIELDS.contains(fieldName)) {
                 // `userRoles: ()` says "only the named users decide"; the roles box stays empty for it.
-                values.put(propertyKey, nilAsBlank(stripQuotes(rawValue)));
-            } else if (TEXT_MODE_CAPABILITY_FIELDS.contains(fieldName)) {
-                values.put(propertyKey, stripQuotes(rawValue));
+                values.put(propertyKey, nilAsBlank(rawValue));
```
NipunaRanasinghe, line 1718: `nilAsBlank` turns `()` into `""`, and `""` isn't a string literal, so `seedCapabilityValue` falls through to the expression branch and switches the roles field (dual-mode TEXT+EXPRESSION, `DurableAgentHumanTaskBuilder.java:206-230`) to the expression editor. The comment says the roles box stays empty; it's now an empty *expression* box.

An early return on blank source in `seedCapabilityValue` would leave the template's mode alone.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentAddActivityBuilder.java
```diff
@@ -263,9 +263,8 @@ private void addActivityIdentityProperties() {
     // The two policy dropdowns, as the workflow Call Activity form has them: an approval before the
     // activity runs, and the engine's retry or review when it fails.
     private void addPolicyProperties() {
-        // A capability's values travel as plain strings, so its wording fields are expressions only.
         ApprovalPolicyForm.addFormProperties(this, ApprovalPolicyForm.NO_APPROVAL_VALUE,
-                ActivityCallBuilder.ReviewFormValues.empty(), false);
+                ActivityCallBuilder.ReviewFormValues.empty());
```
NipunaRanasinghe: With this and the matching change in `DurableAgentRegisterToolBuilder`, nothing passes `dualModeText = false` any more — all four call sites use the 3-arg overload. That leaves the 4-arg overloads of `ApprovalPolicyForm.addFormProperties`, `ActivityCallBuilder.reviewSubProperties` and `addHiddenReviewProperties`, plus the `buildRetrySubProperty` fallback, unreachable.

Their javadoc also now says the opposite of what this PR establishes ("A capability's values reach the form as plain strings with no mode beside them"). Worth deleting them in the same change that made them false.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/RestActivityStrategy.java
```diff
@@ -231,8 +233,9 @@ public String getDescription() {
         return STRATEGY_DESCRIPTION;
     }
 
+    // `activity:callRestAPI` forwards the body on POST, PUT, PATCH and DELETE, and calls
+    // `connection->get(path, headers)` without it — so DELETE carries a body and GET does not.
     private boolean isPayloadMethod(String method) {
-        return METHOD_POST.equalsIgnoreCase(method) || METHOD_PUT.equalsIgnoreCase(method)
-                || METHOD_PATCH.equalsIgnoreCase(method);
+        return !METHOD_GET.equalsIgnoreCase(method);
```
NipunaRanasinghe: The comment names four methods, the code says "anything but GET". Harmless while the method is a fixed dropdown, but the next method added to the module would silently get a body. Since you verified the enumeration against the module, I'd write the enumeration.

Also, the hidden `MESSAGE_KEY` property around line 133 still says "Request body payload (for POST, PUT, PATCH)" — worth updating with the visible one.

packages/ballerina-visualizer/src/views/BI/Forms/ArtifactForm/index.tsx
```diff
@@ -1104,7 +1104,10 @@ export function ArtifactForm(props: ArtifactFormProps) {
                     formFields={fieldsValues}
                     projectPath={projectPath}
                     openRecordEditor={handleOpenTypeEditor}
-                    openFormTypeEditor={handleOpenFormTypeEditor}
+                    // A host that does not allow creating types must not offer the guided editor
+                    // either; it was gated on the type helper's entry alone
+                    // (wso2/product-integrator#2079).
+                    openFormTypeEditor={allowTypeCreation ? handleOpenFormTypeEditor : undefined}
```
NipunaRanasinghe, line 1110: The gate looks right and is scoped correctly — `ArtifactForm` is the only consumer of `FormHostCapabilities.typeCreation`, and the wizard is the only host setting it false.

One question: `openRecordEditor` on line 1106 stays ungated. If the record editor can create a type in the wizard too, the hole is only half closed.

packages/bi-diagram/src/test/__snapshots__/Diagram.test.tsx.snap
```diff
@@ -1,4 +1,4 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
```
NipunaRanasinghe: This and the other two `.snap` files change only the header URL, which comes from a different local jest version. Unrelated to the PR and it'll flap against whatever CI runs — worth reverting.

packages/bi-diagram/src/visitors/NodeFactoryVisitor.ts
hasithaa: You are right on both counts: `startNodeId` is written nowhere in the repo, and `InitVisitor.getDefaultViewState()` would drop it anyway, so that branch never fired. Rewritten in 82025a45ca to key on the marker that does exist. A node that is not the synthetic canvas box gets `updateNodeLinks` unconditionally, so the in-chain statement becomes the chain's tail again; the synthetic box links only when it sits under the start pill, and then without an add button. I kept the `StartNodeModel` check rather than a bare `this.lastNodeModel` so the synthetic box still refuses an incoming link from an ordinary node.

packages/bi-diagram/src/test/durable-agent-run-links.test.ts
hasithaa: Agreed, the suite was pinning the bug. Both fixtures are rebuilt from what the LS actually emits: `agentBox` alone for the in-chain statement, `agentBox` plus `agentDeclarationCanvas` for the synthetic copy, and no hand-injected `startNodeId`. I checked the new in-chain test fails against the old logic before the fix. The private-field reads are still there because the visitor exposes no accessor, but they now read state the pipeline really produces.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Fixed exactly as suggested. `collectKeyedCapabilities` now writes `WorkflowUtil.stringLiteral(entry.name())`, and `durable_agent_task_types.json` regenerates to the quoted form, so the list and keyed goldens agree.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/index.tsx
hasithaa: Confirmed, `withHeldValue` copies only the value, so the sub-field keeps its template mode. Held back: both call sites go back to `addFormProperties(..., false)`, with a comment saying why, and the four goldens regenerate to expression-only. That also keeps the four-argument overloads you flagged in the other thread reachable and their javadoc true.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/capabilityFieldValues.ts
hasithaa: Added the `u` case, so `\\u{1F600}` decodes the way `WorkflowUtil.appendEscaped` does, with a regression test. The decoder is also single-pass now, so an escaped backslash no longer lets the next character be read as an escape. Moving the decode to the LS is the better long-term answer and I would rather do it in the follow-up that reworks the workflow node forms than widen this PR.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Right, blank source was falling into the expression branch. `seedCapabilityValue` now returns early on a blank value and leaves the template's mode alone, covered by a test.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/DurableAgentAddActivityBuilder.java
hasithaa: Resolved the other way: I reverted the two call sites to `dualModeText = false` for the reason in the sub-field thread, so the four-argument overloads and the `buildRetrySubProperty` fallback stay reachable and their javadoc is accurate again.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/RestActivityStrategy.java
hasithaa: Both done. `isPayloadMethod` now names POST, PUT, PATCH and DELETE, and the hidden `MESSAGE_KEY` property carries the same description as the visible one.

packages/ballerina-visualizer/src/views/BI/Forms/ArtifactForm/index.tsx
hasithaa: Good catch, it was half closed. The record editor's default completion is the create-a-type entry point, so `openRecordEditor` is gated the same way now, and `TypeEditor` hides the default completion when no handler is passed instead of calling an undefined one.

packages/bi-diagram/src/test/__snapshots__/Diagram.test.tsx.snap
hasithaa: Reverted all three; only the header URL differed.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/capabilityFieldValues.ts
```diff
@@ -0,0 +1,145 @@
...
+        const escaped = body[++i];
+        if (escaped === "u" && body[i + 1] === "{") {
+            // `\u{1F600}`, the only escape with a body. The language server decodes it too, and a
+            // literal must not read differently depending on which side looked at it.
+            const close = body.indexOf("}", i + 2);
+            const code = close === -1 ? NaN : Number.parseInt(body.slice(i + 2, close), 16);
+            if (close !== -1 && Number.isFinite(code)) {
+                text += String.fromCodePoint(code);
+                i = close;
+                continue;
```
NipunaRanasinghe: The happy path matches the LS now, but none of the guards came across. `WorkflowUtil.appendEscaped` rejects an invalid code point and a lone surrogate by returning 0 and leaving the escape as written; this only checks `Number.isFinite`, which is true for anything `parseInt` manages to read.

Three cases still differ:

- `"\u{110000}"` — `parseInt` gives 1114112, `Number.isFinite` passes, and `String.fromCodePoint` throws `RangeError: Invalid code point 1114112`. Nothing catches it, so seeding the form throws and the panel fails to open. Source that is mid-edit and not yet valid is exactly when this is reachable.
- `"\u{D800}"` — this produces a lone surrogate in the value; Java leaves it as written.
- `"\uabc"` (no brace) — the `default` branch drops the backslash and yields `uabc`; Java keeps the backslash.

Mirroring the Java guard covers all three:

```ts
if (close !== -1 && Number.isInteger(code) && code >= 0 && code <= 0x10ffff
        && !(code >= 0xd800 && code <= 0xdfff)) {
```

with the no-brace case falling through to `text += "\\" + escaped`.

Agreed on moving the decode to the LS later rather than widening this PR — but I'd rather this didn't ship able to throw.

packages/ballerina-side-panel/src/components/editors/TypeEditor.tsx
```diff
@@ -178,7 +178,7 @@ export function TypeEditor(props: TypeEditorProps) {
     }
 
     const handleDefaultCompletionSelect = (value: string | NodeProperties) => {
-        openRecordEditor(true, value);
+        openRecordEditor?.(true, value);
```
NipunaRanasinghe, line 182: This fix is right, but `ActionTypeEditor` has the same shape and didn't get it: line 356 calls `openRecordEditor(true, value)` unguarded from its own `handleDefaultCompletionSelect`, and its `showDefaultCompletion` (line 621) isn't gated on the handler the way line 315 here now is. Its prop is still declared required at line 97.

Nothing will flag it: `EditorFactory` passes `openRecordEditor` as optional into a required prop, and the side-panel tsconfig has `strictNullChecks: false`. And it looks reachable from the path you just gated — `ACTION_TYPE` is a service-model field type (`Value.java:127`), which is what `ArtifactForm` renders.

Worth giving `ActionTypeEditor` the same two lines.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/RestActivityStrategy.java
```diff
@@ -129,7 +131,7 @@ public void setFormProperties(NodeBuilder nodeBuilder, NodeBuilder.TemplateConte
         nodeBuilder.properties().custom()
                 .metadata()
                     .label("Message")
-                    .description("Request body payload (for POST, PUT, PATCH)")
+                    .description("Request body, sent with POST, PUT, PATCH and DELETE")
```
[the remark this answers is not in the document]
hasithaa: Confirmed and fixed in 49933998bd. The hydration path's DELETE entry mapped to an empty set, so reading back a DELETE with a body lost the field the node had written. It carries the message sub-property now, and the label and description are constants on the strategy, shared with both places that used to spell them out.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/capabilityFieldValues.ts
```diff
@@ -0,0 +1,145 @@
...
+            continue;
+        }
+        const escaped = body[++i];
+        if (escaped === "u" && body[i + 1] === "{") {
+            // `\u{1F600}`, the only escape with a body. The language server decodes it too, and a
+            // literal must not read differently depending on which side looked at it.
+            const close = body.indexOf("}", i + 2);
+            const code = close === -1 ? NaN : Number.parseInt(body.slice(i + 2, close), 16);
+            if (close !== -1 && Number.isFinite(code)) {
+                text += String.fromCodePoint(code);
```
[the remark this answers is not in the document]
hasithaa: Fixed. The decoder now rejects a non-hex body, an out-of-range code point and a lone surrogate, and leaves the escape as written in each case, which is what the language server does. Four regression tests cover a valid escape, an out-of-range one, a surrogate and a no-brace escape.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/capabilityFieldValues.ts
hasithaa: All three cases fixed, and the decoder now mirrors `appendEscaped` branch for branch rather than approximating it. An out-of-range code point, a lone surrogate and an escape the syntax does not define all keep their backslash and are left as written. Regression tests for each, plus one for the valid emoji. Agreed on the throw being the part that could not ship.

packages/ballerina-side-panel/src/components/editors/TypeEditor.tsx
hasithaa: Good catch, same shape and the same hole. The action type editor's prop is optional now, its default-completion handler is optional-called, and its default completion is hidden when no handler is passed.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/capabilityApproval.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @return true when a review stands in front of it
+ */
+export function isCapabilityApprovalGated(values: Record<string, unknown> | undefined): boolean {
+    if (!values) {
+        return false;
+    }
+    const policy = asText(values[POLICY]);
+    if (policy !== "") {
+        return policy !== NO_APPROVAL;
+    }
```
samithkavishke: **HIGH — unaddressed coderabbitai comment, quoting bug reintroduced.**

`ApprovalPolicyForm.normalize()` (Java) only strips a module prefix, never surrounding quotes, before comparing against `NoApproval`. So a capability declared as `approvalPolicy: "NoApproval"` (a valid singleton-string literal) hydrates with the quotes still in the source text, and `policy !== NO_APPROVAL` here is `true` for `'"NoApproval"' !== 'NoApproval'` — the badge/shield renders on a capability that actually requires no approval. This is the same class of bug this PR was written to fix, just newly introduced by it, and the review thread was never replied to.

```suggestion
    const policy = stripQuotes(asText(values[POLICY]));
    if (policy !== "") {
        return policy !== NO_APPROVAL;
    }
```

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/capabilityApproval.ts
```diff
@@ -0,0 +1,52 @@
...
+    // A declaration read by an older analysis, or one not yet migrated: the flag gated it unless it
+    // said so itself.
+    const legacy = asText(values[LEGACY_FLAG]);
+    return legacy !== "" && legacy !== "false";
+}
+
+function asText(value: unknown): string {
+    return typeof value === "string" ? value.trim() : "";
```
samithkavishke, line 54: Add the helper the suggestion above relies on, so a quoted literal (`"NoApproval"`) normalizes the same as the bare identifier before comparison:

```suggestion
function asText(value: unknown): string {
    return typeof value === "string" ? value.trim() : "";
}

function stripQuotes(value: string): string {
    return value.length >= 2 && value.startsWith('"') && value.endsWith('"')
        ? value.slice(1, -1)
        : value;
}
```

packages/ballerina-visualizer/src/views/BI/FlowDiagram/capabilityFieldValues.ts
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+    return !escaped;
+}
+
+/**
+ * The text a string literal denotes, with the escapes it carries resolved. Read in one pass: an
+ * escaped backslash consumes the character after it, so `"C:\\new"` is a path and not a line break.
+ */
+function literalText(source: string): string {
```
samithkavishke, line 66: **LOW — hand-ported duplicate decoder, no shared test corpus.**

This literal/escape decoder mirrors Java's `WorkflowUtil.stringLiteralText`/`appendEscaped` by hand. It took two review rounds to converge the two on escape ordering and `\u{...}` validation, and nothing ties them together — a future change to the Java decoder won't be caught until a user hits a mismatch. You already called this a bandaid and flagged the LS-side decode as the real fix for a follow-up, so this is just a marker to make sure that follow-up gets filed/linked rather than lost.

packages/ballerina-side-panel/src/components/editors/TypeEditor.tsx
```diff
@@ -178,7 +178,7 @@ export function TypeEditor(props: TypeEditorProps) {
     }
 
     const handleDefaultCompletionSelect = (value: string | NodeProperties) => {
```
samithkavishke, line 181: **LOW — duplicated gating fix, same miss-one-of-two risk this PR exists to close.**

The optional-call `openRecordEditor` / `showDefaultCompletion` gating fix here is byte-identical to `ActionTypeEditor.tsx:355-358`. Worth a shared helper (or hook) between the two editors — this PR itself exists because an earlier fix to one of these two files was missed in its sibling; duplicating the fix again just resets the clock on that same failure mode.

LakshanWeerasinghe (dismissed):

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/capabilityApproval.ts
hasithaa: The quoted case cannot arise: `NoApproval` is `public const NoApproval = ();` in the module, not a string, so `approvalPolicy: "NoApproval"` would not compile. But the finding points at a real one just beside it. The constant is nil, so a declaration may carry `approvalPolicy: ()`, and both this shield and the Java normalizer compared only against the name, putting a badge on a capability with no gate. Fixed in 801c72c1f7: both accept the nil, and the shield strips a module prefix too, with a guard so a record literal's own colons are not read as one. Three tests cover the nil, the prefixed constant and the record literal.

packages/bi-diagram/src/components/nodes/DurableAgentRunNode/capabilityApproval.ts
hasithaa: Not needed, and it would be misleading: the value is never a quoted string, because the constant is nil. The normalization the value does need, the nil spelling and the module prefix, is in 801c72c1f7 with tests.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/capabilityFieldValues.ts
hasithaa: Agreed, and filed: wso2/product-integrator#2497 tracks moving the decode to the language server, which already knows each property's declared types when it builds the template, so it can send the decoded text with an explicit mode and the client decoder can go.

packages/ballerina-side-panel/src/components/editors/TypeEditor.tsx
hasithaa: Fair, and exactly the failure mode. Both editors now call `canOfferTypeCreation` from a shared module, with its own tests, so the rule lives in one place.

NipunaRanasinghe (approved):
LGTM

PR #1124, opened 2026-09-18
Fix null properties when sorting module flow nodes

LakshanWeerasinghe:
## Purpose
Fix `flowDesignService/getModuleNodes` failing for projects containing a module-level durable agent and HTTP clients, so module connections and variables can be discovered successfully.

Fixes https://github.com/wso2/product-integrator/issues/2481

## Root cause
The durable-agent declaration handler in `CodeAnalyzer` emits an `EVENT_START` node before the agent box. The start node has no property map. Module-node discovery reuses this analyzer and takes its first generated node, allowing the start node into the lists that are sorted by variable name.

The comparator evaluated `node.properties().get(Property.VARIABLE_KEY)` before `Optional.ofNullable` could handle the result, causing a null-pointer exception when `properties()` was null.

## Approach
Use `FlowNode.getProperty`, which safely handles a missing property map, and handle missing variable properties or values with an empty sorting key. Reuse the same comparator for flow-model connections.

## Automation tests
Added an integration fixture to the existing data-driven `ModuleNodesTest` suite, without adding Java test code. The minimal Ballerina project includes a durable agent, model provider, activity, two module-level HTTP clients, and another module variable to exercise sorting. The JSON config checks the complete returned flow model, including client node kinds and ordering.

## Recording

https://github.com/user-attachments/assets/82746cc1-2787-41ef-a161-34dbd03f6ae5

## Summary

- Improve `getModuleNodes` stability when a module-level durable agent produces a node without properties.
- Use `FlowNode.getProperty` with an empty-string fallback when sorting nodes by variable name.
- Reuse the comparator for flow-model connections.
- Add an integration fixture covering durable agents, model providers, activities, HTTP clients, and module variables.
- Validate the returned flow model and node ordering through `ModuleNodesTest`.
- Test results: The existing suite passes with the provided Gradle command.


madushajg (approved):

samithkavishke (approved):
LGTM

PR #1126, opened 2026-09-18
Name the Copilot after the product it is installed in

ravindu25:
## Problem

Fixes wso2/product-integrator#1945

With only the Ballerina extension installed, every Copilot surface still said **"WSO2 Integrator Copilot"** — a product the user hasn't installed. It's also the one mode where the managed WSO2 sign-in isn't available (BYOK only), so the assistant on screen genuinely isn't the Integrator's.

## Fix

`ProductMode` in `ballerina-core` is now the single source of truth for the assistant's name:

- **`INTEGRATOR`** — the host app declares `WSO2_PRODUCT_MODE`, or `wso2.wso2-integrator` is installed alongside → "WSO2 Integrator Copilot"
- **`BALLERINA`** — otherwise → "Ballerina Copilot"

The host seeds the mode into every webview's HTML (`window.productMode`), so panels paint the right name on the first frame instead of flashing the wrong one while an RPC call resolves. `useProductMode` falls back to asking the host; `bi-diagram`, which has no rpc-client dependency, reads the seed directly.

### What changed

- **AI panel** — welcome screen (the surface in the report), login, waiting-for-login, disabled, tab title
- **Status bar, orb, mini chat, overview composer, diagram tooltips**
- **Migration** panel and notifications
- **Agent prompts and the `AGENTS.md` starter template** — so it doesn't introduce itself as a product the user doesn't have
- Error codes, auth toasts, sign-in warnings

`package.json` contributions can't vary at runtime, so the four Copilot-named strings drop the product name ("Open Copilot") rather than pick the wrong one.

## Testing

- New `productMode.test.ts` pins the resolution rules, including an extension installed mid-session
- New case in `CopilotComposer.test.tsx` asserts "Ballerina Copilot" and *not* "WSO2 Integrator Copilot" in Ballerina mode
- Suites green: extension 411, visualizer 644, core 176, bi-diagram 123
- E2E unaffected — setup always installs the Integrator vsix, so those runs stay in Integrator mode

## Summary

- Added product-mode detection for Ballerina and Integrator installations.
- Updated Copilot names, taglines, prompts, notifications, tooltips, authentication messages, and templates to use the active product name.
- Seeded product mode into webviews for correct initial rendering.
- Added hooks and helpers for product-specific assistant labels.
- Added tests for product-mode resolution and Ballerina branding.


packages/ballerina-visualizer/src/hooks/useProductMode.ts
```diff
@@ -0,0 +1,72 @@
...
+            inFlight = undefined;
+            return ProductMode.INTEGRATOR;
+        });
+    return inFlight;
+}
+
+export function useProductMode(): ProductMode {
+    const { rpcClient } = useRpcContext();
+    const resolved = seededProductMode() ?? cached;
+    const [mode, setMode] = useState<ProductMode>(resolved ?? ProductMode.INTEGRATOR);
```
madushajg, line 46: When a webview has no seed yet, the first render shows "WSO2 Integrator Copilot" until the async askHost RPC resolves and flips it. The unseeded fallback should default to `ProductMode.BALLERINA`, matching `getProductMode()`'s fail-safe default in `config.ts` (`isInWI() ? INTEGRATOR : BALLERINA`).

packages/ballerina-core/src/state-machine-types.ts
```diff
@@ -25,6 +25,47 @@ import { DIRECTORY_MAP, ProjectStructureArtifactResponse, ProjectStructureRespon
...
+    if (typeof window === 'undefined') {
+        return undefined;
+    }
+    const seed = (window as unknown as { productMode?: string }).productMode;
+    return seed === ProductMode.BALLERINA || seed === ProductMode.INTEGRATOR ? seed : undefined;
+}
+
+/** The assistant's name for a webview with no RPC client of its own. */
+export function webviewAssistantName(): string {
+    return assistantName(seededProductMode() ?? ProductMode.INTEGRATOR);
```
madushajg, line 66: This is called directly by `bi-diagram`'s three files, which have no RPC fallback of their own. If `window.productMode` is missing or malformed. e.g. an older host, or a webview left open across the update that introduces seeding — this silently defaults to "WSO2 Integrator Copilot" for a Ballerina-only user, with no retry path to correct it later.

ravindu25: fixed

packages/ballerina-visualizer/src/hooks/useProductMode.ts
ravindu25: fixed

madushajg (approved):

SasinduDilshara (approved):

PR #1127, opened 2026-09-18
Add the 0.10 context functions, and wrap long node names

hasithaa:
Closes wso2/product-integrator#2496

Three gaps in the workflow node panel, against the released 0.10 module.

**Long node names wrap instead of being held on one line.** Two columns of a side panel are too narrow for a name like `Send Data to Child Workflow`, so the row cut it after one line. It now wraps onto a second line and the row grows to hold it. A name still longer than that is clipped at two lines, and the row's tooltip carries it in full, as its heading above the description. The measure-and-shrink hook that used to force one line is gone.

**The two 0.10 context functions are in the palette.** `lastHumanTaskCompletion` and `lastReviewDecision` read the most recent human task or review, or a named one. Both are in the Workflow Functions group, both take an optional task name, and both bind a nilable record. The task name carries the expression editor beside its text box, because a call may name the task with a constant rather than a literal.

**Every context function reads back as its own node.** None of them did: opening a workflow that already called `currentTime`, `isReplaying`, `getWorkflowId` or `getWorkflowType` rendered a plain method call rather than the node the palette wrote. All six hydrate now, with the name they bind to and, where the call has one, the task name. A call written as an assignment is written back as an assignment, not as a second declaration of the same variable. A call whose result is not bound stays a generic node, since the form's one field is the name.

**The variable field had no selected mode**, which renders blank. It is selected explicitly, and one helper builds it for both the template and the hydrated node so the two cannot drift.

## Tests

- A node template case for each of the six context functions.
- A source generation case for each, plus the three shapes that round-trip differently: an assignment, a task name held as an expression, and one carrying a tab.
- A diagram fixture that reads all six back from source, including a constant task name, an assignment and a wildcard binding.
- Two palette cases: a long name wraps, and the row carries no second native tooltip.

Full language server suite green locally, plus the side panel and diagram suites and both type checks.

## Summary

- Added `lastHumanTaskCompletion` and `lastReviewDecision` to the Workflow Functions palette.
- Hydrated all six workflow context functions into their corresponding nodes.
- Preserved variable names, task names, assignments, and expression-based task names during hydration and source generation.
- Updated long node names to wrap across two lines and show full names in tooltips.
- Added node template, source generation, hydration, diagram, palette, and long-name test coverage.


packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1042,6 +1043,49 @@ private void populateSleepNodeProperties(MethodCallExpressionNode callNode, Func
...
+                .codedata()
+                    .node(spec.kind())
+                    .org(WORKFLOW_ORG)
+                    .module(WORKFLOW_MODULE)
+                    .object(CONTEXT_CLASS_NAME)
+                    .symbol(spec.methodName());
+
+        String variableName = this.typedBindingPatternNode == null ? spec.defaultVariableName()
+                : this.typedBindingPatternNode.bindingPattern().toSourceCode().strip();
+        WorkflowContextFunctionBuilder.addVariableProperty(nodeBuilder, variableName);
```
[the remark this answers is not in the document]
hasithaa: Confirmed and fixed in 91b322b87b. `visit(AssignmentStatementNode)` leaves the binding pattern unset, so an assigned call read back with the builder's default name and a later save would have renamed the variable. The name now comes off the assignment's left-hand side when there is no pattern, stopping at the enclosing statement so it cannot wander further up. The hydration fixture gained an assignment bound to a name other than the default, which the previous fixture could not have told apart.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1042,6 +1043,49 @@ private void populateSleepNodeProperties(MethodCallExpressionNode callNode, Func
...
+    // The unquoted text of a call argument, so a task name reaches its string field as text.
+    private static String argumentValue(FunctionArgumentNode argument) {
+        ExpressionNode expression = argument instanceof PositionalArgumentNode positional
+                ? positional.expression()
+                : (argument instanceof NamedArgumentNode named ? named.expression() : null);
+        if (expression == null) {
+            return "";
+        }
+        String literal = stringLiteralValue(expression);
+        return literal == null ? expression.toSourceCode().trim() : literal;
```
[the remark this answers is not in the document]
hasithaa: Fixed the same way. The argument is decoded with `WorkflowUtil.stringLiteralText` when it is a string literal, and left as source otherwise, so a name and the form hold the same text. The fixture now carries a task name with a tab in it.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1042,6 +1043,66 @@ private void populateSleepNodeProperties(MethodCallExpressionNode callNode, Func
...
+        }
+        for (Node parent = callNode.parent(); parent != null; parent = parent.parent()) {
+            if (parent instanceof AssignmentStatementNode assignment) {
+                return CommonUtils.getVariableName(assignment.varRef());
+            }
+            if (parent instanceof StatementNode) {
+                break;
+            }
+        }
+        return spec.defaultVariableName();
```
NipunaRanasinghe: Reading an assignment back works, but saving it again writes a declaration, so the variable ends up declared twice.

`toSource` always emits `<type> <name> = ctx.<method>(...)`, and the node's line range is the whole assignment statement (`endNode(assignmentStatementNode)`), which `SourceBuilder` uses as the edit range. So for the fixture this PR adds:

```ballerina
workflow:ReviewDecisionRecord? secondLook = ();
secondLook = ctx.lastReviewDecision("review\ttwo");
```

opening that node and saving replaces line 2 with `workflow:ReviewDecisionRecord? secondLook = ctx.lastReviewDecision("review\ttwo");` — a redeclaration of `secondLook`, which won't compile. There's no `to_source` fixture for an assignment-bound node, so nothing catches this.

A bare call statement (`ctx.currentTime();`, nothing bound) has the same problem from the other direction: `boundName` falls back to the default name and a save quietly introduces a new variable.

Either record on the node whether the source was a declaration or an assignment and branch in `toSource`, or don't hydrate assignments and expression statements into this node kind at all and let them stay generic nodes.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/WorkflowContextFunctionBuilder.java
```diff
@@ -111,7 +185,15 @@ public Map<Path, List<TextEdit>> toSource(SourceBuilder sourceBuilder) {
...
                 .keyword(SyntaxKind.DOT_TOKEN)
                 .name(spec.methodName())
-                .keyword(SyntaxKind.OPEN_PAREN_TOKEN)
+                .keyword(SyntaxKind.OPEN_PAREN_TOKEN);
+        if (spec.takesTaskName()) {
+            sourceBuilder.getProperty(Workflow.CONTEXT_TASK_NAME_KEY)
+                    .map(property -> property.value() == null ? "" : property.value().toString().trim())
+                    .filter(taskName -> !taskName.isBlank())
+                    .ifPresent(taskName -> sourceBuilder.token()
+                            .name(WorkflowUtil.quoteIfPlain(taskName)));
```
NipunaRanasinghe: `quoteIfPlain` is the wrong helper now that the read side decodes escapes.

Two ways it bites:

1. `argumentValue` keeps a non-literal argument as raw source, so `ctx.lastReviewDecision(TASK_NAME)` stores `TASK_NAME` in the field. Saving runs it through `quoteIfPlain` and writes `ctx.lastReviewDecision("TASK_NAME")` — the constant reference silently becomes a string.
2. `"\"quoted\""` in source decodes to `"quoted"` (quotes included). `quoteIfPlain` sees a value that starts and ends with a quote and passes it straight through, so the emitted argument is `("quoted")` and the inner quotes are gone.

The stored value is always decoded text, so the write side wants `WorkflowUtil.stringLiteral(...)`, which escapes unconditionally. That also handles a task name starting with `[`, which `quoteIfPlain` currently emits unquoted.

If a non-literal argument is meant to be supported, the field needs an expression mode; otherwise hydration should leave those calls as generic method-call nodes rather than putting an expression in a string field.

packages/ballerina-side-panel/src/components/NodeList/index.tsx
```diff
@@ -678,15 +682,7 @@ export function NodeList(props: NodeListProps) {
...
-                                        <S.ComponentTitle
-                                            ref={(el) => {
-                                                if (el && el.scrollWidth > el.clientWidth) {
-                                                    el.style.fontSize = "13px";
-                                                    el.style.wordBreak = "break-word";
-                                                    el.style.whiteSpace = "nowrap";
-                                                }
-                                            }}
-                                        >
+                                        <S.ComponentTitle title={node.label}>
```
NipunaRanasinghe: This row is already wrapped in a `<Tooltip content={renderTooltipContent(node.description)}>` a few lines up, so any node with a description now shows two tooltips on the same hover: the styled one with the description, and the browser's native one with the label. Better to fold the label into the `Tooltip` content (as a heading above the description) than to add a second, unstyled tooltip.

NipunaRanasinghe (commented):
Three things, one of them worth fixing before merge.

The main one is that a context function call read back from an assignment can't be written back — `toSource` only knows how to emit a declaration, so saving the node redeclares the variable. The fixture this PR adds is exactly that case.

The other two are the task-name quoting on the write side, and the duplicate tooltip in the side panel.

The rest looks good: the specs moved to the parent and keyed by method name is the right shape, the lookup is properly gated on `Context` plus the workflow module, and all six node kinds are propagated everywhere the existing ones appear (`bi.ts`, `NodeKind`, the builder registry, icons and colour groups). `workflow2.json` is the only available-nodes fixture touching this group, so nothing else goes stale.

packages/ballerina-side-panel/src/components/NodeList/index.tsx
```diff
@@ -191,14 +191,18 @@ namespace S {
         }
     `;
 
+    // Long node names wrap onto a second line instead of being cut: two columns of a side
+    // panel are too narrow for names like "Send Data to Child Workflow" on one line.
     export const ComponentTitle = styled.div`
-        white-space: nowrap;
         flex: 1;
         min-width: 0;
+        font-size: 13px;
```
NipunaRanasinghe: The ref hook this replaces set 13px only when the label actually overflowed. Here it's unconditional, so every node label in the palette drops from the row's 14px to 13px, not just the long ones. That's a global restyle riding along with a clipping fix — worth calling out in the description if it's intended, otherwise drop the declaration and let it inherit.

packages/ballerina-side-panel/src/components/NodeList/index.tsx
```diff
@@ -191,14 +191,18 @@ namespace S {
...
+        line-height: 1.25;
         overflow: hidden;
-        text-overflow: ellipsis;
-        display: block;
-        word-break: break-word;
+        overflow-wrap: anywhere;
+        display: -webkit-box;
+        -webkit-line-clamp: 2;
```
NipunaRanasinghe, line 203: Clamping at two lines still clips: a name that needs three lines gets an ellipsis. So the PR title's "stop cutting long node names" is really "cut at two lines instead of one, with the full name in a tooltip". Fine as behaviour, but the description reads as if nothing is cut any more.

NipunaRanasinghe (commented):
Two follow-ups on the side panel styling, both minor.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Right on both counts, and the round trip was the half I had not checked. Fixed in f2f3ff7876 by the first of your two options: the node carries a hidden flag saying its source assigned an existing variable, and `toSource` then omits the type and writes `name = ctx.<method>(...)`. There is a source-generation fixture for it now, which is what would have caught this. The unbound case takes the second option: a call whose result is not bound is no longer claimed at all, because the form's only field is the name and there would be none to show. In practice the shape that reaches this is `_ = ctx.isReplaying();`, a wildcard assignment, and the fixture covers it.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/WorkflowContextFunctionBuilder.java
hasithaa: Both real, thank you. The write side uses `WorkflowUtil.stringLiteral` now, so decoded text is escaped back unconditionally, with a fixture for a name carrying a tab. And a non-literal argument is meant to be supported, so the field carries the expression editor beside the text box: hydration selects the mode from what the source held, and the write side keeps an expression as written. A fixture covers `ctx.lastReviewDecision(REVIEW_TASK)` going out unquoted.

packages/ballerina-side-panel/src/components/NodeList/index.tsx
hasithaa: Agreed, two tooltips on one hover. The name is the tooltip's heading now, above the description, and the native one is gone. That also means the tooltip is always present, where before a node with no description had none, so a name too long even for two lines still has somewhere to be read in full. A palette test asserts no row carries a native title.

packages/ballerina-side-panel/src/components/NodeList/index.tsx
hasithaa: Not intended, and dropped. The declaration is gone and the name inherits the row's 14px, so only the wrapping changes.

packages/ballerina-side-panel/src/components/NodeList/index.tsx
hasithaa: Fair, the title was overstating it. The pull request is now "Add the 0.10 context functions, and wrap long node names", and the description says a name longer than two lines is still clipped, with the tooltip carrying it in full.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/WorkflowContextFunctionBuilder.java
```diff
@@ -111,7 +224,19 @@ public Map<Path, List<TextEdit>> toSource(SourceBuilder sourceBuilder) {
                 .name(ctxParamName)
                 .keyword(SyntaxKind.DOT_TOKEN)
                 .name(spec.methodName())
-                .keyword(SyntaxKind.OPEN_PAREN_TOKEN)
+                .keyword(SyntaxKind.OPEN_PAREN_TOKEN);
+        if (spec.takesTaskName()) {
+            sourceBuilder.getProperty(Workflow.CONTEXT_TASK_NAME_KEY)
+                    .filter(property -> property.value() != null && !property.value().toString().isBlank())
+                    .ifPresent(property -> {
+                        String taskName = property.value().toString().trim();
```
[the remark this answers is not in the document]
hasithaa: Correct, the spaces are part of the name. Fixed in 85009d4297: the text box's value goes to `stringLiteral` as it stands, so ` review me ` stays that. The trim is kept for the expression mode, where surrounding space carries nothing. There is a source-generation fixture for the spaced name.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/WorkflowContextFunctionBuilder.java
```diff
@@ -76,18 +129,72 @@ public void setConcreteTemplateData(TemplateContext context) {
...
+                .editable(true)
+                .optional(true)
+                .stepOut()
+                .addProperty(Workflow.CONTEXT_TASK_NAME_KEY);
+    }
+
+    // Whether the call was written as an assignment to a variable that already exists, rather than
+    // a declaration. Hidden: the form has nothing to ask, but a save has to write back the same
+    // statement shape or the variable would be declared a second time.
+    public static void addAssignmentProperty(NodeBuilder nodeBuilder) {
```
NipunaRanasinghe, line 183: Non-blocking: when this flag is set, the Variable Name field is still editable. Renaming it on `secondLook = ctx.lastReviewDecision(...)` writes `newName = ctx.lastReviewDecision(...)` against a variable that was never declared, which won't compile. Same for the `_` row. Making the name non-editable when the source assigned an existing variable would match what `HumanTaskBuilder.hide()` does for the same reason.

packages/ballerina-side-panel/src/components/NodeList/index.tsx
```diff
@@ -636,15 +644,16 @@ export function NodeList(props: NodeListProps) {
         }
     }
     
-    const renderTooltipContent = (description?: string): React.ReactNode | undefined => {
+    // The name heads the tooltip so a name too long for its two lines can still be read in full.
+    // It is the whole tooltip when the node carries no description, which is why this never
+    // returns nothing.
+    const renderTooltipContent = (label: string, description?: string): React.ReactNode => {
```
NipunaRanasinghe: Non-blocking: this never returns undefined now, so every palette row gets a tooltip — including nodes with no description, where it just repeats the label already visible in the row. Those rows had none before. Worth a thought, though I don't have a clean suggestion: whether the name is actually clipped isn't knowable without measuring, which is what this PR is getting rid of.

NipunaRanasinghe (dismissed):
Approving. All five points from the earlier rounds are properly fixed.

Two non-blocking notes inline. Also low, and yours to judge: `string|error id = ctx.getWorkflowId();` (legal, no `check`) hydrates and re-saves as `string id = check ctx.getWorkflowId();`, because `toSource` takes `check` from the spec rather than from what the source had.

packages/ballerina-side-panel/src/components/NodeList/index.tsx
hasithaa: Reopening this one: my fix for it broke the end-to-end suite, and I have taken a different route in a7b4a9bc6d.

The toolkit's tooltip portals its content into the document and only toggles visibility, so making the name the tooltip's heading put every node name in the document twice. The helper that opens the node panel matches a node by its name exactly, so it stopped resolving to one element and the panel check never passed. That is what failed group 1 on the last two runs.

The row now carries one tooltip and no more, which was your point: the styled one when the node describes itself, the browser's own with the name when it does not. Nothing is duplicated, and a name clipped at two lines is still readable in the case where there is no description to read instead. Two palette cases hold it, one per shape, each asserting the name appears once in the document.

NipunaRanasinghe (approved):

PR #1130, opened 2026-09-19
[Agent Builder] Fix remaining spots that treat agent tool functions as plain FUNCTION artifacts

dan-niles:
## Summary

This PR fixes operation timeout issues that occur when saving agent tool configuration.

`AGENT_TOOL` is a separate artifact kind from `FUNCTION` (added in #1121), but two more places still assumed `FUNCTION` and needed the same fix as `AgentToolForm`'s save path:

- `findViewByArtifact()` had no `AGENT_TOOL` case, so clicking into a tool function's code (or navigating from the topology diagram) fell through to `null` and bounced to Package Overview instead of the flow diagram.
- `getArtifactDataFromNodeKind()` had no case for the `AGENT_TOOL` node kind used when creating a brand-new tool, so the artifact-notification wait was skipped entirely instead of matching the right notification.


packages/ballerina-visualizer/src/views/BI/AIChatAgent/AgentToolForm.tsx
```diff
@@ -744,9 +744,13 @@ export function AgentToolForm(props: AgentToolFormProps): JSX.Element {
...
                     filePath, flowNode: updatedNode, isFunctionNodeUpdate: true,
+                    artifactData: { artifactType: DIRECTORY_MAP.AGENT_TOOL },
                 });
             } else {
-                response = await rpcClient.getBIDiagramRpcClient().getSourceCode({ filePath, flowNode: updatedNode });
+                response = await rpcClient.getBIDiagramRpcClient().getSourceCode({
+                    filePath, flowNode: updatedNode,
+                    artifactData: { artifactType: DIRECTORY_MAP.AGENT_TOOL },
+                });
             }
```
kaumini: `artifactType: AGENT_TOOL` is wrong here when `hostClass` is set.

The Agent Definition designer renders this form with a host class for the "Custom" tool panel (`AgentDefinitionDesigner.tsx:1479`), so `codedata.data.hostClassName` is set (line 542 above) and `AgentToolBuilder.relocateToolIntoClass` writes the tool as a method **inside** the agent class. `ModuleNodeTransformer.transform(ClassDefinitionNode)` then publishes the class as `AGENT_DEFINITION` with the tool only as a nested child, and `getEntryValue` keeps `AGENT_TOOL` children out of the entry's `resources` — so the `data.type === "AGENT_TOOL"` filter in `ArtifactNotificationHandler.subscribe` matches nothing.

Both waits in `updateSourceCode` then run out: ~10s on the raw-edit wait, then ~10s more before the second one rejects with "Operation timed out", navigates the webview to Package Overview, and leaves `getSourceCode` returning `{ artifacts: [] }` — which makes this function throw "returned no artifacts" and show "Failed to create the agent tool", even though the source edit was applied. Before this PR `artifactData` was `undefined` on this path, so any notification resolved the wait. The new `case 'AGENT_TOOL'` in `rpc-manager.ts` has the same blind spot as a fallback for any AGENT_TOOL node carrying `hostClassName`.

`AGENT_DEFINITION` is what the sibling in-class paths already send (`saveAgentNode`, and the `artifactData` prop passed to `UseAgentToolForm`). With that fixed the two branches differ only by `isFunctionNodeUpdate`, so they collapse into one call:

```suggestion
            }
            // A class-hosted tool is relocated into the agent class by AgentToolBuilder, so the
            // artifact that publishes is the class itself, not a standalone AGENT_TOOL.
            response = await rpcClient.getBIDiagramRpcClient().getSourceCode({
                filePath,
                flowNode: updatedNode,
                ...(isEdit ? { isFunctionNodeUpdate: true } : {}),
                artifactData: {
                    artifactType: hostClass ? DIRECTORY_MAP.AGENT_DEFINITION : DIRECTORY_MAP.AGENT_TOOL,
                },
            });
```

(The leading `}` closes the `if (isEdit) {` block.) The edit path is unaffected either way — in-class edits return early through `handleClassEditSubmit` and `AgentToolFormView` never passes `hostClass`, so `isEdit` always resolves to `AGENT_TOOL`.

PR #1134, opened 2026-09-21
Correct the E2E tests rerun logic

samithkavishke:

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

- Corrects E2E reruns to target failed `test.describe.serial()` blocks.
- Runs reruns only through the manual “Re-run failed jobs” path.
- Generates valid `--grep` patterns and uses the matrix group as a fallback.
- Counts tests as flaky only when an earlier attempt failed.
- Updates related workflow documentation.


.github/scripts/aggregate-e2e-results.js
```diff
@@ -193,8 +193,9 @@ function aggregate(rootDir, expectedGroups) {
       else if (isSkipped) skipped += 1;
       else failed += 1;
       // Flakiness is derived from the merged attempt history rather than either report
-      // file's own test.status: a test re-run via --last-failed spans two separate
-      // Playwright invocations, so no single file's status reflects the merged outcome.
+      // file's own test.status: a test re-run as part of its failing serial block spans
+      // two separate Playwright invocations, so no single file's status reflects the
+      // merged outcome.
       if (isPassed && attempts > 1) flaky += 1;
```
NipunaRanasinghe: **Blocking: this now over-counts flaky tests.**

`isPassed && attempts > 1` only means "flaky" if the re-run report contains *just* the tests that failed. That held for `--last-failed`; it no longer holds once a whole `test.describe.serial()` block is replayed. Every previously-passing test in a replayed block gets a second attempt merged in and is counted flaky.

Reproduced against this branch with a 3-test block where only the third test failed and all three passed on the re-run:

```
Total: 3 - Passed: 3 - Failed: 0 - Skipped: 0 - Flaky (passed after retry): 3
```

`--last-failed` would have reported `Flaky: 1`. The same inflation lands in the NDJSON history rows (`attempts: 2` for tests that never failed), which is the trend data the 6-hourly scheduled run collects — see "Scheduled E2E testing" in the workflows README.

The comment right above is updated to describe the new re-run shape but the accounting underneath it isn't. A fix needs the merge to distinguish a real retry from block context: e.g. carry the source report file onto each result in `mergeGroupReports` and only count flaky when an earlier attempt *for that same test* actually failed, rather than inferring it from `attempts > 1`.

.github/actions/run-e2e-group/action.yml
```diff
@@ -94,11 +95,11 @@ runs:
...
-    # happens on a manual "Re-run failed jobs" (check-results success), which is the
-    # resume path it exists for.
+    # timeout, a second pass rerunning failed serial blocks over platform failures can
+    # spend what is left of the 60-minute budget and get the leg killed before it uploads
+    # anything, which is the same cost BI_E2E_RETRIES: 0 removes from the first attempt.
+    # The re-run still happens on a manual "Re-run failed jobs" (check-results success),
+    # which is the resume path it exists for.
     - name: Run Failed Ballerina e2e Test (Re-run)
       shell: bash
       if: (steps.run-tests-first.outcome == 'failure' && runner.os != 'Windows') || steps.check-results.outcome == 'success'
```
NipunaRanasinghe: **Worth a deliberate decision: the automatic Linux pass is now much more expensive than it was.**

`extract-failed-serial-groups.js`'s own header notes that Playwright's `retries` already replays a whole serial block from the top. Linux runs with `retries: 2` (`BI_E2E_RETRIES` is only forced to `0` on Windows, in `e2e-scheduled.yml`), so by the time a test is reported failed its block has already run three times inside the first attempt. This step adds a fourth full replay — and with `maxFailures: 10` that can be up to ten distinct blocks, where `--last-failed` was bounded by ten individual tests.

The job timeout is still 60 minutes and the README notes that a leg killed by it uploads no artifact at all, so the group disappears from the report rather than failing cleanly. The Windows rationale in the comment above ("a second pass ... can spend what is left of the 60-minute budget and get the leg killed before it uploads anything") applies more strongly to Linux after this change than it did before it.

The manual "Re-run failed jobs" resume path is where this change is clearly a win — there is no in-run retry history to lean on there. Is the automatic Linux pass still worth keeping, given retries already cover it?

.github/scripts/extract-failed-serial-groups.js
```diff
@@ -0,0 +1,86 @@
...
+
+  if (titles.size === 0) {
+    console.error(`No failing test found in ${reportFile}; caller should fall back to the whole group.`);
+    return;
+  }
+
+  process.stdout.write([...titles].map(escapeRegExp).join('|'));
+}
+
+main();
```
NipunaRanasinghe, line 100: **The documented fallback doesn't hold if this throws.**

GitHub's `shell: bash` runs with `-eo pipefail`, so `FAILED_GROUPS=$(node ...)` kills the re-run step outright on any non-zero exit. Only `readFileSync`/`JSON.parse` are guarded here — `fs.readdirSync` in `findLatestReport` and the suite traversal are not. A report file containing the literal `null` parses fine and then `report.suites` on line 76 throws a `TypeError`.

In that case the step dies before running a single test, which contradicts both this file's header ("Prints nothing (and exits 0) if no report could be read") and the fallback comment in `action.yml`. Wrapping the body of `main()` in try/catch and returning empty on any error would make the stated contract true:

```js
function main() {
  try {
    // ... existing body
  } catch (err) {
    console.error(`Failed to extract failed serial groups: ${err.message}; caller should fall back to the whole group.`);
  }
}
```

.github/scripts/extract-failed-serial-groups.js
```diff
@@ -0,0 +1,86 @@
...
+    const last = results[results.length - 1];
+    return last && last.status !== 'passed' && last.status !== 'skipped';
+  });
+}
+
+// A suite whose own specs include a failing test is the nearest serial block ancestor of
+// that failure (every spec sits directly inside its file's single serial block in this
+// suite), regardless of how many describe levels sit above it.
+function collectFailingSuiteTitles(suite, out) {
+  if ((suite.specs || []).some(specFailed)) out.add(suite.title);
```
NipunaRanasinghe: Minor robustness: an empty `suite.title` would make the grep match everything.

`test.describe(fn)` (the anonymous wrappers in `test.list.ts`) serializes with `title: ''`. Nothing puts a spec directly inside one today, so this isn't reachable now — but if a `test()` ever lands outside a spec file's serial block, `escapeRegExp('')` yields `(Real Block|)`, the alternation matches every title, and the whole group re-runs silently rather than the intended block.

One-line guard: `if (suite.title && (suite.specs || []).some(specFailed)) out.add(suite.title);`

.github/scripts/extract-failed-serial-groups.js
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env node
```
NipunaRanasinghe, line 1: Nit: shebang present but the file mode is `100644`, and the only caller invokes it via `node`. Either drop the shebang or `chmod +x` it — `retry.sh` in the same directory is executable, so it's worth being consistent one way or the other.

.github/actions/run-e2e-group/action.yml
```diff
@@ -108,16 +109,29 @@ runs:
...
         TEST_GROUP: ${{ inputs.group }}
       run: |
+        # `--last-failed` would replay only the individually-failed test, skipping the
+        # earlier steps in its test.describe.serial() block that built the tree/webview/
+        # project state it depends on. Instead, find which serial block(s) had a failure
+        # in the latest report and grep for (this matrix group) AND (one of those block
+        # titles), so each failing block reruns in full, from its own beforeAll.
+        FAILED_GROUPS=$(node .github/scripts/extract-failed-serial-groups.js packages/ballerina-extension/e2e-reports)
+        if [ -n "$FAILED_GROUPS" ]; then
+          export RERUN_GREP="(?=.*$TEST_GROUP)(?=.*($FAILED_GROUPS))"
```
NipunaRanasinghe: Nit: `$TEST_GROUP` goes into the regex unescaped, while the block titles from the script are escaped. Harmless for `groupN` and already true of the first-attempt `--grep`, but it's now embedded in a larger pattern where a stray metacharacter would change the meaning of the lookahead rather than just failing to match. Cheap to normalise if the extractor also emitted the escaped group.

.github/scripts/aggregate-e2e-results.js
```diff
@@ -192,10 +192,14 @@ function aggregate(rootDir, expectedGroups) {
...
-      // file's own test.status: a test re-run via --last-failed spans two separate
-      // Playwright invocations, so no single file's status reflects the merged outcome.
-      if (isPassed && attempts > 1) flaky += 1;
+      // Flakiness means THIS test failed on an earlier attempt before finally passing —
+      // not just that it has more than one attempt. Since a whole test.describe.serial()
+      // block is replayed (not just the one failed test), a sibling test that already
+      // passed gets swept into the re-run's report too, giving it a second 'passed'
+      // result with no failure anywhere in its own history; attempts > 1 alone would
+      // wrongly count that sibling as flaky.
+      const hadEarlierFailure = test.results.slice(0, -1).some((r) => r.status !== 'passed');
```
NipunaRanasinghe: This fixes the siblings that run *before* the failure, but not the ones after it, which are the more common case.

When a test fails in a serial block, Playwright skips the rest of the block and records a real result for each of them rather than leaving the array empty. I ran a 4-test serial block failing at t2 on Playwright 1.55.1:

```
{"title":"t3 after failure","status":"skipped","results":["skipped"]}
{"title":"t4 after failure","status":"skipped","results":["skipped"]}
```

So t3 merges to `['skipped','passed']`, `'skipped' !== 'passed'` is true, and it gets counted as flaky. Feeding that through this branch's aggregate script:

```
Total: 4 - Passed: 4 - Failed: 0 - Skipped: 0 - Flaky (passed after retry): 3
```

It should be 1. On Linux with `retries: 2` the first report holds `['skipped','skipped','skipped']` for those tests, which lands the same way.

Ignoring skips as well gives the right answer (verified):

```js
const hadEarlierFailure = test.results
  .slice(0, -1)
  .some((r) => r.status !== 'passed' && r.status !== 'skipped');
```

.github/actions/run-e2e-group/action.yml
```diff
@@ -83,41 +84,54 @@ runs:
...
       run: |
+        # `--last-failed` would replay only the individually-failed test, skipping the
+        # earlier steps in its test.describe.serial() block that built the tree/webview/
+        # project state it depends on. Instead, find which serial block(s) had a failure
+        # in the latest report and grep for (this matrix group) AND (one of those block
+        # titles), so each failing block reruns in full, from its own beforeAll. The
+        # script always prints a valid --grep value (falling back to just the escaped
+        # group when it can't identify a failing block) and never throws, so this can't
+        # abort the step under bash's default -e.
+        export RERUN_GREP=$(node .github/scripts/extract-failed-serial-groups.js packages/ballerina-extension/e2e-reports "$TEST_GROUP")
```
NipunaRanasinghe: Dropping the shell-side fallback makes the bad case worse than it was.

`export` is a builtin, so its own exit status is what bash sees — the command substitution's status is thrown away and `set -e` never fires:

```
$ bash -eo pipefail -c 'export V=$(node -e "process.exit(2)"); echo "survived, V=[$V]"'
survived, V=[]          # outer exit 0

$ bash -eo pipefail -c 'V=$(node -e "process.exit(2)"); echo hi'
                        # outer exit 2
```

The script no longer throws, agreed, but it can still print nothing: `process.exit(2)` when the directory argument is missing, or node failing to start or being OOM-killed. `RERUN_GREP` is then empty, and `--grep ""` matches every test in every group (confirmed with `--list`). Against the 60-minute budget that means a timeout, and a leg killed that way uploads no artifact at all.

The comment above says the script always prints a valid `--grep` value, which covers the throw but not the empty output. Two lines put the guard back:

```bash
RERUN_GREP=$(node .github/scripts/extract-failed-serial-groups.js packages/ballerina-extension/e2e-reports "$TEST_GROUP")
export RERUN_GREP="${RERUN_GREP:-$TEST_GROUP}"
```

NipunaRanasinghe (approved):
LGTM

PR #1135, opened 2026-09-21
Surface existing-function load failures instead of hanging the loader

NipunaRanasinghe:
## Purpose
Follow-up to #1118, raised by @madushajg in https://github.com/ballerina-platform/ballerina-vscode/pull/1118#discussion_r4045742378.

`getExistingFunctionNode` has no rejection handling. If its `getFunctionNode` RPC call fails — LS not ready, transient error, timeout — `isLoading` is never cleared and the edit form is left on a bare spinner with no error and no Retry, recoverable only by closing and reopening the panel. It affects an ordinary open of the edit form for any of the seven artifact kinds this form handles.

To be precise about scope: this hang is **pre-existing**, not introduced by #1118. On the base commit, `getExistingFunctionNode` already had no `try`/`catch` and already called `setIsLoading(false)` after an unguarded await. #1118 added only the sequence claim and the `isCurrentLoad(seq)` guard before the writes.

One narrow case *was* newly reachable after #1118: a create-path load superseded by a rejecting edit-path load, where the create path's `finally` correctly defers to the newer load and the newer load then has no handler to clear `isLoading`. That is fixed here too, as a consequence.

## Goals
Give the edit path the same failure handling the create path got in #1118, so a failed load reports itself and offers Retry instead of hanging.

## Approach
`getExistingFunctionNode` now mirrors `getFunctionNode` exactly:

- `setLoadError(false)` when it claims the load, so a retry starts clean.
- `try` / `catch` / `finally` around the RPC and the post-processing.
- All three state writes gated on the existing `isCurrentLoad(seq)` predicate, so a superseded or unmounted load neither raises the error card nor clears a newer load's spinner.
- `console.log` of the loaded node moved inside the success branch.

No new state, no new component, no change to the rendered card: `loadError` and the existing error card with its Retry button are reused as-is. Retry re-fires the effect through `loadAttempt`, which routes back to `getExistingFunctionNode` when `functionName` is set.

The error card's message was already noun-free after #1118 — `Failed to load the ${formType.current.toLowerCase()}. Please try again.` — so it reads correctly from this path too, for all seven kinds. The comment above it in #1118 said "both loaders share this card"; that was aspirational when written and is now true.

Both loaders then also clear the node on failure — `setFunctionNode(undefined)`, inside the existing `isCurrentLoad(seq)` guard. `functionFields` is derived from `functionNode`, so the form unmounts instead of leaving the error card above a stale form belonging to whatever was open before. This is reachable when the parent swaps `functionName` on a mounted form and the reload fails: without it the previous artifact's fields stay on screen, editable and submittable, under an error. A superseded rejection still clears nothing, because the guard covers this write too.

The package-pulling status stays create-only (`pullsWorkflowPackage` is still gated on `!functionName`), so opening an existing artifact never claims a package is being pulled.

`FunctionForm/index.tsx` also gains a default `React` import. The shared jest preset pins `jsx: 'react'` (`packages/test-config/jest-preset.js`), so the classic runtime needs `React` in scope to render the component in a test; `TypeDiagram`, `PackageOverview` and `GraphqlServiceEditor` all import it the same way.

Most of the line count is re-indentation from wrapping the body in `try`. Under `?w=1` the component change is 19 added and 5 removed lines.

## UI Component Development
- [x] Added reusable UI components to the ui-toolkit — N/A, no new component; reuses the error card added in #1118.
- [x] Use ui-toolkit components wherever possible — no rendering change in this PR.
- [x] Matches with the native VSCode look and feel — unchanged from #1118.

## User stories
As an integration developer opening an existing function, workflow or data mapper, when the language server cannot return the definition I want to be told and offered a retry, rather than watching a spinner that never stops.

## Release note
Show an error with a Retry option when an existing function, workflow or data mapper fails to load in the editor form, instead of leaving a spinner running indefinitely.

## Documentation
N/A — an error state on an existing screen.

## Training
N/A

## Certification
N/A — error handling, no conceptual change to certify.

## Marketing
N/A

## Automation tests
- Unit tests
  > Added `packages/ballerina-visualizer/src/views/BI/FunctionForm/index.test.tsx`, the first test file for this component. Five cases: edit-path rejection, create-path rejection, a later load rejecting after a successful one (node cleared, no stale form under the card), Retry re-firing the load and rendering the form, and a superseded rejection leaving the newer load's node alone.
  >
  > It follows `views/BI/PackageOverview/index.test.tsx`: `jest.mock` of the `@wso2/ballerina-rpc-client` barrel delegating to `src/test/rpcHarness`, `createRoot` + `act`, no new test dependency. The RPC-level failure is injected by giving the fake client a `getFunctionNode` that throws. An earlier version of this description said no harness existed for that; it was wrong.
  >
  > The tests were checked against the unfixed code: reverting `setFunctionNode(undefined)` from both catch blocks fails exactly one case — "clears a previously loaded node when a later load rejects" — and leaves the other four passing, which pin the behaviour #1118 added.
- Integration tests
  > None. Reproducing it end to end still needs the language server to fail a `getFunctionNode` call.

Full `ballerina-visualizer` jest suite: 49 suites, 669 tests, all passing. `tsc --noEmit -p tsconfig.spec.json`: 11 errors, all the pre-existing `Type 'Timeout' is not assignable to type 'number'` ones in other files, unchanged by this PR. No errors in `FunctionForm`.

**Still not verified by running the extension.** The tests assert the card and Retry against a stubbed form and a fake RPC client. Nobody has seen the real error card rendered from this path in a running extension — the same gap called out in #1118.

## Security checks
- Ran FindSecurityBugs plugin and verified report? N/A — TypeScript/React change, no Java code.

## Samples
N/A

## Related PRs
- #1118 — added the pulling status, the error card and the `isCurrentLoad` guard this builds on.

## Migrations (if applicable)
N/A

## Test environment
macOS 15 (Darwin 25.5.0), Node via pnpm workspace. Jest and typecheck only — see Automation tests.

## Learning
Followed `getFunctionNode` in the same file, as restructured in #1118.

## Summary

- Improved `getExistingFunctionNode` error handling for failed RPC requests.
- Added load-sequence guards to prevent stale requests from changing state.
- Reused the existing error card and Retry behavior.
- Added `FunctionForm` tests for load failures, retries, stale requests, and create/edit paths.
- Preserved the create path and component structure.
- TypeScript verification reported only pre-existing errors in other files.


packages/ballerina-visualizer/src/views/BI/FunctionForm/index.tsx
```diff
@@ -379,61 +379,71 @@ export function FunctionForm(props: FunctionFormProps) {
...
+                .getBIDiagramRpcClient()
+                .getFunctionNode({
+                    functionName,
+                    fileName,
+                    projectPath
+                });
+            let flowNode = hideTypeDescriptionField(res.functionDefinition);
+            if (isNpFunction) {
+                /* 
+                * TODO: Remove this once the LS is updated
```
gigara, line 400: Shall we create an issue for this?

madushajg (dismissed):

packages/ballerina-visualizer/src/views/BI/FunctionForm/index.tsx
NipunaRanasinghe: Yes — done: wso2/product-integrator#2508.

Two notes on it.

This block isn't actually changed by this PR. it only appears in the diff because wrapping the body in `try` re-indented it. So I've left the code alone and tracked it separately rather than folding a cleanup into a fix PR.

The issue is scoped wider than the TODO you commented on, because the workaround is a matched pair rather than one block. In this file it's at three places: L311 in `getFunctionNode` and L398 here both fold the advanced properties into `parameters.advanceProperties` on load, and L456 unfolds them before submit. It also reaches `FunctionFormStatic` (12 occurrences), `ballerina-side-panel`'s `editors/utils.ts`, the GraphQL `OperationForm`, `node-property-utils`, and the type itself in `ballerina-core`'s `bi.ts`. Removing a load-side fold without its save-side unfold would write advanced properties back in the wrong shape, so the issue says to remove them together and lists the sites.

Nothing found in either tracker beforehand, so this isn't a duplicate.

madushajg (dismissed):

packages/ballerina-visualizer/src/views/BI/FunctionForm/index.tsx
```diff
@@ -369,6 +369,8 @@ export function FunctionForm(props: FunctionFormProps) {
             // unresolvable-version path. Surface it instead of leaving the loader up forever.
             console.error(`>>> Error fetching the ${kind} node template`, error);
             if (isCurrentLoad(seq)) {
+                // Drop the previously loaded node so a stale form cannot render under the error card.
+                setFunctionNode(undefined);
```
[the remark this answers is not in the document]
NipunaRanasinghe: Added with https://github.com/ballerina-platform/ballerina-vscode/pull/1135/commits/0694fab5b5410fb754c66a89e6b99ff9cc4836d0

madushajg (approved):

NipunaRanasinghe:
@coderabbitai full review

PR #1137, opened 2026-09-21
Teach the Copilot agent that cloning a readonly value returns it unchanged

damsith0:
Copilot seeded a mutable `map<T>` from a `readonly & map<T>` snapshot with `.clone()`. 

`clone()` does not copy immutable subtrees, so the "mutable" map held immutable members and the first in-place field update panicked at run time (wso2/product-integrator#2441).

`readonly` distributes into members, so both `map<T & readonly>` and `map<T> & readonly` hold members of type `T & readonly` even when `T` is a mutable record, and assigning one to a variable declared `T` only widens the apparent type. The compiler cannot report this, so the rule has to come from the prompt.

## Approach

Adds a "Copying values" section to DATA_BINDING_CODING_RULES, already interpolated into the system prompt by getSystemPrompt().

## Summary

- Added a **“Copying values”** section to `DATA_BINDING_CODING_RULES`.
- Documented `readonly` distribution and `clone()` behavior.
- Explained when cloned values can retain immutable members.
- Added guidance to use `cloneWithType()` with a non-readonly target for mutable copies.


packages/ballerina-extension/src/features/ai/agent/data-binding-rules.ts
```diff
@@ -48,6 +54,11 @@ export const DATA_BINDING_CODING_RULES = `## Data binding, type casts and narrow
 - Do not guess a type and cast to it. Check first with \`if x is T { ... }\` (inside the block \`x\` is a \`T\`), or convert with \`T v = check x.ensureType();\`; both return an error instead of panicking.
 - To turn \`json\` into a record use \`T v = check x.cloneWithType();\` — never \`<T>jsonValue\`.
 
+### Copying values
+- \`.clone()\` does NOT copy immutable subtrees: on an already-immutable value it returns that SAME value, still immutable.
+- \`readonly\` distributes into members, so both \`map<T & readonly>\` and \`map<T> & readonly\` hold members of type \`T & readonly\` even when \`T\` is a mutable record. Judge by that type, not by the variable you assign to: \`T? v = m[k].clone();\` is still immutable, type-checks with no diagnostic, and panics on the first in-place update with \`{ballerina/lang.map}InvalidUpdate\`.
```
SasinduDilshara: Lets make this generic

packages/ballerina-extension/src/features/ai/utils/libs/langlibs.ts
```diff
@@ -76,7 +76,7 @@ json result = person.toJson();
 string jsonText = person.toJsonString();
 \`\`\`
 
-When you need to copy data, use clone() to make a modifiable copy:
+When you need to copy data, use clone() to make a modifiable copy (only for a mutable value — on an already-readonly value clone() returns the same immutable value, so use cloneWithType() instead):
```
SasinduDilshara: Lets add two points with conditions

packages/ballerina-extension/src/features/ai/agent/data-binding-rules.ts
```diff
@@ -48,6 +54,11 @@ export const DATA_BINDING_CODING_RULES = `## Data binding, type casts and narrow
 - Do not guess a type and cast to it. Check first with \`if x is T { ... }\` (inside the block \`x\` is a \`T\`), or convert with \`T v = check x.ensureType();\`; both return an error instead of panicking.
 - To turn \`json\` into a record use \`T v = check x.cloneWithType();\` — never \`<T>jsonValue\`.
 
+### Copying values
+- \`.clone()\` does NOT copy immutable subtrees: on an already-immutable value it returns that SAME value, still immutable.
```
SasinduDilshara: Lets say for mutable values it copy

packages/ballerina-extension/src/features/ai/agent/data-binding-rules.ts
```diff
@@ -48,6 +54,11 @@ export const DATA_BINDING_CODING_RULES = `## Data binding, type casts and narrow
...
+- \`readonly\` distributes into members: whenever a container OR its member type is \`readonly\`, the members have type \`T & readonly\` even when \`T\` is a mutable record — \`T[] & readonly\` and \`(T & readonly)[]\`, \`map<T> & readonly\` and \`map<T & readonly>\`, and likewise for tables, records and tuples etc. Judge by that member type, not by the variable you assign to: \`T? v = items[k].clone();\` is still immutable, type-checks with no diagnostic, and panics on the first in-place update with an \`InvalidUpdate\` error.
```
SasinduDilshara: We need to tell 2 points here
 - Immutable inherent for the member types
 - variable type is actually the inherent type, not the static type.

Move this to top

packages/ballerina-extension/src/features/ai/agent/data-binding-rules.ts
```diff
@@ -48,6 +54,11 @@ export const DATA_BINDING_CODING_RULES = `## Data binding, type casts and narrow
...
+- To get a mutable copy of a readonly value use \`T v = check ro.cloneWithType();\`, with a target type that is not itself \`readonly\`.
```
SasinduDilshara: Give an example if not there

SasinduDilshara (approved):

NipunaRanasinghe (approved):

PR #1141, opened 2026-09-21
Fix Try It detection for fast-run services

LakshanWeerasinghe:
### Purpose

Fix Try It failing to detect services started in fast-run mode when the Ballerina project is nested inside the opened workspace.

Fixes https://github.com/wso2/product-integrator/issues/2511

### Changes
  - Use the canonical Ballerina project directory for the fast-run JVM's `-XX:HeapDumpPath`.
  - Preserve parent-directory handling for standalone `.bal` files.
  - Add a regression test that inspects the spawned JVM arguments and verifies that the heap-dump path matches the project root.
  - Ensure the test process is stopped even when an assertion fails.

  ## Root cause

  Try It identifies running Ballerina processes by searching for:

  ```text
  -XX:HeapDumpPath=<project-directory>

  Fast-run previously set this argument using the language server's user.dir. In WSO2 Integrator, this can be the parent workspace rather than the selected integration directory.

  For example:

  Fast-run: -XX:HeapDumpPath=.../test_http_try_it
  Try It:   -XX:HeapDumpPath=.../test_http_try_it/json_read

  Although the Java process was running and listening on the service port, Try It could not associate it with the integration.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
Fixes Try It detection for services started in fast-run mode when the Ballerina project is nested in the workspace.

- Uses the canonical project directory for the JVM `-XX:HeapDumpPath` argument.
- Preserves parent-directory handling for standalone `.bal` files.
- Records the launched command for verification.
- Updates the regression test to verify JVM arguments and stop the test process even when assertions fail.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


packages/ballerina-language-server/langserver-core/src/test/java/org/ballerinalang/langserver/workspace/TestWorkspaceManager.java
```diff
@@ -566,13 +566,28 @@ public void testWSLoadProject() throws WorkspaceDocumentException, ProjectExcept
...
+
+            Path sourceRoot = workspaceManager.projectRoot(filePath);
+            BallerinaWorkspaceManager.ProjectContext projectContext =
+                    workspaceManager.sourceRootToProject.get(sourceRoot);
+            Assert.assertNotNull(projectContext);
+
+            String expectedHeapDumpPath = "-XX:HeapDumpPath=" + projectContext.project()
+                    .sourceRoot().toRealPath();
+            String actualHeapDumpPath = BallerinaWorkspaceManager.getHeapDumpPathArgument(
+                    projectContext.project().sourceRoot());
```
rosensilva: the assertion calls getHeapDumpPathArgument() directly. If command construction reverted to using the language server’s user.dir, this assertion would still pass while Try It broke again. Capture the actual launch arguments through a portable fixture or command capture, and assert the project path there

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/workspace/BallerinaWorkspaceManager.java
```diff
@@ -728,6 +727,11 @@ private List<String> prepareExecutionCommands(RunContext context, Module module,
         return commands;
     }
 
+    private static String getHeapDumpPathArgument(Path projectRoot) throws IOException {
+        Path workingDirectory = Files.isRegularFile(projectRoot) ? projectRoot.getParent() : projectRoot;
+        return HEAP_DUMP_PATH_FLAG + workingDirectory.toRealPath();
+    }
```
NipunaRanasinghe, line 733: `-XX:HeapDumpPath` is a live JVM flag, not only a marker for Try It, so this also changes where OOM heap dumps are written: fast-run dumps now land in the user's integration directory instead of the language server's working directory.

If that is not wanted, `<project>/target` works too. Try It looks for the process with `grep -iF -e "-XX:HeapDumpPath=<project dir>"` (`packages/ballerina-extension/src/features/tryit/utils.ts:160`), which is a substring match, so a path under the project root still matches.

Fine to leave as is if `.hprof` files in the integration directory are acceptable.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/workspace/BallerinaWorkspaceManager.java
```diff
@@ -1799,17 +1804,29 @@ public Optional<Process> process() {
...
             this.process = process;
+            this.launchCommand = List.copyOf(launchCommand);
+        }
+
+        /**
+         * Returns the command line the currently associated process was launched with.
+         *
+         * @return Launch command, or an empty list if no process has been started.
+         */
+        public List<String> launchCommand() {
```
NipunaRanasinghe, line 1820: Minor: `setProcess` requires the project lock to be held, but `launchCommand()` reads the field without it. Harmless while the test is the only reader, and the field is cleared along with the process, so no stale command survives a stop. Worth keeping in mind if production code starts reading it.

packages/ballerina-language-server/langserver-core/src/test/java/org/ballerinalang/langserver/workspace/TestWorkspaceManager.java
LakshanWeerasinghe: Fixed in [commit](https://github.com/ballerina-platform/ballerina-vscode/pull/1141/commits/6cb77a90fb6a978bafae40c416a22fcdcaf616e9).

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/workspace/BallerinaWorkspaceManager.java
LakshanWeerasinghe: No need to change this — fast run is only used in the development environment, so `.hprof` files ending up in the integration directory is acceptable. Keeping the project root as the heap dump path also keeps the Try It match direct rather than relying on a substring under `target`.

packages/ballerina-language-server/langserver-core/src/main/java/org/ballerinalang/langserver/workspace/BallerinaWorkspaceManager.java
LakshanWeerasinghe: Leaving this as is — the launch command is only read from the test, and it is stored as an immutable copy that is cleared together with the process, so there is no stale command to observe. If production code starts reading it we can take the project lock in the getter then. This also keeps the accessor consistent with the existing `process()` getter, which follows the same convention.

NipunaRanasinghe (approved):
LGTM

PR #1142, opened 2026-09-22
[Agent Builder] Stop the trace server when no running project still needs it

dan-niles:
## Summary

Fixes wso2/product-agent-builder#2

https://github.com/user-attachments/assets/a49d998a-8e4c-4c9f-b35d-c952d4c94051

- Stop the shared local tracing OTLP task once the last running project using it has stopped, so it no longer lingers after all runs end.
- Keep it alive while any other running project still uses `idetraceprovider` (a run on `amp` exports remotely and does not count).
- Request the server from the debug session start event. VS Code's debug Restart relaunches from the stored resolved configuration, so the Run/Debug command handlers that normally request the server never run for it, which left tracing dead after every restart.
- Queue a `START_SERVER` that arrives while a stop is still in flight instead of dropping it — on a restart the stop is still tearing down when the new session starts.


dan-niles:
@coderabbit review

packages/ballerina-extension/src/features/project/integration-runner-state.ts
```diff
@@ -94,9 +94,23 @@ function isRunAlive(run: ActiveRun): boolean {
...
     if (!isRunAlive(run)) {
         activeRuns.delete(run.projectPath);
+        stopTraceServerIfUnused();
     }
 }
 
+// Only idetraceprovider consumes the local trace server; amp exports remotely and doesn't keep it alive.
+function stopTraceServerIfUnused(): void {
+    if (!TracerMachine.isServerStarted()) {
+        return;
```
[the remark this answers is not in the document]
dan-niles: Valid, thanks — fixed in 2661d301cd.

The `isServerStarted()` guard was the root of it: during `serverStarting` it returned early, so `STOP_SERVER` was never even sent, and nothing re-checked once the start completed. Reachable in practice whenever a launch dies fast — `BIRunAdapter` fires `TerminatedEvent` immediately for a superseded launch or a working-directory failure.

Two changes:
- Dropped the `isServerStarted()` guard. Sending `STOP_SERVER` in a state with no handler is ignored anyway, so it bought nothing and was the thing suppressing the event.
- `serverStarting` now records `stopRequested` on `STOP_SERVER`, and its `onDone` routes to `serverStopping` instead of `serverStarted`. It still assigns `taskExecution` from the event on that branch so the task it just created is the one torn down. This mirrors the `restartRequested` handling already used for the reverse race in `serverStopping`.

Verified in the Extension Development Host watching port 59500: run → up; restart → stays up; stop → down and stays down; run followed immediately by stop → never comes up, no orphaned task.

packages/ballerina-extension/src/features/tracing/tracer-machine.ts
```diff
@@ -483,6 +518,13 @@ function createTracerMachine(projectPath?: string, childProjectPaths?: string[])
                                         }),
                                     ],
                                 },
+                                START_SERVER: {
+                                    actions: [
+                                        assign({
+                                            restartRequested: true,
+                                        }),
+                                    ],
+                                },
```
kaumini, line 543: **`serverStopping` has no `STOP_SERVER` handler, so a queued restart can outlive the runs that wanted it.**

`START_SERVER` is queued here into `restartRequested`, but a `STOP_SERVER` arriving later in this same state is dropped — no ancestor handles it either (`enabled` only has `ENABLE`/`DISABLE`/`REFRESH`, the root has no `on`). `onDone` then takes the `restartRequested` branch and relaunches the trace server with zero running projects, which is the lingering-server symptom this PR is fixing.

The window is wide enough to hit in practice: `stopServer` waits on `onDidEndTask` → the pseudoterminal's `close()` → `TraceServer.stop()` → `server.close()`, and `trace-server.ts` tracks no sockets and never calls `closeAllConnections()`, so Node waits out existing connections (up to the 5s default `keepAliveTimeout` for an idle one). Two easy triggers: the relaunched run fails immediately (compile error, port still held), or the user hits Stop shortly after Restart.

Reproduced against xstate 4.38.3 with this PR's chart — `START_SERVER` (settle) → `STOP_SERVER` → `START_SERVER` → `STOP_SERVER` → settle stop → ends in `enabled.serverStarting` with nothing running.

```suggestion
                                START_SERVER: {
                                    actions: [
                                        assign({
                                            restartRequested: true,
                                        }),
                                    ],
                                },
                                // A stop that arrives while we are still tearing down cancels the
                                // queued restart — otherwise onDone relaunches a server that no
                                // running project needs.
                                STOP_SERVER: {
                                    actions: [
                                        assign({
                                            restartRequested: false,
                                        }),
                                    ],
                                },
```

packages/ballerina-extension/src/features/tracing/tracer-machine.ts
```diff
@@ -455,6 +480,16 @@ function createTracerMachine(projectPath?: string, childProjectPaths?: string[])
                                         actions: [
                                             assign({
                                                 isDisabling: false,
+                                                restartRequested: false,
+                                            }),
+                                        ],
+                                    },
+                                    {
+                                        target: "serverStarting",
+                                        cond: (context) => context.restartRequested === true,
```
kaumini, line 497: **This guard can fire on a stale flag, because the `onError` below never clears it.**

`onError` (line 500) resets `isDisabling` but not `restartRequested`, so when `stopServer` rejects the machine returns to `serverStarted` with the flag still set. The next legitimate `STOP_SERVER` — the real "last run exited" one — then falls through this branch and restarts the server instead of leaving it stopped. Nothing clears the flag short of an extension reload.

Concrete path to the rejection: `TraceServer.start` assigns `server = app.listen(port)` and rejects on `'error'` **without unsetting `server`** (`trace-server.ts:472-484`), so after an `EADDRINUSE` — a case `trace-server-task.ts` explicitly prints a message for — `server` is left as a non-listening object and the next `TraceServer.stop()` gets `ERR_SERVER_NOT_RUNNING`.

The fix belongs in the `onError` a few lines below; GitHub won't let me attach a one-click suggestion there since it is outside the diff:

```ts
onError: {
    target: "serverStarted",
    actions: [
        assign({
            error: (context, event) => { /* unchanged */ },
            isDisabling: false,
            restartRequested: false,
        }),
    ],
}
```

Separately, `target: "serverStarted"` looks worth a second look on that same path — after an `EADDRINUSE` the server is not actually running, so the machine asserts a state that isn't true.

packages/ballerina-extension/src/features/tracing/tracer-machine.ts
dan-niles: Confirmed and fixed in b67d477b82 — took your suggestion (without the comment, house style).

I reproduced your sequence against the chart at HEAD before changing anything, driving the real `TracerMachine` with xstate 4.38.3 and stubs for `executeTask`/`onDidEndTask`/`TraceServer` so I could hold each async boundary open:

```
START (settle) → STOP → START → STOP → settle teardown
  before: {"enabled":"serverStarting"}   ← server relaunched, nothing running
  after:  {"enabled":"serverStopped"}
```

Your read of the window is right too — nothing cancels the queued restart because neither `enabled` nor the root handles `STOP_SERVER`.

I kept the same harness on the paths that must not regress, all green after the change:
- `START → settle → STOP → settle` ends `serverStopped`
- `START → settle → STOP → START → settle` still ends `serverStarting` (the restart this PR added)
- `START → STOP during serverStarting → settle` ends `serverStopped`

Worth noting the cancel is safe because `stopTraceServerIfUnused` only sends `STOP_SERVER` when no live run needs tracing, so a restart that is still wanted cannot be cancelled by it — once the relaunched session registers in `activeRuns`, the guard returns early.

packages/ballerina-extension/src/features/tracing/tracer-machine.ts
dan-niles: Valid — fixed in b67d477b82, `onError` now clears `restartRequested` alongside `isDisabling`.

I verified your rejection path is real rather than theoretical. `trace-server.ts:479-484` assigns `server = app.listen(port)` and rejects on `'error'` without unsetting it, so after `EADDRINUSE` the next `stop()` calls `close()` on a non-listening server and rejects. Driving the chart with `TraceServer.stop()` rejecting:

```
START (settle) → STOP (rejects) → START → settle → STOP (succeeds) → settle
  before: {"enabled":"serverStarting"}   ← stale flag restarted it
  after:  {"enabled":"serverStopped"}
```

On the `target: "serverStarted"` point — agreed it asserts something untrue after an `EADDRINUSE`, but that transition predates this PR and the real root is in `trace-server.ts`: `start()` leaves `server` set after an error and `stop()` does not clear it on the reject path either, so the module stays wedged until a reload. Fixing that properly means touching the server lifecycle rather than the chart, and it is reachable today without any of this PR's changes, so I would rather not widen this PR's blast radius for it. Happy to raise it as a separate issue if you agree.

kaumini (approved):

PR #1143, opened 2026-09-22
Fix the Workflow deployment section: Config.toml written, and durable agent detection

hasithaa:
Fixes wso2/product-integrator#2527 
Fixes wso2/product-integrator#2528

### The Config.toml the toggle writes

It wrote:

```toml
[ballerina.workflow.management]
enableManagementApi = true
port = 8_234
enableBasicAuth = false
```

None of those keys is a configurable of `workflow.management` (it declares `maxPageSize`, `maxBulkRetrySize` and `reviewActivityAccessRole`); they belong to `workflow.management.rest`, the module that owns the listener, so the block configured nothing. `port` and `enableBasicAuth` were restated at defaults — the latter would have overridden a `true` the module sets on purpose, had it been under the right table — and the serializer grouped the port's digits. Config.toml was parsed and re-serialized, so comments, key order and number formats elsewhere in the user's file were rewritten too.

It now writes:

```toml
[ballerina.workflow.management.rest]
enableManagementApi = true
```

Config.toml is edited as text, so nothing else in the file moves. An existing `enableManagementApi` is flipped rather than duplicated, a file that is already enabled is not rewritten at all, and disabling removes the table only once nothing else is left in it.

Also shortens the panel description, which carried import and configuration-editor detail that belongs in a tooltip or docs page.

### Durable agents count as workflows

A durable agent runs on the workflow engine but is listed under `directoryMap[AGENT]` with `kind: DURABLE_AGENT`, so the explorer can render it in the Agents section. Five places asked for workflow artifacts by reading `directoryMap[WORKFLOW]` alone: the deployment panel's Workflow section, the integration card's chips, the Devant dashboard's deploy eligibility, and both Devant deploy entry points (`handleComponentPushToDevant`, `deployIntegrationInDevant`). None of them recognised a durable-agent project. Adds `hasWorkflowArtifacts` to `ballerina-core` and uses it at all five.

### Review round

- A block the previous toggle wrote under `[ballerina.workflow.management]` (`enableManagementApi`, `port`, `enableBasicAuth` — none a configurable there) is cleared whenever the toggle runs, and that table dropped once empty. Dogfooders who used the old toggle get cleaned up on their next enable or disable.
- A table the author spelled another way (dotted keys) is detected by parsing once and left alone with a logged message, rather than being defined a second time.
- The author's line endings are kept, and a flip replaces the whole old value, so a stray string where a boolean belongs cannot be half-rewritten.

### Testing

`pnpm test:unit` — every package's jest suite plus `scripts/typecheck-tests.js`, as CI runs them, all green. 28 new cases: `workflowConfigToml.test.ts` (table, defaults, formatting and line-ending preservation, trailing comments, dotted-key guard, stale-block cleanup, re-enable, disable paths) and `hasWorkflowArtifacts` as a fixture corpus in `ballerina-core/src/test/identifier-utils.test.ts`.

- Fixed Workflow Management REST API configuration updates. The toggle now writes only the required REST table, preserves existing content, avoids unnecessary writes, and removes stale configuration.
- Added durable-agent detection across workflow visibility, integration types, dashboard eligibility, and Devant deployment.
- Shortened the workflow management description.
- Added 28 test cases for configuration updates and workflow artifact detection.


hasithaa:
Thanks @coderabbitai — two of the three are addressed in 011147b9f6.

**Table-header parsing** — partly real. The CRLF half does not reproduce: ECMAScript counts `\r` as a line terminator, so `$` under `/m` matches before it and a CRLF file already parses (verified). The trailing-comment half was a genuine defect: `[ballerina.workflow.management.rest] # ...` did not match, so enabling appended a *second* table of the same name and left Config.toml invalid. Fixed, with a test. Flipping the key now rewrites the value alone so a comment beside it survives, and an enabled key carrying one is recognised rather than rewritten. The multiline-string case is left alone — a `[` at line start inside a multiline basic string is not a shape a Ballerina Config.toml takes, and tracking string state would cost more than a full parser, which is what this deliberately avoids.

**`as any` in tests** — fixed. The assertions now go through a declared `WorkflowConfig`/`ManagementRestTable` shape.

**Atomic write** — skipped. Every other Config.toml writer in this package (`toml-utils.ts`, `icp/setup.ts`, `tracing/utils.ts`) uses a plain `writeFileSync`; making this one path atomic would be inconsistent and give false assurance about the rest. Worth doing for all of them together if we want it.

packages/ballerina-extension/src/features/devant/activator.ts
```diff
@@ -92,7 +91,7 @@ const handleComponentPushToDevant = async () => {
             });
         }
 
-        if (workflows?.length > 0) {
+        if (hasWorkflowArtifacts(projectStructure)) {
```
NipunaRanasinghe, line 94: There is a fourth call site, and it still has the bug this PR is fixing.

`packages/ballerina-extension/src/rpc-managers/platform-ext/rpc-manager.ts:519`, in `deployIntegrationInDevant`:

```ts
const workflows = project.directoryMap[DIRECTORY_MAP.WORKFLOW];
...
if (workflows?.length > 0) {
    scopes.push(DevantScopes.WORKFLOW);
}
```

That is a separate deploy entry point from `handleComponentPushToDevant` — the comment ~25 lines below it explicitly calls itself one of "the three deploy entry points". A durable-agent-only project deployed from that panel still gets `scopes.length === 0` and falls through to `promptToAddConstruct()`, which is exactly #2528.

It builds `DevantScopes` rather than `SCOPE`, but `hasWorkflowArtifacts` returns a plain boolean, so it drops in unchanged. The PR description says "three places" — it is four.

packages/ballerina-extension/src/utils/workflow-config-toml.ts
```diff
@@ -0,0 +1,112 @@
...
+ * under the License.
+ */
+import * as fs from 'fs';
+import * as path from 'path';
+
+// The configurables live in `workflow.management.rest`, the module that owns the HTTP listener.
+const TABLE = 'ballerina.workflow.management.rest';
+const KEY = 'enableManagementApi';
+
+const TABLE_HEADER = new RegExp(`^[ \\t]*\\[${TABLE.replace(/\./g, '\\.')}\\][ \\t]*(?:#[^\\n]*)?$`, 'm');
```
NipunaRanasinghe: `TABLE_HEADER` only matches the literal `[ballerina.workflow.management.rest]` header, so the dotted-key spelling of the same table is not found and `appendKey` appends a second definition of it. TOML forbids reopening a table that dotted keys already created, so the result does not parse. Verified against `@iarna/toml`:

```
input:  [ballerina.workflow.management]
        rest.enableManagementApi = false

output: ...same... + "\n[ballerina.workflow.management.rest]\nenableManagementApi = true\n"

parse:  FAIL "Can't redefine existing key at row 5, col 27"
```

Same class of bug as the trailing-comment one commit 3 fixed — the regex is the single point of truth for "does this table already exist", and every spelling it misses produces a duplicate.

Rather than widening the regex further (quoted keys and `[a . b . c]` with spaces are also legal), the cheaper guard is: when `findTable` misses, parse the file once and bail out (log and return, as the old code did on a parse error) if `ballerina.workflow.management.rest` already resolves. Appending is then only ever done on a file that genuinely has no such table.

packages/ballerina-extension/src/utils/workflow-config-toml.ts
```diff
@@ -0,0 +1,112 @@
...
+const KEY_LINE = new RegExp(`^[ \\t]*${KEY}[ \\t]*=[^\\n]*$`, 'm');
+const ALREADY_TRUE = /=[ \t]*true[ \t]*(?:#[^\n]*)?$/;
+// Captures up to the `=` so a flip rewrites the value alone, leaving any trailing comment.
+const VALUE = /^([ \t]*[\w.]+[ \t]*=)[ \t]*[^\s#]*/;
```
NipunaRanasinghe: `[^\s#]*` stops the old value at the first space, so a value containing one is only partly replaced and the rest is left on the line:

```
in:  enableManagementApi = "no thanks"
out: enableManagementApi = true thanks"     <- does not parse
```

The precondition is user error (a string where a boolean belongs), but the toggle turns a merely wrong Config.toml into a broken one. Matching to end-of-line minus any trailing comment instead of `[^\s#]*` covers it and keeps the comment-preserving behaviour the test asserts.

packages/ballerina-extension/src/test-support/workflowArtifactDetection.test.ts
```diff
@@ -0,0 +1,56 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Imported from source rather than the package barrel, which pulls in ESM-only LS transports.
+import { DIRECTORY_MAP, ProjectStructure } from "../../../ballerina-core/src/interfaces/bi";
+import { hasWorkflowArtifacts } from "../../../ballerina-core/src/utils/identifier-utils";
```
NipunaRanasinghe: These belong in `ballerina-core`, not here. `packages/ballerina-core/src/test/identifier-utils.test.ts` already exists and already tests `findScope` from this exact module, under `ballerina-core/jest.config.js` (`roots: ['<rootDir>/src/test']`).

As written this is the only file under `ballerina-extension/src/test-support/` that imports across a package boundary — the header comment justifies avoiding the *barrel*, which does not require leaving the package. Moving the suite to the core one removes the `../../../ballerina-core/src/...` escape entirely and puts the test beside the code it covers.

packages/ballerina-extension/src/utils/workflow-config-toml.ts
```diff
@@ -0,0 +1,112 @@
...
+    const configPath = path.join(projectPath, 'Config.toml');
+    const content = fs.existsSync(configPath) ? fs.readFileSync(configPath, 'utf-8') : '';
+    const updated = withManagementApi(content, true);
+    if (updated !== content) {
+        fs.writeFileSync(configPath, updated, 'utf-8');
+    }
+}
+
+/** Removes `enableManagementApi`, and the table too once nothing else is left in it. */
+export function disableWorkflowManagementConfig(projectPath: string): void {
```
NipunaRanasinghe, line 62: Nothing migrates files the previous version wrote. Anyone who already used the toggle keeps a stale

```toml
[ballerina.workflow.management]
enableManagementApi = true
port = 8_234
enableBasicAuth = false
```

and disabling now only touches `...management.rest`, so that block survives both operations forever. Ballerina will warn about the unused config values on every run.

Base is `staging/5.14.0`, so this is probably dogfooders only and may be fine to accept — but if it is accepted, say so, because the obvious reading of "disable" is that the old block goes away too.

packages/ballerina-extension/src/utils/workflow-config-toml.ts
```diff
@@ -0,0 +1,112 @@
...
+    const rest = content.slice(bodyStart);
+    const next = /^[ \t]*\[/m.exec(rest);
+    return { headerStart: header.index, bodyStart, body: next ? rest.slice(0, next.index) : rest };
+}
+
+function appendKey(content: string, table: Table | undefined): string {
+    if (table) {
+        const head = content.slice(0, table.bodyStart);
+        const separator = head === '' || head.endsWith('\n') ? '' : '\n';
+        return head + separator + `${KEY} = true\n` + content.slice(table.bodyStart);
```
NipunaRanasinghe: Nit: the inserted line is hardcoded `\n`, so on a CRLF Config.toml the file ends up with mixed endings:

```
"[ballerina.workflow.management.rest]\r\nenableManagementApi = true\nport = 9999\r\n"
```

TOML accepts it and the parser is happy, so this is cosmetic — but the whole point of the rewrite is that the file comes back looking the way the author wrote it. Detecting `\r\n` in the content once and using it for the inserted line is a two-line change. Same applies to the appended table at line 103.

packages/ballerina-extension/src/utils/workflow-config-toml.ts
```diff
@@ -0,0 +1,112 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import * as fs from 'fs';
+import * as path from 'path';
+
+// The configurables live in `workflow.management.rest`, the module that owns the HTTP listener.
```
NipunaRanasinghe, line 23: This comment is right, but the PR description and commit 1 both say `workflow.management` "declares no configurables". It declares three — `maxPageSize`, `maxBulkRetrySize`, `reviewActivityAccessRole` (`workflow.management/operations.bal:30-44` in `ballerina/workflow:0.10.0`). The old block was inert because it wrote the wrong *keys* there, not because the table has no configurables.

Same paragraph: `enableBasicAuth = false` is described as "overriding a `true` the module sets on purpose". The module default is indeed `true` (`workflow.management.rest/service.bal:114`), but under the wrong table that line overrode nothing. Worth correcting before this lands, since the commit message is the record of why the table changed.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.test.tsx
```diff
@@ -54,7 +54,10 @@ jest.mock("@wso2/ballerina-core", () => ({
         LOCAL_CONNECTORS: "LOCAL_CONNECTORS",
         WORKFLOW: "WORKFLOW",
         ACTIVITY: "ACTIVITY",
+        DURABLE_AGENT: "DURABLE_AGENT",
     },
+    hasWorkflowArtifacts: jest.requireActual(
```
NipunaRanasinghe, line 44: Worth a note: the real `hasWorkflowArtifacts` reads `directoryMap` with the *real* `DIRECTORY_MAP` enum values, while the fixtures and the component use the mocked ones above. This passes only because the mocked strings happen to equal the enum values for `WORKFLOW`/`AGENT`/`DURABLE_AGENT` — and the same mock already gets one wrong (`LOCAL_CONNECTORS: "LOCAL_CONNECTORS"`, where the enum is `"localConnectors"`).

Not a bug today. But the moment a key is added to `DIRECTORY_MAP` whose literal differs from its name, this pattern fails in a way that looks like a logic bug rather than a mock bug. Pulling `DIRECTORY_MAP` from `requireActual` too would keep the two sides in step.

hasithaa:
Thanks @NipunaRanasinghe — all addressed in the two commits ending at a49484fead.

- **Fourth call site** — `deployIntegrationInDevant` now uses `hasWorkflowArtifacts`, and CodeRabbit's find in `DevantDashboard` (the eligibility predicate that shows the empty-artifact message) is fixed too, so it is five sites, not three. PR description corrected.
- **Dotted-key spelling** — when the header regex misses, the file is parsed once; if `ballerina.workflow.management.rest` already resolves, or the file cannot be parsed, it is left untouched with a logged message. Appending only ever happens on a file that genuinely has no such table. Tests for both.
- **Partial value replacement** — the flip now replaces the whole old value (quoted strings included) and keeps only a trailing comment. `"no thanks" # hm` → `true # hm`.
- **Stale block migration** — accepted and implemented rather than documented: the three keys the previous toggle wrote under `[ballerina.workflow.management]` are cleared whenever the toggle runs (enable or disable), and that table dropped once empty. An author's own `maxPageSize` there survives. Tests for both paths.
- **CRLF** — inserted lines use the file's own ending.
- **"declares no configurables"** — you're right; it declares three, and the old block was inert because of the keys, not the table. Fixed in the PR description; the code comment already said the right thing. The commit message stays as history.
- **Test placement** — moved into `ballerina-core/src/test/identifier-utils.test.ts` as a fixture corpus (also CodeRabbit's suggestion); the cross-package import is gone. Both PackageOverview mocks now take `DIRECTORY_MAP` from `requireActual`.

**e2e group4** — the failure is `type.spec.ts:158 › Type Editor Field Diagnostics`, which this PR does not touch. It fails identically on staging's own scheduled e2e runs (09-21 and 09-22, "1 failed, 20 passed") and on a rerun of this PR's previous SHA, so it is pre-existing on `staging/5.14.0`.

packages/ballerina-extension/src/features/devant/activator.ts
hasithaa: Good catch — fixed in 5a4fb2a76c. `deployIntegrationInDevant` uses `hasWorkflowArtifacts(project)`. CodeRabbit also found the `DevantDashboard` eligibility predicate, so the count is five, not three; PR description corrected.

packages/ballerina-extension/src/utils/workflow-config-toml.ts
hasithaa: Fixed in a49484fead, the way you suggest: when the header regex misses, the file is parsed once, and if `ballerina.workflow.management.rest` already resolves — or the file cannot be parsed — it is left untouched with a logged message. Appending only happens on a file that genuinely has no such table. Tests cover the dotted-key spelling (byte-identical, still parses) and an unparseable file.

packages/ballerina-extension/src/utils/workflow-config-toml.ts
hasithaa: Fixed in a49484fead. The flip now replaces the whole old value, quoted strings included, and keeps only a trailing comment: `"no thanks" # hm` → `true # hm`. Test added.

packages/ballerina-extension/src/test-support/workflowArtifactDetection.test.ts
hasithaa: Moved in 5a4fb2a76c into `ballerina-core/src/test/identifier-utils.test.ts`, as a fixture corpus with one `it.each` over the cases (also CodeRabbit's suggestion). The cross-package import is gone.

packages/ballerina-extension/src/utils/workflow-config-toml.ts
hasithaa: Accepted and implemented rather than documented, in a49484fead: the three keys the previous toggle wrote under `[ballerina.workflow.management]` are cleared whenever the toggle runs — enable or disable — and that table is dropped once empty. An author's own key there (`maxPageSize`) survives. Safe because none of the three is a configurable of that module, so they were never anything but warnings. Tests for both paths.

packages/ballerina-extension/src/utils/workflow-config-toml.ts
hasithaa: Fixed in a49484fead — inserted lines use the file's own ending, detected once. Test asserts a CRLF file comes back with no bare `\n`.

packages/ballerina-extension/src/utils/workflow-config-toml.ts
hasithaa: You're right on both counts — it declares three, and the old block was inert because of the keys, not the table; and under the wrong table `enableBasicAuth = false` overrode nothing. Corrected in the PR description. The code comment already said the right thing; the commit message stays as history.

packages/ballerina-visualizer/src/views/BI/PackageOverview/index.test.tsx
hasithaa: Done in 5a4fb2a76c — both PackageOverview mocks take `DIRECTORY_MAP` from `requireActual` now, so fixtures and the real helper agree on the literals (and the `LOCAL_CONNECTORS` mismatch you spotted goes away with it).

hasithaa:
@coderabbitai on the latest round:

- **Malformed file after a valid table / multiline strings** — both answered together in e8c94e171e, without a TOML tokenizer: every edit is now verified by parsing. The file is parsed before anything is touched (unparseable → untouched), and the result is parsed afterwards and must equal the original with exactly this change applied — stale keys gone, `enableManagementApi` set or removed under `…management.rest`, nothing else different. A key line inside a `"""…"""` string, or a table spelled as dotted keys, fails that check and the file is left for the author with a logged message. Tests for both cases.
- **File watcher (outside diff)** — checked: `initStateSubscription` is called once, from `registerPlatformExtRpcHandlers` at activation (`RPCLayer.ts:134`), so the watcher is created once per extension host and the reinitialisation scenario does not arise. It is pre-existing code this PR does not touch; better as its own issue than folded in here.

packages/ballerina-extension/src/utils/workflow-config-toml.ts
```diff
@@ -0,0 +1,237 @@
...
+        return parse(content) as Toml;
+    } catch {
+        return undefined;
+    }
+}
+
+// True when `after` is `before` with the stale keys gone and `enableManagementApi` set (or
+// removed) under the REST table, and nothing else different.
+function isExpectedChange(before: Toml, after: Toml, enabled: boolean): boolean {
+    const expected = JSON.parse(JSON.stringify(before)) as Toml;
```
NipunaRanasinghe: This throws on a Config.toml that contains a large integer. `@iarna/toml` returns `BigInt` for integers outside the safe range, and `JSON.stringify` cannot serialise one:

```
[ballerina.workflow]
big = 9223372036854775807
```
```
TypeError: Do not know how to serialize a BigInt
    at JSON.stringify (<anonymous>)
    at isExpectedChange (workflow-config-toml.ts:119)
    at withManagementApi (workflow-config-toml.ts:100)
    at enableWorkflowManagementConfig (workflow-config-toml.ts:44)
```

The throw escapes into `addWorkflowManagement`'s `catch { console.log(error); return { enabled: false }; }`, so the toggle snaps back to off with nothing shown to the user — and by then `updateSourceCode` (rpc-manager.ts:50) has already written the `workflow.management.rest` import into main.bal. Half-applied: source edited, Config.toml not.

`structuredClone(before)` in place of the JSON round-trip fixes this, and the Date problem below with it.

packages/ballerina-extension/src/utils/workflow-config-toml.ts
```diff
@@ -0,0 +1,237 @@
...
+}
+
+function sameValue(a: any, b: any): boolean {
+    if (a === b) {
+        return true;
+    }
+    if (Array.isArray(a) || Array.isArray(b)) {
+        return Array.isArray(a) && Array.isArray(b) && a.length === b.length && a.every((v, i) => sameValue(v, b[i]));
+    }
+    if (a instanceof Date || b instanceof Date) {
```
NipunaRanasinghe: This branch is correct, but it can never be reached with two Dates, because the clone at line 119 has already turned every `Date` in `expected` into an ISO string. So `sameValue` compares a string against a `Date`, falls in here, and returns false — the verifier rejects a correct edit.

Any TOML date or datetime anywhere in the file is enough:

```
[ballerina.workflow]
startedAt = 1979-05-27T07:32:00Z    -> file unchanged, "declares ... in a form this editor does not rewrite"
day = 1979-05-27                    -> same
```

The toggle silently does nothing on an entirely ordinary Config.toml. `structuredClone` preserves `Date`, after which this branch does what it looks like it does.

packages/ballerina-extension/src/utils/workflow-config-toml.ts
```diff
@@ -0,0 +1,237 @@
...
+            updated = dropTableIfEmpty(content.slice(0, start) + content.slice(Math.min(end + 1, content.length)), TABLE);
+        }
+    }
+
+    if (updated === original) {
+        return original;
+    }
+    const after = parseToml(updated);
+    if (after === undefined || !isExpectedChange(before, after, enabled)) {
+        console.error(`${LOG} Config.toml declares ${TABLE} in a form this editor does not rewrite; `
```
NipunaRanasinghe: Worth deciding rather than a defect: there are now three ways enable can quietly do nothing — an unparseable file (line 74), this rejection, and the clone throwing (line 119, swallowed by the RPC catch). In all three the user sees the checkbox flip back and gets no explanation, while the message telling them what to set by hand goes to a console they are not looking at.

Given the LS has already edited main.bal by this point, the mismatch is worth surfacing. A `window.showWarningMessage` with this same text would cost two lines, and this file is in the extension host so it has access to it.

packages/ballerina-extension/src/test-support/workflowConfigToml.test.ts
```diff
@@ -0,0 +1,283 @@
...
+        const original = '[ballerina.workflow]\nmode = "LOCAL"\n';
+        write(original);
+        const before = fs.statSync(configPath).mtimeMs;
+
+        disableWorkflowManagementConfig(projectPath);
+
+        expect(read()).toBe(original);
+        expect(fs.statSync(configPath).mtimeMs).toBe(before);
+    });
+});
```
NipunaRanasinghe, line 310: The 23 cases here cover the text edit thoroughly, but nothing exercises the verifier against value types other than strings, numbers and booleans — which is where both problems above live. Two cases would have caught them:

```ts
it("enables in a file that carries a datetime", () => {
    write("[ballerina.workflow]\nstartedAt = 1979-05-27T07:32:00Z\n");
    enableWorkflowManagementConfig(projectPath);
    expect(restTable()?.enableManagementApi).toBe(true);
});

it("enables in a file that carries an out-of-range integer", () => {
    write("[ballerina.workflow]\nbig = 9223372036854775807\n");
    enableWorkflowManagementConfig(projectPath);
    expect(restTable()?.enableManagementApi).toBe(true);
});
```

Both fail on this commit: the first leaves the file untouched, the second throws.

packages/ballerina-extension/src/utils/workflow-config-toml.ts
hasithaa: Reproduced exactly as traced, fixed in c7cbebb7e8 with `structuredClone`. Confirmed `@iarna/toml` returns a BigInt for `9223372036854775807` and that `JSON.stringify` throws on it. The half-applied state is what made this worth more than tidiness, so the verifier is also wrapped now: a value it cannot compare leaves the file alone rather than taking the toggle down.

packages/ballerina-extension/src/utils/workflow-config-toml.ts
hasithaa: Right, and worse than you could see from the diff: `structuredClone` alone did not fix it. Two more things were wrong. `isDate` now tests `Object.prototype.toString` rather than `instanceof` — the clone is host-realm, so `instanceof Date` is **false** inside jest's sandbox. And `pruneEmpty` was dropping dates outright: it treated "object with no enumerable keys" as an empty table, which a `Date` also is. Both sides then compared as unequal for a different reason than the clone. Your datetime test is what surfaced each layer.

packages/ballerina-extension/src/utils/workflow-config-toml.ts
hasithaa: Agreed, and done in c7cbebb7e8 — `reportUnchanged` shows a `window.showWarningMessage` naming the table and key to set by hand, on all three paths (unparseable, rejected, and now a verifier that throws). The module is in the extension host as you say, and the jest `vscode` mock already stubs `showWarningMessage`, so it is covered by a test.

packages/ballerina-extension/src/test-support/workflowConfigToml.test.ts
hasithaa: Both added verbatim, and both failed before the fix exactly as you predicted — the datetime one silently left the file unchanged, the integer one threw. They also caught the two further bugs in my note on the date thread, which the `structuredClone` swap alone would have left in. A third case asserts the warning is shown. 26 cases now.

NipunaRanasinghe (approved):
LGTM

PR #1145, opened 2026-09-22
Update the model provider form to the new command palette label

ravindu25:
Related to https://github.com/wso2/product-integrator/issues/2180

## What was missed

The Copilot commands were moved to the `WSO2 Integrator` category, but the default model provider connection form still instructed users to run `> Ballerina: Configure default WSO2 model provider`, which no longer matches anything in the command palette.

## What was changed

Updated the two `InfoBox` call sites in `ConnectionSelector` to the current label, `> WSO2 Integrator: Configure default WSO2 model provider`.


madushajg (dismissed):

packages/ballerina-visualizer/src/components/ConnectionSelector/ConnectionConfig.tsx
```diff
@@ -220,7 +220,7 @@ export function ConnectionConfig(props: ConnectionConfigProps): JSX.Element {
                 <InfoBox
                     text="Configure this model provider using the command palette command:"
                     description={`Open the command palette with ${shortcut}, then run:`}
-                    codeCommand="> Ballerina: Configure default WSO2 model provider"
+                    codeCommand="> WSO2 Integrator: Configure default WSO2 model provider"
```
madushajg, line 223: `ballerina.configureWso2DefaultModelProvider`'s category in `package.json` is still "Ballerina" (not "WSO2 Integrator"). This command was never migrated. This change replaces one wrong label with another; the command palette entry is still `> Ballerina: Configure default WSO2 model provider`. Either update the category in `package.json` too, or revert this string.

packages/ballerina-visualizer/src/components/ConnectionSelector/config.ts
```diff
@@ -73,7 +73,7 @@ export const CONNECTION_SPECIAL_CONFIGS: Record<string, ConnectionSpecialConfig>
         infoMessage: {
             text: "Using the default WSO2 Model Provider will automatically add the necessary configuration values to Config.toml.",
             description: "This can also be done using the VSCode command palette command:",
-            codeCommand: "> Ballerina: Configure default WSO2 model provider"
+            codeCommand: "> WSO2 Integrator: Configure default WSO2 model provider"
```
madushajg, line 76: Same issue as `ConnectionConfig.tsx`. `configureWso2DefaultModelProvider`'s category in `package.json` is still "Ballerina", so this new label doesn't match the actual command palette entry either.

packages/ballerina-visualizer/src/components/ConnectionSelector/ConnectionConfig.tsx
ravindu25: I think this one's already sorted. The category moved in #1114, which merged into staging/5.14.0 on the 17th. On current staging, package.json:589-591 is:

```json
"command": "ballerina.configureWso2DefaultModelProvider",
"title": "Configure default WSO2 model provider",
"category": "WSO2 Integrator"
```

So the palette entry is `> WSO2 Integrator: Configure default WSO2 model provider` now, and this string is the one that matches it. Reverting would put the stale label back.

Easy to miss though, since the category change and this string fix went in as separate PRs and this diff doesn't touch package.json. If you're still seeing "Ballerina" locally, could just be a stale staging checkout.

packages/ballerina-visualizer/src/components/ConnectionSelector/config.ts
ravindu25: Same as the other thread, #1114 already moved the category over to "WSO2 Integrator" in staging. So this label matches what actually shows up in the palette now.

PR #1150, opened 2026-09-22
Let a role field hold several roles, and make every workflow form follow its dropdown

hasithaa:
Fixes wso2/product-integrator#2161, wso2/product-integrator#2114

### A role field could hold one role (#2161)

`userRoles` offered a single text box, so typing `finance, manager` produced one role of that name. The only way to name several was to know to switch to the expression mode and write `["finance", "manager"]` by hand.

The list mode now leads on every field that names people — the human task's `userRoles`, the `users`/`excluded*`/`administrator*` fields beside it, the activity call's review sub-forms, the durable agent's task roles. The expression mode stays second for a reference or a call, and **keeps the type the signature states** (`string|[string, string...]?` on the human task) rather than a hand-written approximation, so the editor does not accept what the compiler will reject. One name writes as `"finance"`, several as `["finance", "manager"]`; a value read back opens in whichever mode can show it.

### A form did not follow its own dropdown (#2114)

`input` was typed from `codedata.symbol` — the workflow chosen when the form *opened* — so changing the dropdown left the expression editor validating against the wrong type. Typed fields now carry `codedata.dependentProperty`, and the form refetches the template for the new choice and retypes from it. The **template** is asked which fields follow, not the current node, because a statement re-read from source does not carry the tag on its variable-type property.

### What the audit turned up

Since these nodes are rarely exercised, I checked every child-workflow and event form against its template rather than only the field the issue named. Re-opening an existing statement gave a different form from creating one:

| Node | On re-open, before |
|---|---|
| Run / Call Child Workflow | no workflow dropdown; raw `childWorkflow` expression, `input` typed `anydata`, leaked `connection`/`checkError` |
| Call Child Workflow | no dropdown **even in its template** |
| Send Data to Child | **no `workflow` field at all**, so its data-event dropdown could never populate |
| Wait / Call Child Workflow | a field labelled literally **`T`** beside "Variable Type" |

All of them now match their templates. The send follows its handle back to the run that bound it to name the workflow, and is served by the same form the top-level send uses (it keys off field names).

### Testing

- Language server, re-run under JDK 25 after merging staging: **2201 tests, every workflow golden passing**. 29 goldens regenerated and reviewed line by line; 6 new (4 child-workflow node templates, 2 diagram configs over a new fixture covering every child operation).
- Unit suites as CI runs them — every `packages/*/jest.config.js` plus `scripts/typecheck-tests.js`: **1819 green**, 0 test-code type errors.
- `compileJava` and `checkstyleMain` clean.

### Also in this PR

`carryTextArrayValue` applies to any field offering a `TEXT_ARRAY` mode, not only role fields: a list of text and the source for it are two shapes of one value, so switching modes carries it either way rather than dropping what was typed.

### Post-merge verification

Staging moved 44 commits under this branch, including the JDK 25 migration, so the goldens were regenerated again on the merged code (JDK 25 installed locally for it). That caught a real problem: the merge conflicted on `workflow_named_steps.json` and `workflow_retry_policy_parsing.json`, and resolving them to this branch's side had silently dropped the default values staging's own golden refresh carried. Both are regenerated properly in `1039f73566`.

Three tests still fail locally and none is a workflow golden — `resource_action_call-persist-get-albumrating`, `types/submodule_case_insensitive`, `types/submodule_description_match`. They fail with `ctxData is null` on compiler-plugin init and `Index 1 out of bounds`, and **the set changes from run to run** (an earlier run failed `local_connector_search_invalid` instead), which is this worktree's package-resolution environment rather than anything here. CI on a clean runner is the arbiter.

Two things I got wrong first and caught by reading the regenerated goldens rather than assuming: the restyle initially **narrowed** the human task's expression type, and the `T` drop did not fire for Call because `type` is attached after that hook — which is why the template, not the node, decides what retypes.

## Summary

- Added list-mode support for multiple roles in workflow forms.
- Preserved expression-mode values and serialized one role as a string or multiple roles as a list.
- Updated workflow and child-workflow forms to use the selected workflow template and retype dependent fields when the selection changes.
- Added text-array parsing and serialization utilities, including support for numeric escapes in role names.
- Added tests for role parsing, capability fields, and dependent-field retyping.


packages/ballerina-core/src/utils/text-array.ts
```diff
@@ -0,0 +1,146 @@
...
+            continue;
+        }
+        const next = body[i + 1];
+        switch (next) {
+            case "n": text += "\n"; break;
+            case "t": text += "\t"; break;
+            case "r": text += "\r"; break;
+            case '"': text += '"'; break;
+            case "\\": text += "\\"; break;
+            default: text += current + (next ?? ""); break;
```
NipunaRanasinghe, line 64: `\u{...}` falls into this default branch and is left as written, so the backslash is escaped again on the way out and the role name changes. The file header says this "mirrors the language server's WorkflowUtil" — it does not, and `WorkflowUtil.appendEscaped` spells out why it cannot:

> `case 'u' ->` // A numeric escape. Decoded because the re-encode cannot reproduce the escape, only the character it names — left as written, its backslash is escaped on save.

It also disagrees with `capabilityFieldValues.literalText`, which already decodes it, and whose comment says it does so "so a literal does not read differently depending on which side decoded it". Both run against the same `userRoles` field in the same form — one seeds it on open, the other carries it across the mode switch. Verified through the real exports:

```
source       ' ["grin \u{1F600}", "\u{41}BC"] '
seedCapabilityValue  -> ["grin 😀", "ABC"]
parseTextArraySource -> ["grin \u{1F600}", "\u{41}BC"]
textArraySource back -> ["grin \\u{1F600}", "\\u{41}BC"]
```

So: open the form, toggle to expression and back, save — the role is now the nine characters `\u{1F600}`. The doubled backslash is permanent. `WorkflowUtilLiteralTest:138` asserts the Java behaviour; this needs the same.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/capabilityFieldValues.ts
```diff
@@ -147,6 +160,63 @@ export function seedCapabilityValue(property: SeedableProperty, source: string):
     }
 }
 
+/**
+ * The names a role value states — one string literal, or a list of them — decoded; undefined for
+ * anything else, which belongs to the expression mode. `()` and nothing name nobody.
+ */
+function literalItems(source: string): string[] | undefined {
```
NipunaRanasinghe, line 173: This is `parseTextArraySource` + `splitTopLevel` from `ballerina-core/src/utils/text-array.ts`, which the same PR adds — same `()`/`[]` handling, same quote-and-depth scanner, same per-item literal check. `ballerina-visualizer` already imports from `@wso2/ballerina-core`, so this can call it.

They have already drifted: `parseTextArraySource("")` returns `[]`, `literalItems("")` returns undefined. And the unicode divergence in the comment above is only possible because there are two copies.

Worth contrasting with `SendDataBuilder.getAvailableWorkflowFunctions` in this same PR, which goes the other way — the inlined loop was pulled out into `WorkflowUtil.workflowFunctionOptions` and the caller now delegates. That is the move this wants too: keep the decoder that handles `\u{...}`, delete the other.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/dependentFields.ts
```diff
@@ -0,0 +1,73 @@
...
+ * carry the tags — the variable's type property is attached after the analysis that would set them
+ * — while the template a choice is fetched with always does, so it is the one asked.
+ */
+export function dependentKeysFromTemplate(template: NodeProperties, changedKey: string): string[] {
+    return Object.entries((template ?? {}) as Record<string, any>)
+        .filter(([, property]) => dependsOn(property?.codedata, changedKey))
+        .map(([key]) => key);
+}
+
+function dependsOn(codedata: { dependentProperty?: string | string[] } | undefined, changedKey: string): boolean {
```
NipunaRanasinghe: The two sides declare this field differently and neither is corrected here:

- `ballerina-core/src/interfaces/bi.ts:347` — `dependentProperty?: string[]`
- `PropertyCodedata` — `String dependentProperty`, and all four emitters (`ChildWorkflowCallBuilder:135,152`, `ChildWorkflowRunBuilder:132`, `WorkflowRunBuilder:143`) pass one key

Accepting both here makes the code work without resolving which is right, and the array branch is unreachable against the current language server — as is the `dependentProperty: ["workflow"]` case the test asserts at `dependentFields.test.ts:29`. Pick one: the wire shape is scalar, so `bi.ts` is the side that is wrong.

Separately, `PropertyCodedata`'s Javadoc still reads "The property that is dependent for this property to be **enabled**". This PR gives it a second meaning — "this property's type follows that property's value" — without saying so there. Whoever reads the record next will believe the old sentence.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/dependentFields.ts
```diff
@@ -0,0 +1,73 @@
...
+            return field;
+        }
+        const property = (template as Record<string, any>)?.[field.key];
+        if (!property) {
+            return { ...field, hidden: true, value: "" };
+        }
+        return {
+            ...field,
+            hidden: false,
+            types: JSON.parse(JSON.stringify(property.types ?? field.types)),
```
NipunaRanasinghe: `JSON.parse(JSON.stringify(undefined))` throws `SyntaxError: "undefined" is not valid JSON`, so this throws whenever neither the template property nor the field carries `types`:

```
retypeFieldsFromTemplate(
  [{ key: "input", value: "x", codedata: { dependentProperty: "workflow" } }],
  ["input"],
  { input: { placeholder: "{}" } })
-> SyntaxError: "undefined" is not valid JSON
```

The `?? field.types` fallback is there because `property.types` may be missing — but `field.types` comes straight from `Property.types`, which `bi.ts:294` declares optional too, so the fallback can be undefined as well. `retypeDependentFields` catches it, which means the symptom is the retype silently not happening: #2114 again, just quieter.

All four child-workflow templates do carry `types` on every tagged property (I checked the goldens), so this is latent rather than live. Worth fixing anyway — #1143 replaced this exact `JSON.parse(JSON.stringify())` idiom with `structuredClone` last week, for the same class of reason.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/dependentFields.ts
```diff
@@ -0,0 +1,73 @@
...
+import { FormField } from "@wso2/ballerina-side-panel";
+import { NodeProperties } from "@wso2/ballerina-core";
+
+/**
+ * A field typed from another field's value — the child workflow's input from the workflow
+ * dropdown, say — carries that field's key as `codedata.dependentProperty`. When the dropdown
+ * changes, the template for the new choice is fetched and the dependent fields take their type,
+ * placeholder and documentation from it, keeping whatever the person had typed.
+ */
+export function dependentFieldKeys(fields: FormField[], changedKey: string): string[] {
```
NipunaRanasinghe: `dependentFieldKeys` is never called. `index.tsx` imports only `dependentKeysFromTemplate` and `retypeFieldsFromTemplate`, and nothing else in the repo references it.

The doc comment above describes the whole retype flow as though this is the entry point, which makes it read as load-bearing. The comment on `dependentKeysFromTemplate` then explains why the template, not the node, is what gets asked — which is the reason this function cannot be used. Delete it and its two assertions at `dependentFields.test.ts:34-35`, and the remaining comment says everything that is true.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/index.tsx
```diff
@@ -882,12 +883,45 @@ export const FlowNodeForm = forwardRef<FormExpressionEditorRef, FlowNodeFormProp
         }
     };
 
+    // The fields typed from a dropdown's choice take their type from the template for the new
+    // choice; the last fetch wins when choices change faster than templates arrive.
+    const retypeRequest = useRef(0);
+    const retypeDependentFields = useCallback(async (fieldKey: string, value: unknown) => {
+        const changed = baseFields.find((field) => field.key === fieldKey);
+        const isSelect = changed?.types?.some((type) => type.fieldType === "SINGLE_SELECT");
```
NipunaRanasinghe: This fires for every SINGLE_SELECT in every flow-node form, not only the workflow dropdown — and `id: { ...node.codedata, symbol: value }` replaces `symbol` with whatever the dropdown holds. SINGLE_SELECT is widespread: the goldens use it under `agents_manager`, `model_provider_manager`, `types_manager` and `chunker_manager` as well as the workflow nodes. So picking a model in an AI agent form now sends a `getNodeTemplate` for `symbol: "gpt-4o"`, and the answer is thrown away because nothing in it is tagged.

The comment argues the cost is fine, and per-request that is true. The part I would push back on is the bogus `symbol`: asking the language server for a template for a symbol that does not exist is not a no-op by contract, it just happens not to hurt today.

A node-kind gate — the four child-workflow kinds plus `WORKFLOW_RUN` — would keep this to the forms it was written for, and is checkable from `node.codedata.node` without needing the tag that a re-read statement is missing.

packages/ballerina-core/src/test/text-array.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        ["an escaped literal", '"He said \\"hi\\""', ['He said "hi"']],
+        ["nil", "()", []],
+        ["nothing", "", []],
+        ["an empty list", "[]", []],
+        ["a reference", "financeRoles", undefined],
+        ["a list holding a reference", '["finance", lead]', undefined],
+        ["a call", 'roles("finance")', undefined],
+    ])("%s reads as %j", (_label, source, expected) => {
+        expect(parseTextArraySource(source)).toEqual(expected);
+    });
```
NipunaRanasinghe, line 52: No case here uses `\u{...}`, which is the one escape the two decoders disagree on. `WorkflowUtilLiteralTest:138` has it on the Java side:

```java
Assert.assertEquals(WorkflowUtil.stringLiteralText("\"grin \\u{1F600}\""), "grin \uD83D\uDE00");
```

The matching case here, plus a round-trip assertion over it, would have caught the comment above. The existing `round-trips what it writes` test only covers escapes that `stringLiteral` itself emits, so it can never exercise one that only appears in hand-written source.

packages/ballerina-core/src/utils/text-array.ts
hasithaa: Reproduced through the built exports exactly as you show, and fixed in 4d3b05ba4d. `stringLiteralText` now decodes `\u{...}`, and leaves an invalid code point or a lone surrogate as written — the same rule `WorkflowUtil` and `capabilityFieldValues.literalText` follow. Your corpus is in the tests, plus a round trip over it; four cases covering the escape, an out-of-range point and a lone surrogate.

packages/ballerina-core/src/test/text-array.test.ts
hasithaa: Added, mirroring `WorkflowUtilLiteralTest:138`: `\u{1F600}`, `\u{41}BC` mid-string, `\u{110000}` and `\u{D800}`, plus a round trip over the first two. You are right that the existing round-trip test could never have caught it — it only covers escapes `stringLiteral` emits, and this one appears only in hand-written source.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/dependentFields.ts
hasithaa: Both done in 4d3b05ba4d. The wire shape is scalar, so `bi.ts:347` is now `dependentProperty?: string` and the reader is a plain equality — the array branch and the test asserting it are gone. `PropertyCodedata`'s javadoc now names both meanings rather than only the original one.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/dependentFields.ts
hasithaa: Deleted, along with its two assertions. You are right that it read as the entry point while the comment underneath explained why it cannot be one; what is left says only what is true.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/dependentFields.ts
hasithaa: Fixed — `structuredClone(property.types ?? field.types ?? [])`, with a test that a template property carrying no `types` does not throw. Same idiom and same reason as #1143, which is a fair thing to have caught twice.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/index.tsx
hasithaa: Agreed, and this was the one worth catching. Gated on node kind in 4d3b05ba4d: `WORKFLOW_RUN`, `CHILD_WORKFLOW_RUN`, `CHILD_WORKFLOW_CALL` — the forms that actually hold a workflow dropdown on this path (the two sends go to `SendEventForm`, and the wait has no dropdown). Your point about the bogus `symbol` is the right one: it is not a no-op by contract, only by luck. A test pins the kinds, including that `AGENT_CALL` is not one.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/capabilityFieldValues.ts
hasithaa: Half agreed, and I could not do it — here is the constraint, because it is not obvious from the diff.

`ballerina-visualizer`'s jest config sets `transformIgnorePatterns: ['node_modules/']` with the comment "these unit tests mock @wso2/ballerina-core and use the pre-built CJS libs". Core ships ESM at `lib/index.js`, so a **source** file that imports the package cannot be parsed by the suite that covers it:

```
FAIL src/views/BI/FlowDiagram/capabilityFieldValues.test.ts
  ● Test suite failed to run
    Jest encountered an unexpected token
```

I tried the import first and hit exactly that. Mapping the specifier would reach 30 test files and 346 source files in that package, which is a bigger change than this PR should carry, and pointing production code at `../../../../ballerina-core/src/...` is the cross-package reach you objected to in #1143.

So the copy stays, but the drift does not: `literalItems` now returns `[]` for `""` as `parseTextArraySource` does, and both decode `\u{...}` identically after the fix on the other thread. Each side has a test pinning that behaviour, and the comment on the copy names its twin and says why it is a copy.

Your contrast with `getAvailableWorkflowFunctions` is fair — that one had no such barrier, which is why it went the other way. Happy to do the jest change as its own PR if you want the duplication gone properly.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -1182,34 +1203,145 @@ public static boolean editsAgentDeclaration(NodeKind nodeKind) {
...
+        List<String> literals = names.stream()
+                .map(item -> item instanceof Map<?, ?> map ? String.valueOf(map.get("value")) : String.valueOf(item))
+                .map(String::trim)
+                .filter(text -> !text.isEmpty())
+                .map(WorkflowUtil::stringLiteral)
+                .toList();
+        if (literals.isEmpty()) {
+            return "";
+        }
+        return literals.size() == 1 ? literals.getFirst() : "[" + String.join(", ", literals) + "]";
```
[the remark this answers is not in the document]
hasithaa: Declining — the premise is out of date. The language server moved to JDK 25 in #1089, which is already on `staging/5.14.0` and merged into this branch: `build.gradle:80` sets `options.release = 25`, with the comment "2201.14.0's libraries are Java 25 artifacts; a lower target fails dependency resolution."

`List.getFirst()` is Java 21+, so it compiles, and it is already the prevailing style in this module — 38 existing `getFirst()`/`getLast()` calls in `flow-model-generator-core` alone (`AiUtils:893`, `AgentsGenerator:619`, `McpClient:250`, …). Changing this one call to `get(0)` would make it the odd one out.

Verified rather than assumed: `compileJava` and the full 2201-test suite both run on this branch under JDK 25.

packages/ballerina-core/src/interfaces/bi.ts
```diff
@@ -344,7 +344,7 @@ export type CodeData = {
     id?: string;
     kind?: string;
     originalName?: string;
-    dependentProperty?: string[];
+    dependentProperty?: string;
```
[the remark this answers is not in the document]
hasithaa: Leaving this scalar, but flagging it for @NipunaRanasinghe to overrule if he disagrees — he asked for exactly this change in the previous round, on the grounds that the wire shape is scalar and `bi.ts` was the side that was wrong.

The reason a compatibility union would not buy anything: `PropertyCodedata` declares `String dependentProperty` and has since it was introduced, so the language server has never emitted an array. The `string[]` in `bi.ts` did not describe a supported shape — it described one that could not arrive. Widening to `string | string[]` would preserve a shape nothing produces and push the union onto every reader.

Nothing in the repo consumed it before this PR: the only references are `bi.ts` itself and the reader added here. So there is no internal migration, and an external consumer written against `string[]` was written against a value the language server never sent — its runtime behaviour does not change either way.

Happy to widen it if you would rather not touch a published type in a bug-fix PR; it is a one-line change.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/index.tsx
```diff
@@ -882,12 +883,47 @@ export const FlowNodeForm = forwardRef<FormExpressionEditorRef, FlowNodeFormProp
...
+            setBaseFields((prev) => retypeFieldsFromTemplate(prev, keys, template));
+        } catch (error) {
+            console.error(">>> Failed to retype the fields that follow", fieldKey, error);
+        }
+    }, [baseFields, fileName, node.codedata, rpcClient, targetLineRange]);
+
     const handleFormChange = useCallback(
         (fieldKey: string, value: any, allValues: FormValues) => {
             setFormDiagnostics(prev => prev.length > 0 ? [] : prev);
+            void retypeDependentFields(fieldKey, value);
```
NipunaRanasinghe, line 937: This may fire when the form opens, not only when the dropdown changes. `Form`'s change effect (`ballerina-side-panel/src/components/Form/index.tsx`, `prevValuesRef = useRef<FormValues>({})`) diffs against an empty object on the first render, so `onChange` runs for every field, `workflow` included.

That would send `getNodeTemplate` on every open of a Run / Call / Workflow Run form. Because the template's `type` carries `dependentProperty: workflow` and `resolveResultType` always returns a non-empty value (`json` by default), `retypeFieldsFromTemplate` would also replace a result type the source already declares. For example, `json r = check ctx->callChildWorkflow(wf, ...)` would be rewritten to `wf`'s return type, and the form would be dirty before anything is typed.

I haven't checked this in the running app. Could you confirm, and if it does happen, only retype when the value differs from the one the node opened with?

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -116,6 +117,26 @@ public static boolean isWorkflowModule(Optional<ModuleSymbol> moduleSymbol) {
      * @param symbol The function symbol to check
      * @return true if the function has @workflow:Workflow annotation, false otherwise
      */
+    /**
```
NipunaRanasinghe, line 114: `workflowFunctionOptions` was inserted between `isWorkflowFunction`'s Javadoc and that method. There are now two `/** */` blocks in a row, and `isWorkflowFunction` has none. Move this method above the `Checks if the given function symbol has the @workflow:Workflow annotation` block.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
```diff
@@ -1402,6 +1537,26 @@ public static void addStepIdProperty(NodeBuilder nodeBuilder) {
      *
      * @param properties the form's properties, edited in place
      */
+    /**
```
NipunaRanasinghe, line 1534: Same as above: `reorderProperties` sits between `markStepIdAdvanced`'s Javadoc and `markStepIdAdvanced`, so that doc now reads as belonging to this method. Move `reorderProperties` above that doc block.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/HumanTaskBuilder.java
```diff
@@ -448,6 +451,14 @@ public static void overlayOptionsLiteral(Map<String, Property> properties,
 
     // Keeps the property but takes it out of the form: not offered, not editable, still emitted
     // by toSource when it holds a value.
+    private static final Set<String> ROLE_KEYS = Set.of(USER_ROLES_KEY, WorkflowUtil.USERS_KEY,
```
NipunaRanasinghe, line 452: `ROLE_KEYS` and `argName` were inserted under the comment that describes `hide()` ("Keeps the property but takes it out of the form…"), which now reads as if it describes `ROLE_KEYS`. Move them above that comment.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -295,6 +298,7 @@ public class CodeAnalyzer extends NodeVisitor {
     public static final String PARAMETERIZED_CALL_QUERY = "sql:ParameterizedCallQuery";
     // Readonly fields
     private final Project project;
+    private List<Option> workflowOptions;
```
NipunaRanasinghe: Nit: this is a mutable, lazily filled cache, but it sits under `// Readonly fields` between `final` fields. It should go with the other mutable state.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/capabilityFieldValues.ts
```diff
@@ -147,6 +160,69 @@ export function seedCapabilityValue(property: SeedableProperty, source: string):
...
+ * The names a role value states — one string literal, or a list of them — decoded; undefined for
+ * anything else, which belongs to the expression mode. `()`, `[]` and nothing name nobody.
+ *
+ * <p>This is `parseTextArraySource` from `@wso2/ballerina-core`, and the two must agree — they
+ * edit the same field, one seeding it on open and the other carrying it across a mode switch. It
+ * is copied rather than imported because this package's jest config does not transform
+ * `node_modules` and the core package ships ESM, so a source file that imports it cannot be
+ * parsed by the suite that covers this module. Change one and change the other.
+ */
+function literalItems(source: string): string[] | undefined {
```
NipunaRanasinghe, line 173: The reason for the copy makes sense. There are now three parsers that must match: this one, core's `parseTextArraySource`, and the Java `WorkflowUtil.roleFieldValue`. The `\u{...}` cases already show how easily they drift. Could the three suites share one table of inputs and expected outputs, in place of the "change one and change the other" note?

packages/ballerina-side-panel/src/components/editors/FieldFactory.tsx
```diff
@@ -245,6 +245,14 @@ export const FieldFactory = (props: FieldFactoryProps) => {
 
         const currentValues = form.getValues();
         const currentFieldValue = currentValues[props.field.key];
+        // A list of text and the expression for it are two shapes of one value: carry it across
+        // the switch, so nothing typed is lost and the new mode has something to show.
+        const carried = carryTextArrayValue(mode === InputMode.TEXT_ARRAY, currentFieldValue);
```
NipunaRanasinghe, line 250: This applies to every field with a `TEXT_ARRAY` mode, not only the role fields, and it returns before the `getEditorConfiguration` path below runs. That is probably right for any text list, but other `TEXT_SET` fields change behaviour too, and the PR description doesn't mention it. Was that intended?

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/index.tsx
hasithaa: Confirmed, and it was worse than the fetch: `Form`'s change effect starts from `prevValuesRef = useRef({})`, so on the first render `!isEqual(undefined, value)` holds for every field and `onChange` fires for all of them, `workflow` included. The retype then ran on open, and because the template's `type` carries `dependentProperty: workflow` with a non-empty value from `resolveResultType`, it rewrote a declared result type — your `json r = check ctx->callChildWorkflow(wf, ...)` example is exactly right, and the form came up dirty.

Fixed in 2eac8c99da the way you suggest: `initForm` records the value each field holds as the form opens, and the retype returns unless the value differs from the last one seen for that key. Keeping it per-key rather than comparing only against the opening value means A → B → A still retypes back to A. Test added.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
hasithaa: Fixed — `workflowFunctionOptions` now sits above that javadoc, which is back with `isWorkflowFunction`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/utils/WorkflowUtil.java
hasithaa: Fixed the same way; `markStepIdAdvanced` has its javadoc back.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/HumanTaskBuilder.java
hasithaa: Fixed — `ROLE_KEYS` and `argName` moved above the comment, which describes `hide()` again.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Moved out of the readonly block, with a one-line note pointing at `workflowOptions()`.

packages/ballerina-side-panel/src/components/editors/FieldFactory.tsx
hasithaa: Intended, yes — it is generic to any field with a `TEXT_ARRAY` mode, not role-specific: a list of text and the source for it are two shapes of one value, and carrying it across the switch is right wherever that mode is offered. It returns before `getEditorConfiguration` only when it has something to carry (a list leaving the mode, or a parsable literal entering it); everything else falls through to the existing path unchanged. You are right that the description did not say so — I have added it.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/capabilityFieldValues.ts
hasithaa: Done for two of the three in 2eac8c99da: `packages/ballerina-core/src/utils/__fixtures__/roleValues.json` holds 14 cases as `{note, source, items}`, and both TS suites iterate it — core's `parseTextArraySource` directly, and this copy through `seedCapabilityValue`. The visualizer suite can read it by relative path (a JSON outside `node_modules` is transformed fine), so the "change one and change the other" note is now enforced rather than requested. The `\u{...}` cases are in the table.

The Java side I have not wired, and I would rather ask than guess. `WorkflowUtil.roleFieldValue` has no test of its own today, and the LS gradle test sets `workingDir targetDir`, so a relative path from the test is fragile — it wants a copy step:

```gradle
tasks.register("copyRoleValueCorpus", Copy) {
    from "$rootDir/../ballerina-core/src/utils/__fixtures__/roleValues.json"
    into "$projectDir/build/resources/test/fixtures"
}
processTestResources.dependsOn copyRoleValueCorpus
```

That adds a build-time dependency from the language server onto a TypeScript package, which is a structural choice rather than a tidy-up. Want it in this PR, or as a follow-up?

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/index.tsx
NipunaRanasinghe: Thanks, the guard looks right. The new test in `dependentFields.test.ts` doesn't exercise it, though. It checks that an empty template has no dependent keys and that a retype takes the template's value, and both still pass with the `retypedForRef` check deleted. The guard lives in `index.tsx`, so a regression here would go unnoticed. Could the "same value as last seen, skip" decision move into `dependentFields.ts` as a small function (say `shouldRetype(lastSeen, key, value)`) with a test? Or add a component test that opens a Call Child Workflow form with a declared result type and asserts it is unchanged and no `getNodeTemplate` call is made.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/capabilityFieldValues.ts
NipunaRanasinghe: The shared table covers the two TypeScript parsers, thanks. The third one, Java `WorkflowUtil.roleFieldValue` / `stringLiteralText`, still has its own cases in `WorkflowUtilLiteralTest`, and that is where the `\u{...}` mismatch came from. Could the Java test read the same `roleValues.json`, or could the fixture move somewhere both sides load it from? Otherwise the language server can still drift from the frontend without any test failing.

packages/ballerina-core/src/interfaces/bi.ts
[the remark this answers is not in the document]
NipunaRanasinghe: Confirming: keep it a single string. `PropertyCodedata.dependentProperty` is a single `String` on the wire, so `string[]` in `bi.ts` was never an accurate type. Nothing in the repo read it as an array after the reader change.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/index.tsx
hasithaa: Moved it: `shouldRetype(lastSeen, key, value)` now lives in `dependentFields.ts` with four tests of its own — first sighting retypes, the same value again does not, a different value does, and A → B → A retypes back to A. Deleting the check fails them.

It also narrows `value` to `string`, so the caller passes it straight on as the symbol.

One thing the move surfaced: skipping now also bumps `retypeRequest.current`, so a fetch still in flight for the previously chosen workflow is stranded rather than landing on the form after the field was cleared.

packages/ballerina-visualizer/src/views/BI/FlowDiagram/capabilityFieldValues.ts
hasithaa: Done — `RoleValueCorpusTest` (flow-model-generator-core, registered in that module's `testng.xml`) reads the same `ballerina-core/src/utils/__fixtures__/roleValues.json` and runs every case through `WorkflowUtil.roleFieldValue` / `stringLiteralText`. All three parsers now agree on the one corpus, `\u{...}` included; the suite passes.

packages/ballerina-core/src/interfaces/bi.ts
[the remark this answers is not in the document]
hasithaa: Agreed, it stays a single `string`.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/index.tsx
```diff
@@ -882,12 +889,55 @@ export const FlowNodeForm = forwardRef<FormExpressionEditorRef, FlowNodeFormProp
...
+    const retypeDependentFields = useCallback(async (fieldKey: string, value: unknown) => {
+        // Only the forms that pick a workflow: elsewhere a dropdown holds something that is not a
+        // symbol, and a template fetched for it would ask for one that does not exist.
+        if (!picksWorkflow(node.codedata?.node) || !fileName) {
+            return;
+        }
+        if (!shouldRetype(retypedForRef.current, fieldKey, value)) {
+            // A value that is gone or unchanged must also strand any fetch still in flight for the
+            // one before it, or its answer would retype fields for a workflow no longer chosen.
+            retypeRequest.current++;
```
NipunaRanasinghe: This bump now runs whenever `shouldRetype` returns false, for **any** field. The dropdown check (`SINGLE_SELECT`) only comes after it, and `handleFormChange` sees every field. So while a template is loading, an edit to a different field throws its answer away:

1. Pick workflow B. The B template request starts and `lastSeen.workflow` becomes `"B"`.
2. Before it returns, clear any field (`""`), edit a role list (an array), or toggle a boolean. `shouldRetype` returns false and `retypeRequest.current++` runs.
3. B's response fails `request !== retypeRequest.current` and is dropped. `input` and `type` keep workflow A's types.
4. Picking B again does nothing, because `lastSeen.workflow === "B"` already. The form stays typed for A until a third workflow is picked.

Suggested fix: run the `SINGLE_SELECT` check first, so only the dropdown's own changes reach `shouldRetype` and the bump:

```ts
const changed = baseFields.find((field) => field.key === fieldKey);
if (!changed?.types?.some((type) => type.fieldType === "SINGLE_SELECT")) {
    return;
}
if (!shouldRetype(retypedForRef.current, fieldKey, value)) {
    retypeRequest.current++;
    return;
}
```

A smaller point in the same area: `shouldRetype` records the value before the request runs. If the request fails (the `catch` below) or its answer is dropped, picking that workflow again won't retry. Recording the value only after the template is applied, or removing it in the `catch` and in the dropped-answer branch, would fix that.

hasithaa: Good catch — both are real. Fixed in `155a96b518`.

**The bump for any field.** The `SINGLE_SELECT` check now runs first, exactly as you suggested, so only the dropdown's own change reaches `shouldRetype` and the counter. Your sequence is worth spelling out because the end state was the bad part: the form stayed typed for A *and* refused to retry B, since `lastSeen.workflow` was already `"B"`. A stuck form, not just a dropped request.

**Recording before the request.** Added `forgetRetype(lastSeen, key, value, previous)`, called in the `catch` and in the dropped-answer branch. I kept the record at check time rather than moving it to after the template is applied, so a repeated report of the same value while a request is in flight does not fire a second one.

It only restores while the key still holds the value it recorded:

```ts
if (lastSeen[key] !== value) {
    return;
}
```

Without that, a stale answer arriving after a newer pick would roll back over the newer attempt's record — B's late answer restoring A while C is in flight. Three tests cover it: retry after a failure, dropping the key when nothing was recorded before, and leaving a newer choice alone.

13 tests in `dependentFields.test.ts` now, 699 in the visualizer suite, typecheck clean.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/test/java/io/ballerina/flowmodelgenerator/core/utils/RoleValueCorpusTest.java
```diff
@@ -0,0 +1,78 @@
+/*
```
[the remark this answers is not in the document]
hasithaa: Done in `d1f8d02ed6` — the header now matches `WorkflowUtilLiteralTest` byte for byte.

hasithaa:
Both CodeRabbit findings from the last review are fixed in `d1f8d02ed6`. Replying here for the outside-diff one, which has no thread to answer in.

**"Include old dependent keys when the new template omits a field"** (`index.tsx:934`) — valid, and the dead-code consequence made it invisible. `dependentKeysFromTemplate` derived keys from the new template alone, so a key the template does not carry could never be returned; `retypeFieldsFromTemplate` was then never reached for it, and its missing-property branch could not fire from the production path at all. The test that covered that branch passed `["input"]` by hand, so it stayed green.

The case is reachable. `CodeAnalyzer.retypeWorkflowInput` drops the property outright when the chosen workflow declares no input:

```java
if (inputType == null) {
    // The workflow function declares no input; drop the library-derived input property.
    currentProps.remove(key);
    return;
}
```

So switching from a workflow that takes input to one that takes none left the old `input` field on the form, still holding the previous value.

`dependentKeys(fields, template, changedKey)` now unions the template's tags with the form's own, deduplicated. The keys are computed inside the `setBaseFields` updater so they come from current fields rather than a captured copy. Two tests added: the template-drops-the-field case asserting the field ends up hidden with its value cleared, and a both-sides-tagged case asserting each key appears once.

**License header** on `RoleValueCorpusTest` — answered in its thread; it now matches `WorkflowUtilLiteralTest` exactly.

Verified: 15 tests in `dependentFields.test.ts`, 701 in the visualizer suite, `typecheck-tests` clean, and `:flow-model-generator-core:checkstyleTest` + `:test` green with `--rerun-tasks`.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/index.tsx
```diff
@@ -882,12 +895,59 @@ export const FlowNodeForm = forwardRef<FormExpressionEditorRef, FlowNodeFormProp
...
+                id: { ...node.codedata, symbol: value },
+            });
+            if (request !== retypeRequest.current || !response?.flowNode) {
+                forgetRetype(retypedForRef.current, fieldKey, value, previous);
+                return;
+            }
+            const template = getFormProperties(response.flowNode) ?? {};
+            setBaseFields((prev) => {
+                const keys = dependentKeys(prev, template, fieldKey);
+                return keys.length === 0 ? prev : retypeFieldsFromTemplate(prev, keys, template);
```
NipunaRanasinghe: `dependentKeys` now returns `input` when the new workflow takes none, and `retypeFieldsFromTemplate` marks it `{ hidden: true, value: "" }`. As far as I can trace, the `""` only reaches `baseFields`, not the values the form saves:

- When `formFields` changes, `Form` rebuilds its values (`ballerina-side-panel/src/components/Form/index.tsx:672`) by first preserving every existing non-empty value. It only uses `field.value` when nothing is held for that key (`:690`), so the old input text stays.
- Only `TYPE` / `ACTION_TYPE` fields are re-read from `field.value` (`:735`). That's why the `type` retype works, and it's the same path that caused the on-open overwrite.
- On save, `updateNodeWithProperties` copies the form values into the properties, and `ChildWorkflowRunBuilder.toSource` / `WorkflowRunBuilder.toSource` write `input` whenever it's non-blank, whether or not the field is hidden.

So switching from a workflow that takes input to one that doesn't would likely produce `ctx->runChildWorkflow(noInputWf, <old input>)`. The field is hidden, so the user can't see or clear it, and the call won't compile. The new test only checks what `retypeFieldsFromTemplate` returns, so it can't catch this.

I haven't run it in the app. A quick check: open a Run Child Workflow with an input, switch to a workflow with no input, save, and look at the source. If it happens, possible fixes are clearing the form value for the keys that become hidden (`form.setValue(key, "")` through the form ref), or dropping hidden dependent fields when building the source.

packages/ballerina-visualizer/src/views/BI/Forms/FlowNodeForm/index.tsx
```diff
@@ -882,12 +895,59 @@ export const FlowNodeForm = forwardRef<FormExpressionEditorRef, FlowNodeFormProp
...
+            });
+        } catch (error) {
+            forgetRetype(retypedForRef.current, fieldKey, value, previous);
```
NipunaRanasinghe: Nit: this writes into whatever `retypedForRef.current` is when the answer arrives. `initForm` replaces that object, so a late answer from the previously open node can restore the old node's `previous` into the new node's record. Saving it in a local before the `await` (`const lastSeen = retypedForRef.current;`) and passing that to `shouldRetype` and both `forgetRetype` calls would avoid it. It's harmless in practice, since at worst it causes one extra retype later.

PR #1151, opened 2026-09-22
Keep module-level value references in the workflow dropdown forms

### LINKED ISSUE #2533: Preserve module-level value references in workflow dropdown form fields (retry policy, REST method)
## Description

The workflow designer's dropdown-driven form fields (`DROPDOWN_CHOICE`) only round-trip values the form itself wrote. When a user binds the same parameter to a **module-level value** — a `const`, or a `final readonly & T` record / array constructed at initialisation — the source-read path does not recognise the value, the form shows the default option, and **saving the form rewrites the user's code**: the argument is either deleted or turned into a different literal.

This is the 20% case (reusing one retry policy or HTTP method across many activities); the 80% case of filling the dropdown by hand is unaffected. It is a data-loss bug, not a rendering glitch.

## Steps to reproduce

```ballerina
final readonly & workflow:AutoRetry STANDARD_RETRY = {maxRetries: 5, retryDelay: 2.0, retryBackoff: 3.0};
final readonly & string[] APPROVERS = ["FINANCE_APPROVER", "MANAGER"];
const string GET_METHOD = "GET";

@workflow:Workflow
function varRefWorkflow(workflow:Context ctx, OrderInput input) returns error? {
    int a = check ctx->callActivity(fetchOrder, {id: 1}, retryPolicy = STANDARD_RETRY);
    int b = check ctx->callActivity(fetchOrder, {id: 2}, retryPolicy = APPROVERS);
    json c = check ctx->callActivity(activity:callRestAPI,
            {connection: httpClient, method: GET_METHOD, path: "/users/1"});
    boolean d = check ctx->awaitHumanTask("approve", userRoles = APPROVERS);
}
```

Open each node in the designer and save without changing anything.

## Observed (verified against the LS `getFlowModel` output)

| Source | Form shows | On save |
|---|---|---|
| `retryPolicy = STANDARD_RETRY` | **No Retry** | argument **deleted** |
| `retryPolicy = APPROVERS` | **No Retry** | argument **deleted** |
| `method: GET_METHOD` | dropdown value `GET_METHOD`, not among the options → blank select | rewritten to `method: "GET_METHOD"` — a string literal, invalid HTTP verb |
| `userRoles = APPROVERS` | Expression mode, value `APPROVERS` | round-trips correctly ✅ |

## Root cause (language server, `flow-model-generator-core`)

- `CodeAnalyzer.addNormalizedRetryPolicyProperties` sniffs the **source string**: `contains("ManualRetry")`, else `startsWith("{")`, else No Retry. An identifier lands in the else branch.
- `ActivityCallBuilder.populateRetryPolicyArg` treats No Retry as the default and emits nothing — read-as-NoRetry plus write-nothing is what deletes the argument.
- `RestActivityStrategy.getCallActivityArgs` emits `method: "<value>"` with unconditional quoting.

`userRoles` works because it is an ordinary multi-type property (`TEXT` + `EXPRESSION`) and `Property.Builder.typeWithExpression` already falls back to selecting `EXPRESSION` when no `SINGLE_SELECT` option matches. `DROPDOWN_CHOICE` does not participate in that fallback.

### Related defects found by the same fixture

1. **Retry Policy options are stale against workflow 0.10.** The dropdown offers No Retry / Auto Retry / Manual Retry, but 0.10 defines `RetryPolicy = AutoRetry|ReviewTaskDefinition|RetryBeforeReview|NoRetry` and has no `ManualRetry` symbol. "Manual Retry" emits `workflow:ManualRetry` (does not compile on 0.10), and any `ReviewTaskDefinition` / `RetryBeforeReview` literal is read as No Retry and deleted on save. Only 0.10+ needs to be supported.
2. `CodeAnalyzer.parseSimpleRecord` splits the record literal on `,`, so a literal containing a nested array or record mis-parses.

## Affected forms

Every node that carries the `retryPolicy` dropdown (user activity call, connection activity call, and the REST / SOAP / Email builtin activities) and the REST activity `method` dropdown. The Send Data function dropdown lists function names (already references) and is not affected. Human task, child workflow and durable agent forms use the generic typed path and round-trip references correctly.

## Proposal

**1. Do-no-harm invariant (first, independent of UI).** Keep the original argument text on the property (`Property.advancedValue` is an available slot) and re-emit it verbatim whenever the form did not recognise the value. Converts silent corruption into a no-op.

**2. "From Expression" dropdown option.** Add one option to each affected dropdown whose dynamic sub-field is a single expression input. The sub-field's `ballerinaType` is the **actual parameter type** (`workflow:RetryPolicy`, `string` for `method`) so the expression editor offers the right completions and diagnostics. The reader selects this option for any value it cannot map to a fixed option (variable reference, function call, unrecognised literal); the writer emits the expression verbatim, never quoted or reshaped.

**3. Type-resolve instead of string-sniff.** Replace `contains(...)` / `startsWith("{")` with `semanticModel.typeOf(arg)` matched against the module's `RetryPolicy` members. This is what makes option 2 round-trip: if the user types `{maxRetries: 3}` into the expression field, on reopen the value resolves to `AutoRetry` and the form renders the Auto Retry sub-fields again; a `ReviewTaskDefinition` literal renders its own form instead of being deleted. It also fixes related defect 1 — the options become the 0.10 union (`No Retry` / `Auto Retry` / `Review Task` / `Retry Before Review` / `From Expression`) and `ManualRetry` goes away.

**4. Apply the same pattern to every workflow dropdown field**, not only `retryPolicy` and `method`, so a reference is never rewritten anywhere in the workflow forms.

Expected result after the fix: the 80% path is unchanged; a module-level `final readonly & workflow:AutoRetry` reference opens as **From Expression → `STANDARD_RETRY`**, saves back unchanged, and a record literal typed in that field reopens as the matching form.

## Environment

- ballerina-vscode `packages/ballerina-language-server`, `flow-model-generator` (workflow module 0.10.0 resolved from the Central cache; Ballerina 2201.13.4)

hasithaa:
Fixes the workflow designer rewriting a user's code when a form parameter is bound to a module-level value.

Resolves wso2/product-integrator#2533.

## The bug

The dropdown-driven form fields only round-tripped values the form itself had written. A parameter bound to a `const`, or to a `final readonly & T` record built at initialisation, was not recognised: the form showed its default, and **saving rewrote the source**.

Verified against the LS `getFlowModel` / `getSourceCode` output:

| Source | Before | After |
|---|---|---|
| `retryPolicy = STANDARD_RETRY` | read as No Retry, argument **deleted** on save | From Expression, written back verbatim |
| `approvalPolicy = FINANCE_GATE` | rendered as a dead-end option of its own | From Expression, written back verbatim |
| `method: GET_METHOD` | rewritten to `method: "GET_METHOD"` — a string literal, invalid HTTP verb | `method: GET_METHOD`, unquoted |
| `cardinality: ONE_SHOT` | stripped to a name no option matched, **dropped** on save | written back verbatim |

## What changed

**A From Expression option** on each dropdown that carries sub-fields (retry policy, approval policy, REST method). Its one sub-field is an expression typed to the parameter's own type — `workflow:RetryPolicy`, `workflow:ApprovalPolicy`, `string` — so the editor completes and checks against it. A value the dropdown cannot show opens there holding its source and is written back exactly as it stands; a record literal typed into it reopens as the option it declares. Selecting the option and leaving the field empty is refused rather than written as nothing.

**The union member now comes from the compiler.** The readers told `AutoRetry` / `ReviewTaskDefinition` / `RetryBeforeReview` apart by the fields a literal carried, which encoded in the form what the module already decides. `semanticModel.typeOf` on the argument names the member exactly — including `{}`, which the fields alone cannot place. The field-presence heuristic remains as the fallback when the type cannot be resolved (the capability-hydration path has only strings), and a member the form has no option for is carried as source rather than filed under whichever option its fields resemble.

A reference still opens under From Expression whatever it resolves to. Showing editable Auto Retry sub-fields for `STANDARD_RETRY` would replace the reference with a literal of whatever the boxes held on the next save.

**Event cardinality** takes the mode-chip idiom (`SINGLE_SELECT` + `EXPRESSION`) rather than a From Expression option: `isDropDownType` makes any select carrying `options` render through the expression editor, where an option's sub-fields are never shown. Its writer decides by value rather than by the field's mode, because a capability's values reach the form as plain strings with no mode beside them.

**Sweep.** These four were the whole of it. Every other workflow select lists identifiers (workflow, agent, activity and event names) and emits the property value verbatim, so an unlisted value already survived.

## Tests

2201 LS-extension tests, 256 LS core tests, 238 form tests across 32 suites. Checkstyle and spotbugs clean.

**Round trips, not halves.** The reader and the writer were already tested apart from each other, which is not where the bug lived — it lived in the pair disagreeing. `RetryPolicyOptionRoundTripTest` runs the real reader into the real writer for each of the eleven shapes a retry policy takes, twice over, and asserts the source comes back as it went in. The approval policy gets the same treatment in `FromExpressionOptionTest`, and both assert their default is written as an absent argument rather than as a value.

**The routing difference, pinned.** Two `EditorFactory` fixtures record that a `DROPDOWN_CHOICE` reaches `DropdownChoiceForm` (which renders an option's sub-fields) while a `SINGLE_SELECT` carrying options is routed to the expression editor and never shows them. This is the constraint that decided how each dropdown got its escape hatch, and the fixture is what would have caught the cardinality field shipping an option with no input under it.

**Form behaviour.** Four tests on the option itself: it is offered, selecting it reveals the expression field, the other options do not, and the value registers under its own key — the root property source generation reads. Two more cover a cardinality whose value is a constant: the select presents no member rather than the wrong one.

**Type resolution.** `RetryPolicyTypeResolutionTest` pins what the semantic model reports for each literal, so a language change that stopped naming the union member surfaces here rather than as forms quietly mis-filing policies.

**Goldens.** New fixtures covering each corruption above, read and write, including the durable agent capability hydration path that nothing reached before. 26 existing goldens regenerated; the only non-additive change is the former dead-end option becoming From Expression with its source captured.

## Notes for review

The issue text also listed a naive comma-split in the record-literal parser. That referred to the old `parseSimpleRecord`; `WorkflowUtil.parseRecordLiteral` on main is already nesting- and comment-aware, so there is nothing to fix there. I have corrected the issue.

One display nuance on cardinality, unchanged by this PR but worth knowing: a capability's values reach its form as plain strings with no mode, so a cardinality given as a constant opens with the select showing nothing until the user switches to the expression mode. The value round-trips correctly either way — it is no longer dropped — and the `name` field of the same form has always behaved this way. Making the mode travel would be a change to how capability forms are seeded, which felt out of scope here.

## Summary

- Preserve workflow dropdown values that reference constants, variables, function calls, or unsupported literals.
- Add **From Expression** options for retry policy, approval policy, and REST method fields.
- Resolve retry-policy members through the compiler, with field-based fallback when resolution fails.
- Preserve event-cardinality expressions and REST method syntax when forms are saved.
- Add language-server, round-trip, type-resolution, and UI tests.


packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/RestActivityStrategy.java
```diff
@@ -160,6 +161,72 @@ public void setFormProperties(NodeBuilder nodeBuilder, NodeBuilder.TemplateConte
...
+        Map<String, Map<String, Property>> methodDynamicFields = new LinkedHashMap<>();
+        methodDynamicFields.put(METHOD_GET, Map.of());
+        methodDynamicFields.put(METHOD_POST, Map.of(MESSAGE_KEY, messageSubProp));
+        methodDynamicFields.put(METHOD_PUT, Map.of(MESSAGE_KEY, messageSubProp));
+        methodDynamicFields.put(METHOD_DELETE, Map.of(MESSAGE_KEY, messageSubProp));
+        methodDynamicFields.put(METHOD_PATCH, Map.of(MESSAGE_KEY, messageSubProp));
+        // An expression may name any method, so the body is offered beside it.
+        Map<String, Property> fromExpressionFields = new LinkedHashMap<>();
+        fromExpressionFields.put(METHOD_EXPRESSION_KEY, FromExpressionOption.subProperty(METHOD_EXPRESSION_LABEL,
+                METHOD_EXPRESSION_DOC, "string"));
```
NipunaRanasinghe: The method expression is typed `string`, but `callRestAPI` declares `RestMethod method`. `activity:RestMethod` is `enum {GET, POST, PUT, DELETE, PATCH}` in both workflow 0.9.0 and 0.10.0. With `string`, the expression editor accepts a plain `string` variable, and the saved call then fails to compile. This should be `activity:RestMethod`, here and in the hidden property at line 227.

The `workflow_var_ref` fixture doesn't catch this: `const string GET_METHOD = "GET"` has the singleton type `"GET"`, so it is assignable to the enum either way.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/RestActivityStrategy.java
```diff
@@ -50,13 +52,22 @@ public class RestActivityStrategy implements BuiltinActivityStrategy {
     public static final String MESSAGE_LABEL = "Message";
     public static final String MESSAGE_DESCRIPTION = "Request body, sent with POST, PUT, PATCH and DELETE";
     public static final String HEADERS_KEY = "headers";
+    // The method dropdown's escape hatch: the method as a string expression, written back unquoted.
+    public static final String METHOD_EXPRESSION_KEY = "methodExpression";
+    private static final String METHOD_EXPRESSION_LABEL = "Method Expression";
+    private static final String METHOD_EXPRESSION_DOC = "The HTTP method as a string expression: a constant or a "
```
NipunaRanasinghe: If the type changes to `activity:RestMethod`, this text should change too. "A string expression" invites a `string` variable, which won't compile. It could also mention `activity:GET`, the usual way to write a method as an expression.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/RestActivityStrategy.java
```diff
@@ -65,54 +76,44 @@ public class RestActivityStrategy implements BuiltinActivityStrategy {
...
+         * @param source the argument source, or {@code null} when absent
+         * @return the selection
+         */
+        public static MethodSelection fromSource(String source) {
+            if (source == null || source.isBlank()) {
+                return template();
+            }
+            String trimmed = source.trim();
+            if (BuiltinActivityStrategy.isBallerinaStringExpression(trimmed)) {
+                String literal = trimmed.substring(1, trimmed.length() - 1).toUpperCase(Locale.ROOT);
```
NipunaRanasinghe: Uppercasing here means `method: "get"` opens as GET and is saved back as `"GET"`. `"get"` isn't a `RestMethod`, so the form silently turns invalid source into valid source, which goes against the rule of writing back what the user wrote. Could this match case-sensitively, so `"get"` goes to From Expression unchanged? If the change of case is intended, a comment saying so would help.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/RestActivityStrategy.java
```diff
@@ -195,16 +262,21 @@ public List<Metadata.AllowedConnector> connectors() {
...
+                : "\"" + method + "\""));
 
         // path — quote if TEXT-typed; only emit when non-default
         BuiltinActivityStrategy.addQuotedArg(args, "path", properties, PATH_KEY);
 
-        // message — every method but GET, whose dispatch in the module ignores it
-        if (isPayloadMethod(method)) {
+        // message — every method but GET, whose dispatch in the module ignores it; an expression
+        // may name any of them, so its body goes along when given
+        if (fromExpression || isPayloadMethod(method)) {
```
NipunaRanasinghe, line 285: Minor: under From Expression, `message` is sent whatever method the expression names, including GET, where the module ignores it. It's harmless, and the comment above covers it, but a body entered under an expression that names GET stays in the source and does nothing.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -1800,13 +1803,33 @@ private void collectCapabilityFields(MappingConstructorExpressionNode config, St
...
+    /**
+     * The cardinality as the form holds it. A member of the enum is written out in source with the
+     * module qualifier ({@code workflow:SINGLE_EVENT}) while the select lists the bare names, so a
+     * member hydrates bare. Anything else — a constant naming one, say — is the field's own source
+     * and hydrates whole, for {@link #cardinalityIsExpression} to put the field in expression mode.
+     */
+    private static void hydrateCardinality(String rawValue, Map<String, String> values) {
+        String trimmed = rawValue.trim();
+        values.put(DurableAgentRegisterEventBuilder.CARDINALITY_KEY,
+                cardinalityIsExpression(trimmed) ? trimmed : WorkflowUtil.stripModulePrefix(trimmed));
```
NipunaRanasinghe, line 1823: `stripModulePrefix` removes any prefix, not only `workflow:`. A constant `other:SINGLE_EVENT` from another module is read as the enum member and saved back as `workflow:SINGLE_EVENT`, which is a different value. It's an edge case, but checking for the `workflow` prefix specifically in `cardinalityIsExpression` would close it.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
```diff
@@ -2469,62 +2498,108 @@ private void populateBuiltinActivityProperties(RemoteMethodCallActionNode callNo
...
+    private String policyMemberName(SeparatedNodeList<FunctionArgumentNode> args, String argName) {
+        for (FunctionArgumentNode arg : args) {
+            if (!(arg instanceof NamedArgumentNode named)
+                    || !argName.equals(named.argumentName().name().text())) {
+                continue;
+            }
+            if (named.expression().kind() != SyntaxKind.MAPPING_CONSTRUCTOR) {
+                return null;
+            }
+            return semanticModel.typeOf(named.expression()).flatMap(TypeSymbol::getName).orElse(null);
```
NipunaRanasinghe, line 2547: For a literal that doesn't type-check, `typeOf` may give the union (`RetryPolicy`) or no name at all. If it's `RetryPolicy`, it isn't in `RETRY_MEMBER_OPTIONS`, so the literal now opens under From Expression. Before, it was guessed from its fields. I think this is the safer result, but it's a change in behaviour and nothing tests it: `RetryPolicyTypeResolutionTest` only pins valid literals. Could you add one case for an invalid literal, or a comment stating the intended result?

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
```diff
@@ -847,19 +856,25 @@ public static void addRetryPolicyFormProperties(NodeBuilder nodeBuilder, String
...
-        if (opaquePolicy) {
-            options.add(new Option(selectedValue, selectedValue));
-        }
+                new Option("Retry, then Review", RETRY_BEFORE_REVIEW_VALUE),
+                FromExpressionOption.option());
+        String requested = retryPolicyValue == null || retryPolicyValue.isBlank() ? NO_RETRY_VALUE : retryPolicyValue;
+        // A caller still passing the policy source as the selection lands on the expression option.
+        boolean known = options.stream().anyMatch(option -> requested.equals(option.value()));
+        String selectedValue = known ? requested : FromExpressionOption.VALUE;
+        String expressionValue = known ? (expression == null ? "" : expression) : requested;
```
NipunaRanasinghe, line 877: Nit: when `retryPolicyValue` isn't a known option, it becomes the expression and any `expression` passed alongside it is dropped. No current caller passes both, so this is fine today. `ApprovalPolicyForm.addFormProperties` has the same pattern.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/RestActivityStrategy.java
hasithaa: Confirmed and fixed in 0962b1bceb. `activity:callRestAPI(http:Client connection, RestMethod method, ...)` with `public enum RestMethod` — so a `string` variable is not assignable and the saved call would not compile. Both the From Expression sub-property and the hidden property now state `activity:RestMethod`. Your reading of why the fixture missed it is right: `const string GET_METHOD = "GET"` has singleton type `"GET"`, which satisfies the enum either way.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/RestActivityStrategy.java
hasithaa: Updated with it: "The HTTP method as an expression: a constant, an enum member such as activity:GET, or a variable of type activity:RestMethod."

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/RestActivityStrategy.java
hasithaa: Agreed, and left as is — under From Expression the method is not known until runtime, so the body has to be offered. The comment above already says the module ignores it on GET.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Fixed. `cardinalityIsExpression` now checks the prefix is the workflow module's own before reading the bare name, so `other:SINGLE_EVENT` stays an expression instead of being saved back as `workflow:SINGLE_EVENT`.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/ActivityCallBuilder.java
hasithaa: Noted — left alone, since no caller passes both and the same shape in `ApprovalPolicyForm.addFormProperties` would want changing with it. Worth doing together if a caller ever needs both.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/model/node/builtin/RestActivityStrategy.java
hasithaa: Made it case-sensitive in 0962b1bceb, and your reasoning is the reason: `"get"` is not a `RestMethod`, so folding it to GET turned source that does not compile into source that does, and saved back a different string than was written.

Worth flagging that this PR's own test asserted the fold — `FromExpressionOptionTest.testMethodSelection` had `fromSource("\"post\"") == MethodSelection("POST", "")`. So the behaviour was deliberate when written, not an oversight, and I have changed it rather than documented it. The test now asserts `"post"` opens under From Expression unchanged, and the stability loop covers it so it reads back the same way. Say if you would rather keep the convenience and document it instead — it is a one-line revert either way.

packages/ballerina-language-server/flow-model-generator/modules/flow-model-generator-core/src/main/java/io/ballerina/flowmodelgenerator/core/CodeAnalyzer.java
hasithaa: Added a test, and it contradicted the supposition in a useful way. With a fixture holding `retryPolicy = {budget: 5}` — matching no member, since `AutoRetry` is closed and the review shapes require an audience — the compiler names **nothing** at all, not `RetryPolicy`:

```
RESOLVED_FOR_INVALID=[null]
```

So an invalid literal never reaches the unknown-member path; it falls back to the fields exactly as before, and there is no behaviour change to worry about. My first version of the test asserted your predicted From Expression and failed with `expected [FromExpression] but found [AutoRetry]`, which is what surfaced it. `testInvalidLiteralFallsBackToTheFields` now pins the null and the unchanged outcome, with a new `retry_policies_invalid.bal` so the valid fixture keeps compiling cleanly.

PR #1153, opened 2026-09-23
Fix non-ASCII characters in Fast Run output

NipunaRanasinghe:
## Purpose

Fast Run shows some non-ASCII characters as replacement characters in the debug console.

Resolves wso2/product-integrator#2540

## Goals

Send the program's output to the client exactly as the program wrote it.

## Approach

Two things were wrong.

The output listener read the process stream in 1024-byte chunks and turned each chunk into a string on its own. A UTF-8 character can be up to four bytes, so a character starting near the end of a chunk lost the rest of its bytes. Reading through an `InputStreamReader` fixes it: the reader keeps the leftover bytes and completes the character on the next read. Output is still forwarded as soon as it arrives, so a `print` without a newline still reaches the console right away.

The launch command also never told the program which encoding to write in. The listener always reads UTF-8, but the JVM writes with the platform's native encoding, which is not UTF-8 on Windows. Added `-Dstdout.encoding=UTF-8` and `-Dstderr.encoding=UTF-8`.

Two tests, because they catch different things:

- `RunExecutorOutputTest` streams known bytes through the listener and checks that what reaches the client matches what went in. It uses a `ByteArrayInputStream` so every read is full and characters always land across a boundary. With a running program a split depends on timing, so the test would pass or fail at random.
- `testWSRunProjectWithNonAsciiOutput` runs a package that prints non-ASCII text and checks it arrives unchanged, covering the encoding of the launched program as well.

## UI Component Development

N/A - no UI changes.

## User stories

N/A

## Release note

Fixed non-ASCII characters being corrupted in Fast Run output.

## Documentation

N/A - bug fix with no documentation impact.

## Training

N/A

## Certification

N/A - internal fix with no impact on certification exams.


packages/ballerina-language-server/langserver-core/src/test/java/org/ballerinalang/langserver/command/executors/RunExecutorOutputTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+        InputStream stdout = new ByteArrayInputStream(programOutput.getBytes(StandardCharsets.UTF_8));
+        new RunExecutor().listenOutputAsync(client, () -> stdout, OUT_CHANNEL);
+
+        await().atMost(5, TimeUnit.SECONDS).until(() -> messagesOn(logCaptor, STOPPED_CHANNEL).size() == 1);
+        return String.join("", messagesOn(logCaptor, OUT_CHANNEL));
+    }
+
+    private static List<String> messagesOn(ArgumentCaptor<LogTraceParams> logCaptor, String channel) {
+        return List.copyOf(logCaptor.getAllValues()).stream()
```
madushajg: `listenOutputAsync` spawns a virtual thread that calls `client.logTrace(logCaptor.capture())` repeatedly while the test thread concurrently polls `logCaptor.getAllValues()` in `await()` and again in the final `messagesOn()` call. Mockito's `ArgumentCaptor` is backed by a plain, unsynchronized `ArrayList` and it's not safe to write from one thread and read from another concurrently. This is a known Mockito anti-pattern and can cause intermittent CI flakiness (`ArrayIndexOutOfBoundsException` or lost captures) since the background thread here is real production code (`Thread.startVirtualThread`), not a mocked/synchronous call. Consider having the mock append to a `Collections.synchronizedList/CopyOnWriteArrayList` via a custom Answer instead of relying on `ArgumentCaptor` across threads.

NipunaRanasinghe: Good catch, you are right. `listenOutputAsync` starts a virtual thread, so the logging happens on a different thread from the one polling in `await()`, and the captor's backing list is a plain `ArrayList`. The `List.copyOf` I had did not help either, since it reads the list while the other thread is still appending.

Fixed in 477c535dde. The mock now appends to a `CopyOnWriteArrayList` through an `Answer`, and nothing touches the captor:

```java
List<LogTraceParams> logs = new CopyOnWriteArrayList<>();
Mockito.doAnswer(invocation -> {
    logs.add(invocation.getArgument(0));
    return null;
}).when(client).logTrace(Mockito.any());
```

One thing worth noting while we are here: `TestWorkspaceManager` uses the same pattern for the fast-run tests. Its `extractLogs` polls an `ArgumentCaptor` that the same virtual threads write to. That one predates this PR. Happy to fix it here as well, or raise it separately if you would rather keep this PR to the encoding fix.

madushajg (approved):

packages/ballerina-language-server/langserver-core/src/test/java/org/ballerinalang/langserver/command/executors/RunExecutorOutputTest.java
madushajg: Yeah, let's address that with a separate PR.

